PR #7, opened 2018-08-03
Confine the moving range of image

zarda:
In case user lose the center of image.


veggiesaurus:
@zarda There are a few other issues related to this PR which we should probably look into. For example, you can currently zoom in/out outside of the range of the image, and you can zoom out further than the FoV of the image.

PR #23, opened 2018-09-10
Add file info, header tab to filebrowser, fix filename sorting issue

duidae:


veggiesaurus:
@duidae This is looking great. Couple of minor issues before a merge:
1. The first time you click on a file, the file info displays the "cannot open file" error until its fetched. We need a separate state for when the frontend is requesting file info.
2. The load button should be disabled when in the above state. At the moment, I can click on file #1, then click on file #2 and quickly click load.

PR #24, opened 2018-09-11
First pass of hooking up additional AST overlay properties to the UI

confluence:
This PR adds a lot more functionality to the overlay dialog. This is still a work in progress, but the dialog is currently usable.

Some known issues still to be fixed:

* We need a nicer input for the number format (perhaps a drop-down of common options with a possibility of entering a custom string).

* If the labelling type is set to "exterior", AST ignores most of the main axis properties. We need to disable them when appropriate and add a tooltip and/or button to make it easy to re-enable them.

* There are other per-axis settings, for example for the grid, which should perhaps be included. This would require some rearrangement of the layout.

* We need to decide which defaults it is sensible to allow AST to autodetect per file, and which ones it is sensible to remember globally.

* Saving and loading presets hasn't been implemented yet.


veggiesaurus:
I've had a look, and it's mostly working nicely. However, there are some issues that need to be fixed:

1. The settings container doesn't resize to fill the available vertical space. Resizing the dialog doesn't get rid of the the scroll bar
2. Some defaults are needed for all the values, so that the user can see what is currently being used before changing them. For values that are AST-dependent, we can work out a way to fetch from AST
3. Toggling axis visibility doesn't do anything if the number or labels are set to visible. 
4. Logging should be removed
5. The numeric input controls should be made wide enough to at least hold the placeholder value
6. Entering anything in the Axes' number format textbar breaks AST.
7. Entering anything in the Axes' label text breaks AST. 
8. It looks like you're switching from uncontrolled to controlled when text is entered.

While I think we need to get some overlay settings merged in soon, some of these need to be fixed before merging. Especially 3, 6 and 7.

confluence:
1. There is an issue with getting scrolling to work correctly in the nested dialog elements -- the current solution of setting the panel to a fixed pixel size is a stopgap.

2. Agreed -- I think that a lot of the defaults should be fetched from AST.

3. I can't reproduce this. Are you sure the labelling is set to Interior when you test this? If it is set to Exterior, the axes are never drawn. There is an initial change in the tick spacing when you toggle the axes switch for the first time, but I believe that is related to one of the default values being overwritten (which should be fixed when the defaults are handled consistently).

4. I've removed the logging statement I added.

5. I have made all the numeric inputs wider. I'm not sure whether these placeholder values will actually end up being seen, if we add defaults for everything.

6 & 7. I need to investigate this further; it looks like we can't get away with a free text entry which auto-updates. I'll make these into dropdowns.

8. This was previously caused by a lack of a default value. Which field?

confluence:
7 is fixed -- the issue was that it doesn't make sense to set the same label text on all axes (and thus it's not a valid AST option).

I'm pretty sure that 6 needs a dropdown. Even if we disable the input auto-updating as the user types, any mistake will cause a crash, which is bad. We could potentially allow arbitrary string entry if we had a way of validating the input, which will require a bit more work.

confluence:
6 is now also fixed (no more format input for all axes).

veggiesaurus:
> 1. There is an issue with getting scrolling to work correctly in the nested dialog elements -- the current solution of setting the panel to a fixed pixel size is a stopgap.
Perhaps try using a flex box layout. Alternatively, we can just remove the fixed panel height, and then scroll up/down the entire settings panel

I guess there's not too much point refining that, because the settings arrangement will change based on what you've mentioned already, and feedback

confluence:
I have removed the dependency of the editability of the numbers and labels subsections on the visibility of the axes, since they aren't really coupled. I have verified that each of these sets of elements can be toggled independently of the others. Does this resolve 3?

(These options will be reorganised further, since their current grouping is a bit arbitrary.)

veggiesaurus:
> I have removed the dependency of the editability of the numbers and labels subsections on the visibility of the axes, since they aren't really coupled. I have verified that each of these sets of elements can be toggled independently of the others. Does this resolve 3?
> 
> (These options will be reorganised further, since their current grouping is a bit arbitrary.)

Yep :+1: 

As for (8), it's when I enter a string in the axis-specific label settings

Issue #38, opened 2018-09-28
AST configuration options


veggiesaurus:
AST config. options:
- if it is RA/DEC, use sexagesimal format by default, rather than in degree format.
		(e.g., R.A. = 12h:23m:43.234s, Dec = -34d:35m:23.123s)
- if it is galactic coordinate, use degree format by default
		(e.g., GLON = 23.123 degree, GLAT = -5.234 degree)
- "WCS(a, b)" in the cursor info bar should follow the above two rules
- default format for "Numbers" as "d.3"

- Overlay settings -> Global -> Color (not functional; this is supposed to change the color of ticks, grid, label, etc. in one shot?)

arusstaylor:
d.3 is no likely to be enough precision for modern images.  Typical JVLA, GMRT and MeerKAT pixels sizes are of order 1 arc or less, which is a ~10^-4 degrees.  Is it possible to link the precision of the display format to the pixel size of the image?  For a VLBI images, for example, the pixels may be milli-arcseconds, which is ~10^-7 degrees.  I think it would be good if the precision of the display were d.3 as minimum but is adapted to the image pixel size such  that the last digit of the display coordinate will change from pixel to pixel.

Russ

veggiesaurus:
As @kswang1029 was the author of the initial issue (I just copied it), hopefully he will comment on that.
 
@arusstaylor What I'd think would be ideal would be to determine the number of decimal places required based on the current image view. So for example, if your view is initially quite wide, 3 decimals might be sufficient, but as you zoom in, it should include more decimal places. That avoids the issue of having a whole number of decimal places that aren't useful, because the field of view is wide.

kswang1029:
@arusstaylor @veggiesaurus Right, modern images can have quite small pixel sizes so that d.3 is not sufficient. As @veggiesaurus suggested, I think what we display is determined by two factors, one is the FOV, which determines the minimum precision, the other is the pixel size, which then determines the maximum precision. As we zoom in, we increase the precision dynamically. 
If this may take some time to implement, perhaps we can set "d.3" as default just for the release. Currently "d.1" is not enough in general cases.

arusstaylor:
If we have to go with a fixed value default for the initial release then I recommend d.5 so we have arc second precision.  Most of the images from MeerKAT will require this.

Russ

kswang1029:
@arusstaylor Sure for d.5 as a temporary fixed default.

veggiesaurus:
[...]

PR #44, opened 2018-10-15
More AST overlay changes

confluence:
This PR fixes issues raised in #38 and introduces various other UI improvements:

* Coordinate system is set from file
* Default number format is calculated and set from file
* Axis number precision is automatically calculated by AST
* Cursor number precision is configurable (not automatically calculated yet)
* Global colour is used by default for all plot elements
* Consistent naming for custom options
* Unused custom options are collapsed


veggiesaurus:
Mostly fine. Few minor issues that should be fixed before merge:
1) The color select's colors are still different before and after loading a file. Probably need to change AST's default palette, or set it on AST load
2) Quite a few of the global options (font, font size, width) change nothing. Perhaps we should just remove them?

confluence:
1. Agreed.
2. Also agreed. The font may be a useful thing to set globally (in which case it should be handled like the colour, but then we should have a more intelligent UI for bolding and italics (which are currently just exposed as separate fonts). Font size and width don't make sense as global options, since they are likely to be different for different elements.

veggiesaurus:
> The font may be a useful thing to set globally (in which case it should be handled like the colour, but then we should have a more intelligent UI for bolding and italics (which are currently just exposed as separate fonts).

I'm happy for you to just remove it for now, and we can add a global font setting later, with a more intelligent UI (we can just have "bold" and "italic" icons I guess).

confluence:
If we want to be really clever we can have a global width setting and set individual element widths as percentages of the global width, but that may be unnecessarily complicated. Changing widths per element seems more intuitive.

Issue #55, opened 2018-10-19
[improvement] overlay setting defaults


kswang1029:
title -> font size to ~18. 24 is a bit too large.
numbers -> font size to ~12. 10 is a bit too small.

please rotate y-tick label by 90 degree so that y-axis label can be displayed independent of the precision of y tick numbers. 

We are almost there! Well done!

Issue #54, opened 2018-10-19
AST options: grid -> custom gap


kswang1029:
frontend version: b9e44e2
In overlay setting/grid, if custom gap is toggled, the coordinate grid is disappeared.

veggiesaurus:
@kswang1029 So this is a tricky one. The "gap" value is in the units of the current file. So if the current file is in degrees, then the gap will be in degrees, and so on. 

When you select "custom gap", the gap value is set from the value in the numeric input below. Depending on the FOV of the image viewer, this value may or may not be sensible, but it's difficult to choose a starting value that works for all images. 

In my mind, that's not a huge problem, but could probably be explained more clearly to the user. What *is* a problem is that AST throws an error when the gap size is too large for any grid lines to be drawn, and that prevents the rest of the AST overlay from being shown. @confluence , I think that's what we should attempt to fix in the short term.

confluence:
Fixed in my branch.

veggiesaurus:
this now behaves correctly. No grid is drawn if the gaps are too large, but the rest of the AST overlay still gets rendered correctly.

Issue #52, opened 2018-10-19
AST: changing coordinate system results in incorrect tick labels


kswang1029:
using IDIA carta with frontend version: b9e44e2

To reproduce:
1) open orion_12co_hera.hdf5
2) we should see correct cursor coordinate and tick label coordinate
![screen shot 2018-10-19 at 9 14 14 am](https://user-images.githubusercontent.com/20819712/47192651-03a22500-d381-11e8-9410-f20c89d383d6.png)

3) now change the coordinate system from FK5 to galactic
4) issue 1: the tick label format should be changed to degree automatically.
![screen shot 2018-10-19 at 9 15 31 am](https://user-images.githubusercontent.com/20819712/47192684-2df3e280-d381-11e8-974a-2489730d5f93.png)

5) issue 2: by manually changing the format to degree, the tick labels (incorrect) are not consistent with the cursor coordinate which should be the correct one. 
![screen shot 2018-10-19 at 9 15 56 am](https://user-images.githubusercontent.com/20819712/47192712-554aaf80-d381-11e8-8d39-a3e039a7432f.png)

6) similarly, with supermosaic.10.hdf5, changing from galactic to FK5 results in the same issue.

veggiesaurus:
@kswang1029 Can you clarify which tick format should be used for each of the coordinate systems? It seems AST's defaults are slightly different to those that you requested. The systems are:

- FK4
- FK5
- Galactic
- Ecliptic
- ICRS

kswang1029:
FK4: hms/dms
FK5: hms/dms
Galactic: degree/degree
Ecliptic: degree/degree
ICRS: hms/dms

confluence:
This has been fixed in my branch. (Cursor format changes are not immediately visible, but become visible as soon as the cursor moves.)

veggiesaurus:
@ad

veggiesaurus:
fixed in #59

kswang1029:
with frontend version: b17a8aa, is the issue 2 supposed to be fixed? It appears the issue is still there. The tick labels and cursor WCS are not consistent. 
<img width="816" alt="screen shot 2018-10-24 at 09 28 04" src="https://user-images.githubusercontent.com/20819712/47400353-66276680-d76f-11e8-8236-70c710d34ad4.png">

veggiesaurus:
@confluence can you look into this? I'm opening the issue again

kswang1029:
[...]

PR #59, opened 2018-10-19
Some AST overlay fixes

### LINKED ISSUE #50: AST config: coordinate systems when multiple frames exist
The coordinate system should remain the same when new image is loaded as new frame.
To reproduce: 
1) open orion_12co_hera.hdf5
2) open supermosaic.10.hdf5 as new frame
3) use animator to switch back to orion
4) check the gridline 
screen shot 2018-10-18 at 20 38 28

screen shot 2018-10-18 at 20 39 00

confluence:
This should fix #50, #52 and #55 (edit: and #54):

* Coordinate system is left as native by default; this means it's automatically read from the file. A manually set system should be persistent.
* Default coordinate system is updated every time the frame changes.
* The default number format is now explicitly set based on the coordinate system, and is updated every time the coordinate system changes. A manually set format should be persistent.
* Y-tick labels were rotated and the font size was increased. The title was removed for now, since it is unused and the rotation of the Y-labels interferes with the positioning of the title. If we need to put it back, we should probably render it ourselves outside the plot area. We may need to do the same with labels, which are currently also sometimes displaced when numbers overflow the edges of their axes.

Known issues:
* Occasional label displacement as described above.
* If the selected coordinate system is not native, AST sometimes ignores requested plot options when they would be difficult or impossible to plot. E.g. at some zoom levels Y-ticks may be rotated back to horizontal or internal labelling may be used. It's unclear whether this behaviour can be changed. It mostly seems to be triggered when there is limited space available because the viewport is small. I can replicate it by opening a RA/DEC image, setting the coordinate system to galactic, and zooming in and out.


veggiesaurus:
merging. Nice work, @confluence

PR #60, opened 2018-10-19
button group for plot type implemented

veggiesaurus:


arusstaylor:
Looks good Angus.
Russ

PR #67, opened 2018-10-23
Confluence/ast final features

### LINKED ISSUE #62: [imageViewer] axis label and tick label
frontend version: a588672

issue 1: 
the gap of axis label needs to be larger
screen shot 2018-10-21 at 11 50 09

issue 2:
if coordinate transform is applied (attached is from native to galactic), y tick label is no longer rotated by 90 degree.
screen shot 2018-10-21 at 11 50 27

confluence:
Fixes #62.

* When a coordinate transform is applied, sometimes the orientation of the axes flips. So *both* axes need to be set to "not upright".

* Added some tweaks to label gap and padding that scale with screen zoom.


veggiesaurus:
@kswang can you give this a test to see if it looks ok? Updated build on dev-viewer

PR #75, opened 2018-10-25
adjusted channel slider step to always include first and last channels

### LINKED ISSUE #68: [improvement] channel index in animator
The channel indices in the animator collide when the panel is narrow and the total number of channels is too large. Attached is an example viewing a cube with a 7680-channel cube on my laptop.
screen shot 2018-10-24 at 12 32 34

Perhaps we can improve this by just showing the first and the last channel indices. Users can estimate where to click to jump to a certain channel with the help of the channel index below the slider. 
Or, the indices can be generated in a more smart way with the consideration of the widget width.

veggiesaurus:
in addition, the number of step labels is limited to 5. 
Fixes #68

### DIFF
--- a/src/components/Animator/AnimatorComponent.tsx
+++ b/src/components/Animator/AnimatorComponent.tsx
@@ -224,8 +224,8 @@ export class AnimatorComponent extends React.Component<WidgetProps> {
 
         // Channel Control
         if (numChannels > 1) {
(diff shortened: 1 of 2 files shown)


veggiesaurus:
@kswang1029 can you have a look with that ~8000 channel image, and any images with channels <= 10?

PR #82, opened 2018-11-08
Add functionality test: FILETYPE_PARSER

zarda:
According to test ICD [carta_ICD_testCases](https://docs.google.com/document/d/1CDa3tpBxKZX8xX2dopKHsYbVwUFZkKHJAYV-ugH4_MQ/edit?ts=5bc715d1), the test "FILETYPE_PARSER" gives a testing helper for the event FILE_LIST_REQUEST & FILE_LIST_RESPONSE on [CARTA ICD](https://docs.google.com/document/d/1V4KzJaV3zQYJoA_ims28iBTit_aixo_xYcAfPiq8hic/edit#heading=h.58ke5sat0rfq).

### DIFF
--- a/src/test/FILETYPE_PARSER.test.ts
+++ b/src/test/FILETYPE_PARSER.test.ts
@@ -137,7 +137,6 @@ describe("FILETYPE_PARSER tests", () => {
             // While receive a message from Websocket server
             Connection.onmessage = (event: MessageEvent) => {
                 const eventData = new Uint8Array(event.data, 36);
-
                 expect(CARTA.FileListResponse.decode(eventData).parent).toBe(expectRootPath);
 
                 done();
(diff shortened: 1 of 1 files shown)


veggiesaurus:
Is the reason you've kept separate tests for files and missing files purely because they're different file types? If so, that's ok, since they're technically testing different things. @kswang1029 this doesn't match your description in the google doc, but I think it's ok, since we're testing different files. Some of the other tests could be combined though. Things like checking the event size and `success = true` would be a candidate for combination into a single test with multiple asserts, as `success` will never be true if the event size is < 40. @kswang1029 thoughts?

kswang1029:
I am not sure technically how FILETYPE_PARSER is implemented. Basically this test is trying to verify what should and should not be seen in the test folder set_QA.

For the assertions in other test cases, I will try to see if trivial tests can be combined into a single assertion.

veggiesaurus:
A single test can have multiple assertions, no need to combine them into one assertion. But we can combine them into one test. For example: "does the backend return the correct file info?" That test would have multiple assertions, but there's no need to test each assertion in a separate test. It just leads to massive code bloat

kswang1029:
I see. That’s what I thought when I wrote this test case.

zarda:
The testing codes were modified. Every file_info_request will be sent after restarting a new connection.

veggiesaurus:
@zarda one minor change before PR merge: can you change all occurances of " is existed" to "exists" please?

PR #89, opened 2018-11-27
 Add testing: RASTER_IMAGE_DATA_PERFORMANCE

zarda:
According to [test ICD,](https://docs.google.com/document/d/1CDa3tpBxKZX8xX2dopKHsYbVwUFZkKHJAYV-ugH4_MQ/edit?ts=5bc715d1) the test "RASTER_IMAGE_DATA_PERFORMANCE" gives a performance testing for the functionality of RASTER_IMAGE_DATA on [CARTA ICD](https://docs.google.com/document/d/1V4KzJaV3zQYJoA_ims28iBTit_aixo_xYcAfPiq8hic/edit#heading=h.sgxhpgbn2sks).

### DIFF
--- a/src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
+++ b/src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
@@ -106,53 +106,7 @@ describe("RASTER_IMAGE_DATA_PERFORMANCE tests", () => {
         ].map(
             function ([testFileName, imageBounds, mip, compressionType, compressionQuality, numSubsets]: 
                     [string, {xMin: number, xMax: number, yMin: number, yMax: number}, number, CARTA.CompressionType, number, number]) {
-                test(`assert the file "${testFileName}" loads info.`, 
-                done => {
-                    // Preapare the message
-                    let message = CARTA.FileListRequest.create({directory: testSubdirectoryName});
-                    let payload = CARTA.FileListRequest.encode(message).finish();
-                    let eventDataTx = new Uint8Array(32 + 4 + payload.byteLength);
-            
-                    eventDataTx.set(Utility.stringToUint8Array("FILE_LIST_REQUEST", 32));
-                    eventDataTx.set(new Uint8Array(new Uint32Array([1]).buffer), 32);
-                    eventDataTx.set(payload, 36);
-            
-                    Connection.send(eventDataTx);
-            
-                    // While receive a message
-                    Connection.onmessage = (event: MessageEvent) => {
-                        let eventName = Utility.getEventName(new Uint8Array(event.data, 0, 32));
-                        if (eventName === "FILE_LIST_RESPONSE") {
-                            let eventData = new Uint8Array(event.data, 36);
-                            expect(CARTA.FileListResponse.decode(eventData).success).toBe(true);
-
-                            // Preapare the message
-                            message = CARTA.FileInfoRequest.create({
-                                directory: testSubdirectoryName, file: testFileName, hdu: ""});
-                            payload = CARTA.FileInfoRequest.encode(message).finish();
-                            eventDataTx = new Uint8Array(32 + 4 + payload.byteLength);
-
-                            eventDataTx.set(Utility.stringToUint8Array("FILE_INFO_REQUEST", 32));
-                            eventDataTx.set(new Uint8Array(new Uint32Array([1]).buffer), 32);
-                            eventDataTx.set(payload, 36);
-
-                            Connection.send(eventDataTx);
-
-                            // While receive a message
-                            Connection.onmessage = (eventInfo: MessageEvent) => {
-                                eventName = Utility.getEventName(new Uint8Array(eventInfo.data, 0, 32));
-                                if (eventName === "FILE_INFO_RESPONSE") {
-                                    eventData = new Uint8Array(eventInfo.data, 36);
-                                    let fileInfoMessage = CARTA.FileInfoResponse.decode(eventData);
-                                    // console.log(fileInfoMessage.fileInfoExtended);
-                                    expect(fileInfoMessage.success).toBe(true);
-                                    done();
-                                } // if
-                            }; // onmessage
-                        } // if
-                    }; // onmessage "FILE_LIST_RESPONSE"
-                }, openFileTimeout); // test
-
+                
                 test(`assert the file "${testFileName}" opens.`, 
                 done => {
                     // Preapare the message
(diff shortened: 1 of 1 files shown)


src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
```diff
@@ -0,0 +1,383 @@
...
+                Connection.send(eventData);
+            } else {
+                console.log(`Can not open a connection.`);
+            }
+            done();
+        };
+    }, connectionTimeout);
+
+    test(`connect to CARTA "${testServerUrl}" & ...`, 
+    done => {
```
veggiesaurus: again, not sure why we need to test these again each time.

src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
```diff
@@ -0,0 +1,383 @@
...
+         ["cluster_16384.fits", {xMin: 0, xMax: 16384, yMin: 0, yMax: 16384}, 46, CARTA.CompressionType.ZFP, 11, 4],
+         ["cluster_32768.fits", {xMin: 0, xMax: 32768, yMin: 0, yMax: 32768}, 92, CARTA.CompressionType.ZFP, 11, 4],
+         ["hugeGaussian10k.fits", {xMin: 0, xMax: 10000, yMin: 0, yMax: 10000},  28, CARTA.CompressionType.ZFP, 11, 4],
+         ["hugeGaussian20k.fits", {xMin: 0, xMax: 20000, yMin: 0, yMax: 20000},  56, CARTA.CompressionType.ZFP, 11, 4],
+         ["hugeGaussian40k.fits", {xMin: 0, xMax: 40000, yMin: 0, yMax: 40000}, 112, CARTA.CompressionType.ZFP, 11, 4],
+         ["hugeGaussian80k.fits", {xMin: 0, xMax: 80000, yMin: 0, yMax: 80000}, 223, CARTA.CompressionType.ZFP, 11, 4],
+        ].map(
+            function ([testFileName, imageBounds, mip, compressionType, compressionQuality, numSubsets]: 
+                    [string, {xMin: number, xMax: number, yMin: number, yMax: number}, number, CARTA.CompressionType, number, number]) {
+                test(`assert the file "${testFileName}" loads info.`,
```
veggiesaurus: same question here: We've already got tests that check for file info. @kswang1029 thoughts on this?

src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
```diff
@@ -0,0 +1,383 @@
...
+                        }
+                    };
+                }, openFileTimeout);
+            }
+        );
+    });
+
+    describe(`prepare the files`, () => {
+        [ 
+         ["cluster_00128.fits", {xMin: 0, xMax:   128, yMin: 0, yMax:   128},  1, CARTA.CompressionType.ZFP, 18, 4],
```
veggiesaurus: it doesn't look like you're using the last three components of this in the test itself (they're hard-coded in the request below)

src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
kswang1029: I suggest in this performance test we do not make assertions on file info. File info assertions should be in the dedicated test. At this moment I only add very basic file info assertions. Once we have settle down the backend, I will add more. Currently the two BEs generate a bit different file info.

src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
kswang1029: I think for each test image in this performance test, we either start a new carta session for each image, or we use a single session for all test images (this can test if backend resources are handled well or not). Probably we should do both?

veggiesaurus: It's fine to start a new test each time, so each test will be independent. But we don't need to explicitly test that the file loads and file info is correct each time.

src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
veggiesaurus: The performance tests should also test how performance scales with subsets, so perhaps some tests should have 1, 2, 3 or 4 subsets.

kswang1029: OK, sure. I will try to make test case for this.

src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
zarda: I suppose to make a single test for one development at one purpose. That is after we make sure in which function the bug happens, the programer only needs to focus on a signle testing item. In the mean while, there is some chance to have other issue about connection and OS. 
And another reason is that there will be too many error messages involved each other at a test which has many assertions. My solution is to run every assertion at a renew process.

src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
zarda: About the performance testing, do we need to have the timeout function to postpone the test after a previous test? In case the tests in my laptop have unstable outcome of performance. The disk IO supposes to be the issue.

src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
veggiesaurus: it's not about renewing the process, that's fine, and that's a sensible idea, since each test will then be independent of the the previous test. However, when testing image performance, or spatial profiles, or any other feature, there's no need to re-test the file loading or file info. We already have separate tests for that.

veggiesaurus: multiple assertions in a single test are also fine, because the test utility will specify which assertion failed.

src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
veggiesaurus: we should prevent multiple tests from running at the same time. Ideally, we should find a way to clear disk cache between tests as well.

src/test/RASTER_IMAGE_DATA_PERFORMANCE.test.ts
zarda: The main purpose of re-testing the file loading and file info is not to assert the programme codes. It may happen there is no supported meta data for loading a certain fits file for instance. I suppose this kind of testing will be removed to other testing about "support file format". I will change the code to not re-test file loading again.

veggiesaurus: in terms of file support, that should be handled by a completely separate test. When the frontend sends a file open event to the backend, the backend will respond with an OPEN_FILE_ACK message, which will contain sufficient information to determine whether the file is supported or not. So there is no neat to combine that with tests of raster image data

zarda: only left the assertion "file opens" to show a based result to compare the benchmark.

zarda: PASS  src/test/CURSOR_Z_PROFILE_PERFORMANCE.test.ts (7.668s)
  CURSOR_Z_PROFILE_PERFORMANCE tests
    ✓ connect to CARTA "ws://localhost:50505" & ... (5ms)
    access directory
      ✓ assert the directory "/Users/zarda/CARTA/Images" opens. (8ms)
      ✓ assert the directory "/Users/zarda/CARTA/Images/QA" opens. (9ms)
    prepare the files
      ✓ assert the file "SgrB2-N.spw0.line.fits" opens. (46ms)
      ✓ assert a random cursor at round 1. (7117ms)
      ✓ assert a random cursor at round 2. (96ms)
      ✓ assert a random cursor at round 3. (94ms)
      ✓ assert a random cursor at round 4. (103ms)
      ✓ assert a random cursor at round 5. (100ms)

  console.log src/test/CURSOR_Z_PROFILE_PERFORMANCE.test.ts:258
    set cursor randomly: returning time = 7037,6,6,8,5 ms. mean = 1412.4 ms. deviation = 2812.3001689008947 ms.

Test Suites: 1 passed, 1 total
Tests:       9 passed, 9 total
Snapshots:   0 total
Time:        7.714s, estimated 11s

veggiesaurus: if we're just measuring Z profile performance, then the file opening etc is not a required performance metric for this test, and should be in a different test

Issue #36, opened 2018-09-28
imageViewer cursor


kswang1029:
imageViewer:
- Need to show velocity, frequency, and Stokes of the current frame somewhere (integrated in the cursor info bar, maybe?). 
Format for velocity is like "**%.3f km/s, LSRK / Radio, Q**". Here "LSRK" can be read  from the header keyword "SPECSYS", while "Radio" can be read from the header keyword "VELREF". "Q" indicates the Stokes parameter of the current frame. Format for frequency is like "**%.3f MHz, LSRK, Q**", where "LSRK" can be read  from the header keyword "SPECSYS"
- as more information is shown in the cursor info bar, we may need a two-line bar by default.

veggiesaurus:
@kswang1029 splitting this into two separate issues. @confluence will handle the AST config options.

veggiesaurus:
almost complete, just requires hi-dpi adjustments

veggiesaurus:
fixed in #65

veggiesaurus:
@kswang1029 are you happy with the functionality in PR #91 in terms of this issue? For post-V1, we can add automatic precision and reference frame conversion.

kswang1029:
Yes. The frequency/velocity display in the cursor info bar works well. Tested all possible frames that casa can transform. Compared with casaviewer which supports a subset of frames and results are consistent.

PR #91, opened 2018-12-05
cursor info freq

veggiesaurus:
This PR adds spectral info to the cursor

- For images without any spectral info, only the channel index is used
- For images with spectral info, the spectral value of the current channel is used
- For images with a valid SPECSYS keyword, the spectral system is also displayed
- For images with a valid rest frequency and a spectral type of "FREQ", conversions to (radio) velocity are displayed
- For images with a valid rest frequency and a spectral type of "VRAD", conversions to frequency


veggiesaurus:
fixes #36

PR #93, opened 2018-12-05
Confluence/save images

confluence:
This PR adds the ability to export the image and overlay as a PNG, from the main menu or from a quick access toolbar. The toolbar also contains shortcuts for zooming to fit the image width or height and toggling the grid and the labels on and off.


veggiesaurus:
I can't reproduce the issue I had yesterday (possibly just my Chrome being weird), and the code looks good, so I'm merging this

PR #94, opened 2018-12-06
Refetor testing to connect server

zarda:
Duo to the backend service has addressed on server. We fix the predefine to adapt the service.
Several refactor and correction are done.


veggiesaurus:
see comment in other PR about not hardcoding the server.

PR #107, opened 2018-12-11
Angus/shortcut modifier macos linux

veggiesaurus:


veggiesaurus:
@kswang1029 can you please adapt the documentation to include these shortcuts?

kswang1029:
Will do

PR #137, opened 2019-02-15
CARTA 1.0.1 patch (data folder changes)

confluence:
These are the frontend changes for the CARTA 1.0.1 patch. **This PR should only be merged once `pam/folder_arguments` is merged into the backend master branch.**

1. The default starting file directory is set to "$BASE"
2. The last directory from which the user opened a file is remembered and used as the starting directory the next time the dialog is used.


veggiesaurus:
@confluence bit out of the loop, is this now ready to merge in?

confluence:
I don't think so -- @pford's branch is not merged into the backend yet.

Issue #140, opened 2019-02-21
[Profiler] broken profiler after scrolled on initially loaded widget


duidae:
reproduce the bug:

open frontend without loading any image -> scrolled on profiler widget -> load an image -> broken profiler

![out-2 ogv](https://user-images.githubusercontent.com/25971696/53163099-54b07700-3608-11e9-995e-8ed2ff3f577c.gif)

kswang1029:
nice catch. It can be fixed by double-clicking to reset the range. But, yes we should prevent this. Would it be possible that we show "No file load" same as the render configuration widget? This should fix this issue right away.

PR #143, opened 2019-02-25
fix broken profiler after scrolled on initially loaded widget

duidae:
Check data existence before to prevent from changing the chart setting when scrolling on it


veggiesaurus:
fixes #140

PR #153, opened 2019-02-28
Jason/draggable spectral profiler

duidae:
this PR mainly solves:
1. draggable marker to change channel in spectral profiler
2. reorganize draggable prop from Rect to Group

### DIFF
--- a/src/components/RenderConfig/RenderConfigComponent.tsx
+++ b/src/components/RenderConfig/RenderConfigComponent.tsx
@@ -263,6 +263,7 @@ export class RenderConfigComponent extends React.Component<WidgetProps> {
                 id: "marker-min",
                 label: this.widgetStore.markerTextVisible ? "Min" : undefined,
                 draggable: true,
+                dragCustomBoundary: {xMax: frame.renderConfig.scaleMax},
                 dragMove: this.onMinMoved,
                 horizontal: false,
             }, {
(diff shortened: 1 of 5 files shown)


kswang1029:
Nice work.

When animation is on, the slider in the animator is disabled (so cannot drag it). In the spectral profiler, it is still draggable when animation is on. Should we make them consistent? I quickly test it when animation is on and it appears fine so far. Will do more detailed tests.

veggiesaurus:
I agree, we should make it consistent. The other thing I've been thinking is that there's a problem with the profiler's marker at the moment: 
- The slider's position is based on the _requested channel
- Dragging the slider changes the _requested_ channel
- The marker's position is based on the _current_ channel
- Dragging the marker changes the _requested_ channel
This leads to jumpy behaviour when the requested channel is not equal to the current channel

markers_comparison

I propose we do the following: 
- Make the _draggable_ marker equivalent to the slider's marker (based purely on the _requested_ channel). This will make the dragging much smoother (see below for demo based on quick hack.. the jumps there were due to me clicking)

- Show another (non-draggable) marker to indicate what the current channel is. If the two are identical, only one will be shown in any case 
- When dragging, hide the "cursor" marker, since it's no longer needed, and will be behind the requested channel marker at all times. 

marker_requested

PR #157, opened 2019-03-04
fixed scaling issues when using renderHiDPI flag and set it enabled by default

### LINKED ISSUE #156: default to using HiDPi resolution for image viewer
the `FrameStore::renderHiDPI` flag controls whether the frontend takes into account the client's `devicePixelRatio` value when requesting data. At the moment, this is set to `false`, so users with HiDPi screens (almost all OSX devices) do not get image data that matches their screen resolution. This leads to blurry images.

This flag should be switched on by default, and the WebGL canvas should be adjusted accordingly to support this flag.

veggiesaurus:
@kswang1029 this is the issue we were discussing previously, where images were slightly blurry, even when viewed at 1.0 resolution. Can you give it a check?

fixes #156


kswang1029:
@veggiesaurus which demo server I could give it a try? idia? Or we need to set it up?

Issue #168, opened 2019-03-11
[spectral profiler] jumping channel indicator while dragging


kswang1029:
with the demo server: CARTA v1.1-alpha | frontend 2b16cf5 | backend bff7bc0
The channel indicator (red line) jumps back and forth while dragging. The following example tries to change the channel from 79 to 165.
![spectralProfiler](https://user-images.githubusercontent.com/20819712/54106604-a717d300-4411-11e9-8975-20f0a8529cad.gif)

veggiesaurus:
@kswang1029 fixing this requires changing the marker to the "requested" channel instead of current channel: https://github.com/CARTAvis/carta-frontend/issues/123#issuecomment-470862572

@duidae can you give this a try?

duidae:
@veggiesaurus okok

PR #174, opened 2019-03-12
Jason/fix 168

duidae:
closes https://github.com/CARTAvis/carta-frontend/issues/168

### DIFF
--- a/src/components/SpectralProfiler/SpectralProfilerComponent.tsx
+++ b/src/components/SpectralProfiler/SpectralProfilerComponent.tsx
@@ -291,7 +291,7 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
                 }
                 linePlotProps.cursorX = {
                     profiler: this.widgetStore.cursorX,
-                    image: this.getRequiredChannelValue(),
+                    image: this.getCurrentChannelValue(),
                     unit: this.getChannelUnit()
                 };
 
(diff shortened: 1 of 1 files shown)


veggiesaurus:
@kswang1029 this solves the jumpy cursor by displaying the _required channel_, rather than the currently displayed channel. However, do you think we should also show the currently displayed one on the graph if they are different?

kswang1029:
Good idea. Maybe with different transparency?

veggiesaurus:
It looks pretty odd with the extra marker: 
show_req_current_channels

Not sure what to do, really. On the one hand, the slider is not great if you use the current channel (jumpy) and on the other hand, if you use the required channel, you aren't displaying a marker indicating which channel is actually displayed, which may lead to confusion. 

The two was around this I can think of are:
1. Display a marker for the current channel as well (see attached, looks a bit weird)
2. Disable dragging of the marker entirely, and just allow the user to click on the channel to move the marker. Which would be annoying, as @duidae has put a fair bit of time and effort into this, and it's mostly great.

@kswang1029 any other ideas?

kswang1029:
I suggest “current” so that the values will be consistent with the values in other widgets (cursor info bar in viewer, xy profilers).

arusstaylor:
Is dragging the marker the only way to change the channel with the marker?  The most common use case for this is to be able to click on a feature in the profile and have the channel change.  Therefore need to be able to change the channel just with a click.

Russ

kswang1029:
@arusstaylor I think now we have both: dragging to change channels and clicking to jump to a certain channel. @veggiesaurus @duidae could you confirm?

veggiesaurus:
@arusstaylor yes, you can left click to move the marker

PR #181, opened 2019-03-15
Jason/fix 111

### LINKED ISSUE #111: [colormap; improvement] reduce number of colormaps
suggestion from @arusstaylor , some colormaps are not that useful for science and we may want to reduce it for v1.1+.

duidae:
closes #111


duidae:
adjusted the css because when setting popover's height to 300, the overflowed items are not selectable

bb

PR #185, opened 2019-03-15
hides region size and rotation columns if region list widget doesn't have enough space

### LINKED ISSUE #184: Region list should remove columns when width is too small
Currently, the region list displays all columns, no matter the size of the widget

As polygon and point shapes don't have rotation or size, these columns can be hidden when the available size is below a certain cutoff.

veggiesaurus:
This PR adds the following:
- If the region list's name field is too small, the region size column is hidden
- If it is _still_ too small, it also hides the rotation column

These two columns are not as important as the others, as point and polygon regions don't have them

closes #184

### DIFF
--- a/src/components/Menu/RootMenuComponent.tsx
+++ b/src/components/Menu/RootMenuComponent.tsx
@@ -23,13 +23,13 @@ export class RootMenuComponent extends React.Component<{ appStore: AppStore }> {
                 <Menu.Item
                     text="Open image"
                     label={`${modString}O`}
-                    disabled={connectionStatus !== ConnectionStatus.ACTIVE || appStore.fileLoading}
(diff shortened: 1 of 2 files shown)


duidae:
a few opinions about this PR (I can help with that):
1. replace constant numbers with define, e.g. 50 => MINIMUM_WIDTH
2. how about using blueprint/table instead of HTML table in blueprint/core? many features available for later enhancement, such as sorting type/area/rotation, directly editable cell for center/rotation (pitty I don't see hide option for columns)

veggiesaurus:
Agree with the minimum width constant. 

I initially wrote it using the blueprint table, but when profiling, found it took about 8 ms to update, which is far to slow if we want to update it when the user moves regions. An HTML table was far more lightweight in the end

duidae:
got that

PR #195, opened 2019-03-25
Angus/region dialog

veggiesaurus:
Adds basic region dialog support for rectangle and ellipse. Some items still need clarifying (the channel min and max sliders), so I've left them out.


duidae:
wonderful dialogs! 2 opinions about the feature:
1. can the values in the dialog also be updated as the region move/rotate, like the values in region list?
out ogv
2. line 29 in src/stores/RegionSetStore.ts can be deleted

Issue #148, opened 2019-02-26
independent clip range per Stokes


arusstaylor:
need to able to set the clip range independently for each Stokes.

veggiesaurus:
In terms of implementation, this requires that each Stokes sub-cube has its own selected percentile  (either a valid percentile or "custom"), and its own clip values if the selected percentile is "custom". 

The idea here being that the user can change the custom clip levels to an appropriate range for each Stokes cube independently, rather than being forced to use (a) the same percentile for each cube or (b) the same absolute clip levels for each cube. 

This is a relatively minor change, but will require some testing. @kswang1029 not sure which release to aim for. 1.2?

arusstaylor:
Is it at all possible to get this into 1.1?  There will be lots of Stokes I viewing from April to June on the MeerKAT side.

veggiesaurus:
implemented in #202

PR #202, opened 2019-03-27
Christy/react editing

ChristyReynolds:
Issue 148: setting persistent scaling for Stokes channels


src/components/RenderConfig/RenderConfigComponent.tsx
```diff
@@ -36,8 +36,8 @@ export class RenderConfigComponent extends React.Component<WidgetProps> {
     private cachedFrame: FrameStore;
 
     @observable width: number;
-    @observable height: number;
-
+    @observable height: number;
```
veggiesaurus: whitespace typo

src/components/RenderConfig/RenderConfigComponent.tsx
```diff
@@ -123,13 +123,15 @@ export class RenderConfigComponent extends React.Component<WidgetProps> {
 
     handleScaleMinChange = (val: number) => {
         if (isFinite(val)) {
-            this.props.appStore.activeFrame.renderConfig.setCustomScale(val, this.props.appStore.activeFrame.renderConfig.scaleMax);
+            const stokes = this.props.appStore.activeFrame.renderConfig.stokes;
```
veggiesaurus: We don't need this variable anymore, do we? Can't see where it's being used

src/components/RenderConfig/RenderConfigComponent.tsx
```diff
@@ -123,13 +123,15 @@ export class RenderConfigComponent extends React.Component<WidgetProps> {
...
+            this.props.appStore.activeFrame.renderConfig.setCustomScale(val, this.props.appStore.activeFrame.renderConfig.scaleMaxVal);
         }
     };
 
     handleScaleMaxChange = (val: number) => {
         if (isFinite(val)) {
-            this.props.appStore.activeFrame.renderConfig.setCustomScale(this.props.appStore.activeFrame.renderConfig.scaleMin, val);
+            const stokes = this.props.appStore.activeFrame.renderConfig.stokes;
```
veggiesaurus: Same as above

src/components/RenderConfig/RenderConfigComponent.tsx
```diff
@@ -306,7 +307,7 @@ export class RenderConfigComponent extends React.Component<WidgetProps> {
             percentileSelectDiv = (
                 <div className="percentile-select">
                     <FormGroup label="Clip Percentile" inline={true}>
-                        <HTMLSelect options={percentileRankOptions} value={frame.renderConfig.selectedPercentile} onChange={this.handlePercentileRankSelectChanged}/>
+                        <HTMLSelect options={percentileRankOptions} value={frame.renderConfig.selectedPercentile[stokes]} onChange={this.handlePercentileRankSelectChanged}/>
```
veggiesaurus: suggestion: we should make a @computed value `selectedPercentileVal` similar to how you've handled the `scaleMinVal` computed values

veggiesaurus:
@kswang1029 can you test this using mix 'n match? To verify it solves #148

kswang1029:
@ChristyReynolds Nice work! 
One issue:
with the default 99.9% clip, the x-range of the plot changes with different Stokes (which is good). However, if the x-range of a given Stokes is changed by users via mouse interactions, then the x-range is fixed to all Stokes. I encountered a case that I zoom in the Stokes=0, then switch to Stokes=1, and the plot is empty. Since it is empty there is no way to use mouse db-click to reset the range. I will need to use the reset range button in the settings menu to cure this. I suggest each Stokes should have its x-range definition. Sounds reasonable? @ChristyReynolds @veggiesaurus

veggiesaurus:
@kswang1029 the same issue occurs if you're zoomed in and change to another channel with a different histogram range. Perhaps we should just reset the histogram range whenever the histogram itself changes? (so, switching channel, switching channel <-> cube or switching Stokes)

kswang1029:
Hmm, right. Let's reset the range when switching channel/cube or Stokes.

veggiesaurus:
> Hmm, right. Let's reset the range when switching channel/cube or Stokes.

@kswang1029 Can you add a new issue for this? It's not directly dependent on this PR, so I'll merge this in and we can look into the issue separately.

kswang1029:
sure. Other than that I think this PR works well.

PR #244, opened 2019-04-09
Jason/fix filelist header

duidae:
trial approach to fix the sortable header th with scrollable tbody in table

out ogv


veggiesaurus:
Very nice! I'd forgotten about `position: sticky` :+1: 

While you're fixing the list, can you try align the sort indicators more clearly. They need to move up and to the right a bit

duidae:
okok

duidae:
@veggiesaurus icon adjustment added, please take a look when available, thanks!

duidae:
closes #245

PR #272, opened 2019-04-17
Jason/fix 73

### LINKED ISSUE #73: [imageViewer, AST] incorrect native coordinate when reopening an image
frontend version: f6daeb9, tested with MBP

to reproduce this issue:
1) open J2000_test.hdf5
2) change the coordinate system from "native" to "galactic"
3) reopen J2000_test.hdf5 (not append image)
4) now the coordinate system of the grid line is still "galactic" which may be okay (ideally maybe we should reset it to "native")
4) change the coordinate system from "galactic" to "native" (should be FK5) and we see
screen shot 2018-10-25 at 12 51 56
The system seems still galactic but its label format changes to FK5.

duidae:
we use implicit system(parameter system="") in overlay store in order to provide native option for users,  however current logic will not append "system" parameter when its value is an empty string, then the AST just use previous coordinate system to render. The solution is to set explicit system (saved in defaultSystem) to AST.

closes #73

### DIFF
--- a/src/components/ImageView/Overlay/OverlayComponent.tsx
+++ b/src/components/ImageView/Overlay/OverlayComponent.tsx
@@ -2,8 +2,8 @@ import * as React from "react";
 import * as AST from "ast_wrapper";
 import * as _ from "lodash";
 import {observer} from "mobx-react";
-import {ASTSettingsString, FrameStore, OverlayStore} from "stores";
-import {CursorInfo, Point2D} from "models";
+import {FrameStore, OverlayStore} from "stores";
+import {CursorInfo} from "models";
 import "./OverlayComponent.css";
 
 export class OverlayComponentProps {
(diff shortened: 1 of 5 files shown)


veggiesaurus:
@confluence, since you wrote most of this stuff, can you give it a check?

src/components/ImageView/Overlay/OverlayComponent.tsx
```diff
@@ -72,10 +72,6 @@ export class OverlayComponent extends React.Component<OverlayComponentProps> {
 
     render() {
         const styleString = this.props.overlaySettings.styleString;
-        const frameView = this.props.frame.requiredFrameView;
```
veggiesaurus: I think those are in there to ensure that changing the frame view, padding or width/height triggers a re-render

duidae: oh! sorry about that, I'll restore those codes

Issue #126, opened 2018-12-19
[Image viewer]Nan value to be transparent or be able to change alpha or opacity


zarda:
Suggestion about: do we need set the "nan" value on canvas to be transparent or able to change its alpha or opacity? Changing the opacity of image can help to compare stacked images.

veggiesaurus:
That might be a nice approach. I think we need to be able to change the color of the NaN values, to make sure that they don't clash with the color map, so being able to change the alpha as well would make sense.

zarda:
Several options: 
1. the color of Nan and opacity can change up to user, 
2. the Nan is transparent and user can change background color,
3. the Nan is transparent and the background color changes with time slowly, 
4. auto-set the color of Nan to be the color not used in color map, 
5. let Nan color change with time slowly,

veggiesaurus:
(1) is what I'd intended, and I think will be the cleanest in terms of UI.

 (4) would be nice, but tricky to calculate for each color map, and users may get confused. Having said that, auto-setting the region color based on color map would be pretty neat too.

veggiesaurus:
@duidae can you add an entry to "default render config" to allow the user to choose a NaN color. Make sure you render the `SketchPicker` with ` disableAlpha={false}` so the user can change the opacity.

Not sure what the default should be. Perhaps transparent? @kswang1029?

duidae:
okok

kswang1029:
I would suggest a grey color so that it is still clear with the default inferno colormap and it fits the entire GUI theme better. Would that work? @veggiesaurus @duidae

kswang1029:
@veggiesaurus @duidae sorry I think a bit more and realized that a grey color is not that obvious  with the "inferno" colormap. Perhaps it is better to still use blue as default. 
![M100_12m_CO_central image mom0 fits-image-2019-9-5-11-28-24](https://user-images.githubusercontent.com/20819712/64309841-797ea180-cfd0-11e9-97d9-4e685ea23689.png)

PR #295, opened 2019-04-22
Patch for issue #126

zarda:
Set opacity of uNaNColor and NaNColor to be 0 so that a non-finite pixel is transparent.


kswang1029:
This needs to be user-configurable. Where is the relevant GUI? Having default as 0 is not a good idea and inconsistent with common UX.

PR #307, opened 2019-04-24
Change number to scientific notation

### LINKED ISSUE #242: improvement of statistics widget
Suggestions from Juergen
- (i) I wonder if the image statistics could give the number of masked pixels, too.  
- (i) stats should have scientific notation, I would argue for 4 significant digits, e.g. 1.3244 x 1e2 Jy/beam 
- (i) stats need units 
- (i) the stats should also show the beam size 
- (b) the numerical value of rms differs. going for only a single pixel, I see the same values for all values (except for sigma which is 0 and sumsq which are both explained), but the rms is numerically different. probably a machine precision issue, will be solved by going for the significant digits as mentioned above

duidae:
partial fix of #242

### DIFF
--- a/src/components/Stats/StatsComponent.tsx
+++ b/src/components/Stats/StatsComponent.tsx
@@ -143,16 +143,12 @@ export class StatsComponent extends React.Component<WidgetProps> {
             StatsComponent.STATS_NAME_MAP.forEach((name, type) => {
                 const index = this.statsData.statistics.findIndex(s => s.statsType === type);
                 if (index >= 0) {
-                    let value;
-                    if (typeof this.statsData.statistics[index].value === 'number' && type !== CARTA.StatsType.NumPixels) {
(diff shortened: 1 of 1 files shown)


src/components/Stats/StatsComponent.tsx
```diff
@@ -144,10 +144,11 @@ export class StatsComponent extends React.Component<WidgetProps> {
                 const index = this.statsData.statistics.findIndex(s => s.statsType === type);
                 if (index >= 0) {
                     const value = this.statsData.statistics[index].value;
+                    const displayValue = (typeof value === "number" && type !== CARTA.StatsType.NumPixels) ? value.toExponential(4) : value;
```
veggiesaurus: should use `isFinite(value)` instead of checking type.

veggiesaurus:
note: this doesn't close #242 entirely, since we still need to add units. I've edited the description to prevent auto-closing.

PR #312, opened 2019-04-26
add unit to stats

### LINKED ISSUE #242: improvement of statistics widget

duidae:
closes #242

### DIFF
--- a/src/components/Stats/StatsComponent.tsx
+++ b/src/components/Stats/StatsComponent.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import {observer} from "mobx-react";
 import {autorun, computed, observable} from "mobx";
-import {ControlGroup, FormGroup, HTMLSelect, HTMLTable, IOptionProps, NonIdealState} from "@blueprintjs/core";
+import {FormGroup, HTMLSelect, HTMLTable, IOptionProps, NonIdealState} from "@blueprintjs/core";
 import ReactResizeDetector from "react-resize-detector";
 import {CARTA} from "carta-protobuf";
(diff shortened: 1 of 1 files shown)


src/components/Stats/StatsComponent.tsx
```diff
@@ -156,12 +155,24 @@ export class StatsComponent extends React.Component<WidgetProps> {
...
+                        } else if (type === CARTA.StatsType.SumSq) {
+                            unitString = `(${unit})^2`;
+                        } else {
+                            unitString = unit;
+                        }
+                    }
+
                     const value = this.statsData.statistics[index].value;
-                    const displayValue = (isFinite(value) && type !== CARTA.StatsType.NumPixels) ? value.toExponential(4) : value;
+                    const valueString = isFinite(value) ? `${(type === CARTA.StatsType.NumPixels) ? value : value.toExponential(4)} ${unitString}` : "";
```
veggiesaurus: this line change now prevents `NaN` values from being shown as `NaN`. It should still display `NaN` or `INF` for non-finite values

PR #327, opened 2019-05-13
Migrate to react-hotkey

### LINKED ISSUE #328: Migrate to react-hotkey

duidae:
closes #328

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
         "react-color": "^2.17.0",
         "react-dom": "^16.8.6",
         "react-flexbox-grid": "^2.1.2",
+        "react-hotkeys": "^1.1.4",
         "react-konva": "^16.8.6",
         "react-resize-detector": "^4.0.5",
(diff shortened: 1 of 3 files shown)


veggiesaurus:
Few issues with this:
- Hotkeys don't seem to respond nicely. For example, load a file and try use the freeze cursor or create region hotkey. They don't work until you click on the UI again
- The `f` and `c` hotkeys still trigger while typing in text boxes
- `Shift + ?` no longer triggers the hotkey popup
- (longer term) we should re-do the hotkey rendering itself so that we can support more complex hotkey indicators, as well as alternative hotkey sequences (eg. `del` _or_ `backspace`)

Not sure how difficult these issues are to fix. If it's just a matter of doing this for the CRA update, perhaps we should just use the `App` component wrapper idea, and work on this issue as a low priority

duidae:
okok, I will try the wrapper instead

Issue #212, opened 2019-03-28
Correcting text highlighting behavior


ChristyReynolds:
Currently, when you double-click around the UI, sometimes it highlights text in odd ways, which is annoying. For example, clicking on the image view too quickly means you end up highlighting the cursor info at the top. The way around this is to specify `user-select: none` in the CSS for components that should have this disabled. We should disable this at the top level and then explicitly enable it using `user-select: text` on only those elements that should be highlightable (cursor text, things like that)

PR #331, opened 2019-05-22
Christy/issue212

creynolds75:
Added user-select settings to SCSS files


veggiesaurus:
@kswang1029 can you test this and see if it solves #212 without interfering with normal functionality?

kswang1029:
Overall I don't see something odd. 
Just a suggestion: it might be still good to enable highlight and copy-paste of the cursor info bar itself on the image viewer. I found it is handy to be able to copy coordinates or pixel value, frequency or velocity.

veggiesaurus:
> Overall I don't see something odd.
> Just a suggestion: it might be still good to enable highlight and copy-paste of the cursor info bar itself on the image viewer. I found it is handy to be able to copy coordinates or pixel value, frequency or velocity.

Agreed, @ChristyReynolds please add this. 

Also, the dialogs are still behaving incorrectly: 
image

src/components/Stats/#StatsComponent.scss#
```diff
@@ -0,0 +1,102 @@
+\1;95;0c@import "../../../node_modules/@blueprintjs/core/lib/scss/variables.scss";
```
veggiesaurus: not sure what this file is for

creynolds75:
Thanks Angus!

PR #336, opened 2019-05-28
Jason/ui customize

duidae:
finish the scaling part


kswang1029:
@veggiesaurus I feel it is clearer if we apply "default settings" on next new session. This preferences dialog should be visited by users less often.

If we sync default setting with the current setting, it has to be one way, e.g, changing default setting in preference dialog *changes* the current setting, but changing current setting *does not* change settings in preference dialog.

veggiesaurus:
@duidae sorry to be picky, but now the tab heading needs to be made larger: 

image


> @veggiesaurus I feel it is clearer if we apply "default settings" on next new session. This preferences dialog should be visited by users less often.

@kswang1029 I'm very glad you said that... makes @duidae's task much easier. 

> 
> If we sync default setting with the current setting, it has to be one way, e.g, changing default setting in preference dialog _changes_ the current setting, but changing current setting _does not_ change settings in preference dialog.

One-way changing would be annoying to implement. Let's leave this out for now? If users give strong feedback on this, we can adjust.

veggiesaurus:
@duidae I see you changed the width, thanks :+1: 

Can you change "Preference" to "Preferences" in the heading as well? Woder if the cog icon wouldn't be a better fit?
image

kswang1029:
@veggiesaurus @duidae  Let's aim for applying settings in the preference dialog to *new* sessions. : )

yes if there are lots of picky users (I really hope not....), we could start to think about how to make those settings sync'd properly.

kswang1029:
@duidae nice job BTW. I briefly tested a bit and it works nicely.

kswang1029:
OK, actually the settings will be applied to any new images (either open as new or append). I think it is okay.

arusstaylor:
I agree that applying default settings to next session is cleaner.  If one is using the current setting to set a default, then by definition the current setting are in place.  There is no need to apply them.

Russ

PR #340, opened 2019-05-30
preference: colormap + percentile + AST color

duidae:


duidae:
color

@kswang1029  about the color selector for AST overlay in Preference, which one will you prefer?
1. original color selector
2. region line selector

kswang1029:
@duidae this is a good question and I guess different users would have different opinions.
I would suggest we stick with the original color selector and listen to users if there are many requests for more colors.

@arusstaylor do you have any preference?

kswang1029:
"WCS overlay" sounds right. 

I would think: 
It is better not to apply any "default" settings immediately. All the "default" settings apply to new images or sessions.

arusstaylor:
agreed.  "WCS overlay".

also agree new "default" setting apply with next image load, refresh or next session.

src/stores/PreferenceStore.ts
```diff
@@ -1,42 +1,73 @@
...
+    };
+
+    @action setASTColor = (newColor: number) => {
+        this.json.astColor = (newColor >= 0 && newColor < AST.colors.length) ? newColor : this.defaultSettings.astColor;
+        localStorage.setItem(PreferenceStore.CARTA_PREFERENCE, JSON.stringify(this.json));
+    };
 
     constructor() {
-        const preference = localStorage.getItem("CARTA_preference");
+        this.defaultSettings = JSON.parse(PreferenceStore.DEFAULT_SETTINGS_JSON);
```
veggiesaurus: why not just construct a default settings object directly, rather than using a default settings JSON and parsing it?

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -1,13 +1,31 @@
...
+
 @observer
 export class PreferenceDialogComponent extends React.Component<{ appStore: AppStore }> {
+    @observable selectedTab: TabId = "renderConfig";
+    @observable scaling = this.props.appStore.preferenceStore.getScaling();
+    @observable colormap = this.props.appStore.preferenceStore.getColormap();
+    @observable percentile = this.props.appStore.preferenceStore.getPercentile().toString();
+    @observable astColor = this.props.appStore.preferenceStore.getASTColor();
+
+    renderPercentilSelectItem = (percentile: string, {handleClick, modifiers, query}) => {
```
veggiesaurus: minor typo

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -18,8 +36,37 @@ export class PreferenceDialogComponent extends React.Component<{ appStore: AppSt
...
+                <FormGroup inline={true} label="Color map">
+                    <ColormapComponent
+                        selectedItem={this.colormap}
+                        onItemSelect={(selected) => { preferenceStore.setColormap(selected); this.colormap = selected; }}
+                    />
+                </FormGroup>
+                <FormGroup inline={true} label="Percentile ranks">
+                    <PercentilSelect
+                        activeItem={this.percentile}
+                        onItemSelect={(selected) => { preferenceStore.setPercentile(selected); this.percentile = selected; }}
```
veggiesaurus: is this done to avoid re-reading the value from local storage on each re-render?

src/stores/PreferenceStore.ts
```diff
@@ -1,42 +1,73 @@
...
+    colormap: "inferno",
+    percentile: 99.9,
+    astColor: 4
+};
 
+export class PreferenceStore {
     // user configurable settings
+    validateScaling(scaling: string) {
+        const value = Number(scaling);
+        return scaling && value !== NaN && RenderConfigStore.SCALING_TYPES.has(value) ? value : null;
```
veggiesaurus: `value !== NaN` is always true, even for `NaN`, since `NaN === NaN` is false. Rather use `isfinite(value)`

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
duidae: if I replace this.percentile with this.props.appStore.preferenceStore.getPercentile().toString() to reduce that observable variable, the content of the selected item will not be updated to UI (it keeps the initially loaded value), so the solution I found is to make it observable and update it in onItemSelect/onChange, is there a better way for this?

PR #342, opened 2019-06-17
Jason/preference AST grid + AST labels + region

duidae:
preference: add AST grid + AST labels + region


kswang1029:
@veggiesaurus I think what you suggested are good ideas and should improve UX. I have modified a bit the requirements (https://docs.google.com/document/d/1VKitCmzXBt7cBQyMU8sprWD4JOnNGgMTS66JWLlovvo/edit#).
Overall, there are three types of settings which are effective at different stages:
1) effective on new session
2) effective on new image (later this should be new WCS group)
3) effective on new region
Hope this won't complicate things and please let me know if there are hidden problems.

duidae:
what about "Default cursor behaviour"? should it be on new image?

kswang1029:
probably yes, but what does that mean? @veggiesaurus 
@duidae do you know what it means?

duidae:
there should be an initially frozen cursor on the center of image when an image is newly loaded, from my understanding

kswang1029:
what happens if the initial cursor is not frozen?

duidae:
if set to freeze
    there will be a frozen cursor in the center of image when image is newly loaded (press F can free the cursor)
else
    the cursor behavior is like what we have now

kswang1029:
I see. Then, yes, it is better to be effective on new image.

PR #346, opened 2019-06-27
Angus/tiled rendering

veggiesaurus:
Note: Companion PR of https://github.com/CARTAvis/carta-backend/pull/250. This should only be merged once both PRs are approved.

Tiled rendering is used when not animating channels / stokes. See [here](https://docs.google.com/document/d/1ryrqjW6f9VFtLpSHB6UiIHNxXb18XCzJX90HzZQY5gc/edit) for details.

There are some known issues, which can be fixed separately after the merge. Those relating to the frontend are:
- Tiles are sometimes not copied to the GPU correctly, resulting in empty squares
- When switching channels, the image is rendered once the first tile arrives, rather than waiting for all required tiles, resulting in flashing between images [feature/bug discussion: #348 ]
- GPU memory leak when unpinning the image view widget.
- Approximate spatial profiles don't work properly. These will be fixed with the profile optimisation feature. 

Additional issues are noted on the backend PR. I'm sure @kswang1029 will find loads more once this is merged as well.


src/models/TileCoordinate.ts
```diff
@@ -0,0 +1,42 @@
...
+    public static Encode(x: number, y: number, layer: number): number {
+        const layerWidth = 1 << layer;
+        if (x < 0 || y < 0 || layer < 0 || layer > 12 || x >= layerWidth || y >= layerWidth) {
+            return -1;
+        }
+
+        return ((layer << 24) | (y << 12) | x);
+    }
+
+    public static Decode(encodedCoordinate: number): TileCoordinate {
```
duidae: is it possible to use some constant variables to replace the numbers for readability? (guess this is for decoding the protobuf message)

arusstaylor:
Traditional render view should be applied for all the animation widget buttons, e.g. next, last, prev, first as well as selecting a channel using the z profile cursor. This is the same use case as play.  The user wants to look at a different channel with the current zoom and pan settings. It is distracting and difficult to perceive changes when the re-rendering is done with tiles.

veggiesaurus:
> Traditional render view should be applied for all the animation widget buttons, e.g. next, last, prev, first as well as selecting a channel using the z profile cursor. This is the same use case as play. The user wants to look at a different channel with the current zoom and pan settings. It is distracting and difficult to perceive changes when the re-rendering is done with tiles.

@arusstaylor The traditional render view is only useful when the user is animating with a predictable interval, which is quite different to when the user selects a specific channel as a once-off operation. I think you're conflating the different rendering and data delivery methods with how this is actually presented to the user. Note that I put the issue you mentioned above as one of the known issues to be fixed:

>When switching channels, the image is rendered once the first tile arrives, rather than waiting for all required tiles, resulting in flashing between images [feature/bug discussion to ensue]

This is not an unsolvable issue, but can be tackled separately, and should be discussed in more detail in a separate issue (I'll file one shortly). There are advantages to rendering the image progressively under some circumstances, so it's important to determine _when_ this should be done. I agree that when switching channels manually, it can be distracting, so this may need to be a user-configurable option.

src/models/TileCoordinate.ts
veggiesaurus: could do, although I'm not sure if it will help with readability. To be consistent, we'd need `const` variables for 12, 24, 19, 7, 4096 and 128, which would make the code quite a mess. 

This is for encoding and decoding a 3-component value (x, y, layer) into a single 32-bit number, in order to (a) reduce data transfer costs and (b) improve hashing behaviour. This is described in the ICD, but perhaps some comments cuold be added to the code

duidae: hmm comments sound good, Angus thanks!

Issue #345, opened 2019-06-26
Preference dialogue - region settings


duidae:
region: Region creation mode (“center-to-corner” vs “corner-to-corner”)
region: Default region color (default: “cyan”)
region: Default region line width
region: Default region dash thickness
region: Default region shape

Issue #344, opened 2019-06-26
Preference dialogue - render config & WCS overlay


duidae:
Color map in render config widget (default: “inferno”)
Color map scaling (default: “Linear”)
Default clip level (default: “99.9%”)
AST Global -> Color (default: “? blue”)
AST Grid -> Visible (default: “yes”)
AST Labels -> Visible (default “yes”)
Default WCS type: Automatic (follows convention based on coordinate system)
Default WCS type: Degrees
Default WCS type: Sexigesimal

PR #347, opened 2019-06-28
Jason/preference: global settings + wcs type

### LINKED ISSUE #343: Preference dialogue - global settings
global: Color theme (“View” -> “Interface” ; default: “Light”)
global: Layout (default: “Cube view”, see the above section “Predefined layouts”)
global: Auto-launch file browser when CARTA is initialized (default: “enabled”)
global: Default cursor behaviour? (frozen, not frozen)
global: Default image zoom level (“zoom to fit” vs “zoom to 1.0x”)

duidae:
closes #343 #344 #345

the behaviors of applying new settings in new session/new image/new region are adjusted according to KS's document https://docs.google.com/document/d/1VKitCmzXBt7cBQyMU8sprWD4JOnNGgMTS66JWLlovvo/edit#

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,7 +8,7 @@ import {Alert, Classes, Colors, Dialog, Hotkey, Hotkeys, HotkeysTarget} from "@b
 import {exportImage, FloatingWidgetManagerComponent, RootMenuComponent} from "./components";
 import {AppToaster} from "./components/Shared";
 import {AboutDialogComponent, ApiKeyDialogComponent, FileBrowserDialogComponent, OverlaySettingsDialogComponent, RegionDialogComponent, URLConnectDialogComponent, PreferenceDialogComponent} from "./components/Dialogs";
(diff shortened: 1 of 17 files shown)


kswang1029:
> Light/dark theme should be effective immediately. There's no reason why not to do this, as far as I'm concerned
Yes, making this effective immediately is more intuitive. 

> I'm not sure I like the slide toggle with text (eg. freeze, unfreeze). I think it's inconsistent with the rest of our UI. Rather use a simple slide toggle for "Auto-lauch File Browser", and either use a simple toggle for "Cursor Position", or use a radio set with "Frozen" and "Unfrozen" (perhaps we should change to "Fixed" and "Tracking"?)
A radio set might be better so that users can directly read what options are. 
Agree that we should use "Fixed" and "Tracking" for better readability. 

> SInce some of these are default values, I think we need to change the working a bit:
"Layout" -> "Default Layout"
"Cursor Position" -> "Initial Cursor Position"
"Image Zoom Level" -> "Initial Zoom Level"
Agree.

PR #353, opened 2019-07-02
fix half-pixel rendering artefacts at tile boundaries

### LINKED ISSUE #350: Rendering artefacts at tile boundaries
There appear to be half-pixel artefacts at tile boundaries. This appears to be related to the vertex shader code.

Screen Shot 2019-07-02 at 4 19 26 PM

veggiesaurus:
Closes #350 

The previous half-pixel artefacts were due to an attempt to solve an existing tiling artefact. However, the fix was applied at the vertex shader level, and ended up cropping off the first and last half-pixels of every tile. The tile has now been moved to the pixel shader level, and fixes both artefacts.

image

### DIFF
--- a/src/components/ImageView/RasterView/GLSL/pixel_shader_float_rgb.glsl
+++ b/src/components/ImageView/RasterView/GLSL/pixel_shader_float_rgb.glsl
@@ -10,6 +10,7 @@ precision highp float;
 #define CUSTOM 7
 
 varying vec2 vUV;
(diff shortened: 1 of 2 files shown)


veggiesaurus:
@kswang1029 please review to ensure artefacts are gone (don't worry about the code changes)

kswang1029:
scanning and staring at few test images, I think the issue is gone. : )

PR #357, opened 2019-07-03
Qi/stokes plot

panchyo0:
1. remove Q +U and Q vs U text from stokes dropdown menu
2. merge with dev branch 
3. reformate


kswang1029:
@panchyo0 @veggiesaurus  
Would it be better to reword "PI" to "Pol. Intensity"(?) and "PA" to "Pol. Angle"(?)?

For "PI" and "PA", y-label of the spectral profiler plot should be changed accordingly.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -62,7 +66,51 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
...
+                vals[i] = Math.sqrt(Math.pow(datas[0].vals[i], 2) + Math.pow(datas[1].vals[i], 2));
+            }
+        }
+        return vals;
+    }
+
+    // Qi, return PI, PA data for ploting
+    private calculate(coordinate: string, statsType: CARTA.StatsType): CARTA.ISpectralProfile {
+        let coordinateData: CARTA.ISpectralProfile;
+        if (coordinate === "PIz" || coordinate === "PAz") {
```
veggiesaurus: I think we should define some of these strings as constants. Perhaps in an enum? I'd rather see something like
````
if (coordinate === StokesCoordinate.PolarizedIntensity || coordiante === StokesCoordinate.PolarizationAngle)
````

File: src/components/SpectralProfiler/SpectralProfilerComponent.tsx [hunk 2]
```diff
@@ -74,7 +122,8 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
         if (frame.regionSet) {
             const region = frame.regionSet.regions.find(r => r.regionId === regionId);
             if (region) {
-                coordinateData = this.profileStore.getProfile(this.widgetStore.coordinate, region.isClosedRegion ? this.widgetStore.statsType : CARTA.StatsType.Sum);
+                // Qi, calculate PI or PA if coordinate is PI or PA
+                coordinateData = this.calculate(this.widgetStore.coordinate, region.isClosedRegion ? this.widgetStore.statsType : CARTA.StatsType.Sum);
```
veggiesaurus: could we call this function something a little less generic than `calculate`? Perhaps `calculateCompositeProfile`?

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -101,7 +150,7 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
             let yCount = 0;
 
             // values are needed to be sorted in incremental order for binary search
-            let values: Array<{ x: number, y: number }> = [];
+            let values: Array<{x: number, y: number}> = [];
```
veggiesaurus: it seems we've previously been inconsistent with the spacing. I'm open to it changing, as long as we're consistent.

src/components/SpectralProfiler/SpectralProfilerToolbarComponent/SpectralProfilerToolbarComponent.tsx
```diff
@@ -49,12 +49,15 @@ export class SpectralProfilerToolbarComponent extends React.Component<{ widgetSt
...
         }
 
+        // Qi, add new stoket plot type PA, PI, Q+U (multiple lines), Q vs U (Scatter plot)
         const profileCoordinateOptions = [
             {value: "z", label: "Current"},
             {value: "Iz", label: "I"},
             {value: "Qz", label: "Q"},
             {value: "Uz", label: "U"},
-            {value: "Vz", label: "V"}
+            {value: "Vz", label: "V"},
```
veggiesaurus: see above, about using enum values. That would make things more maintainable.

src/stores/widgets/SpectralProfileWidgetStore.ts
```diff
@@ -193,7 +207,10 @@ export class SpectralProfileWidgetStore extends RegionWidgetStore {
                 let spectralConfig = regionRequirements.spectralProfiles.find(profiles => profiles.coordinate === coordinate);
                 if (!spectralConfig) {
                     // create new spectral config
-                    regionRequirements.spectralProfiles.push({coordinate, statsTypes: [statsType]});
+                    // Qi, creaet new spectral config according coordinate
+                    this.requiredCoordinate(coordinate).forEach(data => {
```
veggiesaurus: This logic is not quite right: At the moment, if the coordinate is `PIz`, then `spectralConfig` will always be false. We then insert a new entry for both `Qz` and `Uz`. This means that we can end up with duplicate entries for `Qz` and `Uz`.

Instead, you should put the `SpectralProfileWidgetStore.requiredCoordinates(coordinate).foreach` around the entire requirements `find` section. That way, you'll check for the exisitance of `Qz` and `Uz` individually.

veggiesaurus:
@kswang1029 @panchyo0 
> Would it be better to reword "PI" to "Pol. Intensity"(?) and "PA" to "Pol. Angle"(?)?

I'd suggested writing it out in full, which @panchyo0 rightly pointed out is probably too long. "Pol. Angle" is probably a good compromise.

> For "PI" and "PA", y-label of the spectral profiler plot should be changed accordingly.

Since PI is sqrt(Q^2 + U^2), surely it has the same units? Agree that PA should change (to rad)

> When PI or PA is selected, "statistic" menu should be greyed out and we should use "mean" for calculations. Currently if we display PI or PA with other statistic (StdDev, etc) the displayed profile is changed accordingly which should not be the case.

This is a very important point, and makes me wonder if we're actually approaching this the wrong way. The _mean_ of the polarized intensity is not the same as the _polarized intensity_ of the mean (which is what we're currently calculating). @kswang1029 is this a problem? Do we need to check with the science team as to which is the more useful metric? If the mean of the polarized intensity is preferred, it will require changing our approach, unfortunately, and it becomes more of a backend issue.

kswang1029:
Could we add polarization percentage (PI / I * 100) as well?

From a polarization expert, it is more common to use degree for PA, rather than rad.

veggiesaurus:
> Could we add polarization percentage (PI / I * 100) as well?

Can we add that as a widget setting?  (in the side-panel). We should be able to look at Q, U, V or PI as a percentage of I. Perhaps we could call it "Normalize Stokes" or something like that?

> From a polarization expert, it is more common to use degree for PA, rather than rad.

Good idea.

arusstaylor:
I think the toggle from intensity to fractional polarisation should be more visible. It is not really a "setting".  Users will want to see how things change between the two quantities.    I would like to see a visible toggle on the tool itself that the user can use to switch the display from polarised intensity.  The toggle could be something like "fractional pol: on/off.    When it is on all displays use Q/I and U/I as inputs.  When it is off all displays use Q and U as inputs.

veggiesaurus:
image
@arusstaylor would placing it here be ok? It might make things a bit cramped on smaller screens.

arusstaylor:
Yes.

veggiesaurus:
@arusstaylor I've created a separate issue for this (#359), so that we can get the basic features in first, before adding the fractional polarisation (which requires a bit of change, because we need to include Stokes I in the requirements as well).

@panchyo0, I've assigned you to #359, we can get this PR done first before you make the changes outlined there.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -33,7 +33,7 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
         };
     }
 
-    // Qi, required type for PI, PA, Q+U and Q Vs U
+    // required type for PI, PA, Q+U and Q Vs U
     private QUCoordinates = ["Qz", "Uz"];
```
veggiesaurus: these can also make use of the enum

src/stores/widgets/SpectralProfileWidgetStore.ts
```diff
@@ -40,16 +60,16 @@ export class SpectralProfileWidgetStore extends RegionWidgetStore {
         }
     }
 
-    // Qi, add new stoke valid type PI, PA, Qz+Uz, QzVsUz
-    private static ValidCoordinates = ["z", "Iz", "Qz", "Uz", "Vz", "PIz", "PAz"];
+    // add new stoke valid type PI, PA, Qz+Uz, QzVsUz
+    private static ValidCoordinates = ["z", "Iz", "Qz", "Uz", "Vz", "PIz", "PAz", "Qz+Uz"];
```
veggiesaurus: These could also use the enum (and you can type the array as `: Array<StokesCoordinate>` for consistency)

src/stores/widgets/SpectralProfileWidgetStore.ts
```diff
@@ -40,16 +60,16 @@ export class SpectralProfileWidgetStore extends RegionWidgetStore {
...
     private static ValidMultiDataCoordinates = ["PIz", "PAz"];
 
-    // Qi, return regionRequirements spectralProfiles coordinate array
-    private static requiredCoordinate(coordinate: string): Array<string> {
+    // return regionRequirements spectralProfiles coordinate array
+    private static requiredCoordinate(coordinate: string): Array<StokesCoordinate> {
```
veggiesaurus: could `coordinate` also be of type `StokesCoordinate`?

arusstaylor:
Yes indeed.  We need this.  The specification is to be able to toggle switch between plotted quantities as absolute intensity or percent polarisation.  This is an important feature and should be in v1.2.

Russ

kswang1029:
@arusstaylor Could you look at this http://carta.asiaa.sinica.edu.tw/frontend/5edccf7/?socketUrl=wss://carta.asiaa.sinica.edu.tw/socketdev which now has fractional switch?

arusstaylor:
This toggle solution looks good and provides the desired function. 
I wonder for context sensitive layouts and tools whether we not show the fractal button or the Stokes selection drop down if there is no Stokes axes in the data cube.

veggiesaurus:
closing PR, as it will be replaced with another one by @panchyo0

PR #362, opened 2019-07-08
Jason/fix region inconsistent

### LINKED ISSUE #361: inconsistent behavior of region creation/re-scaling
when in corner to corner mode, re-scaling the created region should also be corner to corner, vice versa in center mode

duidae:
fix #361


kswang1029:
@duidae Now working well! Thanks for the fix in such a short time!

Issue #239, opened 2019-04-08
Support for point regions


veggiesaurus:
Region shapes need to be extended to point regions (similar to cursor)
- [x] Create point regions
- [x] Select point regions
- [x] Move point regions

edit: splitting out polygon regions to a separate issue

kswang1029:
Creating/modifying a polygon needs to have a real-time check of validity (no crossing line segments are allowed).

veggiesaurus:
implemented by @zarda :raised_hands:

PR #365, opened 2019-07-09
#239 support point region

zarda:
Add support of point region type to parts of lists on #239 

- Toolbar to add region point
- Move on canvas
- Show on region list
- A box to show the selected point region
- Appearance for editing


kswang1029:
Can we make the region highlighter to be more readable? Making it bigger perhaps?

Some style changes for the custom dialog:
Screen Shot 2019-07-09 at 18 28 25
It should be just "color" and the text format of wcs coordinate should be consistent with the other region types (cf rectangle).

zarda:
Does the region point mean a coordinate which will map to the raw point intensity? It is a matter of if the front will acquire the intensity data from backend. Current region point "is closed" and performed as other regions, it it?

veggiesaurus:
> Does the region point mean a coordinate which will map to the raw point intensity? It is a matter of if the front will acquire the intensity data from backend. Current region point "is closed" and performed as other regions, is it?

A point region is not closed in the sense that it can't be reduced to a single pixel using a specific stats type, since it already corresponds to only a single pixel. It should behave like any other region, in the sense that it can be used in the spectral profile widget to view the spectral profile, but it should not have the stats type enabled.

kswang1029:
Could we use a simpler point selection indicator? Currently it looks quite blocky... 
Screen Shot 2019-07-11 at 3 30 55 PM

I see there is a circle inside a square. Could we just use the square in dashed/dotted line?
Screen Shot 2019-07-11 at 3 29 28 PM

arusstaylor:
How about just a cross hair with the centre clear?


         |
    —   —
         |

Russ

veggiesaurus:
agree we should change the symbol. However, it needs to be changed so that:
1) It's clear when the region is selected or deselected
2) It's clear that it's not the mouse cursor or region creation cursor

kswang1029:
@zarda after checking with the science team members, the consensus is that we will just use anchors to indicate the a region is selected. Please remove the border in dashed line (except point region).

veggiesaurus:
@zarda now that @pford has fixed point regions on the backend ([carta-backend/#279](https://github.com/CARTAvis/carta-backend/pull/279)), can you test your PR and confirm it's working?

zarda:
The patch is tested well. This PR is ready.

kswang1029:
@veggiesaurus sometimes I find it hard to click right on the point too. Is that because the "square" is rotated? Or the invisible clickable area is simply too small?

kswang1029:
Can we have the "statistic" to "mean" to be the same as cursor?

veggiesaurus:
> @veggiesaurus sometimes I find it hard to click right on the point too. Is that because the "square" is rotated (and why?)? Or the invisible clickable area is simply too small?

I think currently only the visible area is clickable.

kswang1029:
> I think currently only the visible area is clickable.

Then perhaps we use square not diamond shape. Although the area is the same but I believe square is easier to click on.

PR #372, opened 2019-07-12
improve rotation wording to P.A and add unit

### LINKED ISSUE #243: improvement of region list widget
Suggestions from Juergen
- (i) Maybe change the word ‘rotation’ to ‘PA’  everywhere 
- (b) looks like I can put a really large number in the ‘line width’ but there seems to be an upper limit on the thickness of the region itself. the number should also hit an upper max to make the behavior consistent. similar for the dash length  
- (i) the region list should also have units for the size (pix) and rotation/PA (deg)

acdo2002:
Close #243

### DIFF
--- a/src/components/Dialogs/RegionDialog/EllipticalRegionForm/EllipticalRegionForm.tsx
+++ b/src/components/Dialogs/RegionDialog/EllipticalRegionForm/EllipticalRegionForm.tsx
@@ -182,9 +182,9 @@ export class EllipticalRegionForm extends React.Component<{ region: RegionStore,
                             </td>
                         </tr>
                         <tr>
-                            <td>Rotation <span className={Classes.TEXT_MUTED}>(deg)</span></td>
+                            <td>P.A. <span className={Classes.TEXT_MUTED}>(deg)</span></td>
                             <td>
-                                <NumericInput {...commonProps} buttonPosition="none" placeholder="Rotation" value={region.rotation} onBlur={this.handleRotationChange} onKeyDown={this.handleRotationChange}/>
+                                <NumericInput {...commonProps} buttonPosition="none" placeholder="P.A." value={region.rotation} onBlur={this.handleRotationChange} onKeyDown={this.handleRotationChange}/>
                             </td>
                         </tr>
                         </tbody>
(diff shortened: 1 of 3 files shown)


kswang1029:
could you remind me which issue is related to this PR?

veggiesaurus:
minor change required: need consistent spacing between label and units (i.e. `P.A. (deg)`)

kswang1029:
do we want to use two lines for P.A. (deg)?
Screen Shot 2019-07-16 at 2 33 22 PM
Screen Shot 2019-07-16 at 2 33 13 PM

acdo2002:
Thanks to Kuo-Song's suggestion. 
The two-line PA problem happens when you zoomin(<90%) the browser.

In **this current commit**, I only add the space in html format.
[video using space](https://drive.google.com/open?id=1qd6MgG5elIecKq5_WpDV8W0Q_AtjBM2j)
Screen Shot 2019-07-16 at 15 28 21
Pros: from the screenshot, the header cell is slightly aligned to the table.
Cons: split two lines in PA.

I also try to use Non-breaking Space in html.
[video using NbS](https://drive.google.com/file/d/15pjYyRuRi2kPy30g-TYmTO4PXlCi9uVu/view?usp=sharing)
Screen Shot 2019-07-16 at 15 27 00
Pros: it will not split into two lines in P.A.
Cons: But the header cell is slightly misaligned to the rest of the table.

Any comment is welcome!!!

kswang1029:
would it work if you shrink the width of "Name" column a little bit?

veggiesaurus:
@acdo2002 please increase the rotation column width a bit (`ROTATION_COLUMN_DEFAULT_WIDTH` should do the trick). 

Note that this code (the original code, I mean, not your changes) needs some reworking, and using a CSS flex approach would probably simplify things considerably

acdo2002:
I decide to use the "space" between label and units rather than using "Non-breaking Space", because another header cell "Pixel Center" used "space".

I slightly adjust the value of ROTATION_COLUMN_DEFAULT_WIDTH and TYPE_COLUMN_DEFAULT_WIDTH, in order to keep the table alignment consistent. It can keep the alignment until zoomin the browser~50%.

About a CSS flex approach, I may need time to study with it.

acdo2002:
I see, I modify the type column width to 80px. Keeping the central column width (120px) and size column width (160px), then it can better display the information of the image.

veggiesaurus:
@acdo2002 approved PR and merged in. Thanks for your contribution :+1:

PR #388, opened 2019-07-19
add event type log filtering to preference

duidae:


kswang1029:
add "check all", "uncheck all" perhaps?

veggiesaurus:
@duidae I've modified the layout a bit, I think this works a little better:

image

Other than that, it looks good :+1: We may want to hide it behind a debug flag sometime, and disable it when using production builds

kswang1029:
with my 13" mbp, some items are not aligned as expected.
Screen Shot 2019-07-19 at 21 20 15

PR #402, opened 2019-07-23
Jason/fix 384

### LINKED ISSUE #384: regression: toolbar drag sources are broken
The toolbar's drag sources seem to have been broken by a recent commit or package upgrade. This means that new widgets can no longer be docked, but must first be created as floating widgets, and then docked.

duidae:
close #384 

dragsource should be created after golden layout object is instantiated, originally created in componentDidUpdate() in ToolbarMenuComponent will not be effective in the very 1st load(golden layout object is instantiated in componentDidMount() in App.tsx, which is later than componentDidUpdate() of ToolbarMenuComponent according to react life cycle)

### DIFF
--- a/src/stores/WidgetsStore.ts
+++ b/src/stores/WidgetsStore.ts
@@ -77,23 +77,6 @@ export class WidgetsStore {
         });
     };
 
-    // create drag source for ToolbarMenuComponent
-    private static createDragSource = (appStore: AppStore, layout: GoldenLayout, widgetConfig: WidgetConfig, elementId: string) => {
-        const glConfig: GoldenLayout.ReactComponentConfig = {
-            type: "react-component",
-            component: widgetConfig.type,
-            title: widgetConfig.title,
-            id: widgetConfig.id,
-            isClosable: widgetConfig.isCloseable,
-            props: {appStore: appStore, id: widgetConfig.id, docked: true}
-        };
-
-        const widgetElement = document.getElementById(elementId);
-        if (widgetElement) {
-            layout.createDragSource(widgetElement, glConfig);
-        }
-    };
-
     constructor(appStore: AppStore) {
         this.appStore = appStore;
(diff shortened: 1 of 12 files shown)


src/stores/WidgetsStore.ts
```diff
@@ -140,7 +152,22 @@ export class WidgetsStore {
         }
     };
 
-    // region Golden Layout Widgets
+    // create drag source for ToolbarMenuComponent
```
veggiesaurus: can this be made `private static`?

duidae: cant be static because the need of appStore instance for props in glConfig. is there a better way for this?

duidae: ah how about I add a new parameter of appStore

duidae: modification added

duidae:
add check to image viewer to for 0 width/height prevention, should work as following:

out-1 ogv

veggiesaurus:
I'm not sure that these changes are right. You've solved #384 , but the changes to the other stuff seems unrelated. Can isolate the fixes to #384 in this branch and discuss the other changes in another PR?

I'd like to keep rendering the `RasterViewComponent` at all times, even if it has zero width and height, since this prevents the WebGL context from being lost, and thus the GPU cache from being cleared. 

I'm also not entirely sure about adjusting the zoom when resizing the view. If I zoom in to 1x, and then resize the view, it jumps back to my auto-scaled zoom level. This would only be considered correct if I was currently in "zoom to fit" zoom. I'd prefer us to respect the user's zoom level choice.

duidae:
reverted. I think we need to add error handling for 0 width/height to RasterViewComponent/OverlayComponent/CursorOverlayComponent/BeamProfileOverlayComponent/RegionViewComponent/, 

since dragsource may stack image viewer to not-major tab, and dynamically changing layout also resize each area to 0 and then re-layout

duidae:
the rest is in  PR #406

PR #407, opened 2019-07-24
AuthN/AuthZ implementation

veggiesaurus:
This PR enables server-side username/password authentication, by sending a `POST` HTTP request to an authorization server using [basic auth](https://en.wikipedia.org/wiki/Basic_access_authentication). As any sensible server would be running HTTPS, this is a sufficiently secure approach. The `authUrl` used when making this request defaults to `${window.location.protocol}//${window.location.hostname}/auth` but can be overridden by specifying `authUrl` in the URL query parameters. 

As described in the server-side auth [doc](https://docs.google.com/document/d/1MZcq2aV0dI8urHs_7sywQpoGHiBmnNQzp7VxCOy9n-A/edit), the authentication server provides a unique token, as well as a `socket`, which indicates which websocket address the frontend should connect to. The token is sent to the backend in the form of a header on connection. This allows us to ensure that the frontend user connects to the correct backend, only after authenticating.  

As the API key feature is no longer required, this has been removed. The "Connect to remote URL" feature has also been removed, as this would be difficult to integrate with authentication. For testing purposes, the frontend can still be instructed to connect to a different url using the URL query params. 

Authentication is turned off by default, and the frontend will simply connect to the default socket without any authentication unless `REACT_APP_AUTHENTICATION` is set to `true` in `.env` or `.env.local`.


veggiesaurus:
@kswang1029 @ajm-asiaa please have a look at the PR description and the linked docs if you get a chance.

duidae:
we can show some server info (ip...etc) in the login dialog for improving UX

veggiesaurus:
> we can show some server info (ip...etc) in the login dialog for improving UX

Can you make a separate issue/question to discuss that addition?

PR #414, opened 2019-07-29
Angus/polygon regions

veggiesaurus:
This PR adds support for polygonal regions. Region creation takes place as follows:
- Left click to create a control point
- Double-click to finish creating the region

Region editing takes place as follows:
- Left click anywhere on a line segment to create a new control point
- Double-click on a control point to delete it
- Click and drag a control point to move it. 
- Click and drag the inside of the region to move the entire region

edit_poly

There are still some issues to fix (notifying the user that a region is complex, for example), but they can be handled in a separate PR


veggiesaurus:
> looks great & polygon editing works wonderfully, a minor suggestion to shrink render() in PolygonRegionComponent by moving anchors/newAnchor out

good idea. I've done so, please re-review.

duidae:
@veggiesaurus thanks!

PR #419, opened 2019-07-30
Qi/stokes widget layout2

panchyo0:
# 374 Responsive layout
# 377 Wiring up real data
# 376 Scatter plot component (still working on equal scale)


veggiesaurus:
@panchyo0 I've fixed the crash issues that were popping up in this branch, and merged `dev` into it as well. Will look at the code next. There is a backend crash that is causing some issues with Stokes profiles, but that's unrelated to your widget's behaviour. 

@arusstaylor you can give the widget a try [here](https://carta.asiaa.sinica.edu.tw/frontend/c96a78e/?folder=public&file=GNPilot.IQU.fits&socketUrl=wss://carta.asiaa.sinica.edu.tw/socketdev)

arusstaylor:
Looking good.  A few points:

- The Q, U profile colour contrast needs to be better.  
-  in the Q, U scatter plot the location of the average value of Q, U should be indicated by a cross hair. 
- should have the tool to change the plot to points, lines, histogram as for the regular profiles.  When one of these options is selected for one profile it should apply to all.

kswang1029:
@arusstaylor quick question: usually how is the color determined in the scatter QU plot? Are we using the full frequency range of the cube? Thinking the case of polarized line emission which is only limited to few channels. Would it be better if we use the x range of the QU spectra/PI/PA plots for the colors? Then if we just zoom in to the polarized lien feature, the scatter plot will be colored properly?

arusstaylor:
Good question.  My though had been that we use the full frequency range of the data to determine the colour mapping to frequency.  When we zoom in on a subrange in frequency then the scatter plot should only show the points in that frequency range.  For the use case that you point of of a polarised spectral line I think the desired behaviour is to have the colour mapping redone for just the frequency range that you are zoomed in on.

veggiesaurus:
@panchyo0 we can leave most of the comments for another PR. A couple of minor ones to fix for this one: 
- The y-axis label for the "Q+U" plot should be "Stokes Value (<unit_name>)"
- The y-axis for the polarization angle should be in degrees, not radians.
- Can you ensure that the min and max of the axes in the scatter plot are the same when auto-scalling? 

i.e for the x axis:
```
const xLimit = Math.max(Math.abs(xMin), Math.abs(xMax))
xMin = -xLimit;
xMax = xLimit;
```
This will ensure that, by default, the origin of the graph is in the center, rather than jumping around like that.

PR #421, opened 2019-07-31
Jason/refactor some

duidae:
Refine the behavior of deleting a layout:
if deleting current layout & it's preference layout => apply & set preference to preset default
if deleting current layout & it's not preference layout => apply preference layout
if deleting other layout which is preference => set preference to preset default(no applying)

& minor refactorings

### DIFF
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -98,7 +98,7 @@ export class PreferenceDialogComponent extends React.Component<{ appStore: AppSt
                     <Switch checked={preference.autoLaunch} onChange={(ev) => { preference.setAutoLaunch(ev.currentTarget.checked); }}/>
                 </FormGroup>
                 <FormGroup inline={true} label="Initial Layout">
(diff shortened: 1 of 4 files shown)


veggiesaurus:
I'm not convinced about this behaviour. I agree that deleting a layout should prevent it from being set as the default in future, but I don't think we should change the current layout. It's just annoying to users. 

For example, suppose they choose a layout, and then fiddle around with it, and plan to save it, but before that, decide to delete the old layout. Then their customised layout will be reset entirely to default. 

@kswang1029 what do you think? Example of what I mean below
delete_layout

kswang1029:
@veggiesaurus agreed with you.

duidae:
agreed with delete without applying new layout immediately, then the question is how to show the current active layout in the layout list?

veggiesaurus:
> agreed with delete without applying new layout immediately, then the question is how to show the current active layout in the layout list? it has no name yet without saving

Realistically, as soon as the layout changes, there's no longer an "active" layout. So if we delete a layout, there should simply be no active layout, as the current layout doesn't match

duidae:
@veggiesaurus  change behavior to no applying, will not affect currently used layout

PR #430, opened 2019-07-31
defer SET_REGION until point is moved when adding a new control point

### LINKED ISSUE #429: adding new anchor to an existing polygon triggers SET_REGION and relevant data update
@kswang1029 commented on [Wed Jul 31 2019](https://github.com/CARTAvis/carta-backend/issues/300)

Not sure where (frontend or backend) we improve this. Guessing we can solve this in the backend by checking area or equivalent quantity. 

If we add a new anchor to an existing polygon, a new SET_REGION will be triggered, and relevant data will be updated. However, the new anchor sits perfectly on a straight line so the area, hence the covered pixels, remains the same. Therefore the relevant data update is not necessary. Allowing this behavior might introduce performance issue when the image is big and/or the polygon region is big.

Jul-31-2019 12-08-27

veggiesaurus:
fixes #429. @kswang1029 please confirm fix with mix 'n match

### DIFF
--- a/src/components/ImageView/RegionView/PolygonRegionComponent.tsx
+++ b/src/components/ImageView/RegionView/PolygonRegionComponent.tsx
@@ -50,10 +50,12 @@ export class PolygonRegionComponent extends React.Component<PolygonRegionCompone
                 this.props.onSelect(region);
             }
 
+            // Add a new control point to the region between two existing control points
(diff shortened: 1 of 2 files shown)


kswang1029:
@veggiesaurus Finally got the chance to test it with mix n match page after it is recovered. 
Yes, it works properly now.

PR #438, opened 2019-08-02
Jason/fix 425

### LINKED ISSUE #425: missing widget in a user customized and saved layout
as in the example, there are four spectral profiler widgets when the layout is saved.
Screen Shot 2019-07-31 at 1 27 32 PM
But once it is re-loaded, three widgets are missing.
Screen Shot 2019-07-31 at 1 28 07 PM

duidae:
closes #425


kswang1029:
here I see multiple identical widgets can be reloaded. 👍 
Screen Shot 2019-08-02 at 12 19 23 PM


Any idea on the tab title of the image viewer? it says "No image loaded".

kswang1029:
Great to see we have good handle of this exception where we have just one widget and it is undocked:
Screen Shot 2019-08-02 at 12 48 40 PM

👍

kswang1029:
Just to confirm that we only allow 10 user-defined layouts? Should be fine in general use cases. @veggiesaurus @duidae ideas?
Screen Shot 2019-08-02 at 12 49 27 PM

duidae:
title issue fixed, and yes I set the maximum number of layouts to 10, can be altered according to requirements

PR #440, opened 2019-08-02
Jason/fix 423

### LINKED ISSUE #243: improvement of region list widget

### LINKED ISSUE #423: initial cursor positions in image appending mode
The initial cursor position is set to the image center. However, this cursor position is also applied to other frames in the appending mode.

Example (check the position of the cross):
Frame 1
Screen Shot 2019-07-31 at 1 01 42 PM

and Frame 2
Screen Shot 2019-07-31 at 1 01 55 PM

duidae:
closes #423 by introducing cursoFrozen &cursoFrozenPoint to FrameStore, letting each frame has its own cursor state

### DIFF
--- a/src/components/ImageView/ImageViewComponent.tsx
+++ b/src/components/ImageView/ImageViewComponent.tsx
@@ -219,7 +219,7 @@ export class ImageViewComponent extends React.Component<WidgetProps> {
                         overlaySettings={appStore.overlayStore}
                         isRegionCornerMode={appStore.preferenceStore.isRegionCornerMode}
                         cursorFrozen={appStore.cursorFrozen}
-                        cursorPoint={appStore.cursorInfo ? appStore.cursorInfo.posImageSpace : null}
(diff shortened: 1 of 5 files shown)


kswang1029:
few issues:
1) the initial cursor position seems off.
Screen Shot 2019-08-02 at 3 33 49 PM
2) when switching to different frames, the cursor info bar needs to be updated accordingly (e.g., coordinates, value, frequency and velocity).
Screen Shot 2019-08-02 at 3 34 09 PM
Screen Shot 2019-08-02 at 3 33 58 PM

duidae:
fixed above issue

Issue #420, opened 2019-07-31
make the breadcrumbs always visible even if scrolling down a long file list


kswang1029:
Currently the breadcrumbs are visible when the file list is scrolled to the top. If the file list is long and if we scroll down, breadcrumbs are not visible. 
![Screen Shot 2019-07-31 at 11 44 25 AM](https://user-images.githubusercontent.com/20819712/62182350-e3c87480-b388-11e9-9843-d1b45beafab1.png)
![Screen Shot 2019-07-31 at 11 44 38 AM](https://user-images.githubusercontent.com/20819712/62182352-e7f49200-b388-11e9-97af-85393a785e5a.png)

Suggest to make breadcrumbs always visible just like the header (File name, type, size) of the file list.

kswang1029:
a relevant cosmetics issue of the scroll bar:
![Jul-31-2019 13-49-01](https://user-images.githubusercontent.com/20819712/62186821-1169e980-b39a-11e9-86f1-3a60aa534f43.gif)

zarda:
![Screen Shot 2019-08-01 at 16 48 27](https://user-images.githubusercontent.com/10767557/62279273-58301000-b47c-11e9-9a63-5ac057d93e59.png)

veggiesaurus:
@zarda if we can just adjust the padding a bit, I think that would be a good solution

kswang1029:
Are we using the space above file info for the breadcrumbs as well? That helps to display path for nested folder structure I think.

veggiesaurus:
> Are we using the space above file info for the breadcrumbs as well? That helps to display path for nested folder structure I think.

We could do that, if you don't think it would look odd. @zarda , perhaps give that a try?

zarda:
![Screen Shot 2019-08-01 at 18 25 45](https://user-images.githubusercontent.com/10767557/62286263-f1b1ee80-b489-11e9-910d-8d10dcf047a5.png)

veggiesaurus:
@zarda I think KS meant use the entire vertical space, not _just_ the info. So something like this pseudocode:

```
<div>
    <breadcrumbs/>
    <div display=flex>
        <FileList/>
        <FileInfo/>
    </div>
</div>
```

kswang1029:
a mockup would be like this
[...]

PR #441, opened 2019-08-02
Hj/#420 layout breadcrumb

zarda:
Fixed the issue #420 and fine tuned the layout to avoid incomplete file list.

### DIFF
--- a/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.scss
+++ b/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.scss
@@ -2,6 +2,10 @@
     min-width: 60%;
     height: 600px;
     user-select: none;
+    .file-path {
+        margin-bottom: 15px;
+        display: flex;
+    }
     .file-info-pane {
         width: 40%;
         margin-left: 5px;
--- a/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.scss
+++ b/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.scss
@@ -14,7 +14,7 @@
             }
             .file-list {
                 overflow-y: auto;
-                height: 94%;
+                height: 93%;
             }
         }
         .file-info-pane {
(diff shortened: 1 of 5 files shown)


veggiesaurus:
@zarda this is looking good. One comment: I like the change you've made to the file list table's header, but I'm not convinced we need `height: 40px` for that
image

Just using the normal height looks more consistent: 
image

Or is it done for alignment purposes? @kswang1029 what do you think?

kswang1029:
@veggiesaurus I think `height: 40px` is for alignment. Following are the two screenshots with different theme. In the file info/file header, we have a horizontal bar as an indicator on what the currently view is. I am not sure how to solve this in a clever way... In addition, in the dark mode, the extra height is even highlighted due to its background color. 
Screen Shot 2019-08-06 at 9 26 25 AM
Screen Shot 2019-08-06 at 9 26 10 AM

kswang1029:
@zarda @veggiesaurus there is an issue with this PR: breadcrumbs do not work after loading an image.
When loading an image for the first time, it works.
Screen Shot 2019-08-06 at 10 49 58 AM
After an image is loaded and if we launch file browser again, the breadcrumbs do not work and we are locked in the current folder.
Screen Shot 2019-08-06 at 10 50 12 AM

kswang1029:
@zarda @veggiesaurus UPDATE: if you click "header" tab, the breadcrumbs show up. If you make the dialog wider, the breadcrumbs show up too.

veggiesaurus:
> @zarda @veggiesaurus UPDATE: if you click "header" tab, the breadcrumbs show up. If you make the dialog wider, the breadcrumbs show up too.

@kswang1029 and this doesn't happen in dev?

kswang1029:
@veggiesaurus no. in dev, it works fine.

zarda:
The Breadcrumbs collapsed while reopening file browser. This issue will be solve whenever opening file browser again or resizing. I do not find any solution but a workaround is to set minVisibleItems=2.

veggiesaurus:
> The Breadcrumbs collapsed while reopening file browser. This issue will be solve whenever opening file browser again or resizing. So fixed it by re-show the file browser after opening.

This is not the right way to solve the problem, because it means a duplicate `FILE_LIST_REQUEST` message gets sent to the backend every time you open the file browser. If the directory has a lot of files, this will effectively double the time it takes to list the files.

veggiesaurus:
@zarda can you try to find another way to solve this? the current solution also breaks appending files

PR #442, opened 2019-08-05
adds simple polygon test for polygonal regions

### LINKED ISSUE #370: support for polygonal regions
Split off from #239

- [x] Create polygonal regions
- [x] Select polygonal regions
- [x] Edit polygonal regions
- [x] Move polygonal regions
- [x] Display polygon region info dialog
- [x] Display polygon region info in region list
- [x] Edit polygon region in info dialog
- [x] Ensure polygon is simple

veggiesaurus:
This PR adds a test for whether a polygonal region is simple. If it is _not_ simple, the polygon is drawn in pink. 

closes #370

### DIFF
--- a/src/components/ImageView/RegionView/PolygonRegionComponent.tsx
+++ b/src/components/ImageView/RegionView/PolygonRegionComponent.tsx
@@ -7,6 +7,7 @@ import {FrameStore, RegionStore} from "stores";
 import {Point2D} from "models";
 import {add2D, average2D, closestPointOnLine} from "utilities";
 import {canvasToImagePos, imageToCanvasPos} from "./shared";
+import {Colors} from "@blueprintjs/core";
 
 export interface PolygonRegionComponentProps {
     region: RegionStore;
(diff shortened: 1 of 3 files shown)


kswang1029:
@veggiesaurus 
One issue:
Aug-06-2019 09-50-49

veggiesaurus:
@kswang1029 I mistakenly assumed that a polygon would have at most one crossing point 😆 I'll fix it so that we only use the optimised check (_O(N)_) when the polygon has previously been determined to be simple. That should solve the issue

kswang1029:
Just as a reference here about what happens with complex polygon with casacore.
Consistent:
Screen Shot 2019-08-06 at 2 49 12 PM
Inconsistent:
Screen Shot 2019-08-06 at 2 50 36 PM

veggiesaurus:
> Inconsistent:

I'm not sure what it should be consistent with, given that a complex polygon has an area that depends on how you define it.

kswang1029:
> > Inconsistent:
> 
> I'm not sure what it should be consistent with, given that a complex polygon has an area that depends on how you define it.

Right. So what we should do is to warn users by pink polygon but still allow the backend to process the complex polygon since casacore can handle it.

veggiesaurus:
@kswang1029 fixed:
simple_polygon

@duidae please review.

kswang1029:
@veggiesaurus tried with `origin/angus/test_simple_polygon e31c4c9 2019-08-06 20:06:49 CST`. NICE!

PR #448, opened 2019-08-06
plot improvements

veggiesaurus:
This PR makes several improvements to plotting in the frontend:
1. The default "tolerance" is reduced to the default value for AST. Should be good enough for visualisation purposes. This is configurable in the overlay settings if users need more accuracy for a "paper-ready" plot. This change speeds up AST plotting considerably (from 16 ms to 8 ms on a MacBook pro in a brief test I conducted with supermosaic.10). This is especially important on devices with smooth scrolling, such as laptops with trackpads. 
2. Instead of automatically adjusting each spatial profile's range whenever the user zooms in on the image, this is now throttled to a maximum of 10 fps. 
3. Switches to lines instead of steps when the profile has more than one point per pixel. The change should be barely noticeable, but line plots have 3 times fewer line segments to plot, so they should be more performant. 
4. Extends the horizontal range of the spatial profile's data by one, which fixes the issue of missing data points at the edge of the plot
5. Adds a padding amount (10%) to the vertical range when autoscaling, to prevent plots from hitting the top and bottom of the plot and being obscured

The combined effect of (1), (2) and (3) should improve usability when zooming in and out of images. 

For a comparison of (4) and (5): 
`dev`:
image

This branch:
image


kswang1029:
for 4) 
> Extends the horizontal range of the spatial profile's data by one, which fixes the issue of missing data points at the edge of the plot

with e371e05, i see
Screen Shot 2019-08-07 at 10 22 39 AM
On the right side there is extra space for the 11th pixel (this is a 11x11 pixel image; index 0~10).
I see the same in dev.

kswang1029:
@veggiesaurus would these improvements apply to the Stokes widget too?

veggiesaurus:
> @veggiesaurus would these improvements apply to the Stokes widget too?

@kswang1029 I will make a separate issue for the Stokes and spectral widgets, as well as the extra pixel issue you mentioned above, once the Stokes widget development has settled down a bit :)

PR #449, opened 2019-08-06
Angus/region locking

### LINKED ISSUE #439: cannot add new anchor to a polygon when it overlaps with other region in some cases
The guider disappears in some circumstances when a polygon overlaps with other region. 
It appears that the issue is related to the relative areas of the two regions.

This works fine:
Aug-02-2019 13-47-47

But not this one:
Aug-02-2019 13-47-29

veggiesaurus:
Fixes #439 by introducing the ability to lock and unlock regions. 

When a region is locked, it is removed from the region hit graph, so no interactions with it can take place. This has three potential uses:
- allowing users to edit other overlapping regions without selecting the locked region (preventing #439 )
- allowing users to keep a specific region fixed, to avoid accidentally moving it
- allowing users to pan within a region without resorting to the middle-click or Ctrl + Click shortcut

There are three ways to lock or unlock regions: 
- The region list has a new column, which shows whether a region is locked or not. Clicking on the icon toggles the lock.
- The region dialog has a small lock icon next to the delete and close buttons. Clicking on the icon toggles the lock.
- Pressing `L` locks the currently selected region. Pressing `shift + L` unlocks all regions. See below for example

out


kswang1029:
> When a region is locked, it is removed from the region hit graph, so no interactions with it can take place.
Does this include resizing? Currently i see a locked region can be resized.

veggiesaurus:
> > When a region is locked, it is removed from the region hit graph, so no interactions with it can take place.
> 
> Does this include resizing? Currently i see a locked region can be resized.
> 
> When saving as png, the locked region keeps the transparency reduced. Would it be better to use the anchor color as a lock indicator?

fixed. @duidae please review

PR #456, opened 2019-08-07
Jason/fix 444

### LINKED ISSUE #444: missing image viewer when reloading a special layout
if there is only one widget docked and the image viewer widget is undocked, and if we save this kind of layout and reload it, the image viewer will be gone and cannot be launched anymore. 
Screen Shot 2019-08-06 at 2 05 38 PM

duidae:
closes #444 & supporting floating widgets

### DIFF
--- a/src/components/FloatingWidget/FloatingWidgetComponent.tsx
+++ b/src/components/FloatingWidget/FloatingWidgetComponent.tsx
@@ -83,8 +83,12 @@ export class FloatingWidgetComponent extends React.Component<FloatingWidgetCompo
             <Rnd
                 className={className}
                 style={{zIndex: this.props.zIndex}}
-                size={{width: widgetConfig.defaultWidth, height: widgetConfig.defaultHeight + headerHeight}}
-                position={{x: widgetConfig.defaultX, y: widgetConfig.defaultY}}
+                default={{
+                    x: widgetConfig.defaultX,
+                    y: widgetConfig.defaultY,
+                    width: widgetConfig.defaultWidth,
+                    height: widgetConfig.defaultHeight + headerHeight
+                }}
                 resizeGrid={[25, 25]}
                 dragGrid={[25, 25]}
                 minWidth={widgetConfig.minWidth}
--- a/src/components/FloatingWidget/FloatingWidgetComponent.tsx
+++ b/src/components/FloatingWidget/FloatingWidgetComponent.tsx
@@ -97,6 +97,10 @@ export class FloatingWidgetComponent extends React.Component<FloatingWidgetCompo
                 bounds={".gl-container"}
                 dragHandleClassName={"floating-title"}
                 onMouseDown={this.props.onSelected}
+                onDragStop={(e, data) => {
+                    widgetConfig["defaultX"] = data.x;
+                    widgetConfig["defaultY"] = data.y;
+                }}
             >
                 <div className={titleClass}>
                     <div className={"floating-title"}>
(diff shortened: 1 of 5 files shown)


veggiesaurus:
@duidae I haven't looked at the code yet, but one thing I noticed is that the floating widgets aren't restored to the correct location. This needs to be saved as well.

duidae:
floatingWidgets in widgetsStore only saves min/default width/height, & tried document.getElementById() but get null object, &  checked the FloatingWidgetComponent/FloatingWidgetManagerComponent, there is no such information either, so wondering is there somewhere else we can get those info?

kswang1029:
In my mind, the minimal requirement is having floating widgets to default size and they are not blocking each others.

veggiesaurus:
@duidae, the `Rnd` component has two useful callbacks: `onResizeStop` and `onDragStop`. Use those to keep track of the widget's current position/size via some additional properties (don't need to be `observable` as far as I can see, since they are purely used for keeping track of things) in `WidgetConfig` maybe?

duidae:
@veggiesaurus oh ok Angus thanks for the info! Let me try it

duidae:
@veggiesaurus  @kswang1029  ok, please review, thx

veggiesaurus:
@duidae this looks _almost _ right. Two issues:
- floating widget resizing seems to only work correctly from right and bottom edges now
- Is it possible to retain compatibility with existing layouts that don't have a floating layout? perhaps we need to start versioning our layout config (on a per-layout basis).

src/components/FloatingWidget/FloatingWidgetComponent.tsx
```diff
@@ -83,20 +83,23 @@ export class FloatingWidgetComponent extends React.Component<FloatingWidgetCompo
...
                 className={className}
                 style={{zIndex: this.props.zIndex}}
-                default={{
-                    // Shift by 5 pixels to compensate for 5px CSS margins
-                    x: widgetConfig.defaultX !== undefined ? widgetConfig.defaultX : widgetsStore.defaultFloatingWidgetOffset + 5,
-                    y: widgetConfig.defaultY !== undefined ? widgetConfig.defaultY : widgetsStore.defaultFloatingWidgetOffset,
-                    width: widgetConfig.defaultWidth,
-                    height: widgetConfig.defaultHeight + headerHeight,
-                }}
+                size={{width: widgetConfig.defaultWidth, height: widgetConfig.defaultHeight + headerHeight}}
```
veggiesaurus: are you sure we need to make this a controlled property? Can't we just simply use the updated values from the events when saving and restoring layouts?

duidae: ah sorry use that prop without fully understanding it's meaning...a little bit confusion actually ToT

duidae:
1 fixed;
about 2, I think versioning is needed for backward compatibility, but I dont have much experience about it, my idea is to add a {verions: 1.0} to CARTA_saved_layouts & add corresponding parser function named with version, i.e. layoutConfigParser1.0(), not sure is that enough? @veggiesaurus can u show some tips about it? 
Btw, can we skip supporting existing layouts without floating widget, since this feature is not released yet, so users dont have those layouts & only exists in our testers, can be deleted by typing localStorage.removeItem("CARTA_saved_layouts") in console

src/components/FloatingWidget/FloatingWidgetComponent.tsx
duidae: adding update size/position in componentDidUpdate() fixes the problem with default prop

PR #459, opened 2019-08-08
Add y-profile widget as createFloatingSpatialProfileWidget

zarda:
Fixed that createFloatingSpatialProfileWidget() did not create y-profile widget.


veggiesaurus:
I'm not sure that was the desired behaviour in the first place. Isn't one button-click supposed to mean one widget? The behaviour is also inconsistent, because clicking and dragging just creates one widget.  The created floating widgets are also on top of each other, so it looks like just one widget is created. Thoughts, @kswang1029?

kswang1029:
What’s the intention of this PR? @zarda Which issue are you referring to?

Issue #450, opened 2019-08-07
make the icon of stokes widget consistent


kswang1029:
The icon in the menu list should be consistent with the one in the widget bar. 
![Screen Shot 2019-08-07 at 9 35 20 AM](https://user-images.githubusercontent.com/20819712/62588015-cea69500-b8f6-11e9-9026-92adcbd0d0fc.png)

Issue #378, opened 2019-07-15
Stokes Widget: Integration between individual plots


veggiesaurus:
The plots within the widget should have a number of shared properties, such as the x axis of the Q&U, PI and PA plots, as well as showing/hiding points in the scatter plot based on the zoom range of the other three plots.

PR #462, opened 2019-08-08
Qi/stokes interaction x

panchyo0:
1. interaction between scatter plots points colour and line plots shared x axis. #378 
2. fixed issue #450 
3. change Q&U y axis label from "Stokes Value" to "Value"
4. calculated PA by (0.5 * (Math.atan(u / q)) ) * (180/pi)

### DIFF
--- a/src/components/Shared/ScatterPlot/ScatterPlotComponent.tsx
+++ b/src/components/Shared/ScatterPlot/ScatterPlotComponent.tsx
@@ -3,6 +3,7 @@ import {observer} from "mobx-react";
 import ReactResizeDetector from "react-resize-detector";
 import {Layer, Stage} from "react-konva";
 import {Colors} from "@blueprintjs/core";
+import {ChartArea} from "chart.js";
 import {PlotContainerComponent} from "components/Shared/LinePlot/PlotContainer/PlotContainerComponent";
 import {ToolbarComponent} from "components/Shared/LinePlot/Toolbar/ToolbarComponent";
 import {LinePlotComponent} from "components/Shared/LinePlot/LinePlotComponent";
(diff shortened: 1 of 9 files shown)


src/components/Shared/ScatterPlot/ScatterPlotComponent.tsx
```diff
@@ -21,17 +24,28 @@ export class ScatterPlotComponent extends LinePlotComponent {
...
+        return ["hsla(", hue, ",100%,50%,", this.opacityInit, ")"].join("");
     }
 
     private fillColor(): Array<string> {
         let scatterColors = [];
-        if (this.props.data) {
-            this.props.data.forEach(data => {
-                let pointColor = this.getScatterColor(data.y, this.props.yMin, this.props.yMax, this.props.colorRangeEnd);
+        if (this.props.data && this.props.data.length && this.props.scatterColorIndex && this.props.scatterColorIndex.length && this.props.interactionBorder) {
+            let yMin = getMinY(this.props.data.filter(data => {return (!isNaN(data.x) && !isNaN(data.y)); }));
```
veggiesaurus: this is a really inefficient way to get the min and max. I'm modifying the `getMinY`  and `getMaxY` functions to just use a normal for loop, and to ignore `NaN` values. It's a little less "neat", but much faster. See [here](https://jsperf.com/minmax-testing/1)

src/components/Shared/ScatterPlot/ScatterPlotComponent.tsx
```diff
@@ -21,17 +24,27 @@ export class ScatterPlotComponent extends LinePlotComponent {
         }
     }
 
-    private getScatterColor(value: number, min: number, max: number, toColor: number): string {
+    private getScatterColor(value: number, min: number, max: number, toColor: number, outRange: boolean): string {
         let percentage = (value + Math.abs(min)) / (Math.abs(min) + Math.abs(max));
         let hue = (percentage * toColor).toString(10);
-        return ["hsl(", hue, ",100%,50%)"].join("");
+        if (outRange) {
+            return ["hsla(", hue, ",0%,50%,", this.opacityOutRange, ")"].join("");
```
veggiesaurus: we should use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for things like this. Fixing.

Issue #467, opened 2019-08-10
Channel indicator for Scatter plot 


panchyo0:
1.  indicator as to which channel is selected in the scatter plot
2. mouse cursor hovered on the other plots, show corresponded point

PR #468, opened 2019-08-10
Qi/stokes scatter indicator

panchyo0:
1. indicator as to which channel is selected in the scatter plot, current channel with a larger point radio.
2. interact with mouse cursor to indicate current hovered channel. 
related to #467


veggiesaurus:
I've made some changes to this branch, @duidae can you have a quick look over the code? Unlike the PR name, it no longer actually adds a scatter indicator, but it does have some other useful changes :)

duidae:
@veggiesaurus refactoring looks good

Issue #435, opened 2019-08-01
cursor info bar: cursor value when the cursor is outside the image


kswang1029:
it would be better to show nothing about the cursor value when the cursor is outside the image to avoid confusion as coordinates are still updating along with the cursor movement. 

![Aug-01-2019 10-50-39](https://user-images.githubusercontent.com/20819712/62262135-3e75d500-b44a-11e9-989b-20db2ce691c0.gif)

Kechil:
leave out whole value field?

PR #469, opened 2019-08-12
Mylin/fix 435 2

acdo2002:
Close the issue #435

### DIFF
--- a/src/stores/FrameStore.ts
+++ b/src/stores/FrameStore.ts
@@ -245,7 +245,7 @@ export class FrameStore {
         }
 
         // Is cursorPos inside the image?
-        var isInsideImg = false;
+        let isInsideImg = false;
         if (cursorPosImageSpace.x + 0.5 < this.frameInfo.fileInfoExtended.width && 
             cursorPosImageSpace.x + 0.5 > 0 &&
             cursorPosImageSpace.y + 0.5 < this.frameInfo.fileInfoExtended.height &&
--- a/src/stores/FrameStore.ts
+++ b/src/stores/FrameStore.ts
@@ -244,9 +244,19 @@ export class FrameStore {
             }
         }
 
+        // Is cursorPos inside the image?
+        var Isinsideimg = true;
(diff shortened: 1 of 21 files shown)


acdo2002:
Hi @kswang1029 ,

I see, many thanks for the example. 
So I should change the codes to fit the following requirements:
(1) KEEP the cursor info bar "always be visible".
(2) Cursor INSIDE the image, the cursor info bar shows 5 items: WCS, Image, Frequency, VALUE & velocity.
Otherwise (cursor outside the image), the cursor info bar only shows 4 items: WCS, Image, Frequency & velocity.

Please let me know whether my understand is correct.

kswang1029:
@acdo2002 yes

veggiesaurus:
> Hi @kswang1029 ,
> 
> I see, many thanks for the example.
> So I should change the codes to fit the following requirements:
> (1) KEEP the cursor info bar "always be visible".
> (2) Cursor INSIDE the image, the cursor info bar shows 5 items: WCS, Image, Frequency, VALUE & velocity.
> Otherwise (cursor outside the image), the cursor info bar only shows 4 items: WCS, Image, Frequency & velocity.
> 
> Please let me know whether my understand is correct.

@acdo2002 currently, if the `activeFrame`'s  `cursorValue` is undefined, the cursor overlay leaves out the `cursor: <cursorVale>` part of the cursor info bar. I think therefore the easiest way to handle this issue is to simply set the `cursorValue` to undefined when the user moves the cursor out of the image

src/stores/FrameStore.ts
```diff
@@ -244,9 +247,19 @@ export class FrameStore {
             }
         }
 
+        // Is cursorPos inside the image?
+        let isInsideImg = false;
+        if (cursorPosImageSpace.x + ORIGIN_POINT_OFFSET < this.frameInfo.fileInfoExtended.width && 
+            cursorPosImageSpace.x + ORIGIN_POINT_OFFSET > 0 &&
```
veggiesaurus: you can just use `Math.round(cursorPosImageSpace.x)` can't you?

PR #473, opened 2019-08-12
Import and Export CRTF

veggiesaurus:
This PR is a companion PR of [carta-backend#312](https://github.com/CARTAvis/carta-backend/pull/312).

It enables region importing and exporting of CRTF files. The existing file browser component has been adapted for displaying responses from region list requests, as well as region info responses.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -6,16 +6,8 @@ import ReactResizeDetector from "react-resize-detector";
 import {Alert, Classes, Colors, Dialog, Hotkey, Hotkeys, HotkeysTarget, Intent} from "@blueprintjs/core";
 import {exportImage, FloatingWidgetManagerComponent, RootMenuComponent} from "./components";
 import {AppToaster} from "./components/Shared";
-import {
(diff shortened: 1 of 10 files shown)


kswang1029:
Some issues:
1) the point region is missing. The crtf syntax is "symbol[[320pix, 200pix], '.']" (I admit it is not a proper syntax)
2) CASA image format is now displayed as "Unknown". It should be "CASA"
Screen Shot 2019-08-13 at 1 51 19 PM

veggiesaurus:
@kswang1029 

> 1. the point region is missing. The crtf syntax is "symbol[[320pix, 200pix], '.']" (I admit it is not a proper syntax)

I've checked, and the frontend is sending the correct region IDs to the backend, meaning this is a backend issue. Can you make a separate issue for it?

> 2. CASA image format is now displayed as "Unknown". It should be "CASA"

Fixed in latest commit

kswang1029:
Will do. Thanks for sorting it out!

PR #478, opened 2019-08-13
Add layout version before beta

duidae:


arusstaylor:
Would be nice to be able to share and import and export layouts.

veggiesaurus:
> Would be nice to be able to share and import and export layouts.

I agree. @arusstaylor can you add an issue for this? I think once we have the layouts stored in a server database, it will be reasonably easy to handle.

duidae:
@veggiesaurus  ok added, please review

Issue #485, opened 2019-08-16
region profile update of an HDF5 cube in the Stokes widget does not have changing line opacity 


kswang1029:
frontend master 3335f5e

When viewing region spectral profile of an HDF5 cube, the profile is updating with changing line opacity. The same feature should be applied to the Stokes widget too.

veggiesaurus:
@panchyo0 can you have a look at how this is handled in the spectral profile widget and use that same approach for the stokes widget?

PR #488, opened 2019-08-17
issue 485

panchyo0:
stokes widget: change line opacity when region updating data #485

### DIFF
--- a/src/components/Shared/LinePlot/PlotContainer/PlotContainerComponent.tsx
+++ b/src/components/Shared/LinePlot/PlotContainer/PlotContainerComponent.tsx
@@ -379,13 +379,21 @@ export class PlotContainerComponent extends React.Component<PlotContainerProps> 
                     data: value,
                     fill: false,
                     lineTension: 0,
-                    borderColor: this.props.multiPlotBorderColor ? this.props.multiPlotBorderColor.get(key) : lineColor,
                     showLine: true,
                     steppedLine: this.props.interpolateLines ? false : "middle",
                     borderWidth: 1,
                     pointRadius: 0,
                     backgroundColor: this.props.dataBackgroundColor ? this.props.dataBackgroundColor : []
                 };
+                if (opacity < 1) {
+                    multiPlotDatasetConfig.borderColor = lineColor;
+                } else {
+                    if (this.props.multiPlotBorderColor) {
+                        multiPlotDatasetConfig.borderColor = this.props.multiPlotBorderColor.get(key);
+                    } else {
+                        multiPlotDatasetConfig.borderColor = lineColor;
+                    }
+                }
                 plotData.datasets.push(multiPlotDatasetConfig);
             });
         }
--- a/src/components/Shared/LinePlot/PlotContainer/PlotContainerComponent.tsx
+++ b/src/components/Shared/LinePlot/PlotContainer/PlotContainerComponent.tsx
@@ -379,7 +379,7 @@ export class PlotContainerComponent extends React.Component<PlotContainerProps> 
                 }
                 const multiPlotDatasetConfig: ChartDataSets = {
                     type: this.props.plotType ? this.props.plotType : "line",
-                    label: key,
+                    label: key[0],
                     data: value,
                     fill: false,
                     lineTension: 0,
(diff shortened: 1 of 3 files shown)


src/components/Shared/LinePlot/PlotContainer/PlotContainerComponent.tsx
```diff
@@ -379,13 +379,21 @@ export class PlotContainerComponent extends React.Component<PlotContainerProps>
...
                     steppedLine: this.props.interpolateLines ? false : "middle",
                     borderWidth: 1,
                     pointRadius: 0,
                     backgroundColor: this.props.dataBackgroundColor ? this.props.dataBackgroundColor : []
                 };
+                if (opacity < 1) {
+                    multiPlotDatasetConfig.borderColor = lineColor;
+                } else {
+                    if (this.props.multiPlotBorderColor) {
+                        multiPlotDatasetConfig.borderColor = this.props.multiPlotBorderColor.get(key);
```
veggiesaurus: why does this not also happen if opacity = 1?

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -565,6 +575,16 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
...
                 paLinePlotProps.data = currentPlotData.paValues.dataset;
                 quScatterPlotProps.data = currentPlotData.quValues.dataset;
+ 
+                const lineOpacity = this.minProgress < 1.0 ? 0.15 + this.minProgress / 4.0 : 1.0;
+                quLinePlotProps.opacity = lineOpacity;
+                piLinePlotProps.opacity = lineOpacity;
+                paLinePlotProps.opacity = lineOpacity;
+                quLinePlotProps.opacity = lineOpacity;
+                if (lineOpacity === 1) {
+                    quLinePlotProps.multiPlotBorderColor.set(StokesCoordinate.LinearPolarizationQ, this.QlinePlotColor);
```
veggiesaurus: not sure why this only applies when the opacity is 1

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -114,10 +117,17 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
...
+                    const minProgress = Math.min(currentData.qProgress, currentData.uProgress);
+                    let currentProgress = "";
+                    if (minProgress === currentData.qProgress) {
+                        currentProgress = "Qz";
+                    } else {
+                        currentProgress = "Uz";
+                    }
                     if (minProgress < 1) {
-                        progressString = `[${(minProgress * 100).toFixed(0)}% complete]`;
+                        progressString = `[${(minProgress * 100).toFixed(0)}% complete ${currentProgress}]`;
```
veggiesaurus: not sure if the user needs to know this. `Qz` an `Uz` are not standard terms.

arusstaylor: I agree.  The labels for the Stokes in the Stokes widget should just be Q, U rather than Qz, Uz.

veggiesaurus: on that note, @panchyo0, we should make sure that the legend displays the colors nicely. In order to do this, we need to set the `backgroundColor` of the plots as well as the border color. I've just added a commit to this branch that does this.

src/components/Shared/LinePlot/PlotContainer/PlotContainerComponent.tsx
panchyo0: just want to make lineColor stay blue like other plots, I will use "hexStringToRgba" for adding opacity.

veggiesaurus: Shouldn’t the line’s RGB color be the same, no matter what the progress is, and only the opacity changed?

panchyo0: will do

veggiesaurus:
This is mostly fine. There's still an issue with how the progress is calculated, but I can't actually think of a proper way of determining this without making larger changes (possibly to the backend as well), because the profiles are delivered one-by-one, so the _old_ Qz profile is still present (with progress @ 100%) while the Uz profile is being delivered. The end result is that progress goes from 0 -> 100% three times

PR #490, opened 2019-08-19
 Fixed #483 : reopen file browser

### LINKED ISSUE #483: hanging file list if we open image file browser without closing region file browser and vice versa
To reproduce the issue, open image file browser, then open region file browser without closing the image file browser dialog. 

Or, open region file browser, then open image file browser without closing the region file browser dialog.

zarda:
Move back getFileList() and only reload file list after opening file browser.

### DIFF
--- a/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
+++ b/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import {observer} from "mobx-react";
-import {computed, observable} from "mobx";
+import {action, computed, observable} from "mobx";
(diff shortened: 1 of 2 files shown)


src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
```diff
@@ -195,7 +195,7 @@ export class FileBrowserDialogComponent extends React.Component<{ appStore: AppS
             lazy: true,
             isOpen: fileBrowserStore.fileBrowserDialogVisible,
             onClose: fileBrowserStore.hideFileBrowser,
-            onOpened: () => fileBrowserStore.getFileList(fileBrowserStore.startingDirectory),
+            onOpened: () => {const _list = fileBrowserStore.fileList; fileBrowserStore.fileList = null; fileBrowserStore.fileList = _list; },
```
veggiesaurus: this should be a private class function instead of an anonymous function. The frontend also doesn't use `_` for any varible names, so rather avoid that and use a descriptive name for it. Maybe `tempList`?

This all looks a bit hacky. `fileBrowserStore.fileList` is an observable, I'd rather we edited it throuh an `@action`

zarda: Fixed it. I will looking into a better solution behind blueprint.

PR #502, opened 2019-08-26
Add check to prevent crash

### LINKED ISSUE #499: Frontend crash when reload page
Using Chrome on MacOS with the asiaa server for the frontend (latest dev), when I restart my local backend then reload the page to reconnect, I see the page as connected, the file browser automatically pops up, then the page goes blank.  I reload the page again, and all is well.  This error is in the dev console output:
Screen Shot 2019-08-21 at 10 11 43 AM

duidae:
quick fix to close #499 

will add the same fix to dev if this is ok


veggiesaurus:
@duidae can you rather open a PR against `dev` for this, and we will backport it once it's done?

duidae:
ok

Issue #371, opened 2019-07-12
Single frame image vanishes when Play is pressed in animator


ajm-asiaa:
If I open a single framed image (for example, aJ.fits) and press Play in the animator, the image vanishes until I press Stop.
On previous versions of CARTA (e.g. the current Master branch), if I press Play in the animator when a single framed image is open, nothing happens; the image remains visible.
I guess the correct behaviour is that nothing should happen.

Would an easy fix be to grey out the Play button (or all animator buttons) if a single framed image is open so that Play just can not be pressed? 
Greyed out animator buttons would double as an extra convenient visual indicator that your image only has a single frame (Although I know when a multichannel image is opened, the Channel slider bar appears so its absence would already indicate a single channel image).

veggiesaurus:
Agreed. At the same time, I would like to combine the play and stop buttons, since there's no reason to have them both visible, and this will free up some valuable real estate.

PR #508, opened 2019-08-26
fixes #496 and #371

### LINKED ISSUE #496: Prevent animation mode from changing during animatoin
This was requested by a few beta testers

veggiesaurus:
- Disables changing animation mode while playing back animations
- Stops animations before opening or appending frame
- Hides Stop button when animation is stopped, hides Play button when animation is playing


kswang1029:
@veggiesaurus tested briefly with multiple images. All work well. 👍

PR #510, opened 2019-09-03
Set default ast grid to false

duidae:


src/stores/PreferenceStore.ts
```diff
@@ -144,7 +144,7 @@ export class PreferenceStore {
 
     private getASTGridVisible = (): boolean => {
         const astGridVisible = localStorage.getItem(PREFERENCE_KEYS.astGridVisible);
-        return astGridVisible === "false" ? false : DEFAULTS.astGridVisible;
```
veggiesaurus: not sure why this line needs to change

duidae: because now default is changed from true to false, so the statement needs to be flipped accordingly, thought about modified to only `return astGridVisible === "true";`, which is better @veggiesaurus  u think? XD

veggiesaurus: if that's the case, then you're not really relying on DEFAULTS.astGridVisible properly... which is not a sensible way to do it. Rather:
`astrGridVisible === "true"? true: (astGridVisible === "false"? false: DEFAULTS.astGridVisible)`. This way, you only resort to the default value if the value is neither "true" nor "false"

veggiesaurus: you might want to make a `parseBoolean(value: string, defaultValue: bool)` utility function for this.

duidae: oh yeah that's a good idea to decouple!  will add it

duidae:
ok added, @veggiesaurus please review

PR #524, opened 2019-09-10
Fixed a warning whenever browsing file

zarda:
A warning:

> [mobx] `observableArray.sort()` will not update the array in place. Use `observableArray.slice().sort()` to supress this warning and perform the operation on a copy, or `observableArray.replace(observableArray.slice().sort())` to sort & update in place

will happen when file browser executes. Fixed the copied data by using .slice() such as the advice from warning.

### DIFF
--- a/src/components/Dialogs/FileBrowser/FileList/FileListComponent.tsx
+++ b/src/components/Dialogs/FileBrowser/FileList/FileListComponent.tsx
@@ -31,8 +31,9 @@ export class FileListComponent extends React.Component<{
         const fileEntries = [];
         const fileList = this.props.files;
         if (fileList) {
-            let sortedDirectories = fileList.subdirectories.slice();
-            if (sortedDirectories) {
+            let sortedDirectories = [];
+            if (fileList.subdirectories && fileList.subdirectories.length) {
+                sortedDirectories = fileList.subdirectories.slice();
                 if (this.state.sortColumn === "name") {
                     sortedDirectories.sort((a, b) => this.state.sortDirection * (a.toLowerCase() < b.toLowerCase() ? -1 : 1));
                 } else {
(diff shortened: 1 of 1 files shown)


kswang1029:
@zarda how to reproduce this? with js console I don't see such warning. Please advice.

zarda:
The waning happens after the frontend receiving the FILE_LIST_RESPONSE and then resort the list on file browser. Opening file browser can reproduce this warning. The devtools needs to turn on Warnings level and there are some other warnings.

src/components/Dialogs/FileBrowser/FileList/FileListComponent.tsx
```diff
@@ -31,12 +31,12 @@ export class FileListComponent extends React.Component<{
         const fileEntries = [];
         const fileList = this.props.files;
         if (fileList) {
-            let sortedDirectories: string[] = [];
-            if (fileList.subdirectories && fileList.subdirectories.length) {
+            let sortedDirectories = fileList.subdirectories.slice();
```
veggiesaurus: this will re-introduce the previous bug, because `fileList.subdirectories` could be undefined (which is why the line above has a check)

src/components/Dialogs/FileBrowser/FileList/FileListComponent.tsx
```diff
@@ -51,19 +51,19 @@ export class FileListComponent extends React.Component<{
                 );
             }));
 
-            let sortedFiles = [];
+            let sortedFiles = fileList.files.slice();
 
-            if (fileList.files && fileList.files.length) {
```
veggiesaurus: same comment as above. `.slice()` of an undefined object is an error

PR #527, opened 2019-09-11
updated packages, migrated to using local packages instead of simple…symlinks for wasm wrappers

veggiesaurus:
This is fairly major update to the package dependencies and how we handle local modules, such as the protobuf definitions and the WebAssembly wrappers. 

- All but `raw-loader` have been updated to their latest version. This does mean that we get those annoying warnings about deprecated react lifecycle methods, but these refer to our dependencies only, not our actual code, so this isn't a problem for now. Hopefully by the next stable release, these will be resolved.
- `raw-loader` has some breaking changes, so I haven't updated this yet, but we should investigate how to adjust to its new loading strategy
- Previously, we were adding symlinks manually during the wrapper build process. This had the major disadvantage that newer version of `npm` would automatically remove these links if you ran `npm install` again, because there wasn't a relevant entry in `package.json`. I've added these entries to `package.json`, so you can now just run `npm install`, followed by `npm run build` to sort everything out.


veggiesaurus:
@duidae I think the build process still needs to be updated to support these changes, so don't merge this into `dev` straight away.

duidae:
ok

Issue #530, opened 2019-09-16
Stokes widget, Information bar display 


panchyo0:
Should we use Q and U, and Q/I and U/I in the information bar at the bottom? @veggiesaurus

Issue #529, opened 2019-09-16
Stokes widget, the shape of scatter plot indicator when interacting with other plots


panchyo0:
Should we just use the highlight circle instead of the grey cross when moving the mouse over the other plots? @veggiesaurus

PR #531, opened 2019-09-16
issues 529 530

panchyo0:
#529 highlight indicator as red, not sure how does that looks like? 
#530


panchyo0:
@veggiesaurus agree, Please check now.

PR #534, opened 2019-09-19
Jason/fix type error

duidae:
in package upgrade branch, tsloader uses typescript 3.6.3 & seems more strict on type check, and then results in crash in this place


duidae:
@veggiesaurus yes & it's done in the drag pan PR, so I didnt add to avoid conflict

veggiesaurus:
> @veggiesaurus yes & it's done in the drag pan PR, so I didn't add to avoid conflict

If its done in the definition, you won't need this change, will you?

duidae:
@veggiesaurus  hmm just a minor reinforcement, could be skipped if not necessary.

PR #536, opened 2019-09-23
Jason/nan value

### LINKED ISSUE #126: [Image viewer]Nan value to be transparent or be able to change alpha or opacity

duidae:
closes #126


duidae:
kk will modify to the suggestions

PR #541, opened 2019-09-30
Qi/stokes zoom interaction

panchyo0:
1.  Scatter plot zoom (click-drag(x and y) and scroll).
2.  Hiding  points in line plots, if point not in the range of scatter plot.
3.  When switching frac pol on or off, widget resets the scatter plot zoom to auto.


src/components/Shared/ScatterPlot/ScatterPlotComponent.tsx
```diff
@@ -382,14 +356,58 @@ export class ScatterPlotComponent extends React.Component<ScatterPlotComponentPr
...
+                const deltaX = (currentPanGraphSpaceX - prevPanGraphSpaceX);
+                const prevPanGraphSpaceY = this.getValueForPixelY(this.panPrevious.y);
+                const currentPanGraphSpaceY = this.getValueForPixelY(currentPanY);
+                const deltaY = (currentPanGraphSpaceY - prevPanGraphSpaceY);
+                this.updatePan(currentPanX, currentPanY);
+                const pinMinX = this.props.xMin - deltaX;
+                const pinMaxX = this.props.xMax - deltaX;
+                const pinMinY = this.props.yMin - deltaY;
+                const pinMaxY = this.props.yMax - deltaY;
+                this.props.graphZoomedXY(pinMinX, pinMaxX, pinMinY, pinMaxY);
```
veggiesaurus: Just check this code, you might have the Y-axis direction mixed up. When you zoom in using the scroll wheel, it zooms centered on the correct X-coordinate, but the wrong Y-coordinate

src/components/Shared/ScatterPlot/ScatterPlotComponent.tsx
```diff
@@ -431,18 +449,127 @@ export class ScatterPlotComponent extends React.Component<ScatterPlotComponentPr
...
+        }
         this.endInteractions();
     };
 
+    onStageWheel = (ev) => {
+        if (this.props.data && this.props.scrollZoom && this.props.graphZoomedXY && this.chartArea) {
+            const wheelEvent: WheelEvent = ev.evt;
+            const chartArea = this.chartArea;
+            const lineHeight = 15;
+            const zoomSpeed = 0.001;
```
veggiesaurus: I think zoom speed should be a user preference. However, we should handle this is a separate PR, and we should make sure it integrates properly with @duidae's changes to the preference handling. 

Can you file an issue for this, so that we can keep track of it? Zoom speed should be adjustable in the user preferences dialog, and should be consistent between the plots and image view

arusstaylor:
This is looking very good.  My one comment is that I think it would be good to dim and gray out the parts of the profile not in the selected portion to give dim trace so one can still see how the highlight parts relate to the unhighlighted parts.

src/components/Shared/ScatterPlot/ScatterPlotComponent.tsx
panchyo0: Thanks, fixed now.

PR #546, opened 2019-10-02
Qi/stokes zoom interaction v2

panchyo0:
1. Scatter plot zoom (click-drag(x and y) and scroll).
2. dim and gray out the line between points in line plots, if point not in the range of scatter plot.
3. When switching frac pol on or off, widget resets the scatter plot zoom to auto.


panchyo0:
@veggiesaurus fixed conflicts. in order to fix the bug (when user zooms in y-axis on line plots. line will draw out of y range.) I changed the zoom default action for y. If points y out of line plot range. Point.y = NaN which will hide the point.

src/components/Shared/LinePlot/PlotContainer/PlotContainerComponent.tsx
```diff
@@ -47,8 +47,57 @@ export class PlotContainerProps {
...
+        function _setColor(newColor: string, plot: any) {
+            plot.dataset._view.borderColor = newColor;
+        }
+
+        if (!colors) {
+            Chart.controllers.line.prototype.draw.call(this, ease);
+            return;
+        }
+
+        for (var i = 2; i <= colors.length; i++) {
```
veggiesaurus: this should be `let`

Issue #513, opened 2019-09-04
Add “jump” or “center” action button in region list to center image view on a region


veggiesaurus:
An action should be added to the region list for each region (placed next to the lock icon), which allows the user to re-center their view on a chosen region. 

- [x] Re-center image view on a specific region's center. 
- [x] For regions other than point regions, adjust the image zoom so that the region takes up a sensible portion of the image view. For example, if we re-center on a 100x100 region, the image view's zoom should be adjusted so that FoV is 300x300.

PR #552, opened 2019-10-07
Add “center” action button in region list

zarda:
Implemented #513, add “center” action button in region list.

### DIFF
--- a/src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
+++ b/src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
@@ -24,7 +24,7 @@ export class RegionDialogComponent extends React.Component<{ appStore: AppStore 
         }
     };
 
-    private handleFocusClicked = () => this.props.appStore.activeFrame.regionSet.selectedRegion.focusCenter(this.props.appStore.activeFrame);
+    private handleFocusClicked = () => this.props.appStore.activeFrame.regionSet.selectedRegion.focusCenter();
 
     public render() {
         const appStore = this.props.appStore;
--- a/src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
+++ b/src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
@@ -24,6 +24,8 @@ export class RegionDialogComponent extends React.Component<{ appStore: AppStore 
         }
     };
 
+    private handleFocusClicked = () => this.props.appStore.activeFrame.regionSet.selectedRegion.focusCenter(this.props.appStore.activeFrame);
+
     public render() {
         const appStore = this.props.appStore;
 
@@ -93,6 +95,17 @@ export class RegionDialogComponent extends React.Component<{ appStore: AppStore 
             }
         }
 
+        let tooltips = region && region.regionId !== 0 && (
+            <React.Fragment>
+                <Tooltip content={`Region is ${region.locked ? "locked" : "unlocked"}`}>
+                    <AnchorButton intent={Intent.WARNING} minimal={true} icon={region.locked ? "lock" : "unlock"} onClick={region.toggleLock}/>
+                </Tooltip>
+                <Tooltip content={"Focus"}>
+                    <AnchorButton intent={Intent.WARNING} minimal={true} icon={"eye-open"} onClick={this.handleFocusClicked}/>
+                </Tooltip>
+            </React.Fragment>
+        );
+
         return (
             <DraggableDialogComponent dialogProps={dialogProps} defaultWidth={600} defaultHeight={450} minHeight={300} minWidth={400} enableResizing={true}>
                 <div className={Classes.DIALOG_BODY}>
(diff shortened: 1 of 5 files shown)


kswang1029:
@zarda can we keep the zoom level, instead of applying a new zoom factor?

kswang1029:
The zoom level needs to be adjusted _only if_ the selected region is too big to be outside the current field of view.

veggiesaurus:
> The zoom level needs to be adjusted _only if_ the selected region is too big to be outside the current field of view.

I agree. Sorry @zarda for telling you the wrong thing...

I think it would also be useful to be able to center on a region when it is open in the region dialog. Due to this, I would suggest you move most of the code you've put in `handleFocusClicked` either to `FrameStore` or `RegionStore`. Once you have done this, can you also add a small icon button (similar to the lock button) to the region dialog as well?

kswang1029:
@zarda thanks for adjusting the features accordingly. Looking at aa905e4, one question: what's the criterion of applying a new zoom level in your implementation? And how the level is determined?

src/stores/RegionStore.ts
```diff
@@ -294,6 +313,17 @@ export class RegionStore {
         }
     };
 
+    @action focusCenter = (activeFrame: FrameStore) => {
+        if (activeFrame) {
+            activeFrame.setCenter(this.center.x, this.center.y);
+            
+            if (activeFrame.renderWidth < activeFrame.zoomLevel * this.boundingBox.x || activeFrame.renderHeight < activeFrame.zoomLevel * this.boundingBox.y) {
```
zarda: If region x/y is larger than width/height of FOV, let reset zoom level.

src/stores/RegionStore.ts
```diff
@@ -294,6 +313,17 @@ export class RegionStore {
...
+                const zoomLevel = FOCUS_REGION_RATIO * Math.min(activeFrame.renderWidth / this.boundingBox.x, activeFrame.renderHeight / this.boundingBox.y);
+                activeFrame.setZoom(zoomLevel);
```
zarda: Reset room level to let region width/height be about a maximum ration FOCUS_REGION_RATIO, which is 40% now.

src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
```diff
@@ -24,6 +24,8 @@ export class RegionDialogComponent extends React.Component<{ appStore: AppStore
         }
     };
 
+    private handleFocusClicked = () => this.props.appStore.activeFrame.regionSet.selectedRegion.focusCenter(this.props.appStore.activeFrame);
```
veggiesaurus: since a region only ever belongs to a single frame, why don't we just add the frame to the region constructor instead, and add a class variable `private frame: FrameStore` to the `RegionStore`? then we don't need to pass this in every time

src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
```diff
@@ -93,18 +95,25 @@ export class RegionDialogComponent extends React.Component<{ appStore: AppStore
             }
         }
 
+        let tooltips = region && region.regionId !== 0 && (
+            <React.Fragment>
+                <Tooltip content={`Region is ${region.locked ? "locked" : "unlocked"}`}>
+                    <AnchorButton intent={Intent.WARNING} minimal={true} icon={region.locked ? "lock" : "unlock"} onClick={region.toggleLock}/>
+                </Tooltip>
+                <Tooltip content={"Focus"}>
+                    <AnchorButton intent={Intent.WARNING} minimal={true} icon={"eye-open"} onClick={this.handleFocusClicked}/>
```
veggiesaurus: this handler should also select the focused region. At the moment, it doesn't center on the correct region, but always on the selected region.

src/components/RegionList/RegionListComponent.tsx
```diff
@@ -22,9 +21,10 @@ export class RegionListComponent extends React.Component<WidgetProps> {
     @observable height: number = 0;
 
     private static readonly LOCK_COLUMN_DEFAULT_WIDTH = 25;
+    private static readonly FOCUS_COLUMN_DEFAULT_WIDTH = 25;
```
veggiesaurus: could we replace these both with `ACTION_COLUMN_DEFAULT_WIDTH`?

src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
zarda: When we select a region on the list, this region should be on the center, shouldn't it?

src/components/RegionList/RegionListComponent.tsx
zarda: Fixed

src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
zarda: Fixed

src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
veggiesaurus: sorry, I was looking at the wrong code. But it seems to be working now :thinking: @kswang1029 do you think clicking on the "center" button in the region list should _also_ select that region?

veggiesaurus:
@zarda some minor merge conflicts with `dev` after @duidae's PR was merged. Can you fix those quickly, and then we can merge this in

src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
kswang1029: Either way should be fine I think. We should consider the case of “active” region in widgets but so far I don’t see issues with it.

kswang1029: Either way should be fine I think. We should consider the case of “active” region in widgets but so far I don’t see issues with it.

Issue #538, opened 2019-09-23
changing coordinate frame of grid overlay


arusstaylor:
Changing the coordinate frame of the grid overlay is too complicated.  It should require only one click.  Currently one must go into the view, selected layouts, select global, select coordinate and then select the coordinate frame.  It should be much easier than this.  The use case for this feature is not that you are customizing overlay settings.  Rather the user wants to be able to dynamically change the grid from equatorial to galactic for instance to check the galactic coordinates of a feature of see how what you are looking at relates to the galaxy.  I suggest we put "coordinate frame" directly in the overlay menu so that is can be changed with one click.

veggiesaurus:
I agree. I think we need to re-think the overlay approach considerably. I see the myriad of options in the complex overlay settings dialog as something that should only be resorted to under two scenarios:
- a user wants to produce a publication-ready plot
- via the scripting interface

Switching between coordinate types should be very easy, and will be especially useful when overlaying contours of a different coordinate type (e.g. overlaying galactic contours on FK5).

@kswang1029 how do you think we should handle this? I also think we should change the nomenclature a bit. Something like coordinate grid overlay?

kswang1029:
How about we borrow the idea from the region button? That is, when the grid button is double clicked, a column of buttons with different coordinate frames shows up (just like regions) so we can switch between different reference frames easier than what we do now.

veggiesaurus:
> How about we borrow the idea from the region button? That is, when the grid button is double clicked, a column of buttons with different coordinate frames shows up (just like regions) so we can switch between different reference frames easier than what we do now.

I like that idea. So something like a button with the following options (tooltip content in brackets):
- Auto (Automatically select the coordinate system based on file headers)
- FK5 (FK5 coordinates, J2000.0 equinox)
- FK4 (FK4 coordinates, B1950.0 equinox)
- GAL (Galactic coordinates)
- ECL (Ecliptic coordinates)
- ICRS (International Celestial Reference System)

veggiesaurus:
@zarda can you have a look at this? it should be relatively simple to add a button group similar to the region button in the image view toolbar, as well as some menu entries for the view menu:

View->Coordinate System->Auto
View->Coordinate System->FK5
...etc

veggiesaurus:
@kswang1029 do we also need entries in the menu?

kswang1029:
[...]

PR #570, opened 2019-10-21
Toolbar selection of grid overlay coordinate

zarda:
Implemented a selection of grid overlay coordinate for issue #538.

### DIFF
--- a/src/components/ImageView/Toolbar/ToolbarComponent.tsx
+++ b/src/components/ImageView/Toolbar/ToolbarComponent.tsx
@@ -4,7 +4,7 @@ import {observer} from "mobx-react";
 import {Button, ButtonGroup, IconName, Menu, MenuItem, Popover, PopoverPosition, Position, Tooltip} from "@blueprintjs/core";
 import {CARTA} from "carta-protobuf";
 import {exportImage} from "components";
-import {AppStore, RegionMode} from "stores";
+import {AppStore, RegionMode, SystemType} from "stores";
 import {toFixed} from "utilities";
 import "./ToolbarComponent.css";
 
@@ -34,6 +34,10 @@ export class ToolbarComponent extends React.Component<ToolbarComponentProps> {
         this.props.appStore.activeFrame.regionSet.setNewRegionType(type);
         this.props.appStore.activeFrame.regionSet.setMode(RegionMode.CREATING);
     };
+
+    handleCoordinateSystemClicked = (coordinateSystem: string) => {
+        this.props.appStore.overlayStore.global.setSystem(coordinateSystem as SystemType);
+    }
 
     render() {
         const appStore = this.props.appStore;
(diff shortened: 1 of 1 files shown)


src/components/ImageView/Toolbar/ToolbarComponent.tsx
```diff
@@ -69,6 +73,14 @@ export class ToolbarComponent extends React.Component<ToolbarComponentProps> {
             </Menu>
         );
 
+        const coordinateSystemMenu = (
+            <Menu>
+                {Object.keys(SystemType).map(key => (
+                    <MenuItem text={key} onClick={() => this.handleCoordinateSystemClicked(key)}/>
```
veggiesaurus: See [here](https://github.com/CARTAvis/carta-frontend/issues/538#issuecomment-534134257), we should put shorthand in, instead of the full names. 

Also, I think we should show the "full" descriptions in the tooltip, using a similar styling approach to the zoom buttons (_Current: 1.0x_)

Given these requirements, I suggest you hard-code the systems menu, or add an additional map object, instead of mapping straight from `SystemType` (although that's a nice elegant way to do so. It's very unlikely that we'll change the system types in the near future.

src/components/ImageView/Toolbar/ToolbarComponent.tsx
```diff
@@ -35,6 +35,10 @@ export class ToolbarComponent extends React.Component<ToolbarComponentProps> {
         this.props.appStore.activeFrame.regionSet.setMode(RegionMode.CREATING);
     };
 
+    handleCoordinateSystemClicked = (coordinateSystem: string) => {
+        this.props.appStore.overlayStore.global.setSystem(coordinateSystem as SystemType);
```
veggiesaurus: this breaks when passing "Native" as the coordinate type, because AST expects an empty string, instead of"Native". 

We should probably also migrate to "Auto" instead of "Native", to be consistent with the rest of the UI's language

src/components/ImageView/Toolbar/ToolbarComponent.tsx
kswang1029: Can we somehow combine the grid button with the reference frame button, similar to the region button?

kswang1029: Can we somehow combine the grid button with the reference frame button, similar to the region button?

src/components/ImageView/Toolbar/ToolbarComponent.tsx
zarda: Fixed

src/components/ImageView/Toolbar/ToolbarComponent.tsx
zarda: Fixed

kswang1029: Ok, let’s try two buttons.

kswang1029: Ok, let’s try two buttons.

zarda: Fixed

Issue #568, opened 2019-10-18
Stokes widget user preferences


panchyo0:
floating settings panel for stokes widget. It will allow user change line and scatter plots settings. like colour, plot style, line wide, WCS value

- [x] stokes setting panel mock up
- [x] floating setting panel which can extend for all widgets
- [x] stokes line settings
- [x] stokes scatter settings

panchyo0:
@veggiesaurus closed with PR #577

PR #577, opened 2019-10-25
Qi/stokes floating options dialog

panchyo0:
#568 : stokes widget user preference.


veggiesaurus:
Another minor change request: The minimum point size for the line plots is 0.5 px. This should be the minimum point size for the scatter plot as well

panchyo0:
@veggiesaurus  Can not reproduce this "There's an issue with the colormap when its first calculated. It seems to be all black until I choose a new colormap. If I actually choose the same one again ("jet"), it works fine as well, not sure where that's cropping up." Please check if this still happens with the new push. Also fixed other minor changes.

src/utilities/color.ts
```diff
@@ -29,4 +33,29 @@ export function hexStringToRgba(colorString: string, alpha: number = 1): RGBA {
         a: alpha
     };
 }
-// end stolen from https://stackoverflow.com/questions/21646738/convert-hex-to-rgba
\ No newline at end of file
+// end stolen from https://stackoverflow.com/questions/21646738/convert-hex-to-rgba
+
+// return color map as Uint8ClampedArray according colorMap
```
veggiesaurus: this comment looks like its in the wrong place

src/utilities/color.ts
```diff
@@ -29,4 +33,29 @@ export function hexStringToRgba(colorString: string, alpha: number = 1): RGBA {
...
+    canvas.height = height; 
+    const ctx = canvas.getContext("2d");
+    return ctx;
+}
+
+export function getColorsForValues (colorMap: string): {color: Uint8ClampedArray, size: number} {
+    const colorMaps = RenderConfigStore.COLOR_MAPS_ALL;
+    const colorMapIndex = colorMaps.indexOf(colorMap);
+
+    // !!!colorMap image size (1024,790)
```
veggiesaurus: this should be re-worded. Perhaps:
`// the source image for colormaps is 1024x790, with each colormap taking a 1024x10 region`

panchyo0: Thanks, fixed now.

Issue #544, opened 2019-10-01
cursor update for profiles stops during zoom


arusstaylor:
When I zoom in on an image,  moving the cursor no longer updates the profiles while the image is waiting for the update from the backend for the new zoom level.  This is very annoying.

arusstaylor:
How about using the unzoomed data still residing on the front end to create profiles while waiting for the new render of zoomed data?

veggiesaurus:
> How about using the unzoomed data still residing on the front end to create approximate profiles while waiting for the new render of zoomed data?

That might go some way to treating the symptom you're seeing, but wouldn't fix the underlying issue (heavy network use, blocking all other received messages). There are a few issues with re-introducing approximate profiles though:

1) They are far more complex to implement with a tiling approach
2) They won't work for spatial profiles of contour images
3) They won't work when we eventually implement HEVC compression

One other option we could explore is requesting tiles more intelligently. At the moment, we request tiles as the user zooms in. This means that during a single "zoom", we might be requesting a large number of tiles that never actually get used. We could instead only request tiles once the user has stopped zooming for some interval (50 ms?). That would mean a slightly longer delay between the user stopping their zoom and the tiles appearing, but the important feature is that the relevant tiles wouldn't be at the end of a long queue of unnecessary tiles

arusstaylor:
The more intelligent tile request sounds like a good thing to do regardless.  Would probably also speed up the responsiveness of the zoom once  it starts.

veggiesaurus:
> The more intelligent tile request sounds like a good thing to do regardless. Would probably also speed up the responsiveness of the zoom once it starts.

Because it wouldn't request tiles until the zoom _ends, two situations will arise:
1) In situations where the network connection is very good, there will be a slightly longer delay between the user zooming and the user seeing the correct tiles
2) In situations where the network connection isn't so good, there will be much less of a delay (than the current implementation) between the user zooming and the user seeing the correct tiles, because all the intermediate layers were not requested

PR #583, opened 2019-10-28
Angus/delay tile requests zooming

### LINKED ISSUE #367: Debounce tile requests
Currently, tiles are requested as soon as they are required. This means that, when zooming in or out, a large number of layers are requested. 

Tiles should only be requested if they are visible for a certain amount of time (200 ms?). 

One way to handle it is to add required tiles to a list, along with a timestamp, and use `setTimeout` with a callback that checks to see if the tiles are still required in the current view.

When using the desktop app or when using CARTA over a good connection, this becomes less necessary, so the timeout should be editable.

veggiesaurus:
Closes #367 and improves the situation describes in #544.

Instead of requesting tiles while the user zooms in, we wait until the zoom level has not changed for a specified amount of time (250 ms) before requesting tiles. This means that if a user zooms continuously  through a number of zoom levels, only the _last_ zoom level will result in tiles being requested. This is a much better approach in situations where:
- the tiles take a long time to calculate (e.g. very large images that need to be downsampled) 
- the tiles take a long time to be delivered (e.g. slow network connection)

The downside of this approach is that higher-resolution tiles do not appear while the user is zooming in. As this may be important to some users, I've added a new preference entry in the "Performance" tab, which allows them to always request tiles by selecting "Stream image tiles while zooming".

zoom_delay_tiles-opt

### DIFF
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -436,6 +436,9 @@ export class PreferenceDialogComponent extends React.Component<{ appStore: AppSt
                         <option key={5} value={1000000}>1M</option>
                     </HTMLSelect>
                 </FormGroup>
+                <FormGroup inline={true} label="Stream image tiles while zooming">
+                    <Switch checked={preference.streamTilesWhileZooming} onChange={(ev) => preference.setStreamContoursWhileZooming(ev.currentTarget.checked)}/>
+                </FormGroup>
             </React.Fragment>
         );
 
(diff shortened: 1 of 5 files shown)


arusstaylor:
Is the latency of the connection something that is known to the front end? Could these options not be set intelligently by the front end based on how good the connection is?

veggiesaurus:
> Is the latency of the connection something that is known to the front end? Could these options not be set intelligently by the front end based on how good the connection is?

The front-end had an approximate knowledge of the latency, but that's only half the story. We have very little idea of the bandwidth. The front-end also doesn't know for sure how long the backend actually takes to crop, downsample and compress the tile

Issue #265, opened 2019-04-16
animator usability enhancement


kswang1029:
referencing https://github.com/CARTAvis/carta-backend/issues/367

Missing features
- [x] playback mode: forward (implemented), backward, bouncing, blinking 
- [x] set start and end ranges
- [x] Adjust step for playback
- [x] display channel index as well as the corresponding frequency and velocity.

veggiesaurus:
#330 implements the start and end playback mechanism and adds the underlying support for most of the remaining functionality, which still needs to be exposed in the UI

veggiesaurus:
the first entry in the list should be implemented before 1.3. @kswang1029 who should we assign this to?

kswang1029:
@duidae Could you look into the first item please?

duidae:
ok

duidae:
it's almost done, however the behavior of play backwards seems incorrect, it does not stop when reaching the first frame and does not loop again when reaching the end, using the message with start frame = last frame, delta = -1, wait for Hengtai to confirm

duidae:
![image](https://user-images.githubusercontent.com/25971696/96689355-deec4800-13b4-11eb-9b75-7e5b3b7e5c8e.png)
added step to UI, btw @kswang1029  would you please elaborate the fourth requirement? thanks.
> display channel index as well as the corresponding frequency and velocity

kswang1029:
@duidae I suggest we replace "Req.... Current....." by the reference frame, frequency and velocity, such as the following mockup.

<img width="930" alt="Screen Shot 2020-10-21 at 4 20 03 PM" src="https://user-images.githubusercontent.com/20819712/96693312-ab5fec80-13b9-11eb-8d28-f3ab5ab28f8b.png">

duidae:
@kswang1029 so we dont have to handle #424 since "Req.... Current....." is no longer needed, right?

kswang1029:
Yes. @veggiesaurus could you confirm this in case I miss something?

kswang1029:
@duidae with origin/jason/issue_265 fbf9ec5, if step=5, we see the sequence is 0, "1", 6, 11, 16, 21. And if we stop and start again, it is "22", 27, 32, ... It seems there is always a single channel step then the step as specified with the spinbox. Could you have a look?

kswang1029:
@duidae @veggiesaurus do you think we can combine the UI of frame rate and step with a dropdown followed by a spinbox? This would save space.

[...]

PR #588, opened 2019-10-31
Jason/issue 265

duidae:
the behavior of reverse animation seems incorrect, however can confirm the UI & bouncing/blink first to see whether it's appropriate

related to #265 

image

### DIFF
--- a/src/components/Animator/AnimatorComponent.scss
+++ b/src/components/Animator/AnimatorComponent.scss
@@ -82,9 +82,12 @@
         width: 40px;
     }
 
+    .playback-mode {
+        margin: 0 3px 0 3px;
+    }
+
     .playback-buttons {
         flex: 3 3 auto;
-        margin-right: 10px;
         height: 20px;
 
         // Re-center button icons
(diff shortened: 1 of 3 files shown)


kswang1029:
@veggiesaurus @duidae For the range slider, since we use radio button to select which to animate (frame, channel, stokes), how about we just the range slider for the selected one? For example, if users want to animate stokes, when the radio button is selected, the range slider shows up right below stokes slider. How about this?

veggiesaurus:
> @veggiesaurus @duidae For the range slider, since we use radio button to select which to animate (frame, channel, stokes), how about we just the range slider for the selected one? For example, if users want to animate stokes, when the radio button is selected, the range slider shows up right below stokes slider. How about this?

That might just work... :+1:

veggiesaurus:
> @veggiesaurus @duidae For the range slider, since we use radio button to select which to animate (frame, channel, stokes), how about we just show the range slider for the selected one? For example, if users want to animate stokes, when the radio button is selected, the range slider shows up right below stokes slider. How about this?

@kswang1029 @duidae I think we should leave this enhancement to a separate issue and PR, because it would require a bit of work. Is that ok? If so, we can leave out blink/bouncing support for frame animations.

kswang1029:
Yes!

duidae:
ok

Issue #495, opened 2019-08-21
Changes to widget options display


veggiesaurus:
Currently, we use a side-bar approach to widget options: commonly used widget options are displayed at the top of the widget, while more options are available by expanding the options bar by pressing the cog:
![image](https://user-images.githubusercontent.com/592504/63413275-aee49600-c3f9-11e9-9a3b-b68c3ad64352.png)
![image](https://user-images.githubusercontent.com/592504/63413185-8066bb00-c3f9-11e9-8d07-f0bc92aa0434.png)

This is feasible for widgets with a small number of options, but:
1) it only works nicely for widgets in landscape mode. In portrait mode, it looks pretty silly (see below). The Stokes widget would also look quite odd with this sidebar
2) we're getting to the stage. where some widgets need a lot more options. For example, the spectral profile widget will eventually have a suite of fitting, smoothing and catalog overlay options. 

![image](https://user-images.githubusercontent.com/592504/63413610-595cb900-c3fa-11e9-974a-818bbb6b8136.png)

Perhaps it's time that we introduce a (floating-only) "options" widget that can be opened by a settings button displayed on the widget's title? The options widget would then be sized appropriately, and contain all the settings for the widget. We then don't have to worry about screen real-estate when adding options. 

@kswang1029 @arusstaylor @low-sky any thoughts?

arusstaylor:
I agree that a floating options widget invoked by a cog wheel in the header bar as a universal approach for widget options would be a better approach. I provides a common approach for all widgets, and frees up real estate within the widget. 

Russ

low-sky:
I agree; a floating window makes sense. The primary thing to watch out for is whether the widget itself is visible while settings are being manipulated.  It's useful to see the effects of interactions with the preferences pane.

veggiesaurus:
> The primary thing to watch out for is whether the widget itself is visible while settings are being manipulated. It's useful to see the effects of interactions with the preferences pane.

I think this is crucial, which is why I'd prefer for the settings to be floating, rather than docked. Floating widgets could easily be rearranged so that the config widget doesn't obscure the actual widget.

I suspect some thought may be required when determining where the config widget actually pops up initially.

panchyo0:
@veggiesaurus since we have a floating settings panel for stokes widgets after #577. Should we start to apply this to all widgets?

veggiesaurus:
@panchyo0 yes, that's a good idea. Lets start with the spectral profiler next

[...]

Issue #413, opened 2019-07-29
highlight value = 0.0 grid line in profiles


arusstaylor:
Have been using the dev viewer to look at some recent MeerKAT MIGTHEE polarization cubes.   Here is a shot of the viewer with Stokes U and Q profiles widgets. Of interest in these profiles is to assess if the signal is significant different from noise around zero. It would be very nice if the grid line at 0.0 were highlighted some way (thicker or darker line).   

![Screen Shot 2019-07-29 at 1 23 26 PM](https://user-images.githubusercontent.com/9055533/62045703-58bf7100-b206-11e9-8588-10fd3e606659.png)

Kechil:
needs to be consistent for all widgets

panchyo0:
closed with PR #592

PR #592, opened 2019-11-01
Qi/spectral floating settings panel

panchyo0:
1. added LinePlotSettingsPanelComponentProps and LinePlotSettingsPanelComponent for line plots settings.
2. added ScatterPlotSettingsPanelComponentProps and ScatterPlotSettingsPanelComponent for scatter plots settings.
3. removed pop up settings panel for histogram, spatial profiler, spectral profiler and render config.
4. added scss margin-right = 5px for Spectral Profiler and histogram .profile-container.
5. added scss overflow = auto for floating settings widgets which will allow scroll for overflow.
6. changed the floating settings widgets id format from "${defaultId}-${nextIndex}" to "${parentId}-${defaultId}-${nextIndex}"
7. changed the default stokes widget size.
8. apply cog button for histogram, spatial profiler, spectral profiler, render config and stokes.
9. hide cog button for unnecessary widgets with golden layout.
10. finished #495 
11. finished #413 by making value =0.0 grid line more thicker.


kswang1029:
We can have multiple spectral profile widgets and thus each can have a floating settings panel (this actually applies to all widgets and their settings panels). We will need a way to let users know the pairs visually.

veggiesaurus:
> We can have multiple spectral profile widgets and thus each can have a floating settings panel (this actually applies to all widgets and their settings panels). We will need a way to let users know the pairs visually.

I agree. Is matching the title sufficient? Perhaps a `*` in the title of the profiler when the settings widget is focused?

panchyo0:
@veggiesaurus Currently spectral settings panel only match the region for title. since the settings panel will allow user custom the line plot for entire widget(stokes Q, U, I, V and Current). Should we add stokes parameter in title? When user change the size of the spectral widget to min, the title UI starts at another line. Should we add "overflow=auto" to solve it?
Screen Shot 2019-11-04 at 10 50 42 AM

veggiesaurus:
> @veggiesaurus Currently spectral settings panel only match the region for title. since the settings panel will allow user custom the line plot for entire widget(stokes Q, U, I, V and Current). Should we add stokes parameter in title? When user change the size of the spectral widget to min, the title UI starts at another line. Should we add "overflow=auto" to solve it?
> Screen Shot 2019-11-04 at 10 50 42 AM

I'd prefer not to change the widget title height, but am open to the idea if it looks ok

PR #599, opened 2019-11-07
Quick fixed #594 and #595

### LINKED ISSUE #594: misalignment in the region list widget
Screen Shot 2019-11-06 at 9 36 42 AM
Screen Shot 2019-11-06 at 9 36 32 AM

### LINKED ISSUE #595: double click on region lock or region to view center
When no region is selected, double click on the region-lock button or region-to-view-center button results in a region setting dialogue saying "no region selected". It would be good to improve this for UX. It would be ideal to ignore double click for these two buttons if possible.

zarda:
Fix #594 by attached a blink icon to avoid the alignment changes.
Fix #595 by hard-code the double click action on table except of icons

### DIFF
--- a/src/components/RegionList/RegionListComponent.tsx
+++ b/src/components/RegionList/RegionListComponent.tsx
@@ -131,14 +131,14 @@ export class RegionListComponent extends React.Component<WidgetProps> {
             if (region.regionId !== 0) {
                 lockEntry = <td style={{width: RegionListComponent.ACTION_COLUMN_DEFAULT_WIDTH}} onClick={(ev) => this.handleRegionLockClicked(ev, region)}><Icon icon={region.locked ? "lock" : "unlock"}/></td>;
             } else {
-                lockEntry = <td style={{width: RegionListComponent.ACTION_COLUMN_DEFAULT_WIDTH}}/>;
+                lockEntry = <td style={{width: RegionListComponent.ACTION_COLUMN_DEFAULT_WIDTH}}><Icon icon={"blank"}/></td>;
             }
 
             let focusEntry: React.ReactNode;
             if (region.regionId) {
                 focusEntry = <td style={{width: RegionListComponent.ACTION_COLUMN_DEFAULT_WIDTH}} onClick={(ev) => this.handleFocusClicked(ev, region)}><Icon icon={"eye-open"}/></td>;
             } else {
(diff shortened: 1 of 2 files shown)


src/components/RegionList/RegionListComponent.tsx
```diff
@@ -80,7 +80,8 @@ export class RegionListComponent extends React.Component<WidgetProps> {
 
         let nameWidth = RegionListComponent.NAME_COLUMN_DEFAULT_WIDTH;
         const availableWidth = this.width - 2 * padding;
-        let fixedWidth = RegionListComponent.TYPE_COLUMN_DEFAULT_WIDTH + RegionListComponent.CENTER_COLUMN_DEFAULT_WIDTH + RegionListComponent.SIZE_COLUMN_DEFAULT_WIDTH + RegionListComponent.ROTATION_COLUMN_DEFAULT_WIDTH;
+        // tslint:disable-next-line: max-line-length
```
veggiesaurus: why not just put this on two lines, instead of disabling tslint for this line?

zarda: Fixed

veggiesaurus:
@zarda the table headers are still mis-aligned (Chrome on Macbook 15")

Screenshot 2019-11-21 at 19 52 10

Also, do you think we should combine the two "action" columns into one? The separator between them looks a bit weird to me. @kswang1029 what do you think?

kswang1029:
@veggiesaurus okay to have just one column for all action icons.

zarda:
A separator is good to UX, either a column line or  a distance between two icons. 
Current mouse action is on table rather than on icon. If a column separator is no good, we could apply the mouse action on a Button with some icon and the button outline should be removed.

veggiesaurus:
> A separator is good to UX, either a column line or a distance between two icons.
> Current mouse action is on table rather than on icon. If a column separator is no good, we could apply the mouse action on a Button with some icon and the button outline should be removed.

Fine to keep the separator columns as you've done them now. Next question: Should we just add the two actions (lock and center) to the cursor as well? The lock icon applied to the cursor region could be used to toggle frozen/unfrozen cursor, and there's no reason why you shouldn't be able to center on the cursor region, in my mind

veggiesaurus:
Approving this and adding separate question issue (#610)

Issue #266, opened 2019-04-16
chart usability enhancement


kswang1029:
Allowing users to customize the displayed x range and y range by text input.
Charts include
- render config widget
- spatial profiler widget
- spectral profiler widget
- histogram

panchyo0:
@kswang1029 can you provide more details? for example the should we match exactly the max and min number for x and y range according data? should we allow the min text input larger than max text input? what accuracy we should use for user input data? Thanks

kswang1029:
I suggest we add a “custom range” toggle and provide text fields as user input in the setting dialogue. The field should provide basic validations (numbers only) and min has to be smaller than max. We can trim user input as the following:

100000.1234
10000.1234
1000.1234
100.1234
10.1234
1.1234
0.1234
0.01234
0.001234
0.0001234

panchyo0:
closed with PR #611

PR #611, opened 2019-11-26
Qi/chart usability enhancement 266

panchyo0:
chart usability enhancement #266


src/utilities/parsing.ts
```diff
@@ -6,4 +6,12 @@ export function parseBoolean(value: string, defaultValue: boolean): boolean {
     } else {
         return defaultValue;
     }
+}
+
+export function parseUndefinedValue(val: number, initVal: number): number {
```
veggiesaurus: I don't think `parseUndefinedValue` is a helpful name. Perhaps `parseNumber`? Also, wouldn't it be better to use `isfinite` instead of checking the type? For example, if `val` is a `string`, this would return the string instead of a number.

src/components/SpectralProfiler/SpectralProfilerSettingsPanelComponent/SpectralProfilerSettingsPanelComponent.tsx
```diff
@@ -79,6 +82,70 @@ export class SpectralProfilerSettingsPanelComponent extends React.Component<Widg
         this.widgetStore.setUseWcsValues(changeEvent.target.checked);
     };
 
+    handleXMinChange = (ev) => {
```
veggiesaurus: could we add explicit types to these `ev` values? It might make maintenance of the functions easier

src/components/SpectralProfiler/SpectralProfilerSettingsPanelComponent/SpectralProfilerSettingsPanelComponent.tsx
```diff
@@ -79,6 +82,70 @@ export class SpectralProfilerSettingsPanelComponent extends React.Component<Widg
...
+        const minX = parseUndefinedValue(widgetStore.minX, widgetStore.linePlotInitXYBoundaries.minXVal);
+        const maxX = parseUndefinedValue(widgetStore.maxX, widgetStore.linePlotInitXYBoundaries.maxXVal);
+        if (isFinite(val) && val !== minX && val < maxX) {
+            widgetStore.setXBounds(val, maxX);
+        } else {
+            ev.currentTarget.value = minX;
+        }
+    };
+
+    handleXMaxChange = (ev) => {
```
veggiesaurus: there's a lot of repetition between these functions. I wonder if it's possible to improve that by extracting a function that takes an event and some parameters? Perhaps that would be needlessly complicated, so treat this as purely a suggestion to investigate

veggiesaurus (changes requested):
some small changes requested, but otherwise it looks good :+1:

veggiesaurus (approved):

PR #616, opened 2019-11-28
Jason/beam overlay

### LINKED ISSUE #118: Add settings for beam profile overlay
The user should be able to change the appearance (or toggle) the beam profile overlay including

- [X] color
- [X] open or solid or (hatched; good to have) 
- [X] line width for open
- [X] position adjustment (especially for overlaying images with different beam sizes)

duidae:
closes #118

### DIFF
--- a/src/components/Dialogs/OverlaySettings/OverlaySettingsDialogComponent.tsx
+++ b/src/components/Dialogs/OverlaySettings/OverlaySettingsDialogComponent.tsx
@@ -564,9 +564,9 @@ export class OverlaySettingsDialogComponent extends React.Component<{ appStore: 
             <div className="panel-container">
                 <FormGroup inline={true} label="Frame">
                     <HTMLSelect
-                        options={[...beam.frameNames, "Default"]}
-                        value={beam.selectedFrame}
-                        onChange={(event: React.FormEvent<HTMLSelectElement>) => beam.setSelectedFrame(event.currentTarget.value as string)}
+                        options={beam.frameNames}
+                        value={beam.selectedFileId}
+                        onChange={(event: React.FormEvent<HTMLSelectElement>) => beam.setSelectedFrame(parseInt(event.currentTarget.value))}
                     />
                 </FormGroup>
                 <FormGroup inline={true} label="Visible">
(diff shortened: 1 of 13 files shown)


veggiesaurus (changes requested):
The use of the "frame" dropdown seems very inconsistent with how we've handled other default values in the overlay. 

@kswang1029 what's the rationale behind being able to change the position of the beam indicator? Would it not be more helpful to just choose which corner/edge it's pinned to? As soon as you resize the image view (or the CARTA window itself), the offsets are likely to be less useful.

Perhaps we only need visibility, display type and color as defaults (in the preferences dialog)?

kswang1029:
That’s for the preparation of displaying multiple beam elements for wcs group. By default all beams should be pinned to the same location. But in some other cases, users would like to apply offsets to each so that it is more readable.

veggiesaurus:
> That’s for the preparation of displaying multiple beam elements for wcs group. By default all beams should be pinned to the same location. But in some other cases, users would like to apply offsets to each so that it is more readable

It seems quite out of place. If I select "default", the beam doesn't move at sll

duidae:
> It seems quite out of place. If I select "default", the beam doesn't move at sll

each frame will have it's own beam setting, and the Default in the drop down menu is for the newly loaded frame to apply, could be removed if its redundant (or I think Default can be moved to preference)

veggiesaurus:
> > It seems quite out of place. If I select "default", the beam doesn't move at sll
> 
> each frame will have it's own beam setting, and the Default in the drop down menu is for the newly loaded frame to apply, could be removed if its redundant (or Default can be moved to preference)

I think moving it to preferences (at least just the colour, visibility and style) makes sense, that would be consistent with the rest of the UI. This seems like something the users are going to get tripped up on, though. 

One other minor issue: the default colour should be one of the colours in the swatch of the colour picker

duidae:
kk will modified to:
1. move Default to preference, including only visible, color, type, width
2. change default color to gray 3 in swatch color

veggiesaurus (approved):
Approved after adding a minor spacing fix (changed `Position(X)` to `Position (X)`.

PR #617, opened 2019-11-28
Angus/session resume

veggiesaurus:
Note: this is a companion PR to [carta-backend/#375](https://github.com/CARTAvis/carta-backend/pull/375), and the two should be merged simultaneously (along with the ICD changes).

Initial resume support when the user's network is disconnected, or the backend crashes and is restarted.

Some work needs to be done to handle resuming sessions when running in authenticated mode, but this can happen at a later stage, once token-based authentication is supported.


duidae (approved):

kswang1029:
@veggiesaurus It happened to me that I woke my laptop up and was asked to resume my CARTA session or not. I clicked yes but the actual backend has been time out so I see the resuming dialogue persist. Should we add another dialogue to warn users that a session cannot be resumed due to backend time out after some time (15s)?

PR #622, opened 2019-12-03
Fix issue #405

### LINKED ISSUE #405: dropdown indicator is not visible in "light" theme
The dropdown indicator that is used when there are too many tabs in a stack is not visible in the light theme:

image

In dark theme, it's visible, although not very noticeable: 
image

`.lm_header .lm_controls .lm_tabdropdown:before` needs to be edited in the Golden Layout theme CSS

TienHao:
Closes #405 
Root cause is that "layout-theme.css" was imported before "layout-base.css", causing light theme tab color(black) was overridden by base tab color(white).

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -49,20 +49,3 @@ body {
     z-index: 999 !important;
 }
 
-.pin-icon {
-    float: left;
-    cursor: pointer;
-    color: $gray3;
-    &:hover {
-        color: black;
-    }
-}
-
-.cog-pined-icon {
-    float: left;
-    cursor: pointer;
-    color: $gray3;
-    &:hover {
-        color: black;
-    }
-}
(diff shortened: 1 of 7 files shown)


veggiesaurus:
the code fix looks fine, adding @kswang1029 as a reviewer to ensure that the issue is fixed and that there are no clearly visible side effects.

veggiesaurus:
@TienHao, here is the reply from @kswang1029 on the issue: 

>Screen Shot 2019-12-03 at 22 06 04
>Screen Shot 2019-12-03 at 22 05 51
>
>Yes, now with light theme, it is visible and clear. With dark theme, it can be improved by using a lighter color identical to the "pin" perhaps? In fact, shall we use same color for all icons?
>
>For light theme, we use the color identical to box and triangle icons. For dark theme, we use the color identical to the pin.

veggiesaurus:
with the latest commit, I think these changes look good. @kswang1029 are you happy with them?

kswang1029:
@veggiesaurus Yes!

PR #623, opened 2019-12-04
Confluence/downsampled region dims rounding fix

confluence:
This is a fix for the off-by-one rounding error in the downsampled tile dimensions. It should be merged at the same time as [the corresponding back-end PR](https://github.com/CARTAvis/carta-backend/pull/386). This needs more thorough testing.

### DIFF
--- a/src/components/ImageView/RasterView/RasterViewComponent.tsx
+++ b/src/components/ImageView/RasterView/RasterViewComponent.tsx
@@ -149,8 +149,8 @@ export class RasterViewComponent extends React.Component<RasterViewComponentProp
 
     private updateTexture() {
         const frame = this.props.frame;
-        const w = Math.floor((frame.currentFrameView.xMax - frame.currentFrameView.xMin) / frame.currentFrameView.mip);
-        const h = Math.floor((frame.currentFrameView.yMax - frame.currentFrameView.yMin) / frame.currentFrameView.mip);
+        const w = Math.ceil((frame.currentFrameView.xMax - frame.currentFrameView.xMin) / frame.currentFrameView.mip);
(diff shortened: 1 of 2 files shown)


veggiesaurus:
@kswang1029 could you give this a test with the associated backend branch on mix 'n match?

confluence:
@kswang1029 To clarify, the intended correct behaviour is for the dimensions to be rounded up after division by the mip factor (they are currently truncated).

kswang1029:
I will do the test during the weekend or so since I need to be in the meetings for the next two full days. @_@

veggiesaurus:
@confluence to demonstrate this issue (and its fix) to @kswang1029, can you do the following: 

- make a 1023x1023 image of all zeros, except for the last column, which has all non-zero values
- view the image using `dev` branch. When not viewing at full resolution, this should trigger the issue, and the tiles will be constructed without the last column's data.

kswang1029:
@veggiesaurus thanks for the heads-up. Now I understand the issue better and can design tests accordingly. : )

confluence:
Here's the test file I made and screenshots of my own test -- as you can see, when you zoom out in dev the stripe on the right disappears; in the branch it remains visible.

[tiletest.zip](https://github.com/CARTAvis/carta-frontend/files/3925792/tiletest.zip)

dev

branch

kswang1029:
@confluence @veggiesaurus I think the fix works. Feel free to try with the images at set_downsampleTest with http://carta.asiaa.sinica.edu.tw/frontend/dev/?socketUrl=wss://carta.asiaa.sinica.edu.tw/socket5034

before:
image_1023_1023_v2 fits-image-2019-12-11-16-11-38
after:
image_1023_1023_v2 fits-image-2019-12-11-16-11-17

veggiesaurus (approved):

PR #624, opened 2019-12-04
Angus/low bandwidth mode

### LINKED ISSUE #603: Low-bandwidth mode
We need to include a low-bandwidth mode that users can enable in the preferences dialog when they are running on a poor (high latency, unreliable or limited bandwidth) network
- [x] Some testing of how much data could be saved
- [x] Add mode toggle in performance section of preferences dialog
- [x] Add low-bandwidth indicator next to connection indicator
- [x] Reduce spatial, spectral and cursor value update speeds to 2 FPS when running in low-bandwidth mode
- [x] Apply layer shift when requesting tiles, in order to reduce data requested by the frontend by a factor of 4 (at least in terms of uncompressed data. Compressed savings will be a bit smaller due to SnR changes between layers)
- [x] Apply mip adjustment during animation, in order to reduce data requested by frontend

veggiesaurus:
closes #603 

This PR adds a low bandwidth mode, as described in #603. One minor additional change is that the connectivity icon's tooltip has been changed to match other tooltips' formatting.

### DIFF
--- a/src/stores/AppStore.ts
+++ b/src/stores/AppStore.ts
@@ -1,6 +1,7 @@
 import * as _ from "lodash";
 import * as AST from "ast_wrapper";
 import {action, autorun, computed, observable, ObservableMap} from "mobx";
+import {IOptionProps} from "@blueprintjs/core";
 import {CARTA} from "carta-protobuf";
 import {
     AlertStore,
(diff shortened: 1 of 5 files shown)


src/stores/AppStore.ts
```diff
@@ -470,6 +464,12 @@ export class AppStore {
     };
 
     public static readonly DEFAULT_STATS_TYPES = [CARTA.StatsType.NumPixels, CARTA.StatsType.Sum, CARTA.StatsType.Mean, CARTA.StatsType.RMS, CARTA.StatsType.Sigma, CARTA.StatsType.SumSq, CARTA.StatsType.Min, CARTA.StatsType.Max];
+    private static readonly CursorThrottleTime = 200;
```
duidae, line 467: just a minor issue: constants better be capitalized. Btw, what is the reason to move those constants to private members?

veggiesaurus: according to the AirBnB style guide, we should only do _exported_ consts in `ALL_CAPS`. I know we're not consistent about this, but perhaps it's better to work toward that. 

We currently use `PascalCase` instead of `camelCase` for a lot of other `static readonly` variables, so I've tried to be consistent with that. Perhaps this needs a bit more thought though. One solution which is somewhat consistent is to use `ALL_CAPS` for `public static readonly`, since those behave like `export const` values, and `camelCase` (or `PascalCase` if we don't want to adjust existing code too much) for `private static readonly`

It would be good to get opinions on that from @panchyo0 as well. 

As for why I've moved the const to private members: they're not used outside of the class, so I thought it would be better to keep them in the class as private static values.

panchyo0: "consistent is to use ALL_CAPS for public static readonly, since those behave like export const values, and PascalCase for private static readonly." This one might better for our next release.

veggiesaurus: > This one might better for our next release.

Ok, I'll make an issue for this and tag it for 1.3. It should be quite easy to fix with an IDE

PR #626, opened 2019-12-05
Fix #443 unable to create polygon if clicking too fast

### LINKED ISSUE #443: cannot create a polygon if users click too fast
I think this is an exceptional case but that may happen.
Aug-06-2019 09-57-15

zarda:
Fix #443 by recording the click position to compute the click distance, which will disable double click while the distance is too long. The distance threshold is 5 px. After this fixing, the polygon creation can be fast and neat.

### DIFF
--- a/src/components/ImageView/RegionView/RegionViewComponent.tsx
+++ b/src/components/ImageView/RegionView/RegionViewComponent.tsx
@@ -11,7 +11,7 @@ import {PolygonRegionComponent} from "./PolygonRegionComponent";
 import {PointRegionComponent} from "./PointRegionComponent";
 import {CursorInfo, Point2D} from "models";
 import "./RegionViewComponent.css";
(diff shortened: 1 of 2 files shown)


veggiesaurus (changes requested):
I think it makes more sense to compare the distance, rather than the x or y component of the distance (i.e. a proper radial distance comparison rather than rectangular)

veggiesaurus (approved):
thanks for fixing this. Approving after a couple of minor changes:
- those two new properties (`mousePreviousClick` and `mouseClickDistance`) don't need to be `@observable`, since they don't need to trigger a React re-render.
- The function @panchyo0 wrote to calculate distance actually calculates the _square_ of the distance, so I've renamed the function accordingly and changed the comparison to `this.mouseClickDistance > DOUBLE_CLICK_DISTANCE * DOUBLE_CLICK_DISTANCE`

PR #627, opened 2019-12-05
Jason/splash screen

### LINKED ISSUE #572: splash screen when initializing frontend
we keep a splash screen displayed until all the basic components (AST, ZFP, etc) have been loaded by the frontend. This would prevent users from loading an image while one of the libraries is still loading

- [X] UI design
- [X] wire up check points

duidae:
closes #572


veggiesaurus (changes requested):
This all looks great. One very minor change: It really bothers me that the logo image pops in while the splash screen is displayed, and then re-flows the rest of the splash screen.

We have two options: 
1. somehow pre-cache the logo file, so that its there already when the splash screen becomes visible
2. Display a placeholder div with the correct size (300x350) so that the reflow doesn't occur

(1) might be tricky, I suggest we go for (2) instead. It might be as simple as wrapping the image element in a fixed-size `div` :thinking:

veggiesaurus (approved):

PR #629, opened 2019-12-05
fixed stokes line plot bugs with chartjs 2.9.3

panchyo0:
1. fixed stokes widget bug (line plots draw line out of chart area) under chartjs 2.9.3
2. fixed stokes widget legend color disappear bug when user choose point style under chartjs 2.9.3


veggiesaurus (approved):
I updated the package versions in `package.json` so that we can use 2.9.3 by default. Thanks for sorting this out, 2.9.3 seems much faster than

Issue #325, opened 2019-05-09
file size column width not big enough in file browser


arusstaylor:
The width of the column that displays files size in the file browser needs to be wider.  The field text is wrapping instead of adjusting to fit the width of the text.

PR #635, opened 2019-12-11
Fix file size column width not big enough(#325)

TienHao:
Add TB into file size column (column is not enough when the file size is over 1000 GB). Set TB with 2 decimals for clear information.


kswang1029:
In some cases (with origin/thchang/modify_file_size_display_325 08069d6 2019-12-11 18:28:05 CST), the width is still not enough. Could you check again?
Screen Shot 2019-12-11 at 21 27 47

TienHao:
I fix this with html style.
It seems that different browsers(Safari) displayed differently with my previous solution. I have only checked chrome before. Now it should fits all browsers.

PR #638, opened 2019-12-16
Jason/issue 632

### LINKED ISSUE #632: Remove comment from unit and frame
The backend now creates standard FITS headers for all images, including comments for e.g. BUNIT and SPECSYS.  These comments should be removed before being displayed in the various widgets.

duidae:
test file: M17_SWex.fits
result: label of vertical axis in X, Y profiler will show no comments 

closes #632


kswang1029:
Looking good👍

veggiesaurus:
This looks fine to me, as long as its safe to assume the actual header value itself wont contain the regex match (space, followed by `/`). @kswang1029 is this a safe assumption?

e.g. `Jy/beam / This is a comment` works fine, but `Jy / beam / This is also a comment` doesn't work so well

kswang1029:
now we use casacore to unify the format of headers from different image formats. So I think the BUNIT format is fixed and it should be safe to use " / " to delimit comments.

PR #640, opened 2019-12-18
Add image info dialog (#417)

### LINKED ISSUE #417: view file information or image header
We should be able to view the image header of the current frame.  One way to do this is to add  "image header" or "image info"  sub menu to the "view" menu.  This would bring up a text widget containing the image header. 

Alternately one could add an info panel that contains the file information.  Not sure which is more useful the file information or the header.  The header is more complete, but the file info is more readable.

TienHao:
closes #417 

* Add image info dialog component

* Add image info dialog css

* Add image info store

* Add "Image Info" in submenu "View"


duidae:
1. issue - when appending multiple frames, the image info should show the info of the active frame
2. instead of saving the lastest in the imageinfo store, i think its better to wrap the image info into frame store as a subsidiary attribute, so that it will be easier to maintain when adding/deleting a frame
3. btw, @veggiesaurus  @kswang1029  will it be better to also have an info button in the tool bar?
image

veggiesaurus:
We already have the contents of the image info in frame (in the extended file info). No need to duplicate it, just use that instead.

@duidae I'd prefer not to add it to the toolbar, it's already getting quite long. I think we should have a separate "dialogs" toolbar at the top near the widgets. We can have buttons for preferences, file info, overlay config, contour config and so on.

src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
```diff
@@ -30,6 +30,10 @@ export class FileBrowserDialogComponent extends React.Component<{ appStore: AppS
         }
 
         if (fileBrowserStore.browserMode === BrowserMode.File) {
+            const imageInfoStore = this.props.appStore.imageInfoStore;
```
veggiesaurus: reuse `appStore.activeFrame.frameInfo.fileInfoExtended` instead, no need to copy it here

src/components/Dialogs/ImageInfo/ImageInfoDialogComponent.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+@observer
+export class ImageInfoDialogComponent extends React.Component<{ appStore: AppStore }> {
+
+    private handleTabChange = (newId: TabId) => {
+        this.props.appStore.imageInfoStore.setSelectedTab(newId);
+    };
+
+    private renderInfoPanel = () => {
+        const imageInfoStore = this.props.appStore.imageInfoStore;
```
veggiesaurus: reuse `appStore.activeFrame.frameInfo.fileInfoExtended` for these values

veggiesaurus (changes requested):
mostly looks ok, but we should reuse information in `FrameStore` instead of duplicating. This will also fix the bug @duidae found.

src/components/Dialogs/ImageInfo/ImageInfoDialogComponent.scss
```diff
@@ -0,0 +1,38 @@
...
+        .bp3-input-group {
+            margin: 10px 5px 5px;
+        }
+
+        .image-info-panes {
+            display: flex;
+            height: 100%;
+            
+            .image-info-pane {
+                width: 95%;
```
veggiesaurus: I don't think these margins are correct. Rather just use a (much smaller) fixed margin and use CSS `calc` to work out the width. I'm not sure why we need a 40px `margin-bottom`.

src/components/Dialogs/ImageInfo/ImageInfoDialogComponent.scss
```diff
@@ -0,0 +1,38 @@
...
+    height: 100%;
+    user-select: none;
+
+    .bp3-dialog-body {
+        height: 100%;
+        overflow: hidden;
+        margin-top: 10px;
+        margin-left: 25px;
+
+        .bp3-input-group {
```
veggiesaurus: is this used?

src/components/Dialogs/ImageInfo/ImageInfoDialogComponent.scss
```diff
@@ -0,0 +1,38 @@
+.image-info-dialog {
+    min-width: 60%;
+    height: 100%;
+    user-select: none;
+
+    .bp3-dialog-body {
+        height: 100%;
+        overflow: hidden;
+        margin-top: 10px;
+        margin-left: 25px;
```
veggiesaurus: don't think we need such a large left margin. In any case, it should be symmetric

src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
TienHao: OK, I will remove ImageInfoStore. It is much simple to implement image info dialog with FrameStore.

src/components/Dialogs/ImageInfo/ImageInfoDialogComponent.scss
TienHao: This is a legacy from FileBrowserComponent. I will adjust it.

src/components/Dialogs/ImageInfo/ImageInfoDialogComponent.scss
TienHao: No, it is not used. This is also a legacy from FileBrowserComponent. I will remove it.

src/components/Dialogs/ImageInfo/ImageInfoDialogComponent.tsx
TienHao: I will remove all ImageInfoStore code.

src/components/Dialogs/ImageInfo/ImageInfoDialogComponent.scss
TienHao: The 40px margin-bottom is for fully seeing the last line of headers information.

src/components/Dialogs/FileInfoDialog/FileInfoDialogComponent.tsx
```diff
@@ -0,0 +1,56 @@
...
+import {DraggableDialogComponent} from "components/Dialogs";
+import {FileInfoComponent, InfoType} from "components/FileInfo/FileInfoComponent";
+import {CARTA} from "carta-protobuf";
+import {AppStore} from "stores";
+import "./FileInfoDialogComponent.css";
+
+@observer
+export class FileInfoDialogComponent extends React.Component<{ appStore: AppStore }> {
+
+    @observable fileInfoExtended: CARTA.FileInfoExtended;
```
duidae: unused variable?

duidae (approved):

src/App.tsx
```diff
@@ -132,6 +133,7 @@ export class App extends React.Component<{ appStore: AppStore }> {
                         {this.renderHotkeys()}
                     </div>
                 </Dialog>
+                <FileInfoDialogComponent appStore={appStore}/>
```
veggiesaurus: I suggest you move this up where the rest of the simple dialogs are

veggiesaurus (changes requested):

src/App.tsx
TienHao: OK

veggiesaurus (approved):

PR #657, opened 2019-12-23
Jason/json schema

### LINKED ISSUE #492: Integrating JSON validator  & support server layout
- [X] integrate ajv
- [X] create json template
- [X] support widget settings of spatial/spectral, stokes, historgram, render config

ajv lib ref: https://github.com/epoberezkin/ajv

duidae:
closes #492: Support layout part for server.

this is a companion PR to https://github.com/CARTAvis/carta-backend/pull/407, and the two should be merged simultaneously.

Note: the preference part of frontend will be in the next PR


src/stores/LayoutStore.ts
```diff
@@ -118,13 +117,15 @@ export class LayoutStore {
...
     @observable private layouts: any;
+    @observable supportServer: boolean;
 
     constructor(appStore: AppStore, alertStore: AlertStore) {
         this.appStore = appStore;
         this.alertStore = alertStore;
         this.dockedLayout = null;
         this.layouts = {};
-        this.initLayouts();
+        this.supportServer = false;
```
veggiesaurus: Minor typo, I think `supportsServer` would make more sense

src/models/LayoutSchema.ts
```diff
@@ -0,0 +1,42 @@
...
+                            "type": "array",
+                            "items": {
+                                "type": "object"
+                            }
+                        }
+                    }
+                },
+                "floating": {
+                    "type": "array",
+                    "items": {
```
veggiesaurus: Do you think we should be stricter with the types of widgets we can store in the schema?

@kswang1029 Currently the schema only specifies layout properties, rather than properties of the widgets themself. Should we store things like plot colours, plot types, etc?

veggiesaurus (changes requested):

src/models/LayoutSchema.ts
duidae: trying to add the new widget properties to layout now

src/stores/widgets/HistogramWidgetStore.ts
```diff
@@ -153,4 +154,52 @@ export class HistogramWidgetStore extends RegionWidgetStore {
     @action initXYBoundaries (minXVal: number, maxXVal: number, minYVal: number, maxYVal: number) {
         this.linePlotInitXYBoundaries = { minXVal: minXVal, maxXVal: maxXVal, minYVal: minYVal, maxYVal: maxYVal };
     }
+
+    public init = (widgetSettings): void => {
+        if (!widgetSettings) {
+            return;
+        }
+        if ("primaryLineColor" in widgetSettings && typeof widgetSettings.primaryLineColor === "string" && isColorValid(widgetSettings.primaryLineColor)) {
```
veggiesaurus: (this applies to all similar logical tests)

if `widgetSettings.primaryLineColor` is undefined, then `typoeof widgetsettings.primaryLineColor` will be "undefined", so you can just use the second test (`typeof widgetSettings.primaryLineColor === "striing"`)

veggiesaurus (changes requested):

veggiesaurus:
Code looks fine. Waiting for @robsimmonds to test before merging this in.

veggiesaurus (approved):
This all looks ok. 

@duidae One thing I noticed is that if I save a custom layout, and then set my default layout to that layout, it doesn't get applied on startup, presumably because it doesn't have the custom layouts at that point. I'm pretty sure this will sort itself out once the preferences are also stored server-side, but it's something to watch out for

duidae:
yes it's in the server preference part

PR #659, opened 2019-12-25
Add file list refresh button in File Browser #557

### LINKED ISSUE #557: reload button for refresh file list in file browser
currently need to re-open the file browser to refresh file list

TienHao:
close [#557](https://github.com/CARTAvis/carta-frontend/issues/557):reload button for refresh file list in file browser 

* add button on the left of breadcrumbs

* modify css

### DIFF
--- a/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.scss
+++ b/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.scss
@@ -6,6 +6,7 @@
     .file-path {
         margin-top: 20px;
         margin-left: 25px;
+        display: flex;
     }
 
     .bp3-dialog-body {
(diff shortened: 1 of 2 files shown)


duidae (changes requested):
works good. some suggestions:
1. I think it should be more intuitive to use "repeat" icon instead of "refresh" since chrome & firefox also use the same icon for refresh button
2. Capitalize the title "Refresh current directory"
3. use minimal prop of button so that the UI style is consistent with bread crumb

duidae (approved):

kswang1029:
we will need a reload button for the region file browser as well. can be in a separate PR.

veggiesaurus (changes requested):
For consistency, we should use the BlueprintJS [tooltip](https://blueprintjs.com/docs/#core/components/tooltip) rather than the HTML tooltip.

veggiesaurus (approved):

PR #670, opened 2020-01-13
Angus/contour dialog

veggiesaurus:
Initial implementation of contour dialog, as described in the image overlay doc [here](https://docs.google.com/document/d/10nG54sC8XoEKhlGNnTmY6ZOHstF3zeiJi22gBjEvE_k/edit#).

Most of the features are implemented (for a single image at a time). One feature that I'd like to add at a later stage is the ability to select whether contours are calculated for the entire FoV (default) or just the current FoV, as this will significantly speed things up when users are working on a particular region of a large file. 

@kswang1029 the default values weren't particularly useful in most of the files I tried, and we may need to refine these after more in-depth tests.


panchyo0 (approved):

src/stores/ContourConfigStore.ts
```diff
@@ -1,87 +1,91 @@
...
-                for (let i = 0; i < levelArray.length; i++) {
-                    levelArray[i] = this.upperBound - i * stepSize;
-                }
-                return levelArray;
-            }
-        }
-
-        return [];
-    }
+    // // Returns computed or manual contour levels
```
duidae: Are those commented code for future use?

veggiesaurus: good catch :) I've removed it

PR #678, opened 2020-01-21
Jason/layer widget

duidae:
PR for initial layer list widget


veggiesaurus (changes requested):
This is mostly looking great. I think we should make the Channel and Stokes columns much smaller. The more important change is that we need to be able to select a row by clicking on any of the cells, I think. 

Finally, for this initial PR, can you highlight the _current_ frame in some way (perhaps bold text?)

duidae:
ok

veggiesaurus (approved):

PR #683, opened 2020-01-22
Add custom icon for contour

### LINKED ISSUE #676: Improve contour icon
As suggested by @low-sky, we should change the contour icon to make it clearer. At the moment, it's more of a heatmap than a set of contours.

The set of existing BlueprintJS icons can be found [here](https://blueprintjs.com/docs/#icons). We should try to use one of the existing ones, as it otherwise requires some graphic design work, and needs to be made consistent with the rest of the UI. 

We could also consider creating an icon from scratch and adding that to BlueprintJS as a pull request, as the wider community using Blueprint would probably benefit from it.

TienHao:
close #676 
@kswang1029: I add 2 contour icons besides heatmap, which one you prefer?
截圖 2020-01-22 下午4 39 57
截圖 2020-01-22 下午4 40 14


veggiesaurus (commented):
it would be good to see if we can use a webpack loader to load SVGs directly, rather than cluttering the source code with SVG icon data

TienHao:
OK, I will use a webpack loader to load SVGs. Thanks

kswang1029:
Let’s use the single blob version🙂

veggiesaurus:
@TienHao progress on this?

TienHao:
Yeah, I'm progressing this one and need some advice.

1. Webpack loader
To introduce a new SVG webpack loader(@svgr/webpack, react-svg-loader...), I have to modify webpack.config.js. However ,since we use react-scripts-ts, the config files are in `node_module\react-scripts-ts\config` but not under `src`. I found two possible solutions to customize webpack, one is `npm run eject` and the other is to use `react-app-rewired`. Either will make great change. do @veggiesaurus  have any suggestion to implement a new webpack loader? 

2. Using react-scripts 
`react-scripts` can import svg as a inline component above version 2.0.0, but the `react-scripts-ts@2.17.0` we use only get `react-scripts@1.0.17`. I have try latest version `react-scripts-ts@4.0.8` but failed, which was realeased one and a half year ago.:joy: `react-scripts-ts` has been deprecated since react-scripts support typescript. Maybe we could consider migrating from react-scripts-ts to react-scripts in future?
https://create-react-app.dev/docs/adding-images-fonts-and-files/
https://vincenttunru.com/migrate-create-react-app-typescript-to-create-react-app/

Considering the upcoming 1.3 release, perhaps cluttering source code of SVG could be a temporary workaround for loading custom SVGs?

veggiesaurus:
I think it's fine to just use the SVG source, then, until we can migrate to `react-scripts` (somewhat related to #527)

edit: Added chore issue #706 for doing this, scheduled for release 1.4. Anyone who feels like undertaking this is welcome to do so...

arusstaylor:
I agree single blob version.

veggiesaurus (approved):

Issue #692, opened 2020-01-26
File browser improvements for v1.4


veggiesaurus:
Epic for tracking a variety of frontend-based file browser improvements for the CARTA 1.4 release

One of the main tasks is to migrate the file browser to [@blueprint/table](https://blueprintjs.com/docs/#table). This should probably be done first, before tackling the rest of the issues

kswang1029:
Is it possible (in terms of time pressure) to use tooltip to show full file name for v1.3?

veggiesaurus:
> Is it possible (in terms of time pressure) to use tooltip to show full file name for v1.3?

Yes, that would be pretty straightforward (I'd estimate it as a `1`). Will you create an issue for it?

PR #695, opened 2020-01-27
show tooltip in file browser

veggiesaurus:
PR adds the file name as a tooltip for situations where the file name is too long to fit in the file column. Temporary fix until #692 is implemented.


kswang1029:
Looking good and also good to have tooltip for file formats including images and region files. 

For MIRIAD format, maybe we should just show it as "MIRIAD format" as there are some other telescopes using MIRIAD image format (SMA, CARMA, ATNF, GMRT, etc).
Screen Shot 2020-01-27 at 2 50 19 PM

veggiesaurus:
> For MIRIAD format, maybe we should just show it as "MIRIAD format" as there are some other telescopes using MIRIAD image format (SMA, CARMA, ATNF, GMRT, etc).

Fixed :+1:

PR #702, opened 2020-01-31
Angus/contour control map

veggiesaurus:
This PR adds the ability to show contours from secondary images overlaid on top of the reference image (The ability to show contours from other images on secondary images will be added in a separate PR). 

While raster images are rendered using an _approximate_ transformation (scaling, rotation and shifting to match a specific point in the image, which is now dynamically calculated from the centre of the current FoV), contours are rendered using a more accurate transformation.  

Using AST to translate every single contour vertex and regenerate the GPU buffers would dramatically increase memory usage and make switching between images pretty slow. Instead, when an image has "spatial matching" enabled, a control map is generated (currently of size 256x256). For each source vertex in the control map, AST is used to transform the vertex to the destination image's coordinate space. 

It is assumed that the exact transformation at any point on the image can be approximated by a cubic  spline interpolation of the nearest control points. Given that the default size of 256x256 means that control points are less than 0.4% of the image size away from each other, this is a fairly safe assumption. ~~In future, we can offer a preference for the desired control map size, if greater accuracy is required.~~ (edit: options of 128x128, 256x256, 512x512 and 1024x1024 have been added)

The control map is then loaded on to the GPU, and the vertex shader performs the cubic interpolation of the control map for each vertex, with no noticeable performance impact. Generating a 256x256 control map generally takes around 50 ms, but is a once-off operation when spatial matching is enabled, and occupies roughly 512 kB of GPU memory. The only downside of this approach is that the number of of control maps required for the full feature (showing contours on secondary images as well) scales like _O(N^2)_, where _N_ is the number of images in the WCS group.

This PR also adds the ability to show or hide the raster or contour images by clicking on the appropriate button in the layer widget.

### DIFF
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -306,7 +306,7 @@ export class PreferenceDialogComponent extends React.Component<{ appStore: AppSt
                 <FormGroup inline={true} label="Beam Visible">
                     <Switch
                         checked={preference.beamVisible}
-                        onChange={(ev) =>  preference.setBeamVisible(ev.currentTarget.checked)}
+                        onChange={(ev) => preference.setBeamVisible(ev.currentTarget.checked)}
                     />
                 </FormGroup>
                 <FormGroup inline={true} label="Beam Color">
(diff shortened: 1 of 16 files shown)


duidae (approved):

duidae:
this PR can be merged first & I'll handle the conflict with server preference

PR #703, opened 2020-02-02
Jason/server preference

### LINKED ISSUE #491: [server] Integrate preference & layout to DB in server version (frontend part)
- [X] integrate layout with server DB
- [X] integrate preference with server DB
- [X] test layout/preference with server DB

duidae:
closes #491 

btw @veggiesaurus could you please specifically review the codes in TileService.ts & related in AppStore.ts, which is the part I lack of confidence, thanks.


src/services/TileService.ts
```diff
@@ -66,13 +63,18 @@ export class TileService {
...
-        this.pendingRequests = new Map<number, boolean>();
 
         // L2 cache: compressed tiles on system memory
         this.lruCapacitySystem = lruCapacitySystem;
+    };
+
+    constructor(backendService: BackendService, lruCapacityGPU: number = 512, lruCapacitySystem: number = 4096) {
+        this.backendService = backendService;
+        this.channelMap = new Map<number, { channel: number, stokes: number }>();
+        this.setCache(lruCapacityGPU, lruCapacitySystem);
```
veggiesaurus: The problem with this is that it will create a large cache in the constructor, and then update it when preferences arrive. We should either just initialise the cache with the lowest possible values, or work out how to avoid re-initialising it altogether.

veggiesaurus: @duidae I've updated your branch with a minor change that solves this, please have a look and see  if you agree with it

duidae: got it, thanks!

veggiesaurus (approved):
Approving this and merging it in

Issue #704, opened 2020-02-02
Contour lines have inconsistent thickness when spatially matched to image with very different pixel size


veggiesaurus:
When using GALFACTS S2 as a spatial reference for `ultravista-dr3-k-band.hdf5`, contour lines look very ugly, due to the accumulation of floating point errors in the shader. 

Note, this only happens when images have very different pixel sizes. pixel size of GALFACTS S2 is  ~400x that of ultravista)

![image](https://user-images.githubusercontent.com/592504/73614809-32dab500-460b-11ea-9580-eda3cdad53b8.png)

PR #705, opened 2020-02-03
shader changes to improve floating point errors

veggiesaurus:
There were two root causes for #704:
* When using spatial matching, but viewing the contours on top of the same image, there was an accumulation of floating point errors that caused the thickness of lines to be very inconsistent (sometimes less than zero, rendering the lines invisible)
* When using spatial matching and a control map to view the contours on top of the reference image, using a "delta" of 0.1 pixels to estimate the transformed normal of the line segment was not an appropriate size, as the image was 48K in width, and thus control points were 190 pixels apart (when using a 256x256 grid)

I've implemented two fixes:
* Rearranged the shader so that more of the floating point calculations that are independent of the vertex positions are done outside of the shader. This has two advantages: Firstly, the accumulated floating point error from these calculations is now independent of vertex position, so any errors are _consistent_, and will just result in a tiny offset for all pixels rendered. Secondly, the calculations are done using double-precision, so the floating point errors are dramatically reduced. Unfortunately, this makes the shader code and the accompanying javascript code a little less readable.
* Instead of using a fixed delta of 0.1 pixels, I now use 10% of the grid size. For the image mentioned above, that would mean a delta of 19 pixels. 

The lines now look much better. They look a little wonky when zooming far in on the reference image, but at that point, the zoom level on the reference image is about 1600x, so I think it's an acceptable solution. Switching the control map resolution to 1024x1024 helps a little bit in this case, so perhaps it's just something users need to be aware of, when comparing images with huge differences in pixel sizes. 

lines_fixed

### DIFF
--- a/src/components/ImageView/ContourView/ContourViewComponent.tsx
+++ b/src/components/ImageView/ContourView/ContourViewComponent.tsx
@@ -1,10 +1,9 @@
 import {observer} from "mobx-react";
 import * as React from "react";
 import {AppStore, ContourDashMode, FrameStore, OverlayStore, RenderConfigStore} from "stores";
-import {getShaderFromString, GL, loadImageTexture} from "utilities";
+import {getShaderFromString, GL, loadImageTexture, rotate2D, scale2D, subtract2D} from "utilities";
 import "./ContourViewComponent.css";
 import allMaps from "static/allmaps.png";
-import {observable} from "mobx";
 
 const vertexShaderLine = require("!raw-loader!./GLSL/vert_line.glsl");
 const pixelShaderDashed = require("!raw-loader!./GLSL/pixel_dashed.glsl");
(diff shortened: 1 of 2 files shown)


kswang1029:
I tried with different pixel size ratios and it appears that if the ratio is 1000, it works fine to a certain high zoom level then breaks down. I think this is acceptable. :)
[diff_1000x.mp4.zip](https://github.com/CARTAvis/carta-frontend/files/4146840/diff_1000x.mp4.zip)

If the ratio is 10000, it breaks initially. 
[diff_10000x.mp4.zip](https://github.com/CARTAvis/carta-frontend/files/4146841/diff_10000x.mp4.zip)

veggiesaurus:
> I tried with different pixel size ratios and it appears that if the ratio is 1000, it works fine to a certain high zoom level then breaks down. I think this is acceptable. :)
> [diff_1000x.mp4.zip](https://github.com/CARTAvis/carta-frontend/files/4146840/diff_1000x.mp4.zip)
> 
> If the ratio is 10000, it breaks initially.
> [diff_10000x.mp4.zip](https://github.com/CARTAvis/carta-frontend/files/4146841/diff_10000x.mp4.zip)

Right... I guess we can always add a higher-resolution control map option if users request it (it's easy to add, but I'd rather not add it unless users specifically request it)

kswang1029:
I think it is (very) good enough now. : )

PR #709, opened 2020-02-04
use -FLT_MAX as a placeholder for NAN values on GPU

### LINKED ISSUE #708: Changes to Intel GPU drivers or shader compiler breaks NaN display on linux
A recent change to _something_ related to Intel GPUs has resulted in NaNs turning into zeros when rendering images. A similar thing happens with the latest iPad Pro GPUs. 

NaNs are pretty hit-or-miss when it comes to GPUs, and it's likely that an overly aggressive shader compiler optimization is mucking with our NaN check. 

One solution is to use `-FLT_MAX` to indicate `NaN` values on the GPU.

veggiesaurus:
closes #708 

(needs testing to make sure it doesn't break _other_ implementations)

### DIFF
--- a/src/components/ImageView/RasterView/GLSL/pixel_shader_float_rgb.glsl
+++ b/src/components/ImageView/RasterView/GLSL/pixel_shader_float_rgb.glsl
@@ -9,6 +9,8 @@ precision highp float;
 #define EXP 6
 #define CUSTOM 7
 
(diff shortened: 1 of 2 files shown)


kswang1029:
Done a brief test and all seem working as usual. Surprisingly iPad now can render images!

PR #717, opened 2020-02-05
Spatial group contours

veggiesaurus:
This PR enables rendering of all contours belonging to the same spatial group, in the correct order. 

contour_order

### DIFF
--- a/src/models/ControlMap.ts
+++ b/src/models/ControlMap.ts
@@ -20,7 +20,7 @@ export class ControlMap {
         this.width = width;
         this.height = height;
 
-        let cleanUpTransform: boolean;
+        let cleanUpTransform: boolean = false;
 
         if (astTransform < 0) {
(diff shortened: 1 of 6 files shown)


src/models/ControlMap.ts
```diff
@@ -20,11 +20,27 @@ export class ControlMap {
         this.width = width;
         this.height = height;
 
+        let cleanUpTransform: boolean;
```
duidae: should cleanUpTransform be initialized to false?

duidae (changes requested):

panchyo0 (approved):

src/models/ControlMap.ts
veggiesaurus: yes, that's probably neater (even though it's implicitly set to `false`. I'll fix shortly

PR #728, opened 2020-02-06
contour rendering in screenshots

### LINKED ISSUE #712: missing contour layers when exporting as png
When exporting an image as png, all contour layers should also be included in the plot.

veggiesaurus:
fixes #712


kswang1029:
This is looking great. Also tested with raster hidden. 👍 

On a side note, we might get complain about the background color not being white for journal paper.

veggiesaurus:
The background is transparent, so it will be white if the journal paper is white, surely?

kswang1029:
you are right, I was fooled by the png viewer of Mac.....

Issue #684, opened 2020-01-22
Feature request: zoom toggle


Jordatious:
It would be cool to have a toggle in the preferences for the behaviour of zoom switching between zooming in on the mouse and zooming in on the central pixel.

veggiesaurus:
This is pretty easy to implement (~2 lines in `FrameStore` and some boilerplate code in the preference dialog). @kswang1029 any suggestions for who to assign to this?

kswang1029:
@TienHao could you investigate this?

TienHao:
I can take this one.

veggiesaurus:
@TienHao just keep @duidae in the loop in terms of the preference schema changes

PR #729, opened 2020-02-07
Add zoom in/out control in preference dialog

TienHao:
meet feature request of #684


duidae (approved):

veggiesaurus (commented):
This looks good, I'm just wondering if we should say "Zoom to" instead of "Zoom in/out on"? @kswang1029 what do you think?

kswang1029:
"Zoom to" reads more natural.

PR #730, opened 2020-02-07
Correct zoom level

### LINKED ISSUE #710: beam size of non-spatial-reference image on screen does not scale correctly with zoom level
beam size of non-spatial-reference image on screen does not scale correctly with zoom level

reference (correct)
Screen Shot 2020-02-05 at 09 14 28

non-reference (problem)
Screen Shot 2020-02-05 at 09 14 37

duidae:
closes #710


veggiesaurus (approved):
Looks good 👍

PR #731, opened 2020-02-07
Angus/close file

veggiesaurus:
This PR adds the ability to remove the currently active image by clicking File->Close Image, or using the shortcut key "{modifier}+W" (so ALT + W for browser, CTRL + W for linux desktop and CMD + W for MacOS)

If the image is currently active, the first remaining image becomes the new active image. If the image is currently the spatial reference for any other images, a confirmation alert is shown, indicating that other images will also be closed. If the user confirms, the first remaining image becomes the new spatial reference.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -97,7 +97,7 @@ export class App extends React.Component<{ appStore: AppStore }> {
                 </Alert>
                 <Alert
                     isOpen={appStore.alertStore.interactiveAlertVisible}
-                    confirmButtonText="Yes"
+                    confirmButtonText="OK"
                     cancelButtonText="Cancel"
                     intent={Intent.DANGER}
                     onClose={appStore.alertStore.handleInteractiveAlertClosed}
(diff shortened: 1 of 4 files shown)


veggiesaurus:
@kswang1029 please give this a try. The other option would be to simply un-match other files, rather than closing them

kswang1029:
> The other option would be to simply un-match other files, rather than closing them

could we support this? I think this is much better than closing all other matched images.

veggiesaurus:
> > The other option would be to simply un-match other files, rather than closing them
> 
> could we support this? I think this is much better than closing all other matched images.

Ok. And we still make the first file the reference?

kswang1029:
yes, the first remaining image is the new reference and no one is matched wrt it. Sounds intuitive to you?

veggiesaurus:
> yes, the first remaining image is the new reference and no one is matched wrt it. Sounds intuitive with you?

Yes, I think that should be fine. In this case, I'll still warn that images will be unmatched, and any regions in the images will be deleted

kswang1029:
👍

PR #741, opened 2020-02-13
modify default region of analysis widgets

### LINKED ISSUE #512: Analysis widgets should default to the "active" region.
When a user creates a new widget, its associated region should default to the "active" region.  This means that users can easily switch from viewing the analysis results of one region to that of another by selection the other region. This has been requested since 1.0, but was blocked by the lack of partial profiles and profile cancellation.

- [ ] An item in the region dropdown needs to be added for "active" region. The active region is the selected region. If there is no selected region, spectral profiles revert to the cursor region, while stats and histogram widgets revert to the image region. 
- [ ] The requirements map checks need to be modified to correctly send requirements when a widget has an "active" region selected.
- [ ] The widget title handling needs to be updated

TienHao:
close #512 : set default region to selected region when user create a analysis widget related to region.

### DIFF
--- a/src/components/Histogram/HistogramComponent.tsx
+++ b/src/components/Histogram/HistogramComponent.tsx
@@ -151,7 +151,7 @@ export class HistogramComponent extends React.Component<WidgetProps> {
         // Update widget title when region or coordinate changes
         autorun(() => {
             const appStore = this.props.appStore;
-            if (this.widgetStore && appStore.selectedRegion) {
+            if (appStore.selectedRegion || !appStore.selectedRegion) {
                 this.widgetStore.syncRegionIdIfActive(-1);
             }
             if (this.widgetStore && appStore.activeFrame) {
--- a/src/components/Histogram/HistogramComponent.tsx
+++ b/src/components/Histogram/HistogramComponent.tsx
@@ -151,6 +151,9 @@ export class HistogramComponent extends React.Component<WidgetProps> {
         // Update widget title when region or coordinate changes
         autorun(() => {
             const appStore = this.props.appStore;
+            if (this.widgetStore && appStore.selectedRegion) {
+                this.widgetStore.syncRegionIdIfActive(-1);
+            }
             if (this.widgetStore && appStore.activeFrame) {
                 let regionString = "Unknown";
                 const regionId = this.widgetStore.regionIdMap.get(appStore.activeFrame.frameInfo.fileId) || -1;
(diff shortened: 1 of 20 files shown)


veggiesaurus (changes requested):
This doesn't fix #512. The default region should be "Active", and when the selected region is changed, the "Active" region should change. At the moment, this requires the user to manually change regions.

src/components/Histogram/HistogramComponent.tsx
```diff
@@ -151,7 +151,7 @@ export class HistogramComponent extends React.Component<WidgetProps> {
         // Update widget title when region or coordinate changes
         autorun(() => {
             const appStore = this.props.appStore;
-            if (this.widgetStore && appStore.selectedRegion) {
+            if (appStore.selectedRegion || !appStore.selectedRegion) {
```
veggiesaurus: this line makes no sense to me...

veggiesaurus: can't the region ID be left as "Active" (perhaps we can use -3 or something for that?) and then the rest of the values can be computed from that, rather than a whole bunch of autoruns?

TienHao: > this line makes no sense to me...

Sorry about that. It was just a temporary workaround for getting image/cursor data. Since `appStore.selectedRegion` always pass `null` when deselect a region so the regionId of a "Active" widget won't change to -1(image) or 0(cursor).

TienHao: > can't the region ID be left as "Active" (perhaps we can use -3 or something for that?) and then the rest of the values can be computed from that, rather than a whole bunch of autoruns?

Yes, it can. I just haven't thought that before...:joy: 
I have made a new commit for what you suggested. It's truly a better way.
Thanks

veggiesaurus (changes requested):
This looks really good. A couple of minor comments:
- I think `regionIdAdjustedWithSelectedRegion` is a little too verbose. Perhaps just `effectiveRegionId`? 
- I think we should probably define -3 as a constant (e.g. `ACTIVE_REGION_ID`), rather than just using a magic number
- We should be consistent with our use of "Active" and "Selected". I think "Active" sounds better, but then we should append `(Active)` instead of `(Selected)` to the widget title
- If there are no defined regions, I should still be able to select between "Active" and "Cursor" (or "Image", depending on the widget). I might explicitly want to select the cursor as my selected region.

veggiesaurus (approved):

Issue #698, opened 2020-01-29
missing ">" in breadcrumbs 


kswang1029:
frontend dev 
origin/dev 5303424 2020-01-28 16:03:37 CST

<img width="1196" alt="Screen Shot 2020-01-29 at 1 37 44 PM" src="https://user-images.githubusercontent.com/20819712/73357481-bbb8c000-429c-11ea-9919-46aa4d644bba.png">

veggiesaurus:
![image](https://user-images.githubusercontent.com/592504/73357664-882e6380-42a5-11ea-81f6-60f0d01f31c9.png)
Our `dev` server still has the arrows. Perhaps a build issue?

veggiesaurus:
@kswang1029 I still can't reproduce this anywhere. Can you test again?

kswang1029:
Hmm, this still happens with the dev server at ASIAA
http://carta.asiaa.sinica.edu.tw/frontend/dev/?socketUrl=wss://carta.asiaa.sinica.edu.tw/socketdev

Also happens in the Mac alpha test package.

veggiesaurus:
Assigning to @ajm-asiaa as it appears to be a build issue. I wonder if it's the version of BlueprintJS being used?

edit: tried 3.19 an 3.23, both look fine

ajm-ska:
@veggiesaurus Did you test it with `npm run start` or `npm run build`? When I do `npm run start` and check localhost:3000, I do see the symbols. But when I do `npm run build` and deploy it properly, then I do not see the symbols. Would you have any idea what could be different?

ajm-ska:
I checked back through all the older frontend dev commits (These are all done with `npm run build`).
It still worked at: `origin/dev c54e057 2020-01-16 13:49:06 CST`
http://carta.asiaa.sinica.edu.tw/frontend/c54e057/?socketUrl=wss://carta.asiaa.sinica.edu.tw/socket

But wasn't working by: `origin/dev 05e27f1 2020-01-22 22:34:01 CST`
http://carta.asiaa.sinica.edu.tw/frontend/05e27f1/?socketUrl=wss://carta.asiaa.sinica.edu.tw/socket

Going through all the commits between those, the point where the problem started appears to be `origin/jason/json_schema 2ffd8c5 2020-01-21 09:18:42 CST` 
http://carta.asiaa.sinica.edu.tw/frontend/2ffd8c5/?socketUrl=wss://carta.asiaa.sinica.edu.tw/socket
[...]

PR #742, opened 2020-02-15
Adding a script to temporarily patch the production build's breadcrumb issue

ajm-ska:
Here is a simple script to fix the bad string of NaNs in the css file of the production build related to issue [#698](https://github.com/CARTAvis/carta-frontend/issues/698)

The script is called in package.json file at the last stage of the "build" script.
```
"build": "npm run git-info; npm-run-all build-wrappers build-protobuf build-css build-js; ./crumbs-fix.sh",
```
The script uses **sed** to find the bad string and replace it with the correct string. It uses a slightly different **sed** command between MacOS and Linux (MacOS sed requires an extra two apostrophes in the command).

The script uses sed "w" so that we can get an "echo" feedback as to whether or not the command ran successfully. Hopefully, if/when the sass-loader bug is fixed, we will see that the script will no longer run, and then we can remove it.

### DIFF
--- a/crumbs-fix.sh
+++ b/crumbs-fix.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+#
+# This script is intended as a temporary fix for the missing ">" breadcrumb symbols in the file browser that only occurs when we build the frontend in production mode.
+# CARTAvis/carta-frontend issue #698
+# It seems to be a sass-loader bug. The problem may eventually go away, but meanwhile this script will fix it. 
+# This script will be run automatically at the last step of "npm run build".
+
+if [ "$(uname)" == "Darwin" ]; then
+
+  sed -i '' "s/M10.71 7.29l-4-4a1.003 1.003 0 0-1.42 1.42 NaNl.46 NaNL5.3 NaNC5.11 NaN 5 NaN 5 NaNa1.003 1.003 0 1.71.71 NaN NaNlNaN NaNcNaN NaN NaN NaN NaN NaN NaN NaN NaN NaN NaN NaNz/M10.71 7.29l-4-4a1.003 1.003 0 00-1.42 1.42L8.59 8 5.3 11.29c-.19.18-.3.43-.3.71a1.003 1.003 0 001.71.71l4-4c.18-.18.29-.43.29-.71 0-.28-.11-.53-.29-.71z/w patch.txt" build/static/css/main*.css
+  if [ -s patch.txt ]; then
+    echo "The production build css file has been patched for the breadcrumbs fix."
+  else
+    echo "The css file has NOT been patched. Maybe the sass-loader/breadcrumbs problem is gone?"
+  fi
+
+fi
+
+if [ "$(uname)" == "Linux" ]; then
+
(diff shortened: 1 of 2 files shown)


kswang1029:
Not sure if this is related: which blueprint version are we using? 3.8.2 or 3.7.1? From catalog discussions, Qi found 3.8.2 is more buggy.

ajm-ska:
We are using 3.8.2: `"@blueprintjs/table": "^3.8.2",`

Actually, it may indeed be related because I noticed the problem only started when `"@blueprintjs/table"` was introduced to the package.json file.

kswang1029:
I guess we can check if the breadcrumbs work with 3.7.1. Guessing at some point the ci updated it to 3.8

ajm-ska:
I just tried `"@blueprintjs/table": "^3.7.1",` but it still has the problem.

But the entire `@blueprintjs/table` was only introduced recently, and that is when the problem started.

Previously we were only using `core,` `icons`, and `select`
```
"@blueprintjs/core": "^3.19.1",
"@blueprintjs/icons": "^3.11.0",
"@blueprintjs/select": "^3.11.1",
```

kswang1029:
I see. Thank you for trying it out. No luck here 😔

veggiesaurus:
asking Adrianna to look into this, since she's better with `sed` etc than  me

crumbs-fix.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+#
+# This script is intended as a temporary fix for the missing ">" breadcrumb symbols in the file browser that only occurs when we build the frontend in production mode.
+# CARTAvis/carta-frontend issue #698
+# It seems to be a sass-loader bug. The problem may eventually go away, but meanwhile this script will fix it. 
+# This script will be run automatically at the last step of "npm run build".
+
+if [ "$(uname)" == "Darwin" ]; then
```
confluence, line 8: I suggest that you only set a command variable inside the conditional, to account for the additional argument to `-i` in BSD sed, instead of repeating everything: `cmd="sed -i"` vs `cmd="sed -i ''"`, and then `$cmd "s/M10..."` outside the conditional.

confluence: Also, where do these values come from, and how likely are they to change in the short term? They seem to be the same in my install. It may be safer to extract them from the files than to hardcode them:

```
wrong=`grep -Po "breadcrumbs>li:after\{.*?\}" main.*.css | grep -Po "d='.*?'"`
right=`grep -Po ".bp3-breadcrumbs > li::after\{.*?\}" main.*.css.map | grep -Po "d='.*?'"`

(later...)

$sedcmdhere "s/$wrong/$right/w somewhere" badfile # note double quotes
```

But this is just more fragile regex; if the values are probably going to be constant, it's overkill. This was tested on Linux; I'm not 100% sure if it needs any modifications for BSD grep.

confluence (approved):
These are suggestions for making the script cleaner and maybe a bit more futureproof, but since this is intended as a temporary fix, it's not that important.

veggiesaurus (approved):

crumbs-fix.sh
ajm-ska: Thanks for the suggestions @confluence Yes. This patch is only intended to be temporary so probably not worth spending too much time to make it too perfect.

PR #744, opened 2020-02-17
Jason/spectral conversion

### LINKED ISSUE #647: conversion of spectral axis (x), such as velocity frame, frequency, velocity, optical/radio definition, etc.(c.f. casaviewer)
sub task of epic #643 

The spectral profile widget should provide conversions in the spectral axis including
- velocity <-> frequency <-> wavelength <-> channel
- velocity frames, such as LSRK, TOPO, etc.
- velocity definition: optical or radio

The conversion should be done either with AST at frontend or with casacore/casacode at backend.

duidae:
closes #647


veggiesaurus:
This is looking really useful. Quick discussion point: This makes the spectral profiler _really_ cramped. @kswang1029 what do you think of moving this out of the dropdown, and to the widget settings (at the top, because they're probably the most important settings)? Once we add the "image" dropdown, it's simply going to be way too crowded.

image

src/stores/widgets/RegionWidgetStore.ts
```diff
@@ -13,7 +13,7 @@ export enum RegionsType {
     CLOSED_AND_POINT
 }
 export class RegionWidgetStore {
-    private readonly appStore: AppStore;
+    public readonly appStore: AppStore;
```
veggiesaurus: probably don't need explicit `public`

wasm_src/ast_wrapper/ast_wrapper.cc
```diff
@@ -80,6 +80,135 @@ EMSCRIPTEN_KEEPALIVE AstFrameSet* initFrame(const char* header)
...
+    (void) snprintf(buffer, sizeof(buffer), "Unit=%s", unit);
+    astSet(specframe, buffer);
+
+    // get Epoch, ObsLon, ObsLat, ObsAlt, RefRA, RefDec, RestFreq, StdOfRest attributes from AST compound frame
+    // due to observation coordinate (ObsLon/ObsLat/ObsAlt) has to be geodetic, RefRA/RefDec has to be FK5, J2000.
+    // RefRA/RefDec will be 0 if skyframe is not provided, & the conversion will be imprecise.
+    const char *epoch = astGetC(frame, "Epoch");
+    if (epoch)
+    {
+        (void) snprintf(buffer, sizeof(buffer), "Epoch=%s", epoch);
```
veggiesaurus: we don't need any of these explicit `(void) ` specifiers.

wasm_src/ast_wrapper/ast_wrapper.cc
```diff
@@ -80,6 +80,135 @@ EMSCRIPTEN_KEEPALIVE AstFrameSet* initFrame(const char* header)
...
+    AstFrameSet* frame = nullptr;
+    int status = 0;
+    if (frame)
+    {
+        astEnd;
+    }
+	astClearStatus;
+    astBegin;
+
+    fitschan = astFitsChan(NULL, NULL, "");
```
veggiesaurus: should use `nullptr` instead (please update my `NULL` values at the same time :)

arusstaylor:
I agree this could go in the widget settings.

duidae:
will do the corresponding modification, btw kswang is on his flight to Taiwan, will be back soon

veggiesaurus (changes requested):
I think the changes work well. But we need the same spectral conversion options in the Stokes widget as well, please.

kswang1029:
Yes, moving those to setting is good. We also need to add spectral reference frame label to x-label. eg “LSRK, radio velocity (km/s)”

veggiesaurus:
> Yes, moving those to setting is good. We also need to add spectral reference frame label to x-label. eg “LSRK, radio velocity (km/s)”

Couldn't we leave the "(km/s)" out, since it would be included in the _value_ quoted itself?

arusstaylor:
I have been trained to always include units in axis labels....

veggiesaurus:
Oh wait I'm misunderstanding what @kswang1029 was suggesting 🙈

veggiesaurus:
this is looking great now. I'm just wondering: Should we increase the default size of the Stokes and spectral config widgets so that the coordinate dropdown lines up nicely? 

image

(it does seem like there's some odd CSS in the widget though, but this can be fixed later)

kswang1029:
how are the default spectral convention and reference frame determined now?
Can we have LSRK Radio velocity (km/s) as default? "m/s" is not a common unit that most of astronomers use. It is used only at certain use cases.

kswang1029:
@duidae please modify "wave length" to "wavelength"

duidae:
the initial default is set to the frame's spectral setting according to it's fits header, can be change to LSRK Radio velocity (km/s) as universal default

veggiesaurus:
> the initial default is set to the frame's spectral setting according to it's fits header, can be change to LSRK Radio velocity (km/s) as universal default

To me, it makes sense to use the fits header for the default, since we're then representing what's actually in the image headers, rather than forcing a specific convention?

kswang1029:
I think at least, if it is velocity we should use km/s. And for frequency, we can use GHz. I agree we should use the reference frame as it is defined in the header. So the requested change is simply to set the units to the commonly used ones.

duidae:
@veggiesaurus  @kswang1029  changed accordingly, please take a look, thanks.

kswang1029:
@duidae found an issue, if we switch to other spectral convention and try to freeze and unfreeze cursor repeatedly, we will see the unit changes (e.g. GHz -> MHz -> kHz -> Hz -> GHz etc).

duidae:
@kswang1029  I cant reproduce that. What file/browser are u using? wait i see it, drop down menu provides a short cut to find wanted option by pressing the initial character of options

kswang1029:
@duidae I am using Chrome with "origin/jason/spectral_conversion 3f0c72e 2020-02-20 13:17:38 UTC".

To reproduce,
1) load set_HD163296/HD163296_CO_2_1.fits
2) we see LSRK, Frequency (GHz)
3) switch to Frequency (kHz)
4) try to freeze/unfreeze cursor repeatedly and observe the unit.

veggiesaurus (approved):

Issue #612, opened 2019-11-27
colour to frequency colour maps for stokes widget


arusstaylor:
Most of the canned colour maps go from blue to red, when applied to the Stokes widget this results in the red colour on the Q,U scatter plot mapping to the high frequency end of the band.  Convention is astronomy is for redshift to correspond to longer wavelength (lower frequency) so this is counterintuitive to the user who expects red to mean longer wavelength.  Having an invert colour map option in the settings widget for the Stokes would allow the user to set an appropriate  colour mapping red to blue for increasing frequency.

kswang1029:
Right, we better use "blue" for high frequency/low velocity and "red" for low frequency/high velocity. In fact, why don't we just provide a single colormap ("jet") and invert it when the display axis is frequency? If the display axis is velocity, we just use the normal "jet" colormap.

arusstaylor:
I think it is nice to have a few options for colour map.  I think as you suggest it would be a good idea to choose a sensible default and invert it for frequency plots.

But also enable an invert colour map option in the widget setting.  Good to offer people to ability to depart from the defaults for strange edge cases..

panchyo0:
is this one works? http://carta.asiaa.sinica.edu.tw/frontend/90e7156/?socketUrl=wss://carta.asiaa.sinica.edu.tw/socketdev
The default setting for "Invert Color Map" is true. User can change it in stokes widget settings panel.

veggiesaurus:
> The default setting for "Invert Color Map" is true. User can change it in stokes widget settings panel.

@arusstaylor surely choosing whether to invert the color map is dependent on the image's `CDELT3` value? (and the units)

arusstaylor:
The logic of inversion is determined by the frequency or velocity direction of the channels.  red to blue for  low frequency to high frequency,   blue to red for low velocity to high velocity.

veggiesaurus:
@kswang1029 can we do this easily by checking `CDELT3` and the `CTYPE3`?

kswang1029:
Yes.

If `CTYPE3` is frequency and `CDELT3` > 0 (ie frequency increasing; velocity decreasing), then it is "RED" to "BLUE". 
If `CTYPE3` is frequency and `CDELT3` < 0 (ie frequency decreasing; velocity increasing), then it is "BLUE" to "RED". 

If `CTYPE3` is velocity and `CDELT3` > 0 (ie velocity increasing; frequency decreasing), then it is "BLUE" to "RED". 
[...]

PR #749, opened 2020-02-18
Qi/invert colour 612

panchyo0:
1. Invert color map setting in settings panel.
2. Auto revert color map according CTYPE #612


veggiesaurus (changes requested):
this PR shouldn't change the protobuf commit, it looks like you're commiting the ctalog protobuf changes by mistake. 

I think we should merge in @duidae's PR first, as there might be some overlap

@kswang1029 can you test this branch and see if it behaves appropriately from a science perspective?

kswang1029:
@panchyo0 Do you only check CTYPE3 for spectral information? If so, please try to check all CTYPE for spectral information since it is not always CTYPE3 carrying spectral info. Usually it is either CTYPE3 or CTYPE4 (if exists). Very rarely, with a permuted cube, the spectral axis can happen with CTYPE1 or CTYPE2. So I would suggest we do a secure check, instead of checking CTYPE3 only.

veggiesaurus:
> @panchyo0 Do you only check CTYPE3 for spectral information? If so, please try to check all CTYPE for spectral information since it is not always CTYPE3 carrying spectral info. Usually it is either CTYPE3 or CTYPE4 (if exists). Very rarely, with a permuted cube, the spectral axis can happen with CTYPE1 or CTYPE2. So I would suggest we do a secure check, instead of checking CTYPE3 only.

Good point. Don't we have a function somewhere that determines which is the spectral axis? If not, we probably _should_

panchyo0:
@veggiesaurus updated with checking all CTYPE. I think we need to merge catalog protobuf and backend branch first.

veggiesaurus:
> @veggiesaurus updated with checking all CTYPE. I think we need to merge catalog protobuf and backend branch first.

This issue is independent of the catalogue branch, so we should merge this one first.

veggiesaurus:
@panchyo0 this has the wrong protobuf commit again, can you fix those changes so we can merge this branch in?

veggiesaurus (approved):

veggiesaurus:
@kswang1029 can you test this?

@duidae is there any overlap with your code that could result in simplification? If so, please go ahead and simplify it before we merge these changes in.

kswang1029:
@panchyo0 It appears that when I load a Stokes cube in velocity, the color is not reverted.
Full view
Screen Shot 2020-02-26 at 8 56 40 AM

Zoom to "negative" velocities, should be colored in blue, but it is in red
Screen Shot 2020-02-26 at 8 56 56 AM

Zoom to "positive" velocities, should be colored in red, but it is in blue.
Screen Shot 2020-02-26 at 8 57 07 AM

I tried with 
http://carta.asiaa.sinica.edu.tw/frontend/2f6fb7c/?socketUrl=wss://carta.asiaa.sinica.edu.tw/socketdev
and use 
set_StokesCube/IRCp10216_sci.spw0.cube.IQUV.manual.pbcor.fits (in freq, working fine)
set_StokesCube/IRCp10216_sci.spw0.cube.IQUV.manual.pbcor.asVelo.fits (in velocity, with issue)

kswang1029:
Just realized an addition complexity here, when users change the x-axis to different spectral convention, shall we reverse the colormap automatically if necessary?

duidae:
@veggiesaurus  removed overlapped part, please have a look. thanks.
@kswang1029  could you please help to confirm my refactoring does not break the function? thanks.

veggiesaurus:
@panchyo0 @kswang1029 what's the status of this PR?

panchyo0:
I think just checking CTYPE is not enough. We also need to consider the x-axis order (increasing or decreasing) with channel, velocity and frequency. I tested with some files and change the x-axis in settings. It looks like our x-axis always in increasing order. Not sure if this apply to all files?

veggiesaurus:
> It looks like our x-axis always in increasing order. Not sure if this apply to all files?    

You can determine that from CDELT3. But perhaps we should rather use AST to solve this in a more systematic manner. Get the frequency of the first channel, get the frequency of the last channel (regardless of what the CTYPE3 is) and then compare

panchyo0:
The CDELT3 seams doesn't matter, since our x-axis always in increasing order. Should our x-axis order depends on CDELT value? 
According @kswang1029, for file `IRCp10216_sci.spw0.cube.IQUV.manual.pbcor.asVelo.fits`, it should follow the fourth case with Red -> Blue color map(CTYPE3 is velocity and CDELT3 < 0), however the x-axis of the spectral plot is increasing so the colormap should be blue -> red. If our x-axis always in increasing order, should we just checking CTYPE?

veggiesaurus:
> The CDELT3 seams doesn't matter, since our x-axis always in increasing order. Should our x-axis order depends on CDELT value?
> According @kswang1029, for file `IRCp10216_sci.spw0.cube.IQUV.manual.pbcor.asVelo.fits`, it should follow the fourth case with Red -> Blue color map(CTYPE3 is velocity and CDELT3 < 0), however the x-axis of the spectral plot is increasing so the colormap should be blue -> red. If our x-axis always in increasing order, should we just checking CTYPE?

`CDELT3` tells you the difference between the value at channel 1 compared to channel 0. But this means the native WCS type of the image is increasing/decreasing. You still may need to convert this to velocity. @duidae what do you think? Easy to just use AST for this?

kswang1029:
Part of the ambiguities comes from the fact that chartjs seems always plot the x axis with increasing value from left to right, regardless it is velocity or frequency.

veggiesaurus:
> > The CDELT3 seams doesn't matter, since our x-axis always in increasing order. Should our x-axis order depends on CDELT value?
> > According @kswang1029, for file `IRCp10216_sci.spw0.cube.IQUV.manual.pbcor.asVelo.fits`, it should follow the fourth case with Red -> Blue color map(CTYPE3 is velocity and CDELT3 < 0), however the x-axis of the spectral plot is increasing so the colormap should be blue -> red. If our x-axis always in increasing order, should we just checking CTYPE?
> 
> `CDELT3` tells you the difference between the value at channel 1 compared to channel 0. But this means the native WCS type of the image is increasing/decreasing. You still may need to convert this to velocity. @duidae what do you think? Easy to just use AST for this?

Here's a bit of code I've used before to check whether a cube's channels are proportional to velocity or not:

```C#
// _zDelt is the floating-point value of the spectral CDELT header
// _zCoord is the string value of the spectral CTYPE
public int VelocityDirection
{
    get
    {
        if (_zCoord != null && _zCoord.Length >= 4 && _zDelt != 0.0)
        {
            var zCoordPrefix = _zCoord.Substring(0, 4);
            switch (zCoordPrefix)
            {
                case "FREQ":
                case "ENER":
                case "WAVN":
                    return -1 * Math.Sign(_zDelt);
                case "VRAD":
                case "WAVE":
                case "VOPT":
                case "ZOPT":
                case "AWAV":
                case "VELO":
                case "BETA":
                    return Math.Sign(_zDelt);
            }
        }
        return 0;
    }
}
```

@panchyo0 perhaps you can use a similar approach? @kswang1029 does this look right? Then you invert the colourmap depending on whether `VelocityDirection` is -1 or not.

kswang1029:
@panchyo0 hope this helps. It is a "visualization" of the above code.
Screen Shot 2021-01-28 at 5 49 31 PM

panchyo0:
@kswang1029 please review again, thanks

kswang1029:
`origin/qi/invert_colour_612 bb92388` this is looking good! One check here, with `Stokes_IQU_FREQ.fits`, I see changing from the default frequency to channel results in a flipped colormap which is not correct. Could you have another look? Once this is fixed, we are ready to go!

panchyo0:
@kswang1029 please check agin, thanks

kswang1029:
@panchyo0 Nice! PR ready merging. @veggiesaurus

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -139,6 +139,20 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
         this.widgetStore.clearScatterPlotXYBounds();
     };
 
+    // true: red->blue, false: blue->red
+    private getColorMapOrder(frame: FrameStore): boolean {    
+        if (frame && frame.channelInfo && frame.channelInfo.fromWCS && frame.channelInfo.channelType && !isNaN(frame.channelInfo.delta)) {
+            let CTYPE = frame.spectralType !== null? frame.spectralType : frame.channelInfo.channelType.code;
```
veggiesaurus: can't we use `??` here?

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -139,6 +139,20 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
...
+            if (CTYPE === SpectralType.CHANNEL) {
+                CTYPE = frame.channelInfo.channelType.code;
```
veggiesaurus: this code is confusing to follow. if ctype is channel, then why do we go looking for the channel info's spectral type?

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -139,6 +139,20 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
...
+            }
+            const inGroup = SPECTRAL_COLORMAP_GROUP.includes(CTYPE);
```
veggiesaurus: I think it would be much easier to not create this array, and just manaully check if it's `===` one of the three options. These are standardised options and will not change

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -512,32 +526,25 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
         return a === b && a === c && a === d && a !== null;
     }
 
-    private getScatterColor(percentage: number, frequencyIncreases: boolean): string {
+    private getScatterColor(percentage: number, redToBlue: boolean): string {
         const colorMap = this.widgetStore.colorPixel.color;
         const mapSize = this.widgetStore.colorPixel.size;
-        if (!frequencyIncreases) {
+        if (redToBlue) {
```
veggiesaurus: I don't think redToBlue is a good name here, because it depends on the colormap itself. Rather just say "reversed"

src/models/SpectralDefinition.ts
```diff
@@ -18,6 +18,13 @@ export const CHANNEL_TYPES: ChannelType[] = [
     {code: "BETA", name: "Beta", unit: ""},
 ];
 
+// FREQ, ENER, WAVN
```
veggiesaurus, line 21: See comment about just just using FREQ ENER WAVN directly

veggiesaurus (changes requested):

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
panchyo0: when user select `Channel` in `conversion`, we use the default colour order from the file according default file spectral type.

veggiesaurus (approved):

PR #754, opened 2020-02-20
Thchang/auto stop animation playback 474

### LINKED ISSUE #474: auto stop animation playback
With the consideration of server performance and resource management, we should apply an auto stop of animation playback for a given amount of time (e.g, 15 or 30 mins). This would prevent the case that users keep the playback on intentionally or unintentionally.

TienHao:
close #474 
I used a timeout to trigger stopAnimation() to stop playback, and the length of timeout can be changed by the selector in Preference > Performance.


kswang1029:
@TienHao @veggiesaurus I am thinking maybe we should show a warning 10s before it times out. If users click "continue", we reset the timer and let the animation continue. If users do not click "continue" or click "stop", we stop the animation playback. Ideas?

veggiesaurus:
> @TienHao @veggiesaurus I am thinking maybe we should show a warning 10s before it times out. If users click "continue", we reset the timer and let the animation continue. If users do not click "continue" or click "stop", we stop the animation playback. Ideas?

that might be overkill for what is effectively a way to prevent users from simply using CPU resources when not needed. I'm not even sure we need a preference setting for this, but since it's been added, I've got no problem with it. 

It's quite easy to restart the animation after it's been automatically stopped, so I'm not sure we need a confirmation.

veggiesaurus (approved):

Issue #755, opened 2020-02-20
Contour rendering breaks when changing layouts


veggiesaurus:
After changing layouts, contours are no longer rendered, due to a WebGL error. A similar fix to that used for #746 is required. The contour WebGL context should be associated with a service, rather than a specific component, which can be unmounted.

veggiesaurus:
@kswang1029 can you confirm this is fixed in `release/1.3` branch and `dev`?

kswang1029:
with 
origin/backport_755 290ff57 2020-03-03 17:10:16 UTC
and dev, the issue is gone. 👍

PR #756, opened 2020-02-20
move WebGL context to a separate WebGL service

veggiesaurus:
This PR prevents the WebGL context from being lost when the `RasterViewComponent` is unmounted during layout changes. This also prevents re-initialisation of texture tiles multiple times, so it should decrease GPU memory usage a little. 

A similar approach will be needed for the contours, as I've just noticed they break when changing layers (see #755). 

@kswang1029 this changes the rendering a bit, please check to see that things are behaving properly.


confluence (approved):
This seems to fix the issue for me.

kswang1029:
tried safari, chrome, Firefox with MBP, all work!

duidae (approved):

panchyo0 (approved):

PR #759, opened 2020-02-20
Add missing check for set spectral requirement about stokes

### LINKED ISSUE #748: empty data in spectral profile initially
reproduce:
1. init CARTA with a stoke widget & a spectral profiler (dont load file yet)
2. set initial cursor position to fixed in preference
3. load a file with spectral axis
4. close the stoke widget, data will come out

duidae:
closes #748

missing check so that backend service sends wrong set spectral requirement message (including stokes info even if the file does not have stoke coordinate) and receive error data


veggiesaurus (changes requested):
should be `hasStokes`. Other than that, looks fine.

PR #765, opened 2020-02-22
Angus/fix 720

### LINKED ISSUE #720: Region display of spatially matched images
Regions need to be scaled, rotated and translated appropriately according to the transform defined by the reference image.

veggiesaurus:
Closes #720 and cleans up code a little bit


kswang1029:
Regions with matched image work well. Great! : 👍

kswang1029:
also tested importing region text files, including ds9 and casa formats, and pixel and world coordinates. Works well. Exporting works too!

duidae (approved):

PR #766, opened 2020-02-22
rotate beam based on transform angle when spatially matched

### LINKED ISSUE #764: Beam indicator does not rotate when enabling spatial matching
When enabling spatial matching, the rendered image and grid may be rotated. However, currently the beam indicator does not rotate:

beam_wcs

veggiesaurus:
Closes #764 
beam_fixed


kswang1029:
looks good! thanks for spotting this. 👍

duidae (approved):

PR #767, opened 2020-02-22
adds checks and warnings when enabling spatial matching

veggiesaurus:
This PR adds basic checks in to handle situations where no valid transform between images can be found. If this occurs, then the image is _not_ spatially matched, and a warning is shown

### DIFF
--- a/src/stores/AppStore.ts
+++ b/src/stores/AppStore.ts
@@ -1029,7 +1029,13 @@ export class AppStore {
         if (frame.spatialReference) {
             frame.clearSpatialReference();
         } else {
-            frame.setSpatialReference(this.spatialReference);
+            if (!frame.setSpatialReference(this.spatialReference)) {
+                AppToaster.show({
+                    icon: "warning-sign",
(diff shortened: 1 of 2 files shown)


duidae (approved):

kswang1029:
tested the obvious case that image to be matched is on the other side of hemisphere. 👍 
Still need to check an edge case that only part of the image is matchable. Need some artificial images for this test. Will work on this shortly.

kswang1029:
I tried an edge case and we see problems.
This is an image as the reference, centered at 0h, 30deg.
Screen Shot 2020-02-23 at 3 41 10 PM
If we load an image centered at 18h, 30deg and try to match it, we will see odd behaviors. 
Feb-23-2020 15-43-04

Probably the reason is that this "18h, 30deg" image has some part that is outside the hemisphere of the reference "0h, 30deg". 

I would consider this as an edge use case, but in reality, users might accidentally trigger this issue when loading multiple different celestial objects located at different part of sky. 

The test is done only with SIN projection which is a converged projection type. I would expect similar would happen when we try to match images with diverged projection type such as TAN.  

To improve this, maybe we can check if all four corners are inside the visible sky of the reference image?

PR #775, opened 2020-02-25
Image frame dropdown in analysis widgets

### LINKED ISSUE #778: add a frame drop down to analytics widgets
Currently the content in any analytics widget is sync'd with the frame slider. So users can view results from one image at a time.

To extend the flexibility to view analytics of different frames (aka different images) at a time, we need a frame dropdown in every analytics widget. This would enable more use cases, especially with WCS features in v1.3.

TienHao:
closes #778 
User can use the frame dropdown in analysis widgets(Histogram, Stats, Spectral and Stoke) to view the analysis data from other inactive frames, instead of changing active frame. The "Current" option in the dropdown stands for the active frame, a little bit like "Active" in #512 .


src/components/SpectralProfiler/SpectralProfilerToolbarComponent/SpectralProfilerToolbarComponent.tsx
```diff
@@ -18,20 +18,24 @@ export class SpectralProfilerToolbarComponent extends React.Component<{ widgetSt
         this.props.widgetStore.setCoordinate(changeEvent.target.value);
     };
 
+    private handleFrameChanged = () => {
+        this.props.widgetStore.setCoordinate("z");
```
veggiesaurus: Can we first check to see if the new file has Stokes parameters? If I select Stokes Q explicitly, and load two images, it would be nice if this didn't result in it resetting back to Stokes I if the file has Stokes Q available. @kswang1029 what do you think?

src/components/StokesAnalysis/StokesAnalysisToolbarComponent/StokesAnalysisToolbarComponent.tsx
```diff
@@ -13,18 +13,22 @@ export class StokesAnalysisToolbarComponent extends React.Component<{widgetStore
         this.props.widgetStore.setFractionalPolVisible(changeEvent.target.checked);
     };
 
+    private handleFrameChanged = () => {
+        this.props.widgetStore.setFractionalPolVisible(false);
```
veggiesaurus: Same as above, I think

src/stores/AppStore.ts
```diff
@@ -1042,6 +1042,12 @@ export class AppStore {
 
     // region requirements calculations
 
+    @observable requiredFrame: FrameStore;
+
+    @action setRequiredFrame(frame: FrameStore) {
```
veggiesaurus: I think this is wrong... doesn't this mean that we can only have one required frame at a time?

src/stores/AppStore.ts
```diff
@@ -1057,11 +1063,11 @@ export class AppStore {
     };
 
     private recalculateStatsRequirements() {
-        if (!this.activeFrame) {
+        if (!this.requiredFrame) {
             return;
         }
 
-        const updatedRequirements = RegionWidgetStore.CalculateRequirementsArray(this.activeFrame, this.widgetsStore.statsWidgets);
+        const updatedRequirements = RegionWidgetStore.CalculateRequirementsArray(this.requiredFrame, this.widgetsStore.statsWidgets);
```
veggiesaurus: this is where the complexity lies with this feature. Previously, we assumed that the only changes to requirements were for the active frame. But now, when we change the effective frame, this will result in _two_ images changing their requirements, won't we?

veggiesaurus (changes requested):
Some comments, and possibly some complications with requirements, give it some thought. @kswang1029 might have run into a bug with this. 

One last thing, can we change to "Active"  instead of "Current", for consistency with the region? Perhaps we should change Stokes as well?

veggiesaurus:
@TienHao can you resolve conflicts in this branch? is it ready for re-review?

TienHao:
Not ready yet. I am working on checking stokes parameters.

veggiesaurus:
I'm not sure if the border should be shown when the region selected is "active". @kswang1029 what do you think? It's a bit annoying to constantly see a spectral profiler outlined. 

The other issue is that the border causes things to reflow when enabled/disabled:
border_reflow

kswang1029:
do you mean the outer broader (the entire widget)? we should remove it and just highlight "image", "region" and "stokes" dropdown when necessary.

veggiesaurus:
> do you mean the outer broader (the entire widget)? we should remove it and just highlight "image", "region" and "stokes" dropdown when necessary.

yes, that sounds better. But then @TienHao, can you make sure the border switches color only (from that colour to transparent) instead of switching on/off. That should prevent the reflow

veggiesaurus (approved):

PR #782, opened 2020-02-25
adjusts min percentage setting and fixes lazy rendering of contour di…

### LINKED ISSUE #671: Improve contour dialog generators' default paremters
The contour dialog's generators work as expected, but some more testing needs to be done to ensure that the default parameters are sensible and useful.

### LINKED ISSUE #781: Contour level settings are lost when I select configuration
In the contour widget, after I set my contour levels the way I want them I select configure to adjust smoothing. When I go back to levels my level settings are lost. The levels are back at the default settings. My setting on the levels should be preserved.

veggiesaurus:
Closes #671. Should also close #781, will as @arusstaylor to test after merging in. 

Also should auto-focus dialogs by default, so that they can be easily closed using the `Esc` key.


veggiesaurus:
@duidae is there any danger to auto-focusing any of the dialogs?

panchyo0 (approved):

PR #789, opened 2020-02-26
Jason/issue 770 2

### LINKED ISSUE #770: Remember spectral system choice when switching active images
Currently, whenever the active image is changed, the coordinate system for a spectral profile widget reverts to the system defined by the image's WCS. Even if I have explicitly selected another system. For example, if I have three images, all with `CTYPE3 = FREQ`, then there's no way to easily compare spectra as velocities, because every time I switch images, the widget defaults to frequency again. 

spectral_coordinate

I propose we add a new option to the dropdown: "WCS". When this is selected (by default), then the system is defined by the image's WCS. If I switch to an image with a different system, then the new image's WCS define the system choice. If I explicitly select another system type, then this persists even after switching images. If the new image doesn't support the given system type, it reverts back to "Channel"

duidae:
closes #770


veggiesaurus (approved):
I think this is a good way of solving #770. We'll see if users complain about this approach, but it makes sense to me. By default, we use WCS.

PR #808, opened 2020-03-02
replace splice with filter operation to trigger component re-render

veggiesaurus:
Filtering instead of splicing `@observable` array seems to solve the issue. I wonder if its because the levels are only used in a private function _called_ by render, and not inside render itself.


kswang1029:
looks good.

duidae (approved):

PR #809, opened 2020-03-03
Angus/fix 755

veggiesaurus:
This PR introduces a contour WebGL service, which creates a WebGL context on startup, allowing us to change layout wihtout losing context when the `ImageViewComponent` is unmounted. 

This PR needs to be backported to `release/1.3` as well on approval


veggiesaurus:
@confluence please check if this also stops all the WebGL warnings on Firefox

duidae (approved):

PR #812, opened 2020-03-04
Jason/issue 793

### LINKED ISSUE #793: Cannot show frequency axis if no conversion to velocity exists
On `dev`, we cannot select anything other than channel:

image

On `master`, we can always choose to use the native WCS: 
image

Example file: GALFACTS_N4_cube_Q_DQA3.3.1.fits (or any other image that doesn't have enough information for FREQ -> VRAD conversion, such as RESTFRQ)

duidae:
closes #793


src/models/SpectralDefinition.ts
```diff
@@ -60,15 +60,15 @@ export enum SpectralSystem {
...
 }
-export const IsSpectralSystemValid = (system: string): boolean => {
+export const IsSpectralSystemSupported = (system: string): boolean => {
     return system && (<any> Object).values(SpectralSystem).includes(system);
 };
 
 export const SPECTRAL_TYPE_STRING = new Map<SpectralType, string>([
-    [SpectralType.VRAD, "Radio velocity"],
-    [SpectralType.VOPT, "Optical velocity"],
+    [SpectralType.VRAD, "Velocity"],
```
veggiesaurus: shouldn't we stick with the WCS Paper III naming conventions for these?

duidae: I am trying to unify the names according to CHANNEL_TYPES(following FITS standard V4.0), since adding "Native WCS" label in dropdown will need to compare the strings of names

veggiesaurus: Yes, that's what I'm suggesting as well:
image

(to be clear, I mean using "Radio velocity" instead of "Velocity"

duidae: hmm I see, will change it

veggiesaurus (approved):
Code looks fine. @duidae please merge into `dev` and create a backport PR (once @kswang1029 has tested and approved)

kswang1029:
I will try to get this tested by tomorrow. need some time to go through possible cases again.

kswang1029:
Overall it works fine. There is a potential issue from casacore that it appears too smart to add missing headers based on existing headers (e.g. if OBSGEO is missing but we have TELESCOP, OBSGEO will be put back). This is a separate issue, not a stopper of this PR. Please proceed.

PR #818, opened 2020-03-05
Backport 793

duidae:


veggiesaurus (commented):
code looks fine, just need @kswang1029 to carefully test with release branch, because there are quite a few changes

kswang1029:
I tested with a set of images with some missing headers (set_spectral_conversion). In general it works fine. A peculiar thing I noticed is that depending on cases, casacore may "smartly" fill up missing headers if those can be computed from other relevant headers (e.g., an image with missing OBSGEO but having TELESCOP: casacore will return headers with OBSGEO recovered). Other thing is with a cube in VRAD but with RESTFRQ missing, the casacore returns in FREQ.

@pford Could you help to look at it? I will file an issue in the backend repo and we can continue there. @veggiesaurus please proceed with this PR.

Issue #795, opened 2020-02-27
Different widget help display depending if CARTA is in dark mode or light mode


ajm-ska:
I wonder if it would be possible to have a different widget help display to show up depending whether CARTA is in Light mode or Dark mode? 

I'm specifically referring to the **Image View** help as it shows some images captured in Light mode. I think it would be a neat touch if when using CARTA in Dark mode, that it could show the tool button images captured in Dark mode, plus have the first image with the Dark mode background colour and the text labels in white.

It is just a suggestion. I know it would mean extra work for @kswang1029 as he would need to create another set of images in dark mode.

kswang1029:
Creating screenshots in dark mode is no problem at all. :)

kswang1029:
I will need some help from @TienHao too

kswang1029:
done with #825

Issue #679, opened 2020-01-22
adding context-sensitive in-app help


kswang1029:
Currently the only way to access to the online user manual is via the "help" menu. For better UX, we should also provide in-app help manual with a question mark icon. For widgets, it should be in the header, next to the "cog". For dialog, it should be next to the close button (X). Once the help button is hit, we can use the blueprint "drawer" to show in-app help. For example, https://blueprintjs.com/docs/#core/components/drawer

The content for each widget and dialogue is TBD (@kswang needs to think about the maintainability of this in-app help and the online user manual).

kswang1029:
A crazy idea: if the position of the drawer can be determined dynamically based on the location of widget/dialogue so that the drawer will not block the widget/dialogue, that would be fantastic!

veggiesaurus:
> A crazy idea: if the position of the drawer can be determined dynamically based on the location of widget/dialogue so that the drawer will not block the widget/dialogue, that would be fantastic!

That shouldn't be _too_ hard. We can have a function that determines the best position (left, right) based on the bounding box of the widget. But the question is whether that would annoy users or not

veggiesaurus:
Setting estimate to 3 for the actual technical work. Adjustments to the docs themselves can be tracked on the doc repo, I suppose

kswang1029:
> That shouldn't be too hard. We can have a function that determines the best position (left, right) based on the bounding box of the widget. But the question is whether that would annoy users or not

indeed, let's not do that for now. "Right" should fit the common UX?

PR #825, opened 2020-03-11
Thchang/improve help drawer

TienHao:
#795 : changing help drawer pictures in different themes (dark theme picture not added yet)
#679 : automatically showing help drawers on left or right depends on widgets location

### DIFF
--- a/src/components/Dialogs/DraggableDialog/DraggableDialogComponent.tsx
+++ b/src/components/Dialogs/DraggableDialog/DraggableDialogComponent.tsx
@@ -24,7 +24,7 @@ export class DraggableDialogComponent extends React.Component<ResizableDialogCom
     componentDidUpdate() {
         const header = this.dd.getElementsByClassName("bp3-dialog-header");
         if (this.props.helpType && header.length > 0 && this.dd.getElementsByClassName("help-button").length === 0) {
-            const helpButton = <Button icon="help" minimal={true} onClick={() => this.props.appStore.helpStore.showHelpDrawer(this.props.helpType)}/>;
+            const helpButton = <Button icon="help" minimal={true} onClick={(ev: React.MouseEvent<HTMLButtonElement>) => this.onClickHelpButton(ev)}/>;
             const helpButtonDiv = document.createElement("div") as HTMLDivElement;
             helpButtonDiv.setAttribute("class", "help-button");
             ReactDOM.render(helpButton, helpButtonDiv);
@@ -35,6 +35,10 @@ export class DraggableDialogComponent extends React.Component<ResizableDialogCom
                 header[0].append(helpButtonDiv);
             }
         }
+    }
+
+    onClickHelpButton(ev: React.MouseEvent<HTMLButtonElement>) {
+        this.props.appStore.helpStore.showHelpDrawer(this.props.helpType, ev.clientX);
     }
 
     render() {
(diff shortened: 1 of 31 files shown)


src/components/Dialogs/DraggableDialog/DraggableDialogComponent.tsx
```diff
@@ -24,7 +24,7 @@ export class DraggableDialogComponent extends React.Component<ResizableDialogCom
     componentDidUpdate() {
         const header = this.dd.getElementsByClassName("bp3-dialog-header");
         if (this.props.helpType && header.length > 0 && this.dd.getElementsByClassName("help-button").length === 0) {
-            const helpButton = <Button icon="help" minimal={true} onClick={() => this.props.appStore.helpStore.showHelpDrawer(this.props.helpType)}/>;
+            const helpButton = <Button icon="help" minimal={true} onClick={(ev: React.MouseEvent<HTMLButtonElement>) => this.onClickHelpButton(ev)}/>;
```
veggiesaurus: can't you just use `onClick={this.onClickHelpButton}` after turning `onClickHelpButton` into an arrow-function? Anonymous functions should be avoided if possible in react, as they always trigger re-rendering

src/components/FloatingWidget/FloatingWidgetComponent.tsx
```diff
@@ -133,7 +137,7 @@ export class FloatingWidgetComponent extends React.Component<FloatingWidgetCompo
                     </div>
                     }
                     {widgetConfig.helpType &&
-                    <div className={buttonClass} onClick={() => appStore.helpStore.showHelpDrawer(widgetConfig.helpType)}>
+                    <div className={buttonClass} onClick={(ev) => this.onClickHelpButton(ev)}>
```
veggiesaurus: same suggestion here

veggiesaurus (changes requested):

veggiesaurus:
@kswang1029 did we resolve to automatically pop up the help drawer on the opposite side to the widget? or let the user decide?

kswang1029:
I think we should try auto.

TienHao:
> @kswang1029 did we resolve to automatically pop up the help drawer on the opposite side to the widget? or let the user decide?

I have implemented automatically left/right popping up by the horizontal location of the help button in widgets and dialogues.

veggiesaurus:
@TienHao what is the logic behind which location it chooses? I'd have thought it would be related to whether the center of the widget is to the left or right of the midpoint, but that's not happening:

image

I see it depends on where you actually _click_ the button, rather than the position of the button itself. I don't think that makes sense. But it's also not switching at the midpoint, so I'm confused as to what the intended logic is.

TienHao:
> I see it depends on where you actually _click_ the button, rather than the position of the button itself. I don't think that makes sense. But it's also not switching at the midpoint, so I'm confused as to what the intended logic is.

@veggiesaurus My previous thought is to use the position of button to decide the help drawer location without blocking the help button.(that did make sense to me at that point. :grin: I think yours better. ) Since the width of drawer is 33%, if the position of help button is over 65% width of browser, the drawer will pop up on left, and less then 65% will pop up on right. 

I have changed the logic to the center of the widget now.

veggiesaurus:
I think we should also consider not applying the dimming overlay, so that you can still see the widget properly

veggiesaurus (approved):
this looks good. Can you also make a PR for `release/1.3`

PR #827, opened 2020-03-12
Add rigorous check (dev)

duidae:
test file: 18h03m_tabular_axis.image

spectral header is invalid for AST to convert data, so should only enable native wcs & channel

### DIFF
--- a/src/components/Shared/SpectralSettings/SpectralSettingsComponent.tsx
+++ b/src/components/Shared/SpectralSettings/SpectralSettingsComponent.tsx
@@ -3,7 +3,7 @@ import {observer} from "mobx-react";
 import {FormGroup, HTMLSelect, IOptionProps} from "@blueprintjs/core";
 import {SpectralProfileWidgetStore, StokesAnalysisWidgetStore} from "stores/widgets";
 import {AppStore} from "stores";
-import {SpectralSystem, SPECTRAL_COORDS_SUPPORTED} from "models";
+import {SpectralSystem} from "models";
 import "./SpectralSettingsComponent.css";
 
 @observer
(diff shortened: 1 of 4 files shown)


veggiesaurus:
@pford does this fix the issue you brought up in the frontend channel?

pford:
Shouldn't it revert to channel as before?

veggiesaurus:
Yes. Is that not what it's doing?

pford:
The spectral profile is empty, Value vs Channel with default ranges 0 - 1.
I tested frontend origin/jason/refine_check_dev and backend origin/dev 22b5f9d (before my FREQ fix)

veggiesaurus:
@duidae please check. I agree with @pford. In fact, You should always be able to view the spectral axis as channel or whatever the native type is, even if AST doesn't recognize the type.

duidae:
kk will resume plot & disable conversion

duidae:
@veggiesaurus @pford @kswang1029  ok, please have a look, thanks.

pford:
@duidae looks good!  Plots the spectral profile vs channel.

arusstaylor:
If spectral frame is empty one should still be able to due basic wcs conversion, i.g convert channel to frequency and wavelength.

pford:
@arusstaylor this fix is for a missing or non-standard CTYPE header for the spectral axis.

veggiesaurus:
> If spectral frame is empty one should still be able to due basic wcs conversion, i.g convert channel to frequency and wavelength.

If the spectral frame is non-standard, the only thing we should be able to do is convert from channel to whatever the spectral frame is (same as DS9). Converting from that frame to another (e.g. freq -> wavelength) should require that the image adheres to WCS standards

duidae:
@veggiesaurus @kswang1029  changed to show only native wcs + channel when cube's spectral axis is invalid (like DS9), please have a look & test, thanks.

kswang1029:
verified. 👍

veggiesaurus (approved):
this looks fine. I'll merge into `dev`, please make a backport for 1.3, which @kswang1029 will test carefully to ensure no other issues crop up

kswang1029:
I tried with cubes 
with CTYPE3 = XYZ and CUNIT3 = Hz (working)
with CTYPE4 = XYZ and CUNIT4 = Hz, and CTYPE3 = STOKES and CUNIT3 = ' ' (backend does not return headers and provides an error message)
with CTYPE3 = XYZ and CUNIT3 = GigaHertz (backend does not return headers and provides an error message)

For the frontend part I think it is good to go. Now the question is do we want to support this kind of exception in the backend header parser?

PR #839, opened 2020-03-24
add SafeNumericInput component

### LINKED ISSUE #615: NaN appears in numeric input
when entering "-1" to various numeric inputs in current CARTA, NaN appears. On ubuntu platform, it can be reset by entering a normal number, however, on Mac, users could be totally stucked.

image

veggiesaurus:
This PR adds a `SafeNumericInput` component, which checks whether the value is finite before calling `onValueChange`. 

This fixes #615 in all of the spots that I could find (please check that I haven't missed any)

### DIFF
--- a/src/components/Animator/AnimatorComponent.tsx
+++ b/src/components/Animator/AnimatorComponent.tsx
@@ -1,9 +1,10 @@
 import * as React from "react";
 import {observer} from "mobx-react";
 import {action, observable} from "mobx";
-import {Button, ButtonGroup, FormGroup, IconName, Menu, MenuItem, NonIdealState, NumberRange, NumericInput, Popover, Position, Radio, RangeSlider, Slider, Tooltip} from "@blueprintjs/core";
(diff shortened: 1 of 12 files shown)


duidae (approved):

kswang1029:
@veggiesaurus NaN is gone. 👍 

I noticed that when a text field is
1) empty
2) "-"
3) "."
4) "+"

the reset default button does not work.
When the text field is one of the above, we can still generate contour levels.

veggiesaurus:
> @veggiesaurus NaN is gone. 
> 
> I noticed that when a text field is
> 
> 1. empty
> 2. "-"
> 3. "."
> 4. "+"
> 
> the reset default button does not work.
> When the text field is one of the above, we can still generate contour levels.

This change shouldn't affect clear-able fields. Can you check with `dev`

kswang1029:
@veggiesaurus ok, you are right, dev has the same thing. OK, for NaN fix, I think it is effective. :)

veggiesaurus:
> @veggiesaurus ok, you are right, dev has the same thing. OK, for NaN fix, I think it is effective. :)

can you file a separate issue for this?

kswang1029:
done: https://github.com/CARTAvis/carta-frontend/issues/840

Issue #834, opened 2020-03-23
make layer list widget more obvious in the default layout 


kswang1029:
Based on the v1.3 beta test results, there are quite some complaints on the intuitiveness of enabling WCS matching. A possible solution (as discussed with @veggiesaurus) to improve this for 1.3 is to make the layer list widget more obvious to users like the following (replacing spatial profile widgets by layer list widget and region list widget). This can save internet bandwidth too since the default cursor is not fixed so users may request lots of spatial profile update unintentionally. 

<img width="1278" alt="Screen Shot 2020-03-23 at 2 45 13 PM" src="https://user-images.githubusercontent.com/20819712/77299491-7c9c6d80-6d27-11ea-8536-81dec70320f3.png">

@arusstaylor do you have concerns on making this change in v1.3? We can review the usability afterward but for v1.3 we have to have a way to let users know how to match images.

duidae:
@kswang1029  is this change specifically for "Default" in presets?

kswang1029:
@veggiesaurus @arusstaylor are you okay with this proposed change specifically for v1.3?

veggiesaurus:
I'm ok with that, because users can easily change their preferred layout or save a new preset

arusstaylor:
Not sure about this. This default only makes sense if you append an image. So could make sense in the context of smart layouts but does not make sense as the default for a single image. 
 Think this is a deeper UI issue.

kswang1029:
@arusstaylor @veggiesaurus thanks for the input. The idea of having this is because some beta testers cannot easily find out how to trigger WCS matching. I agree this is a deeper UI issue but as a quick and dirty improvement just for the coming v1.3 release, this approach could make the layout list widget directly visible to users so that they may trigger WCS matching easily. Surely smart layout could help.

kswang1029:
a temporary solution has been applied in v1.3. This should be revisited in usability review. Close for now.

PR #843, opened 2020-03-26
Switch layer list to active tab in preset default

duidae:
work around for #834


kswang1029:
@arusstaylor @veggiesaurus as an alternative, we just make the layer list widget active by default. It was an inactive tab. Please see if this is acceptable for v1.3.

veggiesaurus:
I like this approach :+1:

veggiesaurus (approved):

PR #844, opened 2020-03-26
Thchang/modify contour parameters ui

### LINKED ISSUE #836: narrow text fields in the contour dialogue
Based on the users feedback of v1.3 beta test, some text fields in the contour dialogue are not wide enough to display numbers properly. The example below is to load an image with values ~ 1E-10.
Screen Shot 2020-03-24 at 10 45 42 AM

As a quick improvement for v1.3 release, some text fields belonging to different level generators are needed to be widen. To support a wide range of numbers, we display it as "%.4E". 

@veggiesaurus @arusstaylor ok to do this for v1.3?
@TienHao when permission granted, could you have a look?

### LINKED ISSUE #837: [question] narrower contour dialogue
based on v1.3 beta test, users complains about the contour dialogue being too big to block widgets. As a quick solution, possibly we can make the contour dialogue narrower by changing the "Parameters" row into two rows, like the following mockup (certainly with proper alignment):
Screen Shot 2020-03-24 at 11 39 22 AM

Except "mean-sigma-list" generator, all other three generators have four fields in "Parameters". So for these three generators, we use two rows to display the four fields (two in each row).

@veggiesaurus @arusstaylor ok to do this for v1.3? Later we can redesign the UI if we have better ideas on using space more efficiently.

TienHao:
Fixed #837 : narrowed contour dialogue and separated parameters into two rows.
Fixed #836 : turn clearable input into scientific format


veggiesaurus (changes requested):
@TienHao this looks good, but one minor issue:
image

It's rather annoying that the controls don't have the same width. It would look much nicer if the controls had the same widths, so that their right edges lined up as well

TienHao:
@veggiesaurus I have modified the css of fields and turned clearable input into scientific format(#836 )
The field will display float number when the input is focused. 

ezgif com-video-to-gif

veggiesaurus (commented):
`ClearableNumericInput.tsx` is no longer needed, is it? Can we remove it

TienHao:
> `ClearableNumericInput.tsx` is no longer needed, is it? Can we remove it

Yeah, it's no longer needed. I have removed it.

veggiesaurus (approved):

PR #847, opened 2020-03-27
fixes #864 by re-centering image before fitting zoom

### LINKED ISSUE #864: Animating of matched images
Images that are matched spectrally need to be animated. In order do to this, changes to the ICD will need to be made.
We will need a way of the backend knowing which channel to switch to when animating across secondary images.

veggiesaurus:
simple fix to re-center images before fitting zoom.


kswang1029:
confirmed 👍

Issue #787, opened 2020-02-26
unify AST usages thoroughly in the frontend codebase


kswang1029:
Currently AST is used semi-independently elsewhere in the frontend codebase. It would be good to review and unify AST usages thoroughly in the frontend codebase for 
1) better code maintainability 
2) consistent AST calculations
3) re-usability for future feature development

 aiming for v1.4

kswang1029:
These are possible AST usages:

FITS header integrity check
spatial reference frame conversion
Spectral reference frame conversion
Intensity unit conversion 
Guard for invalid conversions 
Image FoV coverage check

veggiesaurus:
I also think we should move towards reading the entire FITS header, and then extracting the sky and spectral frames, as @dsberry suggested

veggiesaurus:
additional chore: update to AST 9.1.1 (after bugfix)

kswang1029:
@veggiesaurus need a status update.

PR #852, opened 2020-03-30
update AST to 9.1.1 & simplify AST initialization for spectral conversion

duidae:
this PR includes:
1. update AST to 9.1.1 
2. simplify AST initialization for spectral conversion

related to #787


wasm_src/ast_wrapper/ast_wrapper.cc
```diff
@@ -80,129 +80,47 @@ EMSCRIPTEN_KEEPALIVE AstFrameSet* initFrame(const char* header)
...
+    AstFrameSet* found = static_cast<AstFrameSet*>astFindFrame(frameSet, spectralTemplate, " " );
+    if (!found)
     {
-        cout << "Some AST LIB error, check logs." << endl;
-        astClearStatus;
+        cout << "Spectral frame not found." << endl;
         return nullptr;
     }
-    else if (specframe == AST__NULL)
+    AstSpecFrame *specframe = static_cast<AstSpecFrame*>astGetFrame(found, AST__CURRENT);
```
veggiesaurus, line 89: Why don't we rather load all the headers into a single frame set, and _then_ split it out into a spectral and spatial frame?

veggiesaurus: For example, `initFrame` could init the entire `AstFrameSet`, and then `getSpatialFrame` and `getSpectralFrame` could pull out the appropriate frames

veggiesaurus (changes requested):
This looks much better than our previous approach, but there's a problem with loading images with no WCS info (e.g. `LMCq.fits`). Seems to cause AST to crash.

duidae:
kk, will look into it after vacation(next week)

duidae:
using dev branch, the error also show up in AST ver. 8.7.1:

astGrid(Plot): No numerical labels can be produced for axis 1 using the supplied Frame format string 'd.*'

probably some changes between 8.6.3/8.7.1 about args for astPlot() Format(1)='d.\*', Format(2)='d.\*'

will try to fix it

veggiesaurus:
> using dev branch, the error also show up in AST ver. 8.7.1:
> 
> astGrid(Plot): No numerical labels can be produced for axis 1 using the supplied Frame format string 'd.*'
> 
> probably some changes between 8.6.3/8.7.1 about args for astPlot() Format(1)='d._', Format(2)='d._'
> 
> will try to fix it

One possible solution would be to not apply the `Format` arguments when the WCS is invalid

duidae:
There are 2 issues:
1.  in 8.6.3, AST initFrame() returns invalid after reading the headers of LMCq.fits, however, after ver. 8.7.1, it returns ok, so the dummy frame will not be created.

2. originally 'Format(1)=d.\*' is accepted in  8.6.3, but '\*' seems no longer available after 8.7.1

so a simple fix would change (2) by removing '*', and the axis will be like:

image

and I think for future refactoring, initWCS() & initFullWCS() could be integrated for simplicity.

@veggiesaurus do you think the solution above ok?

veggiesaurus:
> There are 2 issues:
> 
> 1. in 8.6.3, AST initFrame() returns invalid after reading the headers of LMCq.fits, however, after ver. 8.7.1, it returns ok, so the dummy frame will not be created.
> 2. originally 'Format(1)=d.*' is accepted in  8.6.3, but '*' seems no longer available after 8.7.1
> 
> so a simple fix would change (2) by removing '*', and the axis will be like:
> 
> image
> 
> and I think for future refactoring, initWCS() & initFullWCS() could be integrated for simplicity.
> 
> @veggiesaurus do you think the solution above ok?

That sounds sensible. For (1), can we just check whether there's a spatial frame using AST? And if not, update the "validWCS" flag in FrameStore?

veggiesaurus:
@duidae is this ready for a re-review yet?

duidae:
@veggiesaurus  still working on spatial part, will notify you when ready, thanks.

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -187,16 +189,18 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
         if (data.length > 0 && frame && frame.channelInfo) {
             let channelInfo = frame.channelInfo;
             const zIndex = this.matchZindex(x, y, data);
-            let nearestIndex;
+            let nearestIndex = undefined;
```
veggiesaurus: by default, variables are `undefined`

duidae:
just tried David's approach to extract skyframe and replace original wcsinfo with skyframe, however it breaks the AST ticks, not sure what is going on, so I think this PR can only handle the spectral part, spatial part refactoring needs more investigation. (so far the spatial functionality  still works using previous implementation).

original:
image

simply replace wcsinfo with skyframe:
image

src/stores/FrameStore.ts
```diff
@@ -880,6 +863,13 @@ export class FrameStore {
         }
     };
 
+    public invertToNativeWCS = (value: number): number => {
```
veggiesaurus: should this be "convert"?

duidae: Fixed

veggiesaurus (changes requested):
I must have skimmed over your message before, because I see now that the issues for LMCq aren't actually fixed, since it's now just assigning the letter "d" to the ticks, instead of just displaying the ticks as pixel values, which it used to do. For images with no WCS, it should revert to pixel coords, like it used to.

Part of the issue might be that now the WCS is constructed correctly, so no dummy frame is initialised. Technically, it does have a valid transform (from pixels to some unknown type), so perhaps we should use that? 

@kswang1029 the header itself doesn't have a CTYPE or CDELT but casacore seems to be inserting a CDELT of 1.0 by default. Not really sure how to handle situations like this...

duidae:
@veggiesaurus  tried a workaround for LMCq.fits  in commit b5d6e37b0be7eabd56e54db6365817d95612c017, please have a look to see if it's ok, thanks.

veggiesaurus (changes requested):
I think that check might be too specific. I think there's a simper approach: If the FrameSet that AST generates has a `Domain` attribute of `"SKY"`, then it can be displayed with degrees or HMS. If not, then the coordinates should _not_ be formatted as degrees etc.

LMCq for example, gives a domain of `"AXIS1-AXIS2"`

duidae:
changed to avoid sky-missing frame

veggiesaurus (approved):

PR #859, opened 2020-03-30
fixes display of zero in sigma list

veggiesaurus:
Converts numbers to strings to prevent `0` being hidden


veggiesaurus:
(note only one changed line, apart from whitespace changes)

kswang1029:
Looks good!!!

PR #860, opened 2020-03-31
Jason/region cancel

### LINKED ISSUE #511: Region creation should be cancellable by pressing the ESC key
When creating a region, it would be nice to be able to cancel the creation of the region by pressing the ESC key, rather than creating the region and deleting it.

duidae:
closes #511


kswang1029:
for rectangle and ellipse, it looks good. For point, pressing "esc" should bring back the pan mode. For polygon, pressing "esc" will remove all previous anchors but it is still in the region creation mode. I suggest to make all regions have the same behavior, that is, when "esc" is pressed, go back to pan mode.

veggiesaurus:
@duidae is this ready for re-review?

duidae:
still in progress

duidae:
@kswang1029  @veggiesaurus  done, all changed to pan mode after pressing ESC, please have a look, thanks.

kswang1029:
just to double check: for point region, esc key does not switch to pan mode? It appears that esc key works only when we enter the state that a region is _being_ created. For point region there is no such state.

duidae:
for point region, ESC works when the left click of the mouse is not released yet (because the point region is created when left click of the mouse is released)

kswang1029:
I see. 👍

PR #869, opened 2020-04-15
Add tab to spectral/stokes widget settings panel

duidae:
Add tabs to spectral/stokes widget settings panel for supporting smoothing/moments tabs in the future, and TienHao will do the follow up in spatial profiler if this PR is ok.

image


kswang1029:
Screen Shot 2020-04-15 at 2 47 19 PM
Would it be better to have three tabs as "Styling", "Range", and "Tool"? @duidae @veggiesaurus 
Part reason is later there will be multiple profiles in one plot. So in Styling tab, we need the ability to adjust the styling of each profile, which may need some space.

duidae:
@kswang1029 yeah I tried to split "Line Plot" tab, however the line plot settings panel is a shared component among a few widgets, such as spatial profiler, render config, histogram. And it would be related modifications, so need to think about how to do it. Could be an enhancement after this.

veggiesaurus:
> @kswang1029 yeah I tried to split "Line Plot" tab, however the line plot settings panel is a shared component among a few widgets, such as spatial profiler, render config, histogram. And it would be related modifications, so need to think about how to do it. Could be an enhancement after this.

For the spectral profile, it looks a bit odd having a "line plot" tab, since that's the only plot there. Perhaps the spectral profile settings should have "Styling" tab, and the Stokes widget should have "Line Plot Styling" and "Scatter Plot Styling" tabs?

duidae:
changed to Styling/Line Plot Styling/Scatter Plot Styling

veggiesaurus (approved):

Issue #868, opened 2020-04-14
catalog overlay single source highlighting


panchyo0:
Single source selection in catalog table view, scatter view and image view. 
- [x] User select a row in table view, it will highlight the data in table, scatter and image.
- [x] User select a point in scatter plot, it will highlight the data in table, scatter and image.
- [x] User select a point in image overlay, it will highlight the data in table, scatter and image.

veggiesaurus:
This is partially implemented in the latest PR, but I think we still need to think about how to highlight the selected point in the image. I'd prefer to just display an additional outline around the point, rather than hiding all other points.

panchyo0:
User can customer the shape, should we do different outline around for different shape? Should we keep the selection action consistent between scatter plots and image overlay? We can change the color of selected source to make it obvious. since user can custom the catalog color, not sure which color we should use.

Issue #867, opened 2020-04-14
default catalog colour shouldn't clash with default colourmap


veggiesaurus:
The default colour for catalog points is currently red. This makes points very difficult to see when used with the default colourmap (inferno). 

I suggest we use `#00B3A4` (already on the swatch)

![image](https://user-images.githubusercontent.com/592504/79223107-ffad7100-7e58-11ea-858d-16f1734cbfe7.png)

PR #886, opened 2020-04-18
Qi/catalog single source selection

panchyo0:
1. fixed issue [867](https://github.com/CARTAvis/carta-frontend/issues/867).
2. fixed bug when user click refresh button in catalog file browser.
3. single source selection in image view. Issue [868](https://github.com/CARTAvis/carta-frontend/issues/868).


veggiesaurus (approved):
This looks fine, and I'll merge it in, but I think we still need to explore how best to highlight the single sources. It might make more sense to just show them with an additional selection circle

PR #902, opened 2020-04-28
optimisations of WCS transforms and catalog plot data arrays

veggiesaurus:
- makes a single `AST` call for a `Float64Array` of coordinates, rather than transforming one-by-one in separate WebAssembly calls
- stores data for plotly as `Float64Array`, allowing for much faster concatenation using `set`, instead of manually pushing back values to the array.

### DIFF
--- a/src/stores/CatalogStore.ts
+++ b/src/stores/CatalogStore.ts
@@ -9,8 +9,8 @@ type CatalogSettings = {
     color: string,
     size: number,
     shape: CatalogOverlayShape,
-    xImageCoords: Array<number>[],
-    yImageCoords: Array<number>[],
+    xImageCoords: Array<Float64Array>,
+    yImageCoords: Array<Float64Array>,
     selectedPointIndexs: Array<number>,
     showSelectedData: boolean;
 };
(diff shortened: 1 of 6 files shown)


src/stores/FrameStore.ts
```diff
@@ -683,6 +683,20 @@ export class FrameStore {
...
+        //
+        // for (let i = 0; i < N; i++) {
+        //     xWCSValues[i] = 2.6349675579325287 + 0.0001 * i;
+        //     yWCSValues[i] = 0.024208492616605525 + 0.0001 * i;
+        // }
+        //
+        // console.log(xWCSValues);
+        // console.log(yWCSValues);
+        // const results = AST.transformPointArrays(this.wcsInfo, xWCSValues, yWCSValues, false);
+        // console.log(results);
```
panchyo0: This looks like a test for AST.transformPointArrays. Should we remove it?

src/stores/CatalogStore.ts
```diff
@@ -144,15 +144,18 @@ export class CatalogStore {
                 yFraction = Math.PI / 180.0;
```
panchyo0, line 144: I use degree as default unit, if the unit of the column is null in catalog file. Should we keep this way?

panchyo0 (approved):

PR #904, opened 2020-04-29
Fix 887 for 1.3 patch

### LINKED ISSUE #887: incorrect or inaccurate angular size of rectangle and ellipse regions in the region dialogue
In the region dialogue, angular sizes of ellipse region or rectangle region are displayed next to the sizes in image coordinate. These values, however, are not correct for ellipse or are not accurate for rectangle. See the following screenshots. 

Screen Shot 2020-04-20 at 1 19 56 PM

Screen Shot 2020-04-20 at 1 20 09 PM

In addition, the unit of the angular size should be dynamic based on the angular size. For example, in the above cases, it is preferred to use arcsec, instead of arcmin.

duidae:
closes #887

### DIFF
--- a/src/components/Dialogs/RegionDialog/EllipticalRegionForm/EllipticalRegionForm.tsx
+++ b/src/components/Dialogs/RegionDialog/EllipticalRegionForm/EllipticalRegionForm.tsx
@@ -113,7 +113,7 @@ export class EllipticalRegionForm extends React.Component<{ region: RegionStore,
     private getSizeString(size: Point2D) {
         const wcsSize = this.props.frame.getWcsSizeInArcsec(size);
         if (wcsSize) {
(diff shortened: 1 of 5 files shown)


kswang1029 (approved):
Tested and it looks good!

kswang1029:
Oops, found a small issue:
Major axis of an ellipse points to North in astronomical convention (and also CRTF definition). So we need to swap the derived angular sizes in the text.
Screen Shot 2020-04-30 at 10 39 23 AM
Screen Shot 2020-04-30 at 10 39 37 AM

That is, the two text input fields are ordered as major(vertical direction)-minor(horizontal direction). So the following derived angular sizes need to be swapped.

duidae:
@kswang1029 fixed, please have a look, thanks.

kswang1029:
@duidae Looking great! Thanks 👍

Issue #898, opened 2020-04-23
load button needs to be disabled while catalog is loading


veggiesaurus:
At the moment, you can press the `Load Catalog` button as many times as you want, and each time submits another `OPEN_CATALOG_FILE` request. The button should be disabled while loading the catalog, similarly to how the `Load` button is disabled between `OPEN_FILE` and `OPEN_FILE_ACK` messages.

panchyo0:
closed with PR 907

Issue #876, opened 2020-04-16
Additional symbols for catalog shapes


veggiesaurus:
we probably can apply some rotations to these symbol shapes like the attached image. For closed shape, we can set it as filled or not.
<img width="889" alt="Screen Shot 2020-03-04 at 6 28 10 PM" src="https://user-images.githubusercontent.com/20819712/75870663-42396200-5e46-11ea-96bf-d2a7ccf32280.png">

_Originally posted by @kswang1029 in https://github.com/CARTAvis/carta-frontend/issues/814#issuecomment-594442762_

(In addition, the dropdown for selecting the shape should be changed to a BlueprintJS "Select", and should show the symbol shape, rather than the name)

panchyo0:
Should we also update the catalog file dropdown and system dropdown to BlueprintJS "Select"?
UI for syetem dropdown:
<img width="1232" alt="Screen Shot 2020-04-28 at 7 49 24 PM" src="https://user-images.githubusercontent.com/22226157/80554237-ba755d00-8989-11ea-98f6-c1d7faa83e8a.png">
UI for shape dropdown:
<img width="1232" alt="Screen Shot 2020-04-28 at 7 49 01 PM" src="https://user-images.githubusercontent.com/22226157/80554263-ccef9680-8989-11ea-80dd-0ff6f462afdf.png">

We can draw all shapes in Plotly (https://plotly.com/~dmiller123/4/name-circle-name-circle-open-name-circle-dot-name-circle-open-dot-name-square-na/#plot), However Blueprint icon doesn't have all icons (like triangle and hexagon). We might need to create our own SVG icons or using a static png file.

low-sky:
I think(5-pointed) star symbol would be good to include.  

Looking to the future, we'll also want to draw ellipses through this interface but that would be tricky.  For example, lots of astronomical source catalogs fit an ellipse to sources so the symbols should map the major, minor, and position angle from a catalog to the same properties on plotting symbol.  Like so:
<img width="566" alt="Screen Shot 2020-04-29 at 1 33 35 PM" src="https://user-images.githubusercontent.com/1139910/80638657-0b369580-8a1e-11ea-8f88-0826b9d80872.png">

arusstaylor:
agreed.  this is very useful.  But would require is to parse properties from multiple columns of a source catalogue.

veggiesaurus:
[...]

PR #907, opened 2020-05-01
Qi/catalog 876 & 898

panchyo0:
1. fixed issue [898](https://github.com/CARTAvis/carta-frontend/issues/898)
2. added additional symbols for catalog shape. Issue [876](https://github.com/CARTAvis/carta-frontend/issues/876)


veggiesaurus (changes requested):
one minor issue: the popups for the selects should be consistent with other selects in the UI (e.g. `minimal: true` and `position auto-end`)

veggiesaurus (approved):

PR #914, opened 2020-05-06
minor WebAssembly library build adjustments

### LINKED ISSUE #913: Emscripten emcmake/emconfigure
When using emsdk version 1.39.10 (and most likely all later versions) there is a minor issue with the `wasm_libs/build_zfp.sh` script:
```
Building ZFP using Emscripten
error: use `emcmake` rather then `emconfigure` for cmake projects
```
Changing `emconfigure` to `emcmake` as it suggests on line 15, fixes the issue.

veggiesaurus:
closes #913

### DIFF
--- a/wasm_libs/build_ast.sh
+++ b/wasm_libs/build_ast.sh
@@ -14,7 +14,7 @@ cd ast
 patch -i ../0001-New-attribute-TextGapType-for-the-plot-class.patch
 
 echo "Building AST using Emscripten"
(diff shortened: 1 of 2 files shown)


ajm-ska (approved):
I tried on MacOS with two Emscripten versions; An old 1.39.4, and the latest 1.39.14. 
I built the angus/fix_913 branch from scratch up to the `npm start` stage, and it appears to work fine in both cases.

PR #918, opened 2020-05-11
Angus/test debug execution

veggiesaurus:
Implements basic scripting support. Companion PR of https://github.com/CARTAvis/carta-backend/pull/523.

Scripting can also be called without backend support by using the "debug execution" dialog in the toolbar. The scripting works by calling functions that exist relative to the `AppStore`. For example, `openFile("", "C.fits")` will open a file in the base directory with a filename `"C.fits"`

Some adjustments to the typescript config were needed, and changing to a target of `ES6` lead to an issue with the BlueprintJS hotkey wrapper, which required a workaround, included here.


kswang1029:
Short question: where can we get a full set of available commands?

veggiesaurus:
> Short question: where can we get a full set of available commands?

For now: reading the source code... I think we want to make some auto-generated docs though.

kswang1029:
Please let me know if I might help on this part. We will need this info for the user manual for sure. 🙂

veggiesaurus:
> Please let me know if I might help on this part. We will need this info for the user manual for sure.

We need to think about what we're documenting though. The high-level python interface, or the low-level commands?

confluence:
There are two separate but overlapping issues here: end-users of wrappers need to know what high-level wrapped commands they can use in their scripts, and developers of wrappers need to know what low-level commands they can use inside the wrappers. In the case of direct execution in the debug console, there *is* no wrapper, and in a wrapper which deliberately exposes the low-level interface (like the generic Python wrapper) users can use a mix of high-level and low-level commands (when there are no high-level commands to do something that they need).

My feeling is that the low-level commands are mostly a tool for developers and advanced script writers, and that it doesn't make sense to start documenting them manually now. We're still making alterations to the frontend functions to make them compatible with scripting, and I expect that it will be some time before this interface is completely stable.

I do think that it would be a good idea to generate documentation for the frontend automatically, and manually annotate the code in whatever way is necessary to generate readable documentation (I'm not sure how Typescript libraries do this; I'm assuming something similar to Python docstrings). Then developers and advanced users will be able to look up simple things like function signatures without having to dig through the source code to find them, but we will be able to update this resource automatically when we change something.

src/App.tsx
```diff
@@ -225,7 +224,7 @@ export class App extends React.Component<{ appStore: AppStore }> {
             <Hotkey key={0} group={fileGroupTitle} global={true} combo={`${modString}O`} label="Open image" onKeyDown={() => appStore.fileBrowserStore.showFileBrowser(BrowserMode.File)}/>,
             <Hotkey key={1} group={fileGroupTitle} global={true} combo={`${modString}L`} label="Append image" onKeyDown={() => appStore.fileBrowserStore.showFileBrowser(BrowserMode.File, true)}/>,
             <Hotkey key={1} group={fileGroupTitle} global={true} combo={`${modString}W`} label="Close image" onKeyDown={() => appStore.closeCurrentFile(true)}/>,
```
panchyo0: Should we make the key unique? both `Append image` and `Close image` have key = 1

panchyo0 (approved):

duidae (approved):

Issue #880, opened 2020-04-17
[question] multi-selection in the catalog table


kswang1029:
Now users can select multiple sources in the 2d scatter plot of a catalog file and those selected sources will be highlighted in the image and in the table. Is there any way to select multiple sources in the catalog table (for example, with shift key to select consecutive sources and with ctrl/cmd key to select distinct sources on click)? 

<img width="1747" alt="Screen Shot 2020-04-17 at 2 46 07 PM" src="https://user-images.githubusercontent.com/20819712/79540394-ae042100-80ba-11ea-907f-79fc7be27ae9.png">

Issue #916, opened 2020-05-08
make source selected in catalog table all visible in the image view when catalog overlay is enabled


kswang1029:
Currently when catalog overlay is enabled in the image viewer, selecting a source in the image view will highlight the source in the catalog table and the highlighted source is visible in the table view. Similar behavior should be applied in the opposite way:

1. when catalog overlay is not enabled, selecting a source or multiple sources in the catalog table  results in nothing changed in the image view
2. when catalog overlay is enabled, selecting a single source in the catalog table results in changing the image view so that the selected source is centered in the image view without changing zoom level (effective only pan action is applied).
3. when catalog overlay is enabled, selecting multiple sources in the catalog table results in zooming and panning the image so that all selected sources fit in the image view. This can be achieved by deriving a bound box of all selected sources and translate it to have proper pan and zoom factor.

item "3." has a potential issue that the the spatial coverage of a catalog might be larger than the spatial coverage of an image.

@low-sky @Jordatious @veggiesaurus @arusstaylor @panchyo0 suggestions are welcome

veggiesaurus:
I think (3) would be frustrating to the user, because it would adjust the zoom for the image, rather than the selected sources, which may be clustered in one part of the image, or may extend past the image. I don't think there are likely to be issues with performance if we simply calculate the XY bounds (using existing performance-optimised functions), and use those bounds to determine the appropriate center and zoom.

kswang1029:
Using bound is ideal if performance is not an issue with millions of sources.

veggiesaurus:
> Using bound is ideal if performance is not an issue

Based on my previous [jsperf](https://jsperf.com/minmaxarraytesting/), calculating min/max for a 1e6 point array (both X and Y) should take about 27 ms. If we need faster performance, we can also use WebAssembly

kswang1029:
> I think (3) would be frustrating to the user, because it would adjust the zoom for the image, rather than the selected sources, which may be clustered in one part of the image, or may extend past the image. I don't think there are likely to be issues with performance if we simply calculate the XY bounds (using existing performance-optimised functions), and use those bounds to determine the appropriate center and zoom.

have modified the issue accordingly.

veggiesaurus:
@kswang1029 please confirm this is fixed after merging in #931

kswang1029:
confirm fixed after merging in #931

Issue #909, opened 2020-05-04
catalog filtering: use tooltip to show example filter syntax


veggiesaurus:
On a related feature, maybe we should use tooltip to show example filter syntax when hovering over the filter fields. The content of tooltip should dynamically adjusted based on the types of columns (number, string, etc)

Update: on a second thought, why don’t we just use those placeholder texts for filter syntax examples?

_Originally posted by @kswang1029 in https://github.com/CARTAvis/carta-frontend/issues/908#issuecomment-623373275_

veggiesaurus:
> Update: on a second thought, why don’t we just use those placeholder texts for filter syntax examples?

Using a tooltip would allow for a very clear explanation with examples of the different filter types

veggiesaurus:
@kswang1029 I think we could still have clearer explanations and examples in the tooltips, what do you think?

kswang1029:
@veggiesaurus @panchyo0 How about:
for floats:

> Operators: >, >=, <, <=, ==, !=, .., ...
e.g. < 10 (everything less then 10)
e.g. == 1.23 (entries equal to 1.23)
e.g. 10..50 (everything between 10 and 50, exclusive)
e.g. 10...50 (everything between 10 and 50, inclusive)

ps: although we support ... as range filter too, maybe we can hide it as .. servers the same function and it is slightly more concise. 
question: or does ... mean between or equal to? e.g, 10...50 means everything greater than or equal to 10 and less than or equal to 50?

> Filter by substring
e.g. gal (no quotation, entries contain the "gal" string)
e.g. "gal" (with quotation, entries exactly match the "gal" string)

veggiesaurus:
> > Filter by substring
> > e.g. gal (no quotation, entries contain the "gal" string)
> > e.g. "gal" (with quotation, entries exactly match the "gal" string)

I wasn't aware that we had to support exact matches :thinking: 

> question: or does ... mean between or equal to? e.g, 10...50 means everything greater than or equal to 10 and less than or equal to 50?

Isn't the one open and the other closed range?

panchyo0:
`e.g, 10...50 means everything greater than or equal to 10 and less than or equal to 50?`
Yes, `...` means closed range

kswang1029:
@panchyo0 to confirm:
10..50 is 10 < x < 50
10...50 is 10 <= x <= 50
Is the above correct?

panchyo0:
yes

panchyo0:
[...]

Issue #908, opened 2020-05-04
Improve catalog filtering GUI


veggiesaurus:
It is currently not immediately clear to users that columns can be filtered on. I suggest we change the column headers to something like this: 

![image](https://user-images.githubusercontent.com/592504/80951880-00527b00-8df9-11ea-9081-2dd1eeceb9b3.png)

where the placeholder text is shown whenever a user doesn't have a specified filter. We also need to find a way to indicate that the filters have changed, but new results still need to be calculated, otherwise users may expect that the results are immediately available when editing filters.

veggiesaurus:
@kswang1029 has suggested that we make the "Double-click to filter" placeholder slightly more subtle, possibly by adjusting the colour.

kswang1029:
Maybe when filter conditions are changed, we make the text color in the table “grey”?

veggiesaurus:
> Maybe when filter conditions are changed, we make the text color in the table “grey”?

we also need to indicate to the user that they need to perform some action. Perhaps by changing the "Filter" button's name to "Update" and changing the colour?

kswang1029:
On a related feature, maybe we should use tooltip to show example filter syntax when hovering over the filter fields. The content of tooltip should dynamically adjusted based on the types of columns (number, string, etc)

Update: on a second thought, why don’t we just use those placeholder texts for filter syntax examples?

PR #931, opened 2020-05-30
Qi/issue 880 , 882 , 908, 909, 878, 938, 916

panchyo0:
fixed issues #880 , #882 , #908, #909, #878, #938, #916 .
#880  : shift key to select consecutive sources and with ctrl/cmd key to select distinct sources on click


src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -138,18 +135,22 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
     constructor(props: WidgetProps) {
         super(props);
         this.widgetId = "catalog-overlay-0";
-        if (this.widgetStore) {
-            this.catalogFileId = this.widgetStore.catalogInfo.fileId; 
+        const catalogFileId = AppStore.Instance.catalogProfiles.get(this.props.id);
+
+        if (catalogFileId) {
+            AppStore.Instance.catalogProfiles.set(this.props.id, catalogFileId);
```
veggiesaurus: haven't you just retrieved `catalogFileId` from the same place one line before? Why is this `set` needed?
Why not just check if the map `has` `this.props.id`, otherwise setting it to `1`? that would also allow you to remove the `catalogFileId` field, because it's a little confusing to have `this.catalogFileId` and `const catalogFileId`.

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -518,23 +521,23 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
...
-        const selectedData = [];
-        let highlighted = false;
-        if (selectedPointIndices.length === 1) {
-            highlighted = selectedPointIndices.includes(selectedDataIndex);
-        }
-        if (!highlighted) {
-            if (widgetsStore.showSelectedData && selectedPointIndices.length) {
-                selectedData.push(selectedPointIndices[selectedDataIndex]);
+        if (selectedDataIndex.length === 1) {
+            const selectedPointIndexs = widgetsStore.selectedPointIndices;
```
veggiesaurus: typo, should be `Indices`

src/components/CatalogOverlay/CatalogScatterComponent/CatalogScatterComponent.tsx
```diff
@@ -176,9 +180,12 @@ export class CatalogScatterComponent extends React.Component<WidgetProps> {
 
     // Single source selected
     private onSingleSourceClick = (event: Readonly<Plotly.PlotMouseEvent>) => {
+        const appStore = AppStore.Instance;
         const dragmode: DragMode[] = ["select", "lasso"];
         const inDragmode = dragmode.includes(this.widgetStore.dragmode);
-        if (event && event.points && event.points.length > 0 && inDragmode) {
+        if (event && event.points && event.points.length > 0 && inDragmode && appStore.catalogProfiles.size > 0) {
```
veggiesaurus: perhaps we should use the `?` operator here? [optional chaning](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)

src/components/Shared/TableComponent/TableComponent.tsx
```diff
@@ -27,18 +28,27 @@ export class TableComponentProps {
...
 }
 
 @observer
 export class TableComponent extends React.Component<TableComponentProps> {
 
-    private renderDataColumnWithFilter = (columnName: string, columnData: any) => {
+    private getfilterSyntax = (dataType: CARTA.ColumnType) => {
+        switch (dataType) {
+            case CARTA.ColumnType.String || CARTA.ColumnType.Bool:
+                return ("Filter by substring");
```
veggiesaurus, line 43: @kswang1029 please assist with content for these tooltips

src/stores/AppStore.ts
```diff
@@ -774,12 +790,13 @@ export class AppStore {
         this.pendingChannelHistograms = new Map<string, CARTA.IRegionHistogramData>();
 
         this.frames = [];
-        this.catalogs = new Map();
         this.activeFrame = null;
         this.contourDataSource = null;
         this.syncFrameToContour = true;
         this.syncContourToFrame = true;
         this.initRequirements();
+        this.catalogs = new Map();
```
veggiesaurus: better to explicitly specify the map type in the constructor call, to be consistent with the other maps

veggiesaurus (changes requested):
- Why not just disable the "Update" button when the filter has not changed? That might make more sense than changing the color
- The filter tooltip needs to be expanded to provide details on how to use each operator. @kswang1029 can help with the text for this. I think we should format the tooltip as well, so it looks somewhat consistent with the other text-heavy tooltips (e.g. the widget tooltip)
- "Double-click to edit filter" should be a placeholder, and should disappear when the user begins editing, they shouldn't have to manually remove it. It would be nice if they looked and behaved consistently with the other placeholder texts. I wonder if using an input text box instead of an editable cell would solve some of these problems?
- Not sure if the filter row's cells should be marked as "active". Perhaps just those rows with a filter should be set as active. Can they be a bit smaller as well? There seems to be plenty of space below the filter line

veggiesaurus:
@panchyo0 we should also have the ability to clear the sort order via the sort menu (e.g. Sort Asc, Sort Desc, Clear Sort)

veggiesaurus (changes requested):
- I think the "Max" numerical input should be moved down to the footer, because the top row is far too crowded now. It should also say something like "Max rows". @kswang1029 suggests next to the "image overlay" dropdown, I suggest next to the update button. Either is fine with me, lets see which looks better
- We need a way to revert to the default number of rows. Either we need to use a `ClearableNumericInputComponent` or have some other way of reverting to the default value, other than clicking "reset", which resets everything.
- When setting max to `1000`, the text under the table says "showing 1 to 50 of 20000 entries. total 20000 entries". Not sure why we need the total number of rows specified twice, when there's no filter.
- Setting max to `0` prevents any data from being delivered and results in a hard crash `TypeError: Cannot read property 'length' of undefined` in CatalogStore line 120
- When applying sort to a column with a long name, it overflows into two rows:
image

kswang1029:
Regarding the third point, when there is a filter applied, it is better saying “Showing 1 to 50 of 1200 filtered entries. Total 20000 entries.”
If no filter is applied, we show “Showing 1 to 50 of total 20000 entries.”

Would this be clearer?

veggiesaurus:
> Regarding the third point, when there is a filter applied, it is better saying “Showing 1 to 50 of 1200 filtered entries. Total 20000 entries.
> If no filter is applied, we show “Showing 1 to 50 of total 20000 entries.”
> 
> Would this be clear?

that makes sense, yes. Not sure we need that second `total` when no filter applied.

But if you set "max" to 1000, what should it show?

kswang1029:
A very verbose solution:
if max row = 1000, and total row = 20000, we show
"Showing 1 to 50 of top 1000 entries. Total 1200 filtered entries. Total 20000 entires." if n_filtered >= 1000
and show 
"Showing 1 to 50 of 800 filtered entries. Total 20000 entries." if n_filtered < 1000

panchyo0:
@veggiesaurus Next to the update button (current)
Screen Shot 2020-06-16 at 2 06 26 PM
Next to the "image overlay" dropdown
Screen Shot 2020-06-16 at 2 05 44 PM

src/stores/widgets/CatalogScatterWidgetStore.ts
```diff
@@ -58,8 +58,8 @@ export class CatalogScatterWidgetStore {
         const columnsName = this.columnsName;
         if (columnsName.x && columnsName.y) {
             const coords = catalogOverlayWidgetStore.get2DPlotData(columnsName.x, columnsName.y, catalogOverlayWidgetStore.catalogData);
-            this.setXDataset(coords.wcsX);
-            this.setYDataset(coords.wcsY);
+            this.setXDataset(coords.wcsX.slice(0));
```
veggiesaurus, line 61: why does this solve the issue?

veggiesaurus:
> @veggiesaurus Next to the update button (current)
> Screen Shot 2020-06-16 at 2 06 26 PM
> Next to the "image overlay" dropdown
> Screen Shot 2020-06-16 at 2 05 44 PM

@panchyo0 lets go with next to the update button. Let me know when the PR is ready for a re-review

veggiesaurus (approved):

PR #939, opened 2020-06-08
sort frames by fileId numerically instead of alphabetically

veggiesaurus:
minor fix to ensure sorting is done numerically instead of alphabetically.


kswang1029 (approved):
confirmed the fix.

one frontend issue is the indices being crowded 
Screen Shot 2020-06-08 at 4 37 52 PM

veggiesaurus:
> confirmed the fix.
> 
> one frontend issue is the indices being crowded
> Screen Shot 2020-06-08 at 4 37 52 PM

Right... We should probably adjust the tick behaviour above a certain threshold number of files :thinking:

PR #960, opened 2020-06-26
Angus/fix 943

veggiesaurus:
This should allow a wider range of color formats to be supported via the scripting interface, by using the [tinycolor library](https://github.com/bgrins/TinyColor) to convert between rgba, hex and other formats.

This removes the need for our own color validation and conversion functions, so I've removed those. I also noticed a bug in the rendering of NaNs in the raster view shader code, so I fixed that too


confluence (approved):
I've tested this and confirmed that it fixes the issue.

panchyo0 (approved):

duidae (approved):

PR #965, opened 2020-06-29
fixes #945

### LINKED ISSUE #945: initial connection to the backend failed in the electron app
Over the past weeks (probably starting from v1.3) I noticed that _sometimes_ the backend is not launched successfully when we launch the electron app. It is hard to find a robust way to reproduce this but it seems to only happen when the computer is just waken up. 

Screen Shot 2020-06-02 at 10 09 06 AM

log:
/Applications/CARTA.app/Contents/Resources/app/carta-backend/bin/carta_backend: Version 1.3
Listening on port 55131 with root folder /, base folder /Users/kswang, 4 threads in worker thread pool and 16 OMP threads
Session 1 (Tue Jun  2 10:08:01 2020): Client 1 [::1] Connected. Num sessions: 1
Session 1 (Tue Jun  2 10:15:42 2020): Client 1 [::1] Disconnected. Remaining sessions: 1
No remaining sessions.

"10:08:01" is when I launched the electron app and I saw the above screenshot. 
"10:15:42" is when I closed the electron app.

(may consider to mark this issue as a critical bug)

veggiesaurus:
There were two issues at play with #945:
1. The initial connection was getting flagged as a "resumed" connection, because the first attempt to connect failed
2. The failed connection resulted in the `connect` observer being closed (since `observer.error` also closes the subscription)

(also removed some duplicate code in `onStartAnimationAck`

@kswang1029 added you as a reviewer, but not sure how easy it was to reproduce the issue for you :thinking:

### DIFF
--- a/src/services/BackendService.ts
+++ b/src/services/BackendService.ts
@@ -151,7 +151,10 @@ export class BackendService {
         this.connection.binaryType = "arraybuffer";
         this.connection.onmessage = this.messageHandler.bind(this);
         this.connection.onclose = (ev: CloseEvent) => {
-            this.connectionStatus = ConnectionStatus.CLOSED;
+            // Only change to closed connection if the connection was originally active
+            if (this.connectionStatus === ConnectionStatus.ACTIVE) {
+                this.connectionStatus = ConnectionStatus.CLOSED;
+            }
             // Reconnect to the same URL if Websocket is closed
             if (!ev.wasClean && this.autoReconnect) {
                 setTimeout(() => {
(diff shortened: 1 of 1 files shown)


kswang1029:
I only see this issue with a Mac package (easy to reproduce)... Maybe we can test during beta.

veggiesaurus:
> I only see this issue with a Mac package (easy to reproduce)... Maybe we can test during beta.

@ajm-asiaa perhaps you could build a test package for KS?

ajm-ska:
OK @kswang1029. [Here](http://alma.asiaa.sinica.edu.tw/_downloads/CARTA-v1.4.alpha.1.dmg) is what I'm calling v1.4.alpha.1. It is the same a v1.4.alpha.0 but with the `angus/fix_945` frontend.

kswang1029:
thanks @ajm-asiaa I will try it. Will report back in a day or so with some trials.

duidae (approved):

kswang1029 (approved):
Thanks to @ajm-asiaa for making the Mac package for testing. It appears that I cannot reproduce the issue after some trails since last evening.

PR #966, opened 2020-06-29
adds contour properties to resume message

veggiesaurus:
Companion PR of https://github.com/CARTAvis/carta-backend/pull/471

Should allow contours to be used correctly when session is resumed

### DIFF
--- a/src/stores/AppStore.ts
+++ b/src/stores/AppStore.ts
@@ -1211,6 +1211,18 @@ export class AppStore {
                 };
             });
 
+            let contourSettings: CARTA.IContourSettings;
+            if (frame.contourConfig.enabled) {
+                contourSettings = {
+                    levels: frame.contourConfig.levels,
+                    smoothingMode: frame.contourConfig.smoothingMode,
+                    smoothingFactor: frame.contourConfig.smoothingFactor,
+                    decimationFactor: this.preferenceStore.contourDecimation,
+                    compressionLevel: this.preferenceStore.contourCompressionLevel,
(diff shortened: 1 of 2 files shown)


duidae (approved):

kswang1029:
tested with
http://carta.asiaa.sinica.edu.tw/frontend/35b5f3e/?socketUrl=wss://carta.asiaa.sinica.edu.tw/socket5036
frontend: 35b5f3e
backend: 68a06a0

When there is one image cube with contour enabled, contour resume works. After session resume, changing channels results in contour update per channel as expected. 

When there are two image cubes matched both spatially and spectrally, contour resume does not work. After session resume, changing channels only results in raster update per channel. Contour images stay the same as the ones before session resume.

veggiesaurus:
> When there are two image cubes matched both spatially and spectrally, contour resume does not work. After session resume, changing channels only results in raster update per channel. Contour images stay the same as the ones before session resume.

@kswang1029 I have fixed this issue, please retest. I have also added catalog support in a later commit, which should be built shortly in mix 'n match

kswang1029:
@veggiesaurus now contour resume works fine 👍 

do we need a dedicated backend for catalog resume?

veggiesaurus:
> @veggiesaurus now contour resume works fine 
> 
> do we need a dedicated backend for catalog resume?

Frontend commit `1e236a1` should work with catalogs as well

kswang1029:
catalog resume works too. Tried to load two catalog and after resume, I can apply filters to each. 👍

PR #974, opened 2020-07-01
fixed issue with binary search of all-NaN data

veggiesaurus:
This adds a sanity check before returning out-of-bounds values from the binary search. Thanks @duidae for including the unit tests that made fixing this easy :+1:


kswang1029:
Tested and it works. 👍

duidae (approved):

duidae:
didnt realize the cause is Nan, @veggiesaurus thanks for fixing.

PR #975, opened 2020-07-01
skip stop animation action when animation is already stopped

veggiesaurus:
solves issue with missing tiles when resume is called, because `stopAnimation` is called when resuming a session.


kswang1029:
Nice, I think it is fixed after some tests. 👍

kswang1029 (approved):

PR #976, opened 2020-07-02
Angus/file browser blueprintjs

### LINKED ISSUE #606: long filenames in file browser
We usually have pretty long filenames from the NRAO pipeline, e.g. 

VLASS1.2.ql.T14t21.J135502+143000.10.2048.v1.I.iter1.image.pbcor.tt0.rms.subim

They are difficult to read in the file browser, in particular since the prefix of many images is usually the same and only the suffixes are different. They are unfortunately truncated in the CARTA browser and to see them, the entire window plus the browser window need to be rescaled. 

An improvement would be to have a horizontal scrollbar or to make the divider between file information and fline name adjustable.

### LINKED ISSUE #685: Scroll to right for large file names
In the file browser, long file names are cut off, and there is no ability to scroll from left to right. It would be good if this could be implemented, since often the first part of different files are the same.

### LINKED ISSUE #693: Use BlueprintJS table for file browser instead of HTML table
Various file browser improvements require a more feature-complete file list table. BlueprintJS will be used for the catalog viewer, so it would be better to use this for the file browser as well

### LINKED ISSUE #968: frozen when trying to sort a very long file list in the file browser
When there are thousands of image files in a folder, sorting (name, type, or size) is extremely slow and could case UI frozen.

veggiesaurus:
fixes #693, fixes #968 and fixes #685

Switches to a BlueprintJS table instead of an HTML table for the file browser. This greatly improves performance when rendering with lots of files, because the list is virtualised


kswang1029 (approved):
tested with a folder with 25000 files, everything works fine and responsive after the list appeared. 👍

veggiesaurus:
bonus: now supports filtering by regex instead of just substring

panchyo0 (approved):

Issue #962, opened 2020-06-29
catalog: empty table when enable "show only selected sources" toggle


kswang1029:
When a source is selected in the 2D scatter plot and the toggle "show only selected sources" is enabled, we see an empty source table in the catalog widget. After scrolling, the source appears. 

![Jun-29-2020 10-13-25](https://user-images.githubusercontent.com/20819712/85966084-50148380-b9f1-11ea-85ba-22d416a86355.gif)

Issue #951, opened 2020-06-22
update catalog header table "Represent As"


panchyo0:
dynamic change “represent as” menu according catalog plot type.  If it is “as image overlay”, we show ra, dec, none. If it is “as scatter plot”, we show x, y and none. If it is “as histogram”, we show histogram and none.

Issue #950, opened 2020-06-22
catalog widget: make the two tables resizable in vertical direction


panchyo0:
<img width="842" alt="Screen Shot 2020-06-12 at 2 14 22 PM" src="https://user-images.githubusercontent.com/22226157/85323861-0a523b00-b486-11ea-8ea6-4178490a4565.png">
image from @kswang1029

Issue #899, opened 2020-04-24
support pixel coordinate catalogs


veggiesaurus:
Some  source-finding software (e.g. SoFiA) outputs source catalogs in pixel coordinates, rather than WCS. We should be able to select "Pixel" as a coordinate type, which would then allow us to select "X" and "Y" in the "Represent As" 

One issue is that there's ambiguity as to whether the coordinates are zero-based (FITS) or one-based. Should we instead allow two coordinate types? "Pixel (0-based)" and "Pixel (1-based)"?

kswang1029:
Good point and good idea to have pixel-0 and pixel-1 as options

low-sky:
My thoughts here are that pixel coordinates only make sense in the context of the image for which the catalog is produced.  Thus, if it has 0-based coordinates, the catalog is 0-based and 1-based coordinates images would be 1-based catalogs.  I contend that only one Pixel coordinates is needed for most cases.

HOWEVER, I can assume that there are cases where people generate a catalog in 0-based pixels in something like python and then try to put it on a FITS image.  I can see it being a potentially useful edge case.

veggiesaurus:
> My thoughts here are that pixel coordinates only make sense in the context of the image for which the catalog is produced. Thus, if it has 0-based coordinates, the catalog is 0-based and 1-based coordinates images would be 1-based catalogs. I contend that only one Pixel coordinates is needed for most cases.

CARTA uses 0-based pixel coordinates for all file types, rather than adjusting based on the file type. I know of at least one popular use case that will require 0-based and 1-based adjustments: the source finding software package Sofia produces VOTable output in pixel coordinates only, and we will need to explicitly specify the offset correctly to avoid confusion in comparison with other image viewers

veggiesaurus:
@panchyo0 this seems to be working nicely in your latest PR (merged in). I wonder, do we need tooltips in the select menu to explain what `PIX0` and `PIX1` are?

panchyo0:
@veggiesaurus or explain `PIX0` and `PIX1` in the help section? not sure which one is better.

veggiesaurus:
> @veggiesaurus or explain `PIX0` and `PIX1` in the help section? not sure which one is better.

Good point. @kswang1029 which would you prefer?

kswang1029:
I prefer we have both, tooltip and in-app help. The tooltip would allow users know the meanings with less click and less scan though help context.

veggiesaurus:
[...]

PR #979, opened 2020-07-03
Qi/fixed issue 899, issue 950, issue 951 and issue 962

panchyo0:
#899, #950, #951 and #962
#909: updated tooltip according what we have currently


kswang1029 (approved):
looks nice!

one typo in the tooltip of the filter: everything less _than_ 10

veggiesaurus (approved):

PR #981, opened 2020-07-03
Angus/file list time

### LINKED ISSUE #686: Feature request: sort by date added
In the file browser, the files are sorted alphabetically, and it becomes difficult to find files when you have several tens of them. It would be nice to be able to sort by date added (or perhaps date modified, but less ideal) so you can easily find the recent file you're looking for. It would even be nice to be able to sort by file type / extension, so CASA, FITS and HDF5 files could be separated.

veggiesaurus:
closes #686
Companion PR of https://github.com/CARTAvis/carta-backend/pull/550.

This PR adds the ability to sort by modified date. Displayed date is based on the file modification timestamp. If the timestamp is from today, the time is displayed. If older, only the date is displayed. This is similar to the default behaviour in gnome/ubuntu. @kswang1029 and I have checked the behaviour when accessing a server in a different time zone, and the displayed time is based on the _local_ (i.e. client) time.

PR also changes the sorting approach to a more conventional button-based approach, instead of a context menu.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
         "mnemonist": "^0.29.0",
         "mobx": "^5.14.0",
         "mobx-react": "^5.4.4",
+        "moment": "^2.27.0",
         "plotly.js": "^1.52.3",
         "raw-loader": "^0.5.1",
         "react": "^16.10.2",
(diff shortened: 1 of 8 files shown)


panchyo0:
open file browse with hotkey return `There are no files or subdirectories matching the filter expression`

veggiesaurus:
> open file browse with hotkey return `There are no files or subdirectories matching the filter expression`

That's fixed in `dev`, have just merged `dev` into this branch again, so it should be fixed in this branch as well

panchyo0 (approved):

PR #986, opened 2020-07-08
Thchang/profile smoothing

### LINKED ISSUE #651: channel smoothing in the spectral profile widget
sub task of epic #644 

The spectral profile widget should provide the following profile smoothing options:
- hanning
- boxcar
- binning

The smooth profile should be displayed together with the original profile in different colors.

TienHao:
close #651 

[GSL](https://www.gnu.org/software/gsl/) is used to apply profile smoothing. 
Requirement doc: https://docs.google.com/document/d/1-EnbWNeApaLB0HlJ3EUn-mGsLfKD2NeG9TKTWMzR5m4/edit#

demo:
ezgif com-video-to-gif (9)


veggiesaurus (changes requested):
- In order to build GSL, I had to add the argument ` --host=wasm32` to the configure step
- I get a divide by zero runtime error whenever I load a file which requires decimation of spatial profiles:

```
Uncaught RuntimeError: divide by zero
    at filterDecimation (http://localhost:3000/gsl_wrapper.wasm:wasm-function[12]:0x6e9)
    at Object../wasm_src/gsl_wrapper/build/index.js.Module._filterDecimation [as filterDecimation] (http://localhost:3000/static/js/bundle.js:512761:21873)
    at Object../wasm_src/gsl_wrapper/build/index.js.Module.decimation (http://localhost:3000/static/js/bundle.js:512761:27727)
    at ProfileSmoothingStore.getDecimatedValues (http://localhost:3000/static/js/bundle.js:505939:87)
    at SpatialProfilerComponent.get plotData (http://localhost:3000/static/js/bundle.js:490800:62)
```

TienHao:
> * In order to build GSL, I had to add the argument ` --host=wasm32` to the configure step

I removed the argument ` --host=wasm32` because there was a building trouble on RHEL7. (mentioned by @ajm-asiaa)
@veggiesaurus : what OS did you use to build?

veggiesaurus:
@TienHao one quick comment: if you know the length of a large array beforehand, it's much faster to allocate it first, rather than appending to it.

veggiesaurus:
> > * In order to build GSL, I had to add the argument ` --host=wasm32` to the configure step
> 
> I removed the argument ` --host=wasm32` because there was a building trouble on RHEL7. (mentioned by @ajm-asiaa)
> @veggiesaurus : what OS did you use to build?
> 
> 

Ubuntu 20.04. I'll double check what version of emcc I used.

src/stores/widgets/SpectralProfileWidgetStore.ts
```diff
@@ -26,6 +27,7 @@ export class SpectralProfileWidgetStore extends RegionWidgetStore {
     @observable lineWidth: number;
     @observable linePlotPointSize: number;
     @observable linePlotInitXYBoundaries: { minXVal: number, maxXVal: number, minYVal: number, maxYVal: number };
+    @observable smoothingStore: ProfileSmoothingStore;
```
veggiesaurus: this can be a `readonly` variable without an observable decorator, can't it? since it's only assigned once

wasm_src/gsl_wrapper/gsl_wrapper.cc
```diff
@@ -0,0 +1,213 @@
...
+        if (i == inN / decimationWidth) {
+            if (inN % decimationWidth == 0) {
+                break;
+            } else if (inN % decimationWidth == 1) {
+                indexArray[i*2] = i * decimationWidth;
+                break;
+            }
+        }
+
+        size_t* minIndex = new size_t[1];
```
veggiesaurus: should just be `size_t minIndex;` and then pass it in as `&minIndex`, otherwise you're leaking memory.

veggiesaurus (changes requested):
this now builds nicely.

- Can you change all the occurrences where you're pushing a known number of elements to an array to instead pre-allocating and assigning to that array. If we're looking at thousands of points, it might get a bit slow
- It looks to me like you're smoothing the _visible_  part of the profile. This makes it nice and efficient, which is great, but it does mean that you end up with missing data on the sides. I think you need to work out how much extra padding you need on either side and include that in the profile you're smoothing
- Couple of comments in the C++ code made separately

image

TienHao:
> * It looks to me like you're smoothing the _visible_  part of the profile. This makes it nice and efficient, which is great, but it does mean that you end up with missing data on the sides. I think you need to work out how much extra padding you need on either side and include that in the profile you're smoothing

I have added methods to get proper sub-array to perform local smoothing.

TienHao:
GSL has an argument to decide how to handle end points. 
https://www.gnu.org/software/gsl/doc/html/movstat.html#handling-endpoints

I added a dropdown `endType` in the setting panel to change the argument, which had been fixed to `GSL_MOVSTAT_END_PADVALUE`.

`endType` has three options:
1.`none`: set `NaN` to channels which have the window(kernel) needing extra padding for convolution.
2.`zero`: add zero into extra padding (`GSL_MOVSTAT_END_PADZERO`)
3.`value`: add first/last value into extra padding (`GSL_MOVSTAT_END_PADVALUE`)

veggiesaurus:
@kswang1029 which should we use as the default padding? `NaN`?

kswang1029:
NaN sounds natural to adopt as we don’t have any info beyond the edges. In any case either with NaN, zero, or constant extrapolation, I personally won’t treat those smoothed edge channel seriously for analysis or interpretation.

veggiesaurus:
@TienHao is this ready for re-review?

TienHao:
@veggiesaurus : Yes, this is ready for re-review.
@kswang1029 : `NaN` is the default padding type now.

wasm_src/gsl_wrapper/gsl_wrapper.cc
```diff
@@ -0,0 +1,250 @@
...
+
+            gsl_multifit_linear_free (work);
+
+            gsl_matrix_free (X);
+            gsl_vector_free (w);
+            gsl_vector_free (c);
+            gsl_matrix_free (cov);
+        } else {
+            double c0, c1, cov00, cov01, cov11, chisq;
+            double *x = (double *) malloc(wsize * sizeof(double));
```
veggiesaurus: this should be initialised via `new` instead of `malloc`

wasm_src/gsl_wrapper/gsl_wrapper.cc
```diff
@@ -0,0 +1,250 @@
...
+            double *w = (double *) malloc(wsize * sizeof(double));
+            for (size_t s = 0; s < wsize; s++) {
+                x[s] = xInArray[(i - H) + s];
+                w[s] = 0.2;
+            }
+
+            gsl_fit_wlinear (x, 1, w, 1, window, 1, wsize, &c0, &c1, &cov00, &cov01, &cov11, &chisq);
+            yOutArray[i] = c0 + c1 * xInArray[i]; // best fit Y = c0 + c1 * X;
+
+            free(x);
```
veggiesaurus: `delete[]`

veggiesaurus (changes requested):
couple of minor changes required to the code. In addition:
- Please make the settings widget slightly wider by default, so that the select dropdown doesn't get squashed.
- I think `EndType` is a very confusing setting. Just because that's how it's exposed in the library doesn't mean we should use that terminology. @kswang1029 how should we describe it?

kswang1029:
@veggiesaurus @TienHao We should just hide EndType from users and just use it internally in the codebase (ie users do not have a chance to configure it through the GUI).

TienHao:
@kswang1029 I have removed `EndType` from smoothing setting panel.

veggiesaurus:
@TienHao code now looks fine. However, I still can't build properly without `--host=wasm32` as an argument. I think we either need to find the root cause of this, or document it carefully and make it easy for users to run `build_libs` with some sort of command line option that adds that flag

veggiesaurus:
@TienHao let me know if the build script changes work ok for you. Works fine with `npm build-libs-docker` for me

TienHao:
@veggiesaurus : the build script works fine for me too!

PR #988, opened 2020-07-08
Angus/show header comments

### LINKED ISSUE #995: Display header comments
Now that the header comments are sent separately, we should display them in the file info.

veggiesaurus:
After @pford's header PR (https://github.com/CARTAvis/carta-backend/pull/454) was merged, I thought it would be a good idea to make use of the header's comments in a formatted manner:

image

Only the first 1000 headers are formatted. After that, it becomes plain text, to prevent _really_ big headers (that HSP file has over 13K headers)


panchyo0 (approved):

kswang1029 (commented):
@veggiesaurus this looks nice and we should apply the same to "file info".
However, there is performance impact on images with long headers.
Jul-09-2020 09-39-07

without this improvement, switching between the two tabs is quick.

veggiesaurus:
Right, that's why I set max headers to 1000. Perhaps it should be even shorter?

kswang1029:
@veggiesaurus maybe we can experiment with few options and see in general which gives us minimal impact? (The UX with this improvement is boosted significantly :) )

duidae (approved):

veggiesaurus:
I've switched to a virtualised list approach, so we can now render as many headers as we want with little to no performance impact. I still have to fix a couple of CSS issues though.

veggiesaurus:
now working fine with over 13 000 files. @kswang1029 @duidae please re-review

PR #992, opened 2020-07-09
Jason/issue 833

### LINKED ISSUE #833: reload get blank frontend layout when backend is lost
image

the issue is due to commit c43ee944d9ed2fb564b37cdb5370a828b2113d8b

@veggiesaurus  is it ok to put it back?

duidae:
closes #833 

test:
1. terminate backend & start frontend
2. should stay in splash screen while re-connecting

image


kswang1029:
if the server (backend) is not accessible, we should explicitly say it in the splash screen. It feels frustrated when seeing "Connecting..." forever.

duidae:
changed to open connection failed.
image

kswang1029:
change the spinning wheel to something else? eg❗️

duidae:
but frontend is still re-connecting repeatedly, we should keep the user informed of this waiting status

kswang1029:
I see. Then how about showing

> Connecting to server xxx failed.
Retrying...

duidae:
added

image

veggiesaurus:
@kswang1029 what do you think about showing the splash screen again with the same information when the connection drops? that way users won't accidentally continue to interact with the frontend after the connection has dropped.

@duidae this can be enabled by just adding the following in the splashscreen autorun:

```
    else {
                this.showSplashScreen();
     }
```

kswang1029:
Great idea!

kswang1029:
There might be a time gap (few seconds) that the frontend thinks the connection is dropped, right?

veggiesaurus:
> There might be a time gap (few seconds) that the frontend thinks the connection is dropped, right?

If the backend crashes, the underlying TCP connection will be dropped immediately. If it's a network issue, then the frontend won't notice until the timeout is reached (might be ~30 seconds, depending on the network conditions)

kswang1029:
30s sounds about right when I switch off my wifi

PR #996, opened 2020-07-13
Jason/issue 763

### LINKED ISSUE #763: show wcs of cursor in spatial profile
When the option to display wcs coordinates is selected in the spatial profiles, the cursor display value should show the WCS of the cursor in addition to the pixel coord.

duidae:
closes #763

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,8 @@
         "@blueprintjs/icons": "^3.11.0",
         "@blueprintjs/select": "^3.11.1",
         "@blueprintjs/table": "3.7.1",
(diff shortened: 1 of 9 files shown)


veggiesaurus (changes requested):
this works nicely, but if a file doesn't have any valid WCS, then the cursor info disappears completely. 

@kswang1029 do you think we should only show the WCS in cursor when the "Show WCS info" toggle is active?

kswang1029:
It sounds reasonable to do so. @arusstaylor thoughts?

veggiesaurus (approved):

PR #997, opened 2020-07-13
Jason/issue 935

### LINKED ISSUE #935: icon and text inconsistency
There are few icon inconsistencies (including text descriptions) in GUI:

Screen Shot 2020-06-03 at 11 03 03 AM
Screen Shot 2020-06-03 at 11 03 31 AM
Screen Shot 2020-06-03 at 11 04 09 AM
Screen Shot 2020-06-03 at 11 04 26 AM

duidae:
closes #935

image
looks weird using previous approach (because x y z s are luckily the same in upper-case & lower-case, the letters are showed in lower-case), so not applied

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "homepage": "./",
     "dependencies": {
         "@blueprintjs/core": "^3.19.1",
-        "@blueprintjs/icons": "^3.11.0",
+        "@blueprintjs/icons": "^3.19.0",
         "@blueprintjs/select": "^3.11.1",
         "@blueprintjs/table": "3.7.1",
         "@types/tinycolor2": "^1.4.2",
--- a/src/HotkeyWrapper.tsx
+++ b/src/HotkeyWrapper.tsx
@@ -126,8 +126,8 @@ export class HotkeyContainer extends React.Component {
         ];
 
         const animatorHotkeys = [
-            <Hotkey key={0} group={animatorGroupTitle} global={true} combo={`${modString}]`} label="Next frame" onKeyDown={appStore.nextFrame}/>,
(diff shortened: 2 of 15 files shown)


veggiesaurus:
I'm happy with everything except the region list icon. To me, that looks really odd. I'm also wondering whether we shouldn't say "File Header" instead of "File Info"? The "H" suggests header, rather than info (which would be an "i")

One minor change that we should probably make while we're at it: The tooltip for the preferences dialog should be "Preferences", not "Preference"

kswang1029:
I was quite struggling with the icons for region list widget and file info dialogue. Open to any better suggestions. Maybe we will have to make our owns (I was trying not to)?

veggiesaurus:
> image
> looks weird using previous approach (because x y z s are luckily the same in upper-case & lower-case, the letters are showed in lower-case), so not applied

Perhaps we could do this, but just use a smaller font size? or use small caps, or a smaller icon?

duidae:
1. image
2. image

I suggest we apply either 1 or 2 for now, if there is still time available before 8/3, could still refine it.

veggiesaurus:
image
@duidae I played around with the CSS a bit. @kswang1029 does this look ok?

kswang1029:
Nice. Hope this is indicative enough to users 🤞

veggiesaurus:
@duidae I've made some adjustments, and fixed some things relating to #1001. Can you get the region list icon in the menu to look like that of the toolbar? 

Keeping #1001 open for now, because we also need to look at how we are referring to things in the code, due to scripting support. Might need a rather painful refactoring of things like `FrameStore` -> `ImageStore` :nauseated_face:

duidae:
Fixed.
Btw due to this feature is kind of cutting-in-line & not inside 1.4 scope, although it's easy, but adjusting css is quite time consuming to me, I think next time we should have an agreement that satisfying basic requirement is sufficient.

veggiesaurus (approved):

PR #998, opened 2020-07-15
Jason/line id overlay

### LINKED ISSUE #863: Spectral Line ID overlay
- [X] 1. fetch data from splatalogue
- [X] 2. wire up UI
- [X] 3. plot spectral line on spectral widget
- [ ] 4. filtering/sorting
- [x] 5. support layout

ref:
splatalogue of NRAO:
more info about query:
https://www.cv.nrao.edu/php/splat/advanced.php 

top table: "Applications (SLAP Interface)"
https://www.cv.nrao.edu/php/splat/SLAPNotes2.html

some example scripts to make query: ftp://ftp.cv.nrao.edu/NRAO-staff/bkent/slap/

description: https://cdms.astro.uni-koeln.de/classic/predictions/description.html

design doc: https://docs.google.com/document/d/1dzhPc_XptsHFoKQ4fj7bo4TTKf5beGXfWZ0ZljYzdgM/edit

duidae:
closes #863

send PR first to see if there is anything needs to fix beforehand


src/components/Shared/TableComponent/TableComponent.tsx
```diff
@@ -221,39 +287,37 @@ export class TableComponent extends React.Component<TableComponentProps> {
...
-                    enableRowReordering={false}
-                    selectionModes={SelectionModes.NONE}
-                    enableGhostCells={true}
-                    enableRowResizing={false}
-                >
-                    {tableColumns}
-                </Table>
-            );
-        }
+        return table.type === TableType.ColumnFilter ? (
```
veggiesaurus: I think the if/else approach looks a bit neater for such a long return value

src/components/SpectralLineOverlay/SpectralLineOverlayComponent.scss
```diff
@@ -0,0 +1,117 @@
...
+                margin: 0;
+                padding: 0 5px 0 5px;
+                display: flex;
+            }
+
+            .result-table {
+                overflow: auto;
+                height: calc(100% - 50px);
+                width: 100%;
+                padding: 5px 5px 5px 5px;
```
veggiesaurus: can just use `padding: 5px` can't you?

veggiesaurus (changes requested):
One minor change in the code. The only other thing I'd like changed is the dropdown for the spectral profile widget to look like the normal select dropdowns. I think just using `minimal=true` does that

duidae:
fixed, please have a look, thanks.

veggiesaurus (changes requested):
another couple of issues:

- the spectral profile often stays highlighted even after the selection is complete. This seems to happen more often when there's only one widget
- I think we should use a label and a proper BlueprintJS select for the spectral profile dropdown. It should look similar to the data source dropdown in the contour dialog
- Should we not automatically select a particular spectral profile by default? It's quite annoying to have to select a spectral profile widget even if there's only one of them (which will probably be the case fairly often)

duidae:
init selection with the 1st spectral widget & refined dropdown style.
@veggiesaurus what type & version of browser are u using? I cant reproduce the lasting highlight issue on chrome(81.0.4044.138 )/firefox(78.0.1) on ubuntu.

duidae:
added showing line text when zoom-in to appropriate scale
 
ddabe-9169g

kswang1029:
There are a couple of things for improvements:
- [x] the textboxes, from, to, velocity/redshift, should accept floating numbers. redshift >= 0. from/to >=0.
- [x] should also request "Sij mu^2", "Sij" and "Aij" from the spatalogue and display them
Screen Shot 2020-07-21 at 1 22 08 PM

- [x] rename "Freq-MHz(rest frame, redshifted)" to "Rest Frequency"
- [x] rename "Freq Err(rest frame, redshifted)" to "Rest Frequency Error"
- [x] allowing users to resize the two tables in the vertical direction via a horizontal line between the two tables (cf catalog widget)
- [x] rename "Meas Freq-MHz(rest frame, redshifted)" to "Measured Frequency"
- [x] rename "Meas Freq Err(rest frame, redshifted)" to "Measured Frequency Error"
- [x] the shifted frequency column should adopt the following logic: if "Rest Frequency" exists, use it for conversion of "Shifted Frequency". Otherwise, use "Measured Frequency" for conversion of "Shifted Frequency". Some entries do not have rest frequencies and only have measured frequencies. If both exists, use "Rest" for conversion. 
Screen Shot 2020-07-21 at 1 29 51 PM

- [x] add descriptions:
Species: Chemical formula of the species
Chemical name: Chemical name of the species
Shifted frequency: Shifted frequency according to the input velocity or redshift
Rest frequency: Frequency at the rest frame
Rest frequency error: Frequency error at the rest frame
Measured frequency: Frequency measured from laboratories
Measured frequency error: Frequency error measured from laboratories
Resolved QNs: Resolved quantum numbers
CDMS/JPL intensity: log10(intensity) at 300 K
E_L (cm^-1): Lower level energy in wave number
E_L (K): Lower level energy in K
E_U (cm^-1): Upper level energy in wave number
E_U (K): Upper level energy in K
Linelist: Originated catalogue

- [ ] add filtering and sorting functions.
- [x] change to custom icon

veggiesaurus:
>  add filtering and sorting functions.

@kswang1029 sorting should be easy to do, but by filtering, do you mean similar to the catalog stuff? (which is done on the backend currently)

kswang1029:
@veggiesaurus yes, similar to the catalog stuff. Any possibility to reuse the code via wasm?

veggiesaurus:
> @veggiesaurus yes, similar to the catalog stuff. Any possibility to reuse the code via wasm?

It would probably be easier to just handle in JS. I imagine you're not expecting millions of line ID results. But either way it's an added complexity. Could we leave the filtering for a future release, and just include sorting? That should be pretty simple to implement

kswang1029:
what we obtained from spatalogue should be sorted in frequency. 

if it is risky to implement filtering, we can postpone it to v1.5.

duidae:
Mostly fixed.

kswang1029:
I checked the spatalogue user manual and there is a table for description. So I suggest we just follow it. These include:

Sij Mu^2: Dipole-weighted transition dipole matrix elements (Debye^2)
[any chance to show the greek letter /mu ?]

Sij: Line strength

**log10(Aij)**: Base-10 logarithm of the Einstein A coefficient (s^-1)

Species: Descriptive formula of molecular species.

Chemical Name: Common chemical name for species

Resolved QNs: Rotational/rovibrational quantum numbers associated with the transition.

CDMS/JPL intensity: Boltzmann-weighted logarithmic intensity of transition, given at 300 K

Lovas/AST Intensity: Observed astronomical intensity for given transition

duidae:
Refined header & description.

kswang1029:
Screen Shot 2020-07-27 at 11 05 12 AM
@duidae Can we add a space before "v=..."? For example, "CH3CH2CNv=0" should be displayed as "CH3CH2CN v=0". The v=0 here means it is the vibrational ground state.

kswang1029:
@duidae the line labels need to be included in the png output too.

veggiesaurus (approved):

PR #1004, opened 2020-07-19
Jason/editable region wcs 2

### LINKED ISSUE #238: Editing of regions in WCS coordinates
Regions should be editable in WCS as well as pixel coordinates.
An example of this is implemented in DS9:

duidae:
closes #238


duidae:
Note:
1. the newest BlueprintJS 3.30.0 (just released 4 days ago) has some changes about NumericInput and results in un-editable issue, so upgrade only to previous release 3.29.0.
2. added wcs edit for size of rectangular & elliptical region.
3. added difference check to wcs input to avoid unnecessary value changes.

veggiesaurus:
I've resolved the merge conflicts with this. All looks good except one question: 

What does the coord type dropdown actually do? It doesn't seem to change anything in the region dialog

wcs_editing

I thought it would change the coordinate system used in the numerical boxes (same as DS9)

image

kswang1029:
@veggiesaurus that requires an update of region (eg by dragging the region). We should force an update after changing the coordinate system. @duidae

duidae:
the mockup by @kswang1029 :
image

the dropdown is for updating the system of the AST axis labels in imagviewer, for the convenience of editing region when referencing wcs, I think.
However, currently I am solving 2 problems: 
1. when switching the system, the WCS values in region dialog is not updated.
2. when switching the system, it seems that AST is also affected in certain ways, so the WCS value will be changed after a few transformations, not sure why the transformation is not triggered immediately.(ex: when system = auto, the WCS value will be like 12:22:30.4992(sexagesimal), when switching to Galatic, the WCS will not changed to 1.7303(decimal) immediately but will do after a few transformations, like dragging the region)

so I think we probably need to restrict users to input the same value format corresponding to the overlay number setting when editing.

veggiesaurus:
> @veggiesaurus that requires an update of region (eg by dragging the region). We should force an update after changing the coordinate system. @duidae

Currently it requires me to close the dialog and open it again, even if I drag the region around

veggiesaurus:
> 1. when switching the system, the WCS values in region dialog is not updated.

@duidae I think one of the issues is that `frame.wcsInfo` doesn't actually change when you change the WCS settings, since it's still the same AST object.

veggiesaurus:
> > 1. when switching the system, the WCS values in region dialog is not updated.
> 
> @duidae I think one of the issues is that `frame.wcsInfo` doesn't actually change when you change the WCS settings, since it's still the same AST object.

One option might be to encapsulate the AST frameinfo in an object that has an `@observable` timestamp, and manually update the timestamp via some action every time the underlying AST object is expected to change. For 1.5 we could thoroughly rexamine how we "observe" AST changes though

duidae:
@veggiesaurus I spotted when system dropdown changes, the system of frame.wcsInfo is not changed immediately but will do after closing or dragging, ex: when selecting from FK4 to GALATIC using dropdown, the system in frame.wcsInfo is still FK4, but changed to GALATIC after closing & reopening. It should be changed somewhere.

duidae:
seems that AST.plot() will change the system of frame.wcsInfo, so that the wcs values are changed accordingly, but the timing is not early enough for region wcs calculation

duidae:
should be getFormattedCoordinates() changed the settings of frame.wcsInfo for cursor info

duidae:
@veggiesaurus @kswang1029 should be working now, please have a look, thanks.

kswang1029:
@duidae It appears that when i edit the world coordinate of the botton-left corner, the value does not follow my input and it has a small value added. Any ideas? 

Jul-24-2020 23-07-09

duidae:
> @duidae It appears that when i edit the world coordinate of the botton-left corner, the value does not follow my input and it has a small value added. Any ideas?
> 
> Jul-24-2020 23-07-09

we are using AST to do the conversion and there could be precision issue since we are using image coordinate natively, ex:
14.642(Image) transformed to 109.8671(WCS) for display -> edit 109.8671 to 109.8600 ->  transform 109.8600(WCS) to 35.807(Image) & save 35.807 -> transform 35.807(Image) to 109.8604(WCS) for display

kswang1029:
When transforming 109.8600 (world) to 35.807 (image), do we trim the value to “35.80700000...” instead of preserving more significant digits? I think the value used for “calculation” and for “display” should be decoupled, otherwise we will have error propagation problem after multiple transformations.

kswang1029:
When we have calculations involving degrees, it is important to preserve as many significant digits as we could. For example, 0.001 deg is 3.6 arcsec. I suggest we try to keep the accuracy as far as we could down to submilli- to micro-arcsecond scales.

duidae:
precision increased. please have a look, thanks.

kswang1029:
@duidae we need a slightly taller dialog as default.
Screen Shot 2020-07-29 at 11 27 32 AM

duidae:
height extended.

veggiesaurus:
I find the tooltips a bit intrusive when editing. What about just saying `Center (d:m:s)` or `Center (h:m:s)` or `Center (deg)` instead? Wouldn't that be clear to users? Especially considering that the existing values would be in the appropriate format. For Size, we could say `Size (arcsec)`.

@kswang1029 what do you think? If we instead keep the tooltips, they should take longer to appear

src/stores/FrameStore.ts
```diff
@@ -248,6 +251,19 @@ export class FrameStore {
         return null;
     }
 
+    public getImageValueFromArcsec(arcsecValue: number): number {
```
veggiesaurus, line 254: @kswang1029 this assumes the projection is constant, because it gets image distance from arclength based on the CDELT. Is that ok?

kswang1029: Yes, that’s fine. 
In general, especially wide field images with some projection types, the angular size per pixel changes everywhere in the image. If we want to do it right, we should see a rectangle or an ellipse changing its size (but still a rectangle or an ellipse) while moving in the image. We might consider to support this but I think we need more thoughts and need input from the science team.

duidae:
> I find the tooltips a bit intrusive when editing. What about just saying `Center (d:m:s)` or `Center (h:m:s)` or `Center (deg)` instead? Wouldn't that be clear to users? Especially considering that the existing values would be in the appropriate format. For Size, we could say `Size (arcsec)`.
> 
> @kswang1029 what do you think? If we instead keep the tooltips, they should take longer to appear

x, y could be in different formats, ex: `x in h:m:s` & `y in d:m:s`
unless we show `Center (h:m:s, d:m:s)`

kswang1029:
What about
```
"Format: degrees"
"Format: h:m:s"
"Format: d:m:s"
```

for size we need tooltip too, such as:
```
" : arcsec
' : arcmin
deg: degrees
```

veggiesaurus:
Ok, I see the complications. In that case let's just increase the tooltip hover time substantially

duidae:
tooltip refined. please have a look, thanks.

kswang1029:
looks good 👍

src/components/Dialogs/RegionDialog/EllipticalRegionForm/EllipticalRegionForm.tsx
```diff
@@ -256,26 +256,30 @@ export class EllipticalRegionForm extends React.Component<{ region: RegionStore,
...
-                <NumericInput
-                    allowNumericCharactersOnly={false}
-                    buttonPosition="none"
-                    placeholder="Semi-minor"
-                    disabled={!this.props.wcsInfo}
-                    value={this.sizeWCS ? this.sizeWCS.y : ""}
-                    onBlur={this.handleMinorAxisWCSChange}
-                    onKeyDown={this.handleMinorAxisWCSChange}
-                />
+                <Tooltip content={"Format: arcsec(\"), arcmin('), or degrees(deg)"} position={Position.BOTTOM}>
```
veggiesaurus: Can we increase the `hoverOpenDelay` for these tooltips? Otherwise they seem a bit invasive

duidae:
extended the hover open time to 300ms, sorry didnt get the point before.

veggiesaurus (approved):

PR #1019, opened 2020-07-20
Qi/catalog histograms

panchyo0:
1. update plotlyjs form 1.52.3 -> 1.54.6 in package json
2. support histogram plot in catalog overlay widget


src/components/CatalogOverlay/CatalogSubplotComponent/CatalogSubplotComponent.tsx
```diff
@@ -104,28 +134,43 @@ export class CatalogScatterComponent extends React.Component<WidgetProps> {
...
-    private genProfilerInfo = (): string[] => {
+    @computed get genProfilerInfo(): string[] {
         let profilerInfo: string[] = [];
         const widgetStore = this.widgetStore;
         const column = widgetStore.columnsName;
         const indicatorInfo = widgetStore.indicatorInfo;
         if (indicatorInfo) {
-            profilerInfo.push(column.x + ": " + indicatorInfo.x + ", " + column.y + ": " + indicatorInfo.y);   
+            if (widgetStore.plotType === CatalogPlotType.D2Scatter) {
+                profilerInfo.push(column.x + ": " + indicatorInfo.x + ", " + column.y + ": " + indicatorInfo.y);
```
veggiesaurus: probably cleaner to use ``    `${colum.x}: ${indicatorInfo.x}, ${column.y}: ${indicatorInfo.y}` `` and so on.

(p.s. working out how to get the character literal ` in code was fun)

src/components/CatalogOverlay/CatalogSubplotComponent/CatalogSubplotComponent.tsx
```diff
@@ -134,14 +179,33 @@ export class CatalogScatterComponent extends React.Component<WidgetProps> {
...
-                    xMin: isFinite(event["xaxis.range[0]"]) ? event["xaxis.range[0]"] : widgetStore.border.xMin,
-                    xMax: isFinite(event["xaxis.range[1]"]) ? event["xaxis.range[1]"] : widgetStore.border.xMax,
-                    yMin: isFinite(event["yaxis.range[0]"]) ? event["yaxis.range[0]"] : widgetStore.border.yMin,
-                    yMax: isFinite(event["yaxis.range[1]"]) ? event["yaxis.range[1]"] : widgetStore.border.yMax
-                };
-                this.widgetStore.setBorder(border);   
+            if (widgetStore.plotType === CatalogPlotType.D2Scatter) {
+                if (isFinite(event["xaxis.range[0]"]) || isFinite(event["yaxis.range[0]"])) {
+                    const scatterBorder: Border = {
+                        xMin: isFinite(event["xaxis.range[0]"]) ? event["xaxis.range[0]"] : widgetStore.scatterborder.xMin,
```
veggiesaurus: I think you should store `event["xaxis.range[0]"]` and so on as variables

src/components/CatalogOverlay/CatalogSubplotComponent/CatalogSubplotComponent.tsx
```diff
@@ -164,10 +228,19 @@ export class CatalogScatterComponent extends React.Component<WidgetProps> {
             const catalogFileId = this.widgetStore.catalogOverlayWidgetStore.catalogInfo.fileId;
             AppStore.Instance.updateCatalogProfiles(catalogFileId);
             let selectedPointIndices = [];
-            const points = event.points;
+            const points = event.points as any;
             for (let index = 0; index < points.length; index++) {
                 const selectedPoint = points[index];
-                selectedPointIndices.push(selectedPoint.pointIndex);
+                if (this.widgetStore.plotType === CatalogPlotType.D2Scatter) {
+                    selectedPointIndices.push(selectedPoint.pointIndex);
```
veggiesaurus: since we _know_ the length of `selectedPointIndices` beforehand, I think you should pre-allocate it and assign it instead of using `push`, because this can be much slower

src/components/CatalogOverlay/CatalogSubplotComponent/CatalogSubplotComponent.tsx
```diff
@@ -164,10 +228,19 @@ export class CatalogScatterComponent extends React.Component<WidgetProps> {
...
-            const points = event.points;
+            const points = event.points as any;
             for (let index = 0; index < points.length; index++) {
                 const selectedPoint = points[index];
-                selectedPointIndices.push(selectedPoint.pointIndex);
+                if (this.widgetStore.plotType === CatalogPlotType.D2Scatter) {
+                    selectedPointIndices.push(selectedPoint.pointIndex);   
+                }
+    
+                if (this.widgetStore.plotType === CatalogPlotType.Histogram) {
```
veggiesaurus: might be better to move the if statement out of the loop, and then you can also assign the proper type to `points` instead of `any`

veggiesaurus (changes requested):
some minor comments, as well as the issue we discussed on slack

src/components/CatalogOverlay/CatalogSubplotComponent/CatalogSubplotComponent.tsx
panchyo0: the current types we use for plotly is not support the histogram, that's why I use any for it.

src/components/CatalogOverlay/CatalogSubplotComponent/CatalogSubplotComponent.tsx
```diff
@@ -227,22 +235,35 @@ export class CatalogSubplotComponent extends React.Component<WidgetProps> {
...
+                const points = event.points;
+                selectedPointIndices = Array(points.length);
+                for (let index = 0; index < points.length; index++) {
+                    selectedPointIndices[index] = points[index].pointIndex;
                 }
-    
-                if (this.widgetStore.plotType === CatalogPlotType.Histogram) {
-                    for (let i = 0; i < selectedPoint.pointIndices.length; i++) {
-                        selectedPointIndices.push(selectedPoint.pointIndices[i]);   
+            }
```
veggiesaurus: shouldn't this be an `else if`?

Issue #888, opened 2020-04-20
include catalog overlay when export as png


kswang1029:
Catalog overlay should also be included when exporting as png

veggiesaurus:
@panchyo0 can you implement this by taking the same approach used for contour and raster images?

kswang1029:
looks like the Retina display issue also applies to the exported png. The catalog layer looks blurry. 
![paws-pdbi+30m-12co10-6as-mom0 fits-image-2020-7-21-14-55-27](https://user-images.githubusercontent.com/20819712/88022713-5ca87980-cb62-11ea-891a-e44c3b1ab6b7.png)

Issue #889, opened 2020-04-20
blurry catalog overlay on Retina display


kswang1029:
It appears that the rendered catalog overlay is blurry on Retina display. On a non-retina display, the catalog overlay is sharp.

veggiesaurus:
It looks like plotly is not taking into account the `devicePixelRatio`: https://github.com/plotly/plotly.js/issues/3246#issuecomment-439175019

I tried to solve the issue by doubling the desired size and then transforming using CSS. It looked great, but it didn't seem to work nicely with the single point selection stuff. or with zooming. @panchyo0 if you want to give it a try, here's a patch of what I tried:

```
Index: src/components/ImageView/CatalogView/CatalogViewComponent.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/components/ImageView/CatalogView/CatalogViewComponent.tsx	(revision Local Version)
+++ src/components/ImageView/CatalogView/CatalogViewComponent.tsx	(revision Shelved Version)
@@ -40,7 +40,7 @@
             data.marker = {
                 symbol: catalog.shape, 
                 color: catalog.color,
-                size: catalog.size,
+                size: catalog.size * devicePixelRatio,
                 line: {
                     width: 1.5
                 }
@@ -100,8 +100,8 @@
 
     render() {
         const frame = this.props.appStore.activeFrame;
-        const width = frame ? frame.renderWidth || 1 : 1;
-        const height = frame ? frame.renderHeight || 1 : 1;
+        const width = frame ? frame.renderWidth * devicePixelRatio || 1 : 1;
+        const height = frame ? frame.renderHeight * devicePixelRatio || 1 : 1;
         const padding = this.props.overlaySettings.padding;
         let className = "catalog-div";
         if (this.props.docked) {
@@ -160,6 +160,7 @@
                     layout={layout}
                     config={config}
                     onClick={this.onClick}
+                    style={{transform: "scale(0.5)", transformOrigin: "top left"}}
                 />
             </div>
         );
```

panchyo0:
Using `transform: scale()` will cause this [bug](https://github.com/plotly/plotly.js/issues/888).
Example to show the offset with css scale() https://codepen.io/anon/pen/ZXdYdb. 
Since we are using the default onClick() for source selection, this will effect us.

veggiesaurus:
[...]

PR #1020, opened 2020-07-21
export image with catalog overlay; add pix0 and pix1 tooltip

panchyo0:
issue #899 and #888

### DIFF
--- a/src/components/ImageView/ImageViewComponent.tsx
+++ b/src/components/ImageView/ImageViewComponent.tsx
@@ -39,7 +39,7 @@ export const getImageCanvas = (padding: Padding, backgroundColor: string = "rgba
         regionCanvas = regionQuery[0] as HTMLCanvasElement;
     }
 
-    const catalogQuery = $(".catalog-plotly").children().children().children(".gl-container").children(".gl-canvas-context");
+    const catalogQuery = $(".catalog-plotly")?.children()?.children()?.children(".gl-container")?.children(".gl-canvas-context");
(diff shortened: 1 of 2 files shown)


src/components/ImageView/ImageViewComponent.tsx
```diff
@@ -38,6 +39,11 @@ export const getImageCanvas = (padding: Padding, backgroundColor: string = "rgba
         regionCanvas = regionQuery[0] as HTMLCanvasElement;
     }
 
+    const catalogQuery = $(".catalog-plotly").children().children().children(".gl-container").children(".gl-canvas-context");
```
veggiesaurus: should use `?` before each of the `.children()` values just in case

veggiesaurus (changes requested):

src/components/ImageView/ImageViewComponent.tsx
panchyo0: Thanks

veggiesaurus (approved):
looks fine to me. @kswang1029 please check. We can keep the blurry image issue constrained to #889

kswang1029:
Hmm, problem when export as png. 
h_m51_i_s05_drz_sci fits-image-2020-7-22-9-14-28

On screen I see
Screen Shot 2020-07-22 at 9 14 39 AM

It feels like the deviceRatio issue? The above png is generated with a _non-retina_ display (iMac27 2K screen).

veggiesaurus:
can confirm latest commit fixes issue :+1:

PR #1022, opened 2020-07-21
Thchang/add stokes info in cursor value bar 743

### LINKED ISSUE #743: add Stokes to information in cursor value bar for Stokes cubes
For images that have a Stokes axis we should at Stokes to the information in the cursor value bar.  Below is an image of Stokes U.  It would be good to have the Stokes displayed with the cursor information, otherwise there is no indicator of what Stokes is being viewed.  Eg:  Stokes: U

Screenshot 2020-02-17 at 09 58 09

TienHao:
close #743 
Use frame header to get current stokes information.


kswang1029 (commented):
should just show "I" or "Q" or "U" or "V", not "Iz", "Qz", "Uz" or "Vz".

veggiesaurus (approved):

PR #1059, opened 2020-07-22
default to building WebAssembly scripts in a docker container

veggiesaurus:
I don't know why we didn't do this sooner...

The worst part about building the frontend (in fact, the only really challenging part) is the WebAssembly library and wrapper build process. This can be made _so_ much easier if we rely on a docker container for building them. 

@ajm-asiaa please check if this works ok for you. As far as I can see, this means that the only _actual_ dependency for building the frontend is now NodeJS. So there should be no need to build the entire frontend itself from within a container, since NodeJS is available everywhere


ajm-ska:
@veggiesaurus That's a great idea! Assuming everyone has access to Docker.

I found a recently updated pre-built Docker container here: https://hub.docker.com/r/trzeci/emscripten/
Its 'latest' version is using 1.39.18 but that might change to the troublesome 1.39.19 soon, so we better make sure to use a specific version; `trzeci/emscripten:1.39.18-upstream`

It appears we can actually use it to build the entire frontend, so we actually don't need to worry about NodeJS either.

I propose the simplest way is we just log in to the container with the carta-frontend directory mounted inside it, follow our normal build procedure, and then exit the container:

> cd carta-frontend
> **docker run -ti -v $PWD:/src trzeci/emscripten:1.39.18-upstream /bin/bash**
> git submodule init
> git submodule update
> npm install
> cd wasm_libs
> ./build_libs.sh 
> cd ..
> cd wasm_src
> ./build_wrappers
> cd ..
> cd protobuf
> ./build_proto.sh
> cd ..
> npm run build
> exit

The only potential issue is that I ran out of memory when doing `npm run build` first time:

> The build failed because the process exited too early. This probably means the system ran out of memory or someone called `kill -9` on the process.

I see npm uses `Using 1 worker with 2048MB memory limit`. My default Docker memory setting is 2GB. I'm not sure if that is exactly 2GB (2000MB) or not. I dialed the Docker resources up slightly to 2.25GB and then it worked!

So if we are going by what I feel is the easiest route to also use `npm` inside Docker, we should mention that the user may need to slightly increase their assigned Docker resources to at least 2.25GB.

veggiesaurus:
@ajm-asiaa I think that's a good idea, but we should probably provide two options: build WebAssembly with docker and build the entire frontend with docker. Reason being that debugging things outside of a docker container is often a lot easier than from within, and for dev builds (i.e. `npm run start`) it would be helpful to be able to run that outside of the docker container. Maybe we need to just add an `npm run build-docker` script, that performs the WebAssembly scripts from inside docker?

That way, if running from outside docker, you'd just run `./build_libs_docker.sh` and then `npm run build-docker` but if running inside docker already, you'd just run `npm run build`

ajm-ska:
@veggiesaurus Good point. 
Actually I haven't been able to test your changes. Did you forget to commit the new `build_wasm_wrappers_docker.sh` script as I can not find it?

ajm-ska (approved):
@veggiesaurus Looks good to me. It certainly simplifies the build procedure a lot. Both the new Docker and Emscripten build steps worked for me to produce a production build. The README is also a lot easier to follow than before. Nice work!

PR #1061, opened 2020-07-23
Angus/singularity build scripts

veggiesaurus:
Extends the docker WebAssembly approach to Singularity as well, since this means we don't need `sudo` privileges.
Currently still pulling the image from docker hub. I haven't found an appropriate image for singularity.


ajm-ska:
@veggiesaurus I tried on ubuntu 18.04 and it works. It was using singularity version `2.6.1-dist` that I installed from the NeuroDebian repository.

However, when I tried on CentOS7, there was problem. The easiest way to install singularity on CentOS7 seems to be the EPEL repository and that installs singularity version `3.6.0-1.el7`. 
The first problem is that this version of singularity does not use the "check" keyword so the `build_wasm_libs_singularity.sh` script fails. 

> Error: unknown command "check" for "singularity"

I can't seem to find common commands between the two versions of singularity that could work instead. For the CentOS7 version, "search" or "verify" might work, but those don't exist for the Ubuntu version. It is strange that there is such a big difference between these two versions of singularity.

Also, I don't how it happens, but I even have trouble with the singularity pull command:

> singularity pull docker://trzeci/emscripten:1.39.18-upstream
> FATAL:   Image file already exists: "emscripten_1.39.18-upstream.sif" - will not overwrite

I need to use -F

> singularity pull -F docker://trzeci/emscripten:1.39.18-upstream

But then for some reason it produces a **.sif** file:

> INFO:    Creating SIF file...
> Building WebAssembly libraries using singularity
> FATAL:   could not open image /home/carta-jenkins/carta-frontend/emscripten-1.39.18-upstream.simg: failed to retrieve path for /home/carta-jenkins/carta-frontend/emscripten-1.39.18-upstream.simg: lstat /home/carta-jenkins/carta-frontend/emscripten-1.39.18-upstream.simg: no such file or directory

Plus the created .sif file has an underscore after emscripten for some reason: `emscripten_1.39.18-upstream.sif`

So I need to change 
`singularity exec emscripten-1.39.18-upstream.simg npm run build-libs` 
to 
`singularity exec emscripten_1.39.18-upstream.sif npm run build-libs`

The build procedure then works for a little while, then I get this issue:

> OSError: [Errno 30] Read-only file system: '/emsdk_portable/.data/cache.lock'
> Checking for Zstd bitcode...
> Not found!

So somehow `singularity version 3.6.0-1.el7` on CentOS7 is quite troublesome. I should see if I can install a version closer to Ubuntu's version 2.6.1, or we recommend not to use singularity on CentOS7/RHEL. 
What version of singularity are you using on Ubuntu 20.04?

I haven't tried it on MacOS yet as the installation looks complicated requiring Virtual Box and Vagrant installed first.

veggiesaurus:
@ajm-asiaa good point. I was using singularity 2.6, I'll make it compatible with both 2.x and 3.x

ajm-ska (approved):
Adding Singularity support is a great idea. The latest fixes for singularity 3.x appear to work perfect for version `3.6.0-1.el7` on CentOS7 while still working fine for 2.x, i.e. version `2.6.1-dist` on Ubuntu. Excellent work!

PR #1062, opened 2020-07-25
Qi/update catalog interactions with different image files

panchyo0:
1. associated each image file to a catalog file list. when user switch image file, only display associated catalog table view and plots view.
2. open image will clear all catalog data, if we already have image file opened.
3. close file will close all associated catalog file data, keep UI alive.
4. refactor  catalog widget store.
5. fixed issus #984


veggiesaurus (changes requested):
it would be good to have `catalogStore` as a public member of `AppStore`, so that the instance can still be accessed via `appStore.catalogStore`. This is particularly useful for scripting, and for situations where a component _already_ has an `AppStore` instance. 

that way you can also keep using `this.catalogStore` in the appstore functions

veggiesaurus (approved):

PR #1073, opened 2020-07-28
Thchang/toast message improvements

### LINKED ISSUE #925: display warning/error toast longer in time
As an example, when exporting regions to a file, we use toast to display the state of request and error messages if failed. It would be good to have different display time for different states. If it is success (green toast), the display period can be as it is now. For orange and red toast, we need longer display time so that users can have sufficient time to read (or copy-paste) error messages. It would be good to have a "dismiss" button in red toast so that users can read the error messages with no pressure.

TienHao:
close #925 
Added `copy to clipboard` button in the error and warning toasts.


veggiesaurus (approved):
I think we might need an "info" toast type as well in future, but this is fine for now :+1:

PR #1077, opened 2020-07-29
clamp input values correctly when out of bounds values are entered

veggiesaurus:
Sorts out an issue where incorrect values were being sent to the animator when manually entered values were out of the prescribed range.


kswang1029:
One special case is 0. Need to clamp it too (to 1).

PR #1079, opened 2020-07-29
Jason/spectral line limit

### LINKED ISSUE #1066: spectral line query: performance impact when rendering too many line overlay
Roughly the GUI becomes less responsive when more than a thousand line labels are rendered. With the spatalogue query, it is very easy to get many lines with a narrow frequency range. From the usability point of view, rendering too many labels at one time does not help much in the analysis and result in poor UX when zooming the profile. I suggest we limit the total number of lines up to 1000 to be rendered, so that only the first 1000 _selected_ lines will be rendered.  @duidae @veggiesaurus thoughts?

duidae:
closes #1066

image

### DIFF
--- a/src/components/SpectralLineQuery/SpectralLineQueryComponent.tsx
+++ b/src/components/SpectralLineQuery/SpectralLineQueryComponent.tsx
@@ -348,7 +348,14 @@ export class SpectralLineQueryComponent extends React.Component<WidgetProps> {
             className += " dark-theme";
         }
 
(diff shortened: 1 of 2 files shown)


kswang1029:
I suggest we show that 1000 line limitation in the tooltip _only_ when users select more than a 1000 lines. 

In addition, if there is no spectral profiler, we probably need to use the tooltip to remind users we need at least one spectral profiler widget to overplot line labels.

veggiesaurus (approved):

PR #1085, opened 2020-08-05
Thchang/beam elements update

### LINKED ISSUE #804: provide options to plot beam elements of contour layers
As a common practice, beam elements are shown per image layer. For example:
Screen Shot 2020-03-02 at 2 03 39 PM
One is for the raster image, and the other is for the contour image.

TienHao:
closes #804 
Image viewer now shows beams of contour layers which are referencing active image. The center points of referencing beam are align with the center point of referenced beam(raster image).

Screenshot 2020-08-05 18:11:53


veggiesaurus (changes requested):
- Crashes when loading a file without beam info
- Warnings displayed in dev console, because each node returned from `plotBeam` needs to have a unique key. I suggest you use the frame ID

kswang1029:
this is what the console says when crashing

Screen Shot 2020-08-07 at 2 42 32 PM

veggiesaurus (changes requested):
beams for the active image are no longer shown. It's now only showing the beams of the contoured spatial siblings

veggiesaurus (approved):

PR #1086, opened 2020-08-06
Added cog icon for image viewer

### LINKED ISSUE #1003: Create cog for changing settings in image viewer
Other widgets have a cog to change settings. The image viewer needs to be made consistent with this.

TienHao:
closes #1003


kswang1029 (approved):
this looks nice. 
Just need to make the overlay settings dialogue a bit taller so that options in different tabs can be displayed properly without being cut off

veggiesaurus (changes requested):
This works well, but I think we should now do the following:
- Remove the overlay settings icon from the menu and toolbar
- Move the overlay settings from a separate dialog to a settings widget  (e.g. spectral profiler widget). @kswang1029 and I think it'd look better to keep the vertical tabs though
- Rename it to Image view settings

PR #1096, opened 2020-08-10
adds basic fuzzy search

### LINKED ISSUE #1092: Feature request: case insensitive fuzzy search in file browser
While the file browser is open filter matches by fuzzy search while typing.

veggiesaurus:
Closes #1092 

Adds fuzzy search as default search mode. For unix search, start a string with `+`. For regex, start and end with `/`

(`package.json` automatically alphabetised by running `npm install`)


kswang1029:
@veggiesaurus I feel the unix style is now slightly harder to use. For example, with dev
Screen Shot 2020-08-11 at 8 52 59 AM
`*tabular*.image` gives use one result.
With this branch, it does not:
Screen Shot 2020-08-11 at 8 50 00 AM
instead, we need an extra + sign
Screen Shot 2020-08-11 at 8 50 14 AM

This becomes not consistent with the unix style search. (e.g. `ls *tabular*.image`)

kswang1029:
@veggiesaurus I suggest we revise the hint text a bit, such as 
`Filter by filename with fuzzy search. To use unix style search, do +<expression>. To use regular expression, do /<expression>/`.

It is a bit lengthy but it is clearer to let user know we support different modes with the fuzzy one as default.

duidae (approved):

PR #1100, opened 2020-08-11
Angus/matched contour animation

veggiesaurus:
Sends a map of file IDs and matching channels, in order to automatically update the channels of spatially matched images when animating. Companion PR of https://github.com/CARTAvis/carta-backend/pull/582


src/stores/AnimatorStore.ts
```diff
@@ -171,6 +192,10 @@ export class AnimatorStore {
         return 1000.0 / clamp(this.frameRate, this.minFrameRate, this.maxFrameRate);
     }
 
+    @computed get serverAnimationActive() {
+        return this.animationState === AnimationState.PLAYING && this.animationMode !== AnimationMode.FRAME;
```
duidae, line 196: the original logic in animate() is 
this.animationState === AnimationState.PLAYING && this.animationMode === AnimationMode.FRAME
is this typo?

veggiesaurus: good catch! you're right. Line 151 shouldn't be replaced by `serverAnimationActive`. I've updated

duidae (approved):

PR #1102, opened 2020-08-11
show image filename as context menu header in the image list

veggiesaurus:
image

To avoid accidentally closing images, the filename is now shown as a header in the context menu.

bonus: added ability to close other images or all images


kswang1029 (approved):
this looks nice

PR #1104, opened 2020-08-12
Qi/issue 1098 & 1099

### LINKED ISSUE #1098: blinking tooltip of the filters in the catalog widget
This only happens when there is no entries after filtering. When we want to redefine a filter, the tooltip is blinking resulting in an unpleasant UX.

Aug-11-2020 11-00-29

### LINKED ISSUE #1099: add catalog filename in the File dropdown of the catalog widget
Currently the File dropdown only shows a number for a file. It would be good to show something like
```
1: a.xml
2: b.votable
3: c.fits
```
when the dropdown is expended. 

Screen Shot 2020-08-11 at 11 10 26 AM

panchyo0:
1. fixed #1098, replace `Tooltip` with `Popover` with minimal style (with `pointer-events: none` not solving the problem)
2. fixed #1099, since we have current file name displayed as title, only show file name in dropdown menu. This also save some  space.


kswang1029 (changes requested):
when a catalog is just loaded, the catalog widget is a floating window with the catalog file name in the title. However, once we dock it, the file name disappears (only showing "Catalog"). 
Screen Shot 2020-08-12 at 2 39 27 PM

kswang1029:
about the "tooltip" styling, this new approach works but it is not consistent with the tooltip in other places. Not sure which style we should adopt. @veggiesaurus ?

veggiesaurus:
> about the "tooltip" styling, this new approach works but it is not consistent with the tooltip in other places. Not sure which style we should adopt. @veggiesaurus ?

We should be consistent with the tooltip used elsewhere

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -155,6 +156,7 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
         if (!CatalogStore.Instance.catalogProfiles.has(this.props.id)) {
             CatalogStore.Instance.catalogProfiles.set(this.props.id, 1);
         }
+        this.catalogFiles = new Map<number, string>();
```
veggiesaurus: This should be renamed to `catalogFileNames`

veggiesaurus (approved):

Issue #1103, opened 2020-08-12
cant save layout


duidae:
layout is not saved in current dev branch

veggiesaurus:
@duidae could you provide more details? layout saving works for me with dev branch

duidae:
@veggiesaurus checked again with my dev & mix'n match, the saved layout is gone after reload: 
![55vup-fux7a](https://user-images.githubusercontent.com/25971696/89988537-49488400-dcb2-11ea-9121-f35e929eec24.gif)

veggiesaurus:
I can't reproduce this with mix 'n match. Can you check what's happening in your local storage? Could your new layout be failing the JSON schema check?

PR #1081, opened 2020-08-03
Angus/database service

veggiesaurus:
Closes #912, closes #911 

This PR introduces an API service, which is used for preferences or layouts in an agnostic manner. For the server version, HTTP requests are used. For the desktop application (or simple server installations without authentication), local storage is used, preferences are stored as a single JSON string and validated against a JSON schema using AJV, removing the need for manual validator functions. Layouts are stored in the same way as before, but are validated against a much more comprehensive JSON schema, again removing the need for manual validation. 

When encountering a local storage without the `preferences` item, this item is automatically created, and older versions of the preferences (in separate untyped key-value pairs) are upgraded to this new approach.

This PR also adopts the new authentication approach, which allows us to remove all mongo and JWT behaviour from the backend itself.

Issue #980, opened 2020-07-03
broken layout issues associated with catalog widget and scatter plot widget


kswang1029:
To reproduce:
1) load an image and a catalog and enable 2D scatter plot
2) dock the 2D scatter plot and save the layout (eg cat_scatter1)
3) undock the 2D scatter plot (as a floating widget) and save the layout (eg cat_scatter2)
4) reload CARTA

Now there are two issues:
5) once CARTA is just reloaded, switch to cat_scatter1 and we will see the catalog widget is disappeared (scatter plot widget too)
6) once CARTA is just reloaded, switch to cat_scatter2 and we will see the frontend crashes with a blank page.

veggiesaurus:
@duidae will you work with @panchyo0 to expand the layout support to include catalogs? @panchyo0 it would also be great if we could include the (percentage) position of the vertical divider in the layout

duidae:
ok

duidae:
@veggiesaurus @panchyo0 @kswang1029 , about supporting saving scatter plot widget in layout, there may be some issues needed to be considered:

1. Once I dock both catalog overlay widget & scatter plot widget and save this layout, seems something breaks when restart frontend.
2. Scenario of saving ONLY scatter plot widget: do we need to support this? because so far it is not feasible since closing catalog overlay widget will also close scatter plot widget.

duidae:
@panchyo0 @veggiesaurus I've added the layout support for catalog/scatter plot in **jason/layout_support_catalog** branch, and since catalog widget has a special design for the widget id & scatter plot, the rest may need your help for the id part & deciding whether to support scatter plot or not in layout.

veggiesaurus:
we should add a separate issue for improving the catalog support in layouts after adjustments to the catalog UI are made

PR #1105, opened 2020-08-12
fixed support for floating widgets in layout

veggiesaurus:
Adjusts floating widget saved config to be consistent with docked widgets

### DIFF
--- a/src/models/LayoutConfig.ts
+++ b/src/models/LayoutConfig.ts
@@ -165,7 +165,8 @@ export class LayoutConfig {
         // 2. handle floating widgets
         appStore.widgetsStore.floatingWidgets.forEach((config: WidgetConfig) => {
             let floatingConfig = {
-                type: config.type,
+                type: "component",
+                id: config.type,
                 defaultWidth: config.defaultWidth ? config.defaultWidth : "",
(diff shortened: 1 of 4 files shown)


duidae:
thanks for fixing, btw it's better tested before merge to ensure the functionality correctness (seems that @kswang1029 did not approve the previous PR #1081?)

kswang1029 (changes requested):
when there is a docked catalog widget in the layout, saving is fine but in a new session we cannot see the saved layout.

veggiesaurus:
> when there is a docked catalog widget in the layout, saving is fine but in a new session we cannot see the saved layout.

This is just designed to fix floating layout widgets. Catalog widgets (#980) will be handled by @duidae in a separate PR

kswang1029:
@veggiesaurus ok. in that sense, it works well

kswang1029 (approved):

veggiesaurus:
@kswang1029 @duidae I've made a couple of changes, so that older layouts should still work. Please re-review

duidae (approved):

kswang1029 (approved):
yes old layouts still load

duidae:
closes #1103

PR #1110, opened 2020-08-12
Angus/fix 1109

veggiesaurus:
Uses the _reference image_ for region information, rather than recalculating when the active frame changes. That way, WCS and pixel information is consistent. The region's associated file name is now shown in the title bars of the region list and region editing dialog.

image


duidae (approved):

kswang1029 (commented):
The works fine. I suggest we change the default coordinate system from Image to World so that the inconsistency in image coordinate is less subtle as the primary focus usually is the text input fields.

veggiesaurus:
Ok, I'll make that change and then merge

PR #1111, opened 2020-08-12
adds default region on click

veggiesaurus:
Adds similar behaviour to DS9: Ellipse and rectangular regions can be created with a default size by simply clicking instead of clicking and dragging. 

default_regions

The default region radius can be configured in the preferences->region panel. As with DS9, the created region's size depends on this default value and the zoom level.

### DIFF
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -402,6 +402,15 @@ export class PreferenceDialogComponent extends React.Component {
                         {regionTypes}
                     </HTMLSelect>
                 </FormGroup>
+                <FormGroup inline={true} label="Region radius" labelInfo="(px)">
+                    <SafeNumericInput
+                        placeholder="Region radius"
+                        min={1}
+                        value={preference.regionRadius}
+                        stepSize={1}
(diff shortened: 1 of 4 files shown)


veggiesaurus:
note no breaking changes with preferences schema, so no updated version number is required

duidae (approved):

kswang1029 (commented):
This works fine. 
The term "Region radius (px)" is a bit confusing for rectangle.

If I set the zoom level to 1x (432 px on screen) and I create a circle, we have a circle with a radius of 432 screen px as expected.
Screen Shot 2020-08-13 at 9 59 19 AM

For square, same thing happens (a bit confusing).
Screen Shot 2020-08-13 at 9 58 34 AM

Is it possible to use the short side (in screen px) of the image viewer with a factor (0.2?) as the size of the region? Then we don't need the preferences possibly. 

Screen Shot 2020-08-13 at 10 09 29 AM

veggiesaurus:
Perhaps I should use "region size" instead?

kswang1029:
@veggiesaurus that could work. I think this is not a big deal in general (regarding the default size on screen). Just need to make sure the term in preference is less ambiguous

veggiesaurus:
If we use the term "size" to mean width of the square and diameter of the circle, that would probably be consistent

PR #1113, opened 2020-08-13
Fix warning

duidae:
`fix warning of <ul> cant be descendant of <p>` 

image


kswang1029:
There might be other help contents with same problem... need to double check.

duidae:
yeah there are a few

kswang1029:
Thanks for fixing 🙈

That’s I didn’t follow the HTML rules or it is react doesn’t like it? Or else?

duidae:
`<p></p>` tag can only contain inline elements according to W3C spec.

veggiesaurus (approved):

PR #1122, opened 2020-08-18
Disabling data source select of contour dialog when animating

### LINKED ISSUE #1120: disabling the data source dropdown of the contour config dialogue while animating
As spotted by @zarda 

There are three places that can be used to switch to other images
1. the "image" slider of the animator
2. the image list widget
3. the "data source" dropdown of the contour config dialogue

When animating, 1. and 2. are disabled so that we cannot switch to other images. The same constraint should be applied to 3. At the moment, we can select other images with the dropdown but the image viewer does not respond to it (which is correct). We will just need to disable the dropdown when animating. 

@veggiesaurus backport 1.4-final?

TienHao:
closes #1120


kswang1029 (approved):
it works 👍

veggiesaurus:
@TienHao this looks good. I'll merge in a backport as well

TienHao:
> @TienHao this looks good. I'll merge in a backport as well

Thanks!

PR #1124, opened 2020-08-19
improvements of the colormap in stokes scatter plot

### LINKED ISSUE #990: allowing a custom range for rendering the scatter plot in the Stokes widget
Currently, the range for rendering the scatter plot in the Stokes widget based on a colormap is determined by the full spectral range of the cube. This is nice when there is only one feature in the spectral coverage since the full colormap is mapped to mostly full feature in the spectral coverage. However, possibly there are multiple features in one cube. Then only a subset of the colormap is used to render a feature, potentially resulting in a scatter plot with much less contrast, such as the following example.
Full range:
Screen Shot 2020-07-09 at 11 49 15 AM
Zoom to the strongest line:
Screen Shot 2020-07-09 at 11 50 45 AM

To improve this, shall we allow users to customize the spectral range for color mapping? That is, in the above example, we can use the full colormap to render the line feature only. Or, the spectral range of color mapping is automatically determined by the zoomed range of the QU/PI/PA plots?

@arusstaylor @veggiesaurus @panchyo0 thoughts?

TienHao:
closes #990 
The color mapping of Q, U scatter plot is determined by the zoomed range of the QU/PI/PA plots.


panchyo0:
when user click on scatter plot, should the indicator (channel) move to user selected one? @kswang1029

kswang1029:
Do you mean if we click a point on the scatter plot, the image changes to the corresponding channel? This is not possible if the smoothing type is binning. Or do you mean we draw a vertical line as the indicator on the smoothed profile? Then yes, I think we need the indicator.

kswang1029:
@arusstaylor could you give it a try with 
http://carta.asiaa.sinica.edu.tw/frontend/ca48fa7/?socketUrl=wss://carta.asiaa.sinica.edu.tw/socketdev
using the images in set_Stokescube? Thanks.

veggiesaurus:
> Do you mean if we click a point on the scatter plot, the image changes to the corresponding channel? This is not possible if the smoothing type is binning. Or do you mean we draw a vertical line as the indicator on the smoothed profile? Then yes, I think we need the indicator.

How do we handle the situation for the spectral profiler? Surely we should be consistent

veggiesaurus:
currently, clicking on a particular point in the scatter plot results in the channel being set to 0. This seems like an independent issue though. @kswang1029 is this PR ok? Code looks fine to me

kswang1029:
@veggiesaurus yes the color part is fine. The red indicator does not.
Screen Shot 2020-09-02 at 9 12 51 PM

TienHao:
@kswang1029 : I find that other profile widgets(spatial, spectral) with indicator also have this problem. Should we file another issue for this?
indicator

kswang1029:
Hmmm.... is that a regression? @TienHao please help to file an issue.

PR #1158, opened 2020-08-24
only load google API when required

### LINKED ISSUE #1157: Google APIs are loaded even when not using Google authentication
As discovered by @lh-astro, the Google API javascript code is loaded by the frontend even when not required. It should only be loaded once the frontend determines that Google auth is being used for the server. For LDAP-based servers, there should be no need to load the Google APIs.

veggiesaurus:
fixes #1157


ajm-ska (approved):
I have only tested the case where the Google API _is_ required. I used an older version of carta-node-server and had its frontend symlink pointing towards this built branch. Our carta-node-server setup only uses Google login. It worked. It logged in and this frontend loaded and worked fine. So I assume everything is OK.

duidae (approved):

PR #1160, opened 2020-08-25
fixes #1117

### LINKED ISSUE #1117: file browser forgot its pwd after moment generation
(not sure if this is due to frontend or backend)

To reproduce:
1) load an image cube (and remember the path)
2) generate a moment image
3) launch file browser and it should point to $root (should be the path in 1))

veggiesaurus:
keeps the current file browser directory constant instead of reverting to the home directory when generating moment images


kswang1029 (approved):
confirmed working 👍

duidae (approved):

PR #1161, opened 2020-08-25
ensures animation is stopped before closing any files

### LINKED ISSUE #977: frontend crashes or behaves abnormally if closing an image while animation playback is on
to reproduce
1) load set_M17/M17_SWex.fits
2) set fps to 15
3) enable animation playback
4) after few seconds close the image
5) launch the file browser
6.1) the frontend might crash (entire GUI gone with an empty page)
6.2) if the frontend survived, the entire UI behaves abnormally when a new image is loaded (no histogram, spatial profile, etc)
6.3) potentially the backend might crash 

Possibly we need to send a STOP_ANIMATION message before CLOSE_IMAGE if animation playback is on?

veggiesaurus:
fixes #977

### DIFF
--- a/src/stores/AppStore.ts
+++ b/src/stores/AppStore.ts
@@ -406,8 +406,6 @@ export class AppStore {
     };
 
     @action openFile = (directory: string, file: string, hdu: string) => {
-        // Stop animations playing before loading a new frame
-        this.animatorStore.stopAnimation();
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
yes this seems to work. We can load new images without problems. All things appear functional.

duidae (approved):

PR #1162, opened 2020-08-25
fixes #1053

### LINKED ISSUE #1053: selecting texts with cursor in the file info box has an offset
In other words, it seems we cannot select the correct row of texts, where the cursor is pointing at.
Jul-22-2020 11-11-40

veggiesaurus:
Not entirely sure why I added the 15px padding to start with... @kswang1029 please check for weird rendering behaviour on Safari. Looks fine to me on Chrome and FF

### DIFF
--- a/src/components/FileInfo/FileInfoComponent.scss
+++ b/src/components/FileInfo/FileInfoComponent.scss
@@ -18,9 +18,8 @@
         white-space: pre;
 
         .header-entry {
-            padding-left: 15px;
+            padding-left: 5px;
             padding-right: 15px;
-            padding-top: 15px;
             .header-name {
                 color: $blue2;
                 font-weight: bold;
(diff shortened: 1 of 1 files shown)


kswang1029 (commented):
This fixes the selection problem. I think with padding it looks nicer however.
was
Screen Shot 2020-08-26 at 9 41 36 AM
now
Screen Shot 2020-08-26 at 9 41 58 AM

veggiesaurus:
> This fixes the selection problem. I think with padding it looks nicer however.

Unfortunately the padding is what was causing the selection offset. I'll see how else I can fix it :thinking:

veggiesaurus:
@kswang1029 please retest with `0150649`

kswang1029 (approved):
wonderful 🚀

PR #1163, opened 2020-08-25
fixes #1121

### LINKED ISSUE #1121: Image slider deselected in animator when image appended
When a new image is appended, the image/frame slider appears in the animator. However, it is deselected, and then animator buttons don't work unless you first select it. It should be automatically selected.

veggiesaurus:
- If appending a 2D image, slider changes to "image"
- If _opening_ a 3D image (i.e. no other images open), slider changes to "channel"
- If _opening_ a 2D+1 (2D Stokes cube), slider changes to "stokes"

### DIFF
--- a/src/stores/AppStore.ts
+++ b/src/stores/AppStore.ts
@@ -352,7 +352,7 @@ export class AppStore {
             this.animatorStore.setAnimationMode(AnimationMode.FRAME);
         } else if (this.frames.length === 1 && newFrame.frameInfo.fileInfoExtended.depth) {
             this.animatorStore.setAnimationMode(AnimationMode.CHANNEL);
-        } else if(this.frames.length === 1 && newFrame.frameInfo.fileInfoExtended.stokes) {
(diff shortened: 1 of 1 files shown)


kswang1029 (changes requested):
if we 
1) load a 2D image
2) append a 3D cube
we should see channel is selected, instead of image.

if we 
1) load a 2D+Stokes cube [8064, 8064, 1, 4]
we should see the stokes radio button is selected. now it is not.

I would prefer we try to keep channel selected in any cases, and if not possible (not exist), then select Stokes, and if not possible (not exist), then select image. Animating channels is the primary use case of the animator I think (for cubes). Animating images/stokes (to be precise, "blinking" in this case) should be the secondary choice and be triggered by users intentionally.

veggiesaurus:
@kswang1029 is this behaviour ok? (commit `d5887b8`)

kswang1029:
@veggiesaurus yes~

PR #1167, opened 2020-08-26
Jason/spectral line filter

### LINKED ISSUE #1236: Support filtering and sorting in the spectral line query table
after spectral line query and displaying the result as a table, the table should provide filtering and sorting function so that it is much easier to select target lines for line ID overlay.

duidae:
closes #1236 

companion backend PR https://github.com/CARTAvis/carta-backend/pull/614


duidae:
bugs fixed, please have a look, thanks.

kswang1029:
@duidae would three issues, please have a look and let me know if you cannot reproduce them.

- when the resulting total number of lines is 0, we will not be able make a new query by refining a filter. To reproduce:
1) make a query from 348700 to 349500 MHz with intensity limit -3.5. we should see 626 lines
2) in chemical name column, set "Glycine" and click the filter button. We should see 16 lines. (ps. We can still scroll down indefinitely beyond row 16)
3) in E_U (K) column, set <500 and click the filter button. We should see 6 lines and the filter button is active.
4) in E_U (K) column, change <500 to <300 and click the filter button. We now see 0 line and the filter button is disabled.
5) no additional filter can be applied even if we edit existing filters. The only way out is to click the reset button.

---

- issue as described in 2) of the above issue.

---

- when we click the plot button, we should plot _all_ selected lines, not the selected lines in a filtered list. To reproduce:
1) make a query from 348700 to 349500 MHz with intensity limit -3.5. we should see 626 lines
2) in Species column, set CH3CNv=0 and click the filter button. We see 11 lines.
3) select them all and we see the text says "Selected 11 line(s)".
4) Click the reset button to remove the filter and we see in the text "Selected 11 line(s)".
5) in Species column, set CH3OH and click the filter button. We see 3 lines.
6) select the 3rd line with the checkbox. Now the text says "Selected 1 line(s)".
7) click the plot button and we see 1 label.
8) click the reset button to remove the filter and we see "Selected 12 line(s)".
9) click the plot button and we see 12 labels.

The expected behavior is at step 6, we should mention there are total 12 lines selected and when we click plot, we should see 12 labels. Think this behavior is like a shopping cart. We apply _different_ filters to select different lines and put them into a shopping cart. Whenever plot button is clicked, we plot all lines in the shopping cart.

duidae:
> @duidae would three issues, please have a look and let me know if you cannot reproduce them.
> 
> * when the resulting total number of lines is 0, we will not be able make a new query by refining a filter. To reproduce:
> 
> 1. make a query from 348700 to 349500 MHz with intensity limit -3.5. we should see 626 lines
> 2. in chemical name column, set "Glycine" and click the filter button. We should see 16 lines. (ps. We can still scroll down indefinitely beyond row 16)
> 3. in E_U (K) column, set <500 and click the filter button. We should see 6 lines and the filter button is active.
> 4. in E_U (K) column, change <500 to <300 and click the filter button. We now see 0 line and the filter button is disabled.
> 5. no additional filter can be applied even if we edit existing filters. The only way out is to click the reset button.
> 
@kswang1029 Not sure what's the expectation of the result? line filtering is copying the filtering behavior of catalog widget, and seems catalog widget filters the filtered result again for multi layer filter.

> * issue as described in 2) of the above issue.

do u mean forbid infinite scrolling?

> * when we click the plot button, we should plot _all_ selected lines, not the selected lines in a filtered list. To reproduce:
> 
> 1. make a query from 348700 to 349500 MHz with intensity limit -3.5. we should see 626 lines
> 2. in Species column, set CH3CNv=0 and click the filter button. We see 11 lines.
> 3. select them all and we see the text says "Selected 11 line(s)".
> 4. Click the reset button to remove the filter and we see in the text "Selected 11 line(s)".
> 5. in Species column, set CH3OH and click the filter button. We see 3 lines.
> 6. select the 3rd line with the checkbox. Now the text says "Selected 1 line(s)".
> 7. click the plot button and we see 1 label.
> 8. click the reset button to remove the filter and we see "Selected 12 line(s)".
> 9. click the plot button and we see 12 labels.
> 
> The expected behavior is at step 6, we should mention there are total 12 lines selected and when we click plot, we should see 12 labels. Think this behavior is like a shopping cart. We apply _different_ filters to select different lines and put them into a shopping cart. Whenever plot button is clicked, we plot all lines in the shopping cart.

working on select all issue.

veggiesaurus:
@duidae I think the main thing is that there's no way to clear a single existing filter. For example, if I filter with "Glycine", I get down to 626 entries. If I then add the E_U < 300 filter, it goes to zero entries (this is correct, since there are no entries with both E_U < 300 _and_ chemical name Glycine). But now I can't remove the "Glycine" filter without resetting completely. I should be able to remove the string and click "filter" again to now apply just the E_U < 300 condition. 

Actually, there might be two separate issues here:
1) Can't change filters when the number of _filtered_ rows = 0
2) Can't remove filters by simply erasing the filter text

kswang1029:
Would it be possible to modify the filtering logic so that once a new filtering condition is made, we always use all unfiltered data for filtering with the new filters? This is less efficient than the current implementation but might be still ok from performance point of view. 

With the current more efficient filtering mechanism, then we will need to check the state of the filters. For example, if we have applied filters A and B, and users add C with A and B unchanged, we can reuse the filtered data (with A and B applied) and apply filter C. However, If users remove and/or modify A or B, we will then use all unfiltered data to do filtering. 

@duidae @veggiesaurus Could you provide some insight on how we proceed for v1.4?

duidae:
basically the change is too many for 1.4, almost 1000 lines, pretty risky & inappropriate.

veggiesaurus:
> basically the change is too many for 1.4, almost 1000 lines, pretty risky & inappropriate.

Agreed. It's something that should be beta tested.

kswang1029:
frontend commit: 11e02a9
Unable to unselect all checkboxes. 

Sep-09-2020 14-12-30

duidae:
it's not done yet, still in development

> frontend commit: [11e02a9](https://github.com/CARTAvis/carta-frontend/commit/11e02a97f05bfa8e92a65a00b478cee8d1dbcb3e)
> Unable to unselect all checkboxes.
> 
> Sep-09-2020 14-12-30

kswang1029:
remove backport tag after risky evaluation. aiming for 1.5

duidae:
@kswang1029  @veggiesaurus this PR is ready, please have a look, thanks.

kswang1029:
Tested with frontend 38ffc8e and backend 2c37433

The filtering function works nicely in general. 👍 

Some issues (mostly UX related) I spotted:

1. in the line table, is it possible to adjust the default width of each column so that long column title can be displayed properly? (eg with the current default column width, we see two "Rest Frequency" but in fact the second one is "Rest Frequency Error")
2. The equal operator `==` in the filter does not work. In the following example, the "rest frequency error" of the first two rows is 10.4286. By setting `==10.4286` we should see these two rows survived after filtering. 

https://user-images.githubusercontent.com/20819712/108585105-69ec4780-7381-11eb-8357-cba358f94509.mp4

3. Is sorting expected to be working?
4. I suggest we rename "Reset" button to "Reset filter" and add a new one next to it as "Reset selection". At the moment to remove all selected lines (selected via different filters) we need to click the checkbox at the head of the first column twice (select all then unselect all).
5. Is it possible to scroll to the top after a filter is applied? At the moment after filtering the filtered entries are displayed outside the view.

https://user-images.githubusercontent.com/20819712/108585380-475b2e00-7383-11eb-8a20-4baf3c0d3f8d.mp4

6. Is it possible to show how many filters are applied too? eg `Showing 9 filtered line(s) of total 2695 line(s). Applied 2 filter(s)`

kswang1029:
Retested :
1. The column width of "Measured Frequency" and "Measured Frequency Error" needs to be wider. Other columns look good.
Screen Shot 2021-03-23 at 1 26 32 PM

2. Confirmed the fix. 👍 

3. on hold

4. Now with a single click on the checkbox of the first column can remove all previously selected lines. Confirmed the fix. 👍 

5. It seems to show from the top but not starting from the first entry. The example below shows it starts from the 4th entry.

https://user-images.githubusercontent.com/20819712/112099281-30a43300-8bde-11eb-9ebf-708a806b4c7d.mp4

6. The filter counter should be updated _after_ new filter set is applied. At the moment it updates right after users edit a filter.

https://user-images.githubusercontent.com/20819712/112099532-ac05e480-8bde-11eb-8c97-19083d2846b6.mp4


@duidae @veggiesaurus I have done another test run. Please check.

kswang1029:
@duidae @veggiesaurus if you think proper this PR can be merged without sorting function and we can have another issue for that (ideally as part of 2.0) if you agree I will approve this PR after the remaining minor issues are fixed

duidae:
> @duidae @veggiesaurus if you think proper this PR can be merged without sorting function and we can have another issue for that (ideally as part of 2.0) if you agree I will approve this PR after the remaining minor issues are fixed

ok, thanks.

duidae:
@veggiesaurus  @kswang1029 the above issues are fixed except sorting, please have a look, thanks.

kswang1029 (approved):
Confirmed 1 and 6 are fixed. Issue 5 persists. 

https://user-images.githubusercontent.com/20819712/113836616-a05e1480-97bf-11eb-8ab1-33608ecb70ce.mp4

As issue 5 is not a showstopper, we can deal with that later.

kswang1029:
Sorry just spotted that "CDMS/JPL intensity" and "Lovas/AST intensity" columns need to be widen.

duidae:
> Sorry just spotted that "CDMS/JPL intensity" and "Lovas/AST intensity" columns need to be widen.

fixed, please have a look, thanks.

veggiesaurus (approved):

PR #1183, opened 2020-08-31
fixes #1177 and fixes #1178

### LINKED ISSUE #1177: regression: locked region can still be resized
reported by a beta tester.
to produce:
1) create a region and keep it active
2) click the lock button and we see the region is dimmer but still with control points active.
3) resize the region with the control point. We should not be able to resize a locked region.

This does not happen for v1.3.1.

### LINKED ISSUE #1178: region duplication check ignores rotation
reported by a beta tester
to reproduce:
1) load two images and match them and keep the secondary image as the active image
2) create an ellipse with longer axis in x direction
3) unmatch the secondary image (the secondary image is still the active image)
4) rotate the ellipse region on the _was-secondary_ image
5) match the image again and we see the rotated ellipse is replaced by the unrotated one.

veggiesaurus:
fixes #1177 and fixes #1178

### DIFF
--- a/src/components/ImageView/RegionView/SimpleShapeRegionComponent.tsx
+++ b/src/components/ImageView/RegionView/SimpleShapeRegionComponent.tsx
@@ -427,7 +427,7 @@ export class SimpleShapeRegionComponent extends React.Component<RegionComponentP
 
         let anchors: React.ReactNode[];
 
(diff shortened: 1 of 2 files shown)


kswang1029 (approved):
this works nicely.

duidae (approved):

PR #1185, opened 2020-09-01
minor column width adjustments

### LINKED ISSUE #1082: unable to hide a raster layer at a certain condition
When the Layers column in the image list widget is not wide enough (default is not wide enough) to show "R" and "C", "R" is not clickable. 
Screen Shot 2020-08-04 at 9 37 56 AM

We will need to manually widen the column to make "R" (and "C") clickable.

veggiesaurus:
should fix #1082


kswang1029 (approved):
yes, confirmed with Chrome, Firefox, and Safari.

PR #1189, opened 2020-09-02
removes old auth component and fixes default socket URL

veggiesaurus:
@ajm-asiaa please check that the modified approach to calculating the socket URL works properly for the desktop version


duidae (approved):

ajm-ska (approved):
I tried it on both Ubuntu and MacOS. 
I just went as far as producing the production build, `npm run build`, with the appropriate `.env.local` in place. i.e. `.env.local` containing:
```
REACT_APP_TARGET=linux
REACT_APP_DEFAULT_ADDRESS_PROD=ws://localhost:54140
```

Then added Electron with:
`meteor npm install`
`meteor npm start`

Electron still connects successfully with the carta-backend, so the Desktop versions should be fine.

However, before merging, another change you could make would be to remove:
`REACT_APP_AUTHENTICATION=false` from the `.env` file.

PR #1190, opened 2020-09-03
Thchang/export data improvements

### LINKED ISSUE #1144: add world coordinate as region properties in the tsv files
relevant  widgets:
- spectral profiler
- stokes analysis

Currently only region properties in image coordinate are included in the tsv file. We will need to add the properties in world coordinate as well.

TienHao:
closes #1144 
adding region properties in world coordinate in export data of Spectral profiler and Stokes analysis.


src/components/Shared/LinePlot/LinePlotComponent.tsx
```diff
@@ -54,7 +54,7 @@ export class LinePlotComponentProps {
     width?: number;
     height?: number;
     data?: { x: number, y: number, z?: number }[];
-    comments?: string[];
+    comments?: () => string[];
```
veggiesaurus: I'm not sure why this needs to change. Can't we still use a `get` function?

src/components/Histogram/HistogramComponent.tsx
```diff
@@ -109,7 +109,7 @@ export class HistogramComponent extends React.Component<WidgetProps> {
         return null;
     }
 
-    @computed get exportHeaders(): string[] {
+    private exportHeaders = (): string[] => {
```
veggiesaurus: @TienHao could this be kept as a get function? I'm a bit confused by the change

src/components/Shared/LinePlot/LinePlotComponent.tsx
TienHao: previously, I found that the wcs region properties won't change to correct coordinate if I change system(eg. FK4 -> Galatics) after latest update of profile with `get` function, so I came up with this solution. However, It seems OK right now with `get` function.

src/components/Histogram/HistogramComponent.tsx
TienHao: yes, I have reverted this change.

veggiesaurus (approved):

PR #1192, opened 2020-09-03
Jason/png spectral line

### LINKED ISSUE #1168: spectral profiler: add line labels in exported png
When molecular line labels are plotted on top of the profile, they should be also added to the saved png file.

duidae:
closes #1168


kswang1029 (approved):
this looks good.

One minor issue (zero impact to UX or use cases I believe) is that when there are lots of line, we hide the text  labels in the widget. The exported png, however, still contains text labels. If we insist WYSIWYG, then this needs to be addressed. Otherwise we can leave it as it is now (again zero impact).

veggiesaurus (approved):

PR #1194, opened 2020-09-07
Thchang/fix 1147

### LINKED ISSUE #1147: the region dropdown of the moments tab should hide cursor and point regions
In the example below, region 1 is a point region. In this dropdown, cursor and point region should be hidden. And if the Active region is a point region, we request full image moments. 

Screen Shot 2020-08-22 at 7 35 30 PM

TienHao:
closes #1147 
The options of cursor and point regions in the region dropdown of moment generator are disabled.
The moment generator will request **Image** data when both the region is "Active" and there is no region being selected.


veggiesaurus:
@TienHao minor merge conflicts from the other PRs

veggiesaurus (approved):

PR #1195, opened 2020-09-07
Thchang/implementing wcs in region list

### LINKED ISSUE #1123: long file name in the title of the region list widget
Screen Shot 2020-08-19 at 12 22 02 PM
Screen Shot 2020-08-19 at 12 05 35 PM

TienHao:
closes #1123 

Displaying region properties in world coordinate and angular size in the region list widget.

Note:
Region list shows wcs info of `cursorInfo` in the cursor row, since the wcs precision of Region Dialogue(fixed to 10) is different from `cursorInfo` (from 1 to 25). 
The wcs info of `cursorInfo` won't be update immediately after changing the system coordinate(eg. FK4 --> GAL). It only changes after the cursor moves. This also happens in the wcs info of image viewer banner.


veggiesaurus (approved):
I'm happy with the code. @kswang1029 are you ok with the functionality and the number of decimals?

One minor thing I thought we might want to try is to see if we can get the icons vertically centered as well. Not sure how it will look though
image

kswang1029:
We can try to make the decimals smart in 1.5. We will need to support down to microarcsecond scale so it is not that straightforward.

kswang1029:
@TienHao @veggiesaurus maybe we need to center **name**, **type**, and **P.A.** too? It might look better that way.🤔

veggiesaurus:
yes, I think this looks better :+1:

PR #1197, opened 2020-09-07
Add streaming data status of all spectral widgets

### LINKED ISSUE #1196: Disable all moment generate buttons while making a spectral profile
It is better to disable *all* `generate` buttons when opening multiple moment generator widgets (which are with respect to different regions or a cursor) and making a spectral profile at the same time. Otherwise, the enqueued moment tasks will keep going and can not be canceled if you frequently click the generate and cancel buttons in a short time. Like the following plot:
截圖 2020-09-07 下午3 50 20

duidae:
closes #1196


markccchiang (approved):

kswang1029 (approved):
this works well. thanks for fixing this quickly. @veggiesaurus backport this too?

veggiesaurus (approved):
This looks good. For 1.5 I suggest we show the loading indicator in the top right when streaming spectral data

PR #1211, opened 2020-09-16
update catalog UI

panchyo0:
1. remove "represent as", replaced with dropdown, make dropdown and buttons as two different rows. 
2. add column type to catalog header table.
3. add settings panel (global and image overlay) for catalog widget. 
4. move image overlay color, shape and size to image overlay panel. add highlight color to overlay settings 
5. table separator position as implicit preference.
6. add default displayed columns to catalog settings and preferences dialog.
7. add filter to each column name dropdown list (catalog widget and catalog plot widget) with fuzzy search.
8. show tooltip when user hover on column name in catalog data table.
9. support catalog settings for layouts 
10. fixed issue #889


kswang1029:
@panchyo0 this new UI feels nice and cleaner! I will have more tests later on.

Just wonder where we should relocate the cosmetic options for the image overlay (color, shape, size)? It is a bit inconsistent to have those at the top of the catalog widget. One potential solution is we have a setting dialogue and put those there. We probably will need a setting dialogue for some preferences (e.g. default number of displayed columns, etc) I guess.

In addition, it might be good to have a responsive layout for the dropdowns and buttons at the bottom of the widget. When it is narrow, we can have two rows. 

 @veggiesaurus @low-sky @panchyo0 ideas?

panchyo0:
setting dialogue sounds good. Will move overlay settings in there.

panchyo0:
@kswang1029 please review again. Thanks.

kswang1029:
I suppose we still need the description column for the header table? It will help users to judge if a column should be displayed or not. At the moment users need to blindly enable a column and hover over to see the description.

kswang1029:
@panchyo0 when we load two catalog files and trigger image overlay rendering from each, the file dropdown menus in the settings and in the widget are not sync'd. Thus we cannot simply use the file dropdown in the settings to adjust the colors of each image overlay independently. Could you have a look?

veggiesaurus:
@kswang1029 please check this again

kswang1029:
Tested again and it works well as far as I can see! After fixing the following minor issue, we are ready to merge. Great work @panchyo0 !

We need to trim the displayed file name for the file dropdown menu when the menu is collapsed. At the moment, it wraps. 
Screen Shot 2021-01-08 at 10 07 15 AM

panchyo0:
@kswang1029 fixed the dropdown menu issue, please check. thanks

kswang1029:
@panchyo0 yes the issue is gone~ 👍 Ready for merging.

kswang1029 (approved):

PR #1216, opened 2020-09-23
Jason/issue 883

### LINKED ISSUE #883: incorrect formatting on rms in  spatial profile windows
The formatting of the rms in the mean/rms display in profile windows is incorrect. 

I have set display mean and rms option in the y profile window in this screen capture.  The values are reported as -1.30e-03/0.01.  

The rms values should also be reported in 'e' format in the same way as the mean and the data value.  As currently displayed the number is not precise enough to be useful. This seems to be an issue only with the spatial profiles.  The z-profile has it right.

rmsvalue

duidae:
closes #883 

1. change mean/rms to scientific notation
2. change profiler info to single line for space utilization

image

### DIFF
--- a/src/components/Shared/ProfilerInfo/ProfilerInfoComponent.tsx
+++ b/src/components/Shared/ProfilerInfo/ProfilerInfoComponent.tsx
@@ -7,11 +7,9 @@ interface ProfilerInfoComponentProps {
 
 export const ProfilerInfoComponent: React.FC<ProfilerInfoComponentProps> = (props) => {
     let infoString = "";
-    if (props.info?.length > 0) {
-        props.info.forEach((info, index) => {
-            infoString = index === 0 ? info : (infoString + ", " + info);
-        });
-    }
+    props.info?.forEach((info, index) => {
+        infoString = index === 0 ? info : (infoString + ", " + info);
+    });
 
     return (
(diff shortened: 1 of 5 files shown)


veggiesaurus (approved):

kswang1029 (approved):
Tested and this works fine. : )

src/components/Shared/ProfilerInfo/ProfilerInfoComponent.tsx
```diff
@@ -6,18 +6,16 @@ interface ProfilerInfoComponentProps {
 }
 
 export const ProfilerInfoComponent: React.FC<ProfilerInfoComponentProps> = (props) => {
-    let infoTags = null;
-    if (props.info && props.info.length > 0) {
-        infoTags = props.info.map((infoString, index) => <tr key={index}><td><pre>{infoString}</pre></td></tr>);
+    let infoString = "";
+    if (props.info?.length > 0) {
+        props.info.forEach((info, index) => {
```
veggiesaurus: can't we just use `props.info?.forEach`?

duidae: Refined.

veggiesaurus (approved):

PR #1217, opened 2020-09-28
Jason/issue 1014

### LINKED ISSUE #1014: Hide pin in image view
The image view has a pin with tooltip 'detach', but it seems non-functional. (Webgl thing - needs to be examined).

duidae:
closes #1014


kswang1029 (approved):
this works well.

PR #1221, opened 2020-10-05
Jason/issue 435

### LINKED ISSUE #435: cursor info bar: cursor value when the cursor is outside the image

duidae:
closes #435


kswang1029 (approved):
this works well with matched images with different sizes.

PR #1226, opened 2020-10-14
Integrate active tab

### LINKED ISSUE #1083: active tab in saved layout
When we save a layout like this:
Screen Shot 2020-08-04 at 9 42 51 AM
(e.g. The animator co-located with the x spatial profile with the _animator tab_ active)
and load it back, we see 
Screen Shot 2020-08-04 at 9 43 03 AM
(e.g. The animator co-located with the x spatial profile with the _x spatial profile tab_ active)

The active tab should be saved too.

duidae:
closes #1083


src/models/layout_schema_2.json
```diff
@@ -432,6 +432,9 @@
                 },
                 "height": {
                     "$ref": "#/definitions/percentage-value"
+                },
+                "activeItemIndex": {
+                    "type": "integer"
```
veggiesaurus: shall we add a min value of zero here?

src/models/LayoutConfig.ts
```diff
@@ -277,6 +280,9 @@ export class LayoutConfig {
                         type: child.type,
                         content: []
                     };
+                    if (child.type === "stack" && child.activeItemIndex && child.activeItemIndex >= 0) { // load active tab
```
veggiesaurus: shouldn't we also check whether the index is less than the number of children?

kswang1029 (approved):
This works well. 👍

PR #1227, opened 2020-10-14
showing selected statistic string in y-axis label of spectral profiler

### LINKED ISSUE #256: spectral profiler change y-axis label to reflect statistic
When you change the statistic being displayed the label of the y axis on the display should change to match. Instead of generic “value” replace with “mean” or “RMS” or whatever stat is being display. “Value” is the appropriate label only for “cursor”.

TienHao:
closes #256


kswang1029 (approved):
Tested and it looks good. 

UPDATE: found a bug please refer to the following comment.

kswang1029 (changes requested):
When we have a point region and a rectangle region with FluxDensity as statistic, switching between these two region results in the wrong unit in the point region profile (should be Jy/beam but it is Jy).

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -380,13 +380,25 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
...
+                let yLabelName = "Value";
+                let yLabelUnit = "";
+                let region: RegionStore;
+                if (frame.regionSet) {
+                    region = frame.regionSet.regions.find(r => r.regionId === this.widgetStore.effectiveRegionId);
+                    if (region.regionType !== CARTA.RegionType.POINT) {
+                        yLabelName = SpectralProfileWidgetStore.StatsTypeString(this.widgetStore.statsType);
+                    }
+                }
+                if (region && region.regionType === CARTA.RegionType.POINT) {
```
veggiesaurus: couldn't this be more elegantly combined with the `if` statement above?

veggiesaurus:
@kswang1029 are you happy with these changes?

kswang1029:
@veggiesaurus @TienHao tested 6bb934b and it works well. 👍

PR #1235, opened 2020-10-26
report verbose session info in hidden div

### LINKED ISSUE #1228: Scripting: machine-parseable host, gRPC port and session ID
The scripting wrapper now has support for starting a new session in a selenium-controlled browser (headless or not). Selenium has to parse the backend host and the session ID from the existing page DOM. The easiest way to do this is to parse the contents of the log widget, but even this comes with various complications and race conditions, and is a fragile solution.

(Parsing the console log was considered but abandoned, because different browser drivers handle this in non-standard ways.)

It would be ideal if this data were available in an easily retrievable page element which is rendered early (it doesn't have to be visible in the GUI) -- for example, a series of invisible spans with keys as IDs and values as text content.

In addition to this I would like to request that the gRPC port be logged by the frontend in the same way (which may require a change in the backend so that it is reported to the frontend in the first place) so that the wrapper user does not have to provide it as an additional parameter.

veggiesaurus:
Closes #1228.Companion PR of https://github.com/CARTAvis/carta-backend/pull/646

This adds an invisible div with element ID `hidden-status-info`, which can be used to get the session ID, GRPC port or server URL. For example, to get the session ID:

`document.getElementById("info-server-url").innerHTML`

### DIFF
--- a/src/components/Menu/RootMenuComponent.scss
+++ b/src/components/Menu/RootMenuComponent.scss
@@ -47,4 +47,8 @@
             }
         }
     }
+
+  #hidden-status-info {
+    display: none;
+  }
 }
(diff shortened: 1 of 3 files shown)


confluence (approved):
Looks good to me!

Bikeshedding: the `-span` in the span IDs is redundant and also unnecessarily referring to an implementation detail. I'd use `property-` or `info-` or `api-` or some other prefix or suffix referring to the semantic meaning of these values.

duidae (approved):

panchyo0 (approved):

PR #1245, opened 2020-10-27
Angus/match image scaling

### LINKED ISSUE #1089: Feature request - match color scale between images
One feature request I have for CARTA that would save me a lot of time is a button to match the scale limits between frames/images. I’m often comparing images of the same field (e.g. self-calibration loops) and wanting to see them with the same transfer function. So I have to copy and paste between the min and max boxes for all frames, which takes an annoying amount of time. It would be useful to have a single button to do this. Perhaps within the match column in the image list.

veggiesaurus:
closes #1089 

Adds the ability to match render config properties (clip min, clip max, scaling type, gamma, alpha) by enabling "render config matching" in the image list widget

@Jordatious please use [Mix 'n Match](http://carta.asiaa.sinica.edu.tw/builds/) for testing (branch `angus/match_image_scaling`)

### DIFF
--- a/src/components/LayerList/LayerListComponent.scss
+++ b/src/components/LayerList/LayerListComponent.scss
@@ -14,6 +14,10 @@
         cursor: pointer;
     }
 
+    .bp3-table-column-name {
+        font-size: 12px;
+    }
+
     .bp3-button {
         font-size: 11px;
         &.bp3-small {
--- a/src/components/LayerList/LayerListComponent.tsx
+++ b/src/components/LayerList/LayerListComponent.tsx
@@ -179,22 +179,22 @@ export class LayerListComponent extends React.Component<WidgetProps> {
         }
 
         let renderConfigMatchingButton: React.ReactNode;
-        if (appStore.renderConfigReference) {
+        if (appStore.rasterScalingReference) {
             let tooltipSubtitle: string;
-            if (frame === appStore.renderConfigReference) {
-                tooltipSubtitle = `${frame.frameInfo.fileInfo.name} is the current render config reference`;
+            if (frame === appStore.rasterScalingReference) {
+                tooltipSubtitle = `${frame.frameInfo.fileInfo.name} is the current raster scaling reference`;
             } else {
-                tooltipSubtitle = `Click to ${frame.renderConfigReference ? "disable" : "enable"} matching to ${appStore.renderConfigReference.frameInfo.fileInfo.name}`;
+                tooltipSubtitle = `Click to ${frame.rasterScalingReference ? "disable" : "enable"} matching to ${appStore.rasterScalingReference.frameInfo.fileInfo.name}`;
             }
             renderConfigMatchingButton = (
-                <Tooltip position={"bottom"} content={<span>Render config matching<br/><i><small>{tooltipSubtitle}</small></i></span>}>
+                <Tooltip position={"bottom"} content={<span>Raster scaling matching<br/><i><small>{tooltipSubtitle}</small></i></span>}>
                     <AnchorButton
-                        className={frame === appStore.renderConfigReference ? "outlined" : ""}
+                        className={frame === appStore.rasterScalingReference ? "outlined" : ""}
                         minimal={true}
                         small={true}
-                        active={!!frame.renderConfigReference}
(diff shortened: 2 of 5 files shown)


duidae (approved):

Jordatious:
Looks great! So it matches the limits and scaling? Not sure if "render config matching" makes it sound like the whole render config is matched (e.g. also colormap and inversion). 

Can we also have a global preference to match upon append? 

If I match two of the same image with a percentile button selected on the reference image, it doesn't seem to know I'm using a percentile on the matched image, but it has "custom" selected.

If I deselect the render config matching, do you think it makes more sense to keep it at the value it was when matched, or what it was before matching?

veggiesaurus:
> Looks great! So it matches the limits and scaling? Not sure if "render config matching" makes it sound like the whole render config is matched (e.g. also colormap and inversion). 

Good point. I'm not sure what to call it. Raster scaling?

> 
> Can we also have a global preference to match upon append? 

I think that would be a useful addition, but I might handle it in a separate PR, because it will involve an updated preferences schema

> If I match two of the same image with a percentile button selected on the reference image, it doesn't seem to know I'm using a percentile on the matched image, but it has "custom" selected.

That's because you can't match to multiple image's percentiles, since their histograms are different (unless you're loading two identical images)

> If I deselect the render config matching, do you think it makes more sense to keep it at the value it was when matched, or what it was before matching?

I think it makes sense to not adjust any levels when unmatching.

Jordatious:
Ok cool, thanks. Not sure what to call it. Maybe 'match scale'? It's not really a problem anyway. Yes, when loading two identical images, it does this.

Jordatious (approved):

veggiesaurus:
adjusted from "render config" matching to "raster scaling" matching. @kswang1029 please test and let me know if it's ok

kswang1029:
@veggiesaurus tested with match and unmatch actions with multiple matched/unmatched images. It works nicely! 👍

keflavich:
I was about to request this as a feature; can I access this in any available version, or do I need to wait for the next release?

keflavich:
...ok, I already requested it in https://github.com/CARTAvis/carta/issues/42.  Is there a way to access dev builds?  do you distribute them?

veggiesaurus:
> ...ok, I already requested it in https://github.com/CARTAvis/carta/issues/42.  Is there a way to access dev builds?  do you distribute them?  

We don't currently build dev desktop builds, our CI builds frontend and backend separately. Although @ajm-asiaa may want to put together a dev desktop build soon to test out Big Sur compatibility. Perhaps we could pass that on to you?

ajm-ska:
> > ...ok, I already requested it in [CARTAvis/carta#42](https://github.com/CARTAvis/carta/issues/42).  Is there a way to access dev builds?  do you distribute them?
> 
> We don't currently build dev desktop builds, our CI builds frontend and backend separately. Although @ajm-asiaa may want to put together a dev desktop build soon to test out Big Sur compatibility. Perhaps we could pass that on to you?

OK

ajm-ska:
Here is a [MacOS CARTA v2.0.0-dev.2](https://drive.google.com/file/d/1C-iff_FS-8aNrkiuInlrO4N1sk-ICxqm/view?usp=sharing) Desktop version.
It is using carta-frontend dev commit 494460a8, carta-backend dev commit 043b972, and Electron 11.0.0-beta.22.
(I haven't tried it on Intel based Big Sur yet, but it can run on Apple Silicon Big Sur under Rosetta 2)

veggiesaurus:
> Here is a [MacOS CARTA v2.0.0-dev.2](https://drive.google.com/file/d/1C-iff_FS-8aNrkiuInlrO4N1sk-ICxqm/view?usp=sharing) Desktop version.
> It is using carta-frontend dev commit [494460a](https://github.com/CARTAvis/carta-frontend/commit/494460a8cb99cbe9d0af0d81f5c5bf749f087c08), carta-backend dev commit 043b972, and Electron 11.0.0-beta.22.
> (I haven't tried it on Intel based Big Sur yet, but it can run on Apple Silicon Big Sur under Rosetta 2)

thanks @ajm-asiaa 

Seem to work fine on my Intel MBP with Big Sur

Issue #1046, opened 2020-07-21
display Stokes value instead of numeric index in animator


Kechil:
When loading a Stokes cube, it is better to display I, Q, U, or V instead of 0, 1, 2, or 3 which are meaningless and misleading due to the classic 0-base or 1-base indexing issue.

The FITS standard document defines how numeric values in the header map to Stokes parameters and other polarization parameters. Table 29:
![Screen Shot 2021-06-14 at 3.04.14 PM.png](https://images.zenhubusercontent.com/590bdf9f51aa2c30577d5260/be903618-0859-48c7-b579-2d39a909bc74)

We will need to show the string in the "Symbol" column based on the "Value" column.

kswang1029:
At the moment, if we load a cube with Stokes QU (the header is set correctly), the Stokes column in the image list widget still starts with 0. Numeric index should be 1 but we better just show alphabetic index as Q.

Kechil:
Medium Priority

PR #1248, opened 2020-11-02
Jason/issue 265 2

### LINKED ISSUE #265: animator usability enhancement

duidae:
closes #265

### DIFF
--- a/src/components/Animator/AnimatorComponent.scss
+++ b/src/components/Animator/AnimatorComponent.scss
@@ -29,9 +29,10 @@
             .slider-info {
                 overflow-x: hidden;
                 flex: 1 1 180px;
-            }
-
-            .channel-slider {
+                pre {
+                    margin: 0px;
+                    font-size: x-small;
+                }
             }
 
             &.range-slider {
--- a/src/components/Animator/AnimatorComponent.scss
+++ b/src/components/Animator/AnimatorComponent.scss
@@ -76,19 +76,17 @@
                 margin-top: 10px;
             }
         }
+
+        .bp3-numeric-input {
+            .bp3-input-group {
+                width: 40px;
+            }
+        }
     }
 
     .playback-framerate {
         min-width: 150px;
         flex: 1 1 150px;
-    }
-
-    #framerate-numeric {
-        width: 40px;
-    }
-
-    #step-numeric {
-        width: 40px;
     }
 
     .playback-mode {
--- a/src/components/Animator/AnimatorComponent.scss
+++ b/src/components/Animator/AnimatorComponent.scss
@@ -59,49 +59,48 @@
         margin-right: 10px;
         display: flex;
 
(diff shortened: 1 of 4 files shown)


kswang1029:
@duidae is it possible to show the reference frame, frequency and velocity with one per line (ie three lines)? It may look nicer and be easier to read. 

Update: suggesting the order as velocity, frequency and reference frame

kswang1029:
@duidae I will think about how to determine a proper upper limit for the step spinbox. Apparently 5 is not sufficient. We might need to use the range slider to dynamically determine a proper upper limit. For example, if the width of the range slider is 101 channel, we may use 50 (ie width//2) as the upper limit.

kswang1029:
@duidae regarding the upper limit for step, let’s just try 50 for now. I guess this should work mostly and is an easy solution

duidae:
i think the upper limit can be dynamic, since it should be easy to have the frame number

kswang1029 (approved):
Tested with different steps with different playback modes. It works well. 👍 

Just a note here that when a step is defined, the playback sequence is defined based on the range slider. This ensures that from the 2nd loop onward, uses see _same_ set of channels in the animation. However, the 1st loop may have a different set of channels depending on the starting channel index. This is fine I think. 

The rest minor improvements are:
1) set the upper limit of step spinbox to 50.
2) display spectral info text in three lines with the following order: Velocity, Frequency, and Reference frame.

duidae:
@kswang1029  fixed & please have a look, thanks.

kswang1029:
Screen Shot 2020-11-03 at 4 04 01 PM
I suggest to have consistent font type for image/channel/Stokes. In addition, the spectral info should not be displayed for Stokes slider.

duidae:
@kswang1029 Refined, please have a look, thanks.

kswang1029:
@duidae f9ce122 looks good. The Stokes string display will be dealt with https://github.com/CARTAvis/carta-frontend/issues/1046

veggiesaurus (changes requested):
this looks great. One very minor issue: this seems to break the responsive layouts, because the dropdown is slightly larger. 

image

kswang1029:
@veggiesaurus @duidae to save some space, how about changing "frame rate" to "rate"? We could add tooltip if it is necessary. "Rate" should be already clear enough. 

ps I tried to find two icons from blueprint to represent frame rate and step for the responsive layout but I failed. -_-

duidae:
the responsive issue is fixed, I think we dont need extra manipulations.

veggiesaurus (approved):

PR #1249, opened 2020-11-05
Thchang/unifying color selectors

### LINKED ISSUE #341: unify color selection widget
so far we have 2 kinds of color selectors, better to unify them to the advanced one

color

TienHao:
closed #341.
Replacing AST color selectors with new color picker components, by turning AST palette(`Module.colors`) into a color set of AST plotting elements(`OverlayXXXSetting`: global, title, grid, border, ticks, axis, number, label).


kswang1029 (commented):
this looks mostly good. Two minor issues:
1) could you add the default "blue" as the preset colors at the bottom of the color selector? Possibly modify the first preset color to match the default "blue".
2) I see a slightly different color selector for NaN color which has an additional transparency control. Shell we make it more consistent elsewhere?
Screen Shot 2020-11-06 at 10 24 27 AM
Screen Shot 2020-11-06 at 10 24 45 AM

veggiesaurus:
> 2. I see a slightly different color selector for NaN color which has an additional transparency control. Shell we make it more consistent elsewhere?

Do any other colours actually _need_ a transparent selector in the swatch?

kswang1029:
> > 1. I see a slightly different color selector for NaN color which has an additional transparency control. Shell we make it more consistent elsewhere?
> 
> Do any other colours actually _need_ a transparent selector in the swatch?

I cannot think of any, even for NaN. Maybe we can remove it to keep it simple?

kswang1029:
Oh wait, transparency is needed if we want to support transparent NaN rendering. It might be still useful to have it in some cases. So I suggest we just accept the current implementation with NaN with an additional control for transparency. Others then don’t have transparency control.

src/stores/AppStore.ts
```diff
@@ -971,9 +968,19 @@ export class AppStore {
         this.activeLayer = ImageViewLayer.RegionMoving;
 
         AST.onReady.then(runInAction(() => {
-            AST.setPalette(this.darkTheme ? nightPalette : dayPalette);
             this.astReady = true;
             this.logStore.addInfo("AST library loaded", ["ast"]);
+            const colors = [
```
veggiesaurus: this is a nice way of solving this issue!

veggiesaurus:
@kswang1029 please re-test

kswang1029:
The default color in "image view settings" -> "Global" -> "Color" should be (R, G, B) = (18, 86, 146) instead of (20, 104, 175) which is lighter and not consistent with the entire UI. Other than this, we are ready to go. @TienHao Could you help to check the default color and make the change, please?

kswang1029:
@TienHao could you check the default color then we can merge the PR and close this issue?

TienHao:
@kswang1029  I have changed the default color in "image view settings" -> "Global" -> "Color". Please take a look.

kswang1029:
@TienHao Please set the default color to the following based on the color themes.

> Light theme
> R, G, B = 16, 107, 163
> 
> Dark theme
> R, G, B = 43, 149, 214

The main issue here is we need the same color in the following boxed color panel so that users can have a way to restore the default color.
Screen Shot 2021-02-17 at 6 30 11 PM

TienHao:
the auto color picker component has been added, which uses color string: `<auto-XXX>` to display color depending on current theme. Colors set [here](https://blueprintjs.com/docs/#core/colors) (eg. `auto-blue` is `blue2` when using light theme, and `blue4` when using dark theme) 

https://user-images.githubusercontent.com/20379662/109773061-bbc28680-7c39-11eb-81f6-66ab7849589b.mov


auto color pickers are applied in 
- Image Settings(AST, Beam)
- Histogram
- Catalog
- Spectral Profiler
- Spectral Profiler
- Stokes Analysis
- Profile Smoothing

original color pickers are remained in
- Region color
- NaN color
- Contour

src/stores/PreferenceStore.ts
```diff
@@ -107,12 +107,12 @@ const DEFAULTS = {
         contourColormap: "viridis",
     },
     WCS_OVERLAY: {
-        astColor: 4,
+        astColor: "auto-blue",
```
veggiesaurus, line 112: I think we'll need to create a new preference schema (version 2) and will need to also add an upgrade path for going from v1 (astColor: number) -> v2 (astColor: string)

veggiesaurus: It _might_ be easier to rename this setting to overlayColor in the new schema, as this would prevent conflicts

veggiesaurus (changes requested):

kswang1029:
Screen Shot 2021-03-05 at 2 15 27 PM
can the dropdown element be narrower to be consistent with others? Otherwise in the above case we need a wider default window size.

kswang1029:
minor CSS adjustment: apply padding just below "Other:" Suggest the size to be the same as the horizontal one. 
Screen Shot 2021-03-05 at 2 17 39 PM

kswang1029:
> original color pickers are remained in
> 
> * Region color
> * NaN color
> * Contour

@TienHao Just curious, any reason why these three keep the same as before?

veggiesaurus:
> Screen Shot 2021-03-05 at 2 15 27 PM
> can the dropdown element be narrower to be consistent with others? Otherwise in the above case we need a wider default window size. 

Why do we need a dropdown here instead of just using a colour picker? Isn't that what #541 is describing?

veggiesaurus:
> > original color pickers are remained in
> > 
> > * Region color
> > * NaN color
> > * Contour
> 
> @TienHao Just curious, any reason why these three keep the same as before?

surely the region and contour colours are independent of the theme, since they're on a colourmapped backdrop?

kswang1029:
> surely the region and contour colours are independent of the theme, since they're on a colourmapped backdrop?

Right. Then it is fine to have those three as before. 👍

veggiesaurus:
> > surely the region and contour colours are independent of the theme, since they're on a colourmapped backdrop?
> 
> Right. Then it is fine to have those three as before.

@TienHao can you revert the change you made to those three?

TienHao:
> Screen Shot 2021-03-05 at 2 15 27 PM

> 
> Why do we need a dropdown here instead of just using a colour picker? Isn't that what #341 is describing?

@veggiesaurus 
The dropdown is for setting auto color string `auto-<xxx>`, because the color picker components of react-color can only accept hex, rgb, hsl, and "transparent".
Therefore, I combined the original two color pickers into a new one to meet the need of theme-aware colors.

veggiesaurus:
> @veggiesaurus
> The dropdown is for setting auto color string `auto-<xxx>`, because the color picker components of react-color can only accept hex, rgb, hsl, and "transparent".
> Therefore, I combined the original two color pickers into a new one to meet the need of theme-aware colors.

I see the complication. Hmm. Well, for the auto-picker, surely we con't need the swatch as well?
Also, could we not just have the text "Other", rather than an additional button? Having a button inside of a select seems a little clunky, and not needed, since changing the value in the colour picker updates the value displayed in the select anyway. Alternatively, the picker popup should pop up from the location you click, not the top of the select

https://user-images.githubusercontent.com/592504/110614042-9b934a00-819a-11eb-8e0d-3741c9fff178.mp4

TienHao:
> I see the complication. Hmm. Well, for the auto-picker, surely we con't need the swatch as well?

The color pickers in the _CatalogOverlaySettingPanel_ and _LinePlotSettingsPanel_ have "transparent" in the presetColors. Should we keep the swatch in the auto-color picker?

kswang1029:
> > I see the complication. Hmm. Well, for the auto-picker, surely we con't need the swatch as well?
> 
> The color pickers in the _CatalogOverlaySettingPanel_ and _LinePlotSettingsPanel_ have "transparent" in the presetColors. Should we keep the swatch in the auto-color picker?

With those if we select transparent "color" what we actually see is black. This is same in v1.4. Wonder if this is due to Plotly scattergl? In any case, I think this PR is ready to go.

kswang1029 (approved):

veggiesaurus (approved):

PR #1274, opened 2020-12-09
Jason/issue 633

### LINKED ISSUE #633: Remove FITS hdus from file list and add to file info
create corresponding frontend issue for https://github.com/CARTAvis/carta-backend/issues/357

duidae:
closes #633 , needs backend support of https://github.com/CARTAvis/carta-backend/issues/357

### DIFF
--- a/src/components/Animator/AnimatorComponent.scss
+++ b/src/components/Animator/AnimatorComponent.scss
@@ -29,10 +29,8 @@
             .slider-info {
                 overflow-x: hidden;
                 flex: 1 1 180px;
-                pre {
-                    margin: 0px;
-                    font-size: x-small;
-                }
+                font-size: x-small;
+                word-wrap: break-word;
             }
 
             &.range-slider {
--- a/src/components/Animator/AnimatorComponent.tsx
+++ b/src/components/Animator/AnimatorComponent.tsx
@@ -261,7 +261,7 @@ export class AnimatorComponent extends React.Component<WidgetProps> {
                             disabled={appStore.animatorStore.animationState === AnimationState.PLAYING}
                         />
                         <div className="slider-info">
-                            <pre>{activeFrame.filename}</pre>
+                            {activeFrame.filename}
                         </div>
                     </React.Fragment>
                     }
--- a/src/components/Animator/AnimatorComponent.tsx
+++ b/src/components/Animator/AnimatorComponent.tsx
@@ -261,7 +261,7 @@ export class AnimatorComponent extends React.Component<WidgetProps> {
                             disabled={appStore.animatorStore.animationState === AnimationState.PLAYING}
                         />
                         <div className="slider-info">
-                            <pre>{activeFrame.frameInfo.fileInfo.name}</pre>
+                            <pre>{activeFrame.filename}</pre>
                         </div>
                     </React.Fragment>
                     }
--- a/src/components/Dialogs/ContourDialog/ContourDialogComponent.tsx
+++ b/src/components/Dialogs/ContourDialog/ContourDialogComponent.tsx
@@ -149,7 +149,7 @@ export class ContourDialogComponent extends React.Component {
         if (!frame) {
             return null;
         }
-        return <MenuItem text={frame.frameInfo.fileInfo.name} onClick={handleClick} key={frame.frameInfo.fileId}/>;
+        return <MenuItem text={frame.filename} onClick={handleClick} key={frame.frameInfo.fileId}/>;
     };
 
     private renderHistogramSelectItem = (isCube: boolean, {handleClick, modifiers, query}) => {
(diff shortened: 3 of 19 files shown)


kswang1029:
@pford The frontend part is ready. Could you issue PR for the backend part?

kswang1029 (commented):
Shall we apply the same thing to the "file header dialogue"? This would be useful to know which HDU is loaded and displayed. In fact, I wonder if we should append the HDU name to the image list widget as well as the title of the image viewer tab? At the moment there is no way to tell which HDU is loaded/displayed. @veggiesaurus @duidae thoughts?

veggiesaurus:
> Shall we apply the same thing to the "file header dialogue"? This would be useful to know which HDU is loaded and displayed. In fact, I wonder if we should append the HDU name to the image list widget as well as the title of the image viewer tab? At the moment there is no way to tell which HDU is loaded/displayed. @veggiesaurus @duidae thoughts?

Good point! Although I suggest we only show HDU name if there are multiple HDUs. Otherwise it clutters things unnecessarily

kswang1029:
This is the embarrassing issue about the loooong HDU name. In our demo server we do have such example images as seen in the screenshot. Is there a way to define a maximum width of the dropdown? The text can be truncated when the dropdown is collapsed. When expanded, we show the full name.

Screen Shot 2020-12-10 at 13 46 46

duidae:
> Shall we apply the same thing to the "file header dialogue"? This would be useful to know which HDU is loaded and displayed. In fact, I wonder if we should append the HDU name to the image list widget as well as the title of the image viewer tab? At the moment there is no way to tell which HDU is loaded/displayed. @veggiesaurus @duidae thoughts?

@kswang1029  is it possible to have some mock-ups for implementation? thanks

kswang1029:
Is it possible to modify the file name with some additional strings _after_ it is loaded? Using the following screenshot as an example, the HDU 1 "image" is loaded. So the file name changes from `spire500_ext.fits` to `spire500_ext.fits.HDU_1_image`.

Screen Shot 2020-12-11 at 2 54 32 PM

kswang1029:
I just realized the dev has the same issue so the proposed approach is a _new_ improvement (aka off the scope of this hdu improvement) 🙈 but we do need this for better UX.

duidae:
> Is it possible to modify the file name with some additional strings _after_ it is loaded? Using the following screenshot as an example, the HDU 1 "image" is loaded. So the file name changes from `spire500_ext.fits` to `spire500_ext.fits.HDU_1_image`.
> 
> Screen Shot 2020-12-11 at 2 54 32 PM

@kswang1029  @pford so far frontend can add hdu to those required fields, but extension name is missing in the computed entries of open file ack, may also need backend to return extension name in open file ack.

pford:
The computed entries in fileInfoExtended for the file browser and for open file ack now begin with:
* Name
* HDU ("0" -> "Primary")
* Extension name

In general, if the HDU or EXTNAME is empty, it is not added to computed entries.

Let me know if "Extension name" needs to be first as before, but this order makes more sense to me and is the same order as KSW's suggested "spire500_ext.fits.HDU_1_image".

duidae:
@pford  thanks for the great help. Now the full name is showed in image viewer tab/image list/image list widget(since Pam added extname field to file info widget, so no changes here), @kswang1029 please have a look, thanks.
image
image

pford:
@duidae for images which have HDU 0 (jcmts20091208_00085_450.fits, any HDF5 image), the frontend is using the "Shape" for the extension name since there is no extension name. I also think I should change the backend to send HDU="0" since "Primary" is a convention for FITS but not HDF5, which I had not considered for this feature.

duidae:
> @duidae for images which have HDU 0 (jcmts20091208_00085_450.fits, any HDF5 image), the frontend is using the "Shape" for the extension name since there is no extension name. I also think I should change the backend to send HDU="0" since "Primary" is a convention for FITS but not HDF5, which I had not considered for this feature.

@pford thanks for spotting that, @kswang1029 could you help elaborate what we should do with HDU="0"?

kswang1029:
If a FITS image has only one HDU ("0"), I suggest we don't modify the displayed file name. If there are more than one HDUs, we modify the displayed file name. If there are multiple HDUs and if an HDU has no extension name, we leave it blank (e.g. xxx.fits_HDU_0)

This is far more complicated than I expected 🙈 as there are a number of places where file name is used including:
1. the image slider of the animator
2. data source dropdown in the contour config dialogue
3. data source dropdown in the moments tab of the spectral profiler settings
4. All profile tsv export. The first line includes the file name.
5. image list widget
6. the title of the image viewer
7. the title of the region config dialogue

kswang1029:
@pford @duidae would it be simpler if in OPEN_FILE_ACK we apply the file name modification? Then the frontend can display it as it is.

duidae:
> If a FITS image has only one HDU ("0"), I suggest we don't modify the displayed file name. If there are more than one HDUs, we modify the displayed file name. If there are multiple HDUs and if an HDU has no extension name, we leave it blank (e.g. xxx.fits_HDU_0)

@kswang1029  About this, I suggest we unify the name to "xxx.fits" for HDU="0" no matter this file has multiple hdus or not, because after opening this frame, we have no way to know whether it has other hdu siblings or not, the whole hdu list is not preserved in open file ack of the opened frame.

> 
> This is far more complicated than I expected  as there are a number of places where file name is used including:
> 
> 1. the image slider of the animator
> 2. data source dropdown in the contour config dialogue
> 3. data source dropdown in the moments tab of the spectral profiler settings
> 4. All profile tsv export. The first line includes the file name.
> 5. image list widget
> 6. the title of the image viewer
> 7. the title of the region config dialogue

kswang1029:
to recap:
with the frontend 9871d66 and the backend pam/357_fits_hdu_info f975aff, if HDU 0 does not have a name, then on the file browser it is not displayed correctly. As long as name exists for a given HDU, it works nicely (cf the image list in the screenshot). 

Screen Shot 2021-01-06 at 3 26 27 PM

duidae:
should be fixed, @kswang1029 please have a look, thanks.

> to recap:
> with the frontend [9871d66](https://github.com/CARTAvis/carta-frontend/commit/9871d66c159be050fcd192c1aed781afb4f4e8e7) and the backend pam/357_fits_hdu_info f975aff, if HDU 0 does not have a name, then on the file browser it is not displayed correctly. As long as name exists for a given HDU, it works nicely (cf the image list in the screenshot).
> 
> Screen Shot 2021-01-06 at 3 26 27 PM

kswang1029 (approved):
Tested with some other FITS images with multiple HDUs. This works nicely! 👍

veggiesaurus:
Note that this still requires a backend change for double-clicking (and now, multiple-selections as well)

Issue #1281, opened 2020-12-22
Select all logged Event Types


acdo2002:
Current situation: If the testers would like to see the Event Types on the console.log, they should select the Event Type items under File->Preference->Log Events. 

Improvement: Many thanks to Jason's suggestion!!!
So far my method is adding a button, it only shows up on the Log Events tab and it can select all the Event Types at once (no matter you already selected or not). 
The original "Restore defaults" can remove all the click. 
Any suggestion is welcome!!!

Video memo is [here](https://drive.google.com/file/d/1VCeWay3gDQEh71R3sloOyTi_7hABqF2m/view)
Hope this can help the testers that can select Event Types in a relatively efficient way.

PR #1282, opened 2020-12-22
add a button to select all Event Types

acdo2002:
Refer to #1281 
I add few lines (carta-frontend/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx) to make a button that can select all the Event Types... any better suggestion is welcome!!!

### DIFF
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -586,6 +586,9 @@ export class PreferenceDialogComponent extends React.Component {
                 </div>
                 <div className="bp3-dialog-footer">
                     <div className="bp3-dialog-footer-actions">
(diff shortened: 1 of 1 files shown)


kswang1029:
@acdo2002 we should grey out the "select all" button when _all_ messages are selected (either via the select all button or manually).

duidae:
this implementation is a little bit strange comparing to conventional "select all", I'll suggest we add the "select all" checkbox like gmail next to the event type header

s7h99-ybdkv

kswang1029:
@acdo2002 What if we make "Enable logged event type" always visible as a header regardless the scrolling state of the list below, then we can add a checkbox (or a toggle) next to it with the same effect as @duidae suggested?

Screen Shot 2020-12-23 at 3 16 04 PM

acdo2002:
I am ok with this. My initial motivation is very simple and naive, just would like to have a button that can select all :)

duidae:
send a PR #1284 for select all, so close this.

Issue #1259, opened 2019-04-04
Multi-panel view with image viewer


Kechil:
science requirements: https://docs.google.com/document/d/1W228FbvwE9J9oTc9_eKOuiYnHiGfeKJugDJQAVf2GHg/edit#

Issue #1203, opened 2020-09-11
channel map view (frontend)


kswang1029:
as one of the NRAO priorities.

This would allow users to view multiple channels of an image cube in the image viewer. @jott3077 please help to add more descriptions. 

Associated with https://github.com/CARTAvis/carta-backend/issues/1186

Requirement doc: https://docs.google.com/document/d/1bdjwuGSatu1ZhmW2YbxKvVGea1i20B6W0luzDZotdao/edit#heading=h.gjdxovqft8g1

jott3077:
We need a proper google doc for this. It should be part of the larger multi-display effort.

kswang1029:
requirement doc is attached: https://docs.google.com/document/d/1bdjwuGSatu1ZhmW2YbxKvVGea1i20B6W0luzDZotdao/edit#heading=h.gjdxovqft8g1

PR #1288, opened 2021-01-06
Angus/tile cache changes

veggiesaurus:
This PR makes a major change to the GPU tile cache (L1 in our caching scheme, with L2 being compressed system memory). Instead of using a single GPU cache for the active image only, it instead uses a shared GPU cache for all open images. this allows users to switch between active images without encountering a tile cache miss. 

This is a pre-requisite for #1259 and #1203, since this will require viewing _multiple_ images at once.


veggiesaurus:
@kswang1029 I've opened this _draft_ PR so that we can keep track of what works and what doesn't work with this branch. Can you do some testing? I'm pretty sure there are still loads of issues, but I've forgotten what they are :roll_eyes:

kswang1029:
The play button of 592c33e does not work (playback is not triggered). The 9075ed1 works. If no zoom/pan action is applied while animating the "image" axis, it seems to work nicely and responsive. However, if there is zoom/pan action while animating, the tile cache is messed up. See the following video for demos.


https://user-images.githubusercontent.com/20819712/104985416-db1ba080-5a4b-11eb-9672-44dab4192d91.mp4

kswang1029:
just as a note which may not be related to this implementation directly: with a single cube with multiple channels, while stopping the channel animation playback, occasionally we can see missing tiles (https://github.com/CARTAvis/carta-frontend).

veggiesaurus:
Some of the missing/invalid tile issues can be reproduced via a script:

https://user-images.githubusercontent.com/592504/123682876-e02b1a00-d84b-11eb-882a-a8cea40ca49d.mp4

This should help me track down what's going on

veggiesaurus:
@kswang1029 can you give this a test and see if it works any better now?

kswang1029:
@veggiesaurus I tested 38bac75 with the dev backend and I found three issues:
1. if we load a cube and animate it, we sometimes see incorrect tiles (transient) as a "cross" centered at the cursor position when zooming.

https://user-images.githubusercontent.com/20819712/124532879-f9cbf300-de43-11eb-8551-b449e9a6e46c.mov

2. If we load multiple single channel images and match them, we see images not updating at certain zoom levels while animating different images.


https://user-images.githubusercontent.com/20819712/124533045-3b5c9e00-de44-11eb-9eec-5d52905a1853.mov

3. while testing 2) I saw the backend log

> [2021-07-06 10:16:55.171] [debug] Draining WebSocket backpressure: client 1 [127.0.0.1]. Remaining buffered amount: 403564 (bytes).
[2021-07-06 10:16:55.476] [debug] Draining WebSocket backpressure: client 1 [127.0.0.1]. Remaining buffered amount: 357853 (bytes).
[2021-07-06 10:16:55.701] [debug] Draining WebSocket backpressure: client 1 [127.0.0.1]. Remaining buffered amount: 100429 (bytes).
[2021-07-06 10:16:55.848] [debug] Draining WebSocket backpressure: client 1 [127.0.0.1]. Remaining buffered amount: 137128 (bytes).
[2021-07-06 10:16:55.981] [debug] Draining WebSocket backpressure: client 1 [127.0.0.1]. Remaining buffered amount: 0 (bytes).
[2021-07-06 10:16:56.064] [debug] Draining WebSocket backpressure: client 1 [127.0.0.1]. Remaining buffered amount: 22054 (bytes).
[2021-07-06 10:16:56.188] [debug] Draining WebSocket backpressure: client 1 [127.0.0.1]. Remaining buffered amount: 0 (bytes).
[2021-07-06 10:16:56.370] [debug] Draining WebSocket backpressure: client 1 [127.0.0.1]. Remaining buffered amount: 0 (bytes).

Not sure if this is related to this PR. What I did is I tried to zoom the images while multiple of them are animating at (requested) 15 fps (so tiles from different images are streaming quickly). I also noticed that the actual fps reaches 15 only when we see 0 bytes buffer. Before that, the fps is small and unstable.

veggiesaurus:
moving back to draft in order to address issues raised by @kswang1029 

Underlying issue is that the frontend is requesting multiple tiles over and over again when animating too quickly.

veggiesaurus:
Issue (1) can be dealt with separately if necessary. Issue (3) seems related to the underlying issue with (2)

src/components/ImageView/CatalogView/CatalogViewGLComponent.tsx
```diff
@@ -37,12 +38,13 @@ export class CatalogViewGLComponent extends React.Component<CatalogViewGLCompone
         // dummy values to trigger React's componentDidUpdate()
         /* eslint-disable @typescript-eslint/no-unused-vars */
         const appStore = AppStore.Instance;
-        const baseFrame = appStore.activeFrame;
+        const baseFrame = this.props.frame;
```
veggiesaurus, line 41: these changes are in preparation for #1259, because we will need to render frames other than the active one.

kswang1029 (approved):
Tested with some usual and unusual use cases. This is a big step forward.

duidae (approved):

PR #1289, opened 2021-01-06
Angus/fix 1250

### LINKED ISSUE #1250: Feature request: Open multiple files via file browser
I can open and append multiple files via the debug menu with something like:

```
openFile("B6/fullcubes_12m/G337.92_B6_spw4_12M_spw4.image")
appendFile("B6/fullcubes_12m/G337.92_B6_spw4_12M_spw4.residual")
appendFile("B6/fullcubes_12m/G337.92_B6_spw4_12M_spw4.model")
```

It would be nice if we could select these with the usual shift-click or control-click interface in the file browser menu.

Comparing image, residual, and model is a common use case when inspecting clean cubes.

veggiesaurus:
closes #1250 

This PR adjusts the file table to allow users to select multiple files to load at once. Currently, only image and region files are supported (because `AppStore::appendCatalog` isn't an `async` function)

edit: catalogs should work as well now

### DIFF
--- a/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
+++ b/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
@@ -8,7 +8,7 @@ import {FileInfoComponent, FileInfoType} from "components/FileInfo/FileInfoCompo
 import {FileListTableComponent} from "./FileListTable/FileListTableComponent";
 import {DraggableDialogComponent} from "components/Dialogs";
 import {TableComponentProps, TableType} from "components/Shared";
-import {AppStore, BrowserMode, CatalogProfileStore, FileBrowserStore, FileFilteringType, HelpType, PreferenceKeys, PreferenceStore} from "stores";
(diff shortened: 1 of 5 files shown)


veggiesaurus:
@kswang1029 please test. I will try to fix the async issues with catalog files in the mean time

kswang1029:
some issues:
ISSUE1: If we use shift+click to select a set of consecutive files, then use cmd+click to hit some selected files, those are being selected _again_. Thus we will have duplicated load. The expected behavior should be unselect instead.

https://user-images.githubusercontent.com/20819712/103843815-779c8500-50d3-11eb-904d-87364204826b.mp4

ISSUE2: the button name and tooltip of the load button in the image/region/catalog file browser should change accordingly when  multiple files are selected. E.g Load->Load selected, E.g, in tooltip, load this image -> load selected images

some questions:
1. when a file is selected, shall we support shift+arrow up/arrow down to select consecutive files?
2. UX: how do we let users know that they can select multiple files now? (not necessary to deal with this in this PR, just point this our as a potential UX issue)

duidae (approved):

veggiesaurus:
some minor adjustments, @duidae please re-review

panchyo0 (approved):

PR #1290, opened 2021-01-06
use boolean instead of animation state enum

### LINKED ISSUE #1125: [2.0 suggestion] add computed animating status to appstore for others to access & refactor accordingly
since animation status is commonly used in quite some places: 

@computed get isAnimating() = {
return this.animatorStore.animationState === AnimationState.PLAYING;
}

veggiesaurus:
I see no need to use an enum instead of a boolean, so I've refactored the animation store a bit. This closes #1125


kswang1029 (approved):
this works fine after the refactoring. 👍

duidae (approved):

PR #1291, opened 2021-01-06
guard against WebGL initialisation errors

### LINKED ISSUE #971: Error reporting and suggested course of action when WebGL fails to initialise
Currently, if WebGL fails to initialise, the frontend does not display any error notifications. Since it is a critical failure (we can't display images correctly at all), it would make sense to display a detailed error message and some suggested course of action. 

If running CARTA in a browser, the best thing to do is to check WebGLReport with the same browser in order to check if it is a CARTA-specific issue or a browser-wide issue. If running the application, we should mention issues with X11 forwarding and VNC.

@kswang1029 what do you think? See https://github.com/CARTAvis/carta-backend/issues/549 for the sort of issue that could be more easily cleared up by this addition

veggiesaurus:
closes #971 

Displays a simple error message indicating that CARTA requires a WebGL-compatible browser. If the user continues after this, no hard-crashes occur, but images and contours are not displayed.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -33,10 +33,11 @@ export class App extends React.Component {
         return (
             <div className={className}>
                 <UIControllerComponent/>
-                <Alert className={appStore.darkTheme ? "bp3-dark" : ""} isOpen={appStore.alertStore.alertVisible} onClose={appStore.alertStore.dismissAlert} canEscapeKeyCancel={true}>
+                <Alert icon={appStore.alertStore.alertIcon} className={appStore.darkTheme ? "bp3-dark" : ""} isOpen={appStore.alertStore.alertVisible} onClose={appStore.alertStore.dismissAlert} canEscapeKeyCancel={true}>
                     <p>{appStore.alertStore.alertText}</p>
                 </Alert>
                 <Alert
+                    icon={appStore.alertStore.alertIcon}
                     className={appStore.darkTheme ? "bp3-dark" : ""}
                     isOpen={appStore.alertStore.interactiveAlertVisible}
                     confirmButtonText="OK"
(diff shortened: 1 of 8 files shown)


ajm-ska:
I put it into a quick linux Electron desktop AppImage for testing, although I know I could have simply used Mix ' N Match with a web browser. 

Anyway when running it through a VNC session, as expected, I see the warning message and the images do not display.
I also see the warning when running on an old Parallels VM where the images do not display for some reason (If I start it with the --disable-gpu flag, the warning does not appear and the images display properly).
However, I do not see the warning message when running it on a newer Parallels VM where the images _do_ display properly. So overall it seems to work perfectly. Nice work!

But perhaps the message could be modified? 

At the moment it says "CARTA requires a browser with WebGL support". But I'd say the more common use case of getting this message would be through using the Electron desktop version over VNC, or even a web browser over VNC, so it might be worth saying something like "CARTA requires WebGL support. WebGL may not function through a VNC connection"?
(I know we are discouraging usage over VNC, so I'm not suggesting that we mention the solution of the --disable-gpu flag for the Desktop versions into the warning message).

veggiesaurus:
the other common case is when users don't update their GPU drivers frequently enough and Chrome's WebGL stops working. 

I'm a bit hesitant to mention VNC in particular. Perhaps this would be more general:

> "Could not load WebGL. If you are using CARTA through a browser, please ensure your browser supports WebGL. If you are using the CARTA standalone app, please ensure you are not running in a remote desktop."

kswang1029:
I think we just tell users "don'ts" and not give workaround as workaround seems to encourage users do those "don'ts" The above sentence fits the idea I think.

ajm-ska:
It all looks fine to me. It is great to finally detect and report when this is an issue.
Screenshot 2021-01-08 at 12 16 01

ajm-ska (approved):

Issue #1279, opened 2020-12-20
Resizing preferences window does not expand content


nbrunett:
When I first open the preferences window, there is too much content to show all at once, like below
![Screen Shot 2020-12-20 at 11 42 13 AM (2)](https://user-images.githubusercontent.com/7672534/102718880-a683c000-42b8-11eb-9cdc-68e6c28292fd.png)

However, if I stretch out the window by dragging the bottom edge, the content of the window stays in the same small space like this
![Screen Shot 2020-12-20 at 11 42 23 AM (2)](https://user-images.githubusercontent.com/7672534/102718901-c7e4ac00-42b8-11eb-8a8c-c9550cfccaba.png)

My expectation was that the content would expand to fill the larger window and remove the need for the scroll bar if the window was made tall enough.

I am accessing a CARTA server (v1.4.0) that is running on a linux machine from a browser on my Mac.

PR #1308, opened 2021-02-03
adjust height of tab panels

YuHsuan-Hwang:


YuHsuan-Hwang:
This PR fixes https://github.com/CARTAvis/carta-frontend/issues/1279.

duidae:
image

the log events become a single row, if that's acceptable, then this PR is ok.

kswang1029:
> image
> 
> the log events become a single row, if that's acceptable, then this PR is ok.

not sure if this is related to screen resolution or something else. With dev frontend, iMac27 2K, and MBP13 I see two columns but with MBP16 or a 2K external monitor, I see single column. In any case, I don't have preference to have it as single or multi-columns.

kswang1029 (approved):

duidae (approved):

Issue #727, opened 2020-02-06
Allow user to generate contours only for the currently visible FoV


veggiesaurus:
For large files, contours can take a very long time to generate. We should allow users to generate contours for the currently visible portion of the file only, through a dropdown in the contour dialog. This will require some ICD and backend changes as well.

kswang1029:
would the area be determined by the area of the composited tiles (which might be larger than visible FoV)?

veggiesaurus:
I think a bounding box including all visible tiles (assuming #587 is done properly) should be the safest option. It doesn't' really matter if we end up contouring 25% more of the image than necessary, as long as it's not the entire image. Especially for giant (64K x 64K) images!

arusstaylor:
This should be the default behaviour.

veggiesaurus:
> This should be the default behaviour.

Possibly this should be in a preference, because it's unusual, when compared to other viewers

arusstaylor:
I am working on a large mosaic image from MeerKAT.  Making contours in one polarization to overlay on other polarization states.  The contours are fairly low level.   This means that there are a large number of contours over the entire mosaic. Even though I have only the object of interest in the viewport, CARTA grinds to a halt when I try to apply the contours.  It is borderline unworkable.    It would be good to get this feature implemented asap.  Large mosaic images will be coming out of MeerKAT more and more.

arusstaylor:
This feature being an option in preferences is fine.  Or, in the spirit of a smart system, it could be applied by default if the image is larger than some size.

arusstaylor:
Just purchased a mac studio with a skucom GPU.  It is much better now.  Workable at least.

kswang1029:
@arusstaylor what’s the image size in pixels? I would like to perform some tests to identify performance bottlenecks.

arusstaylor:
Hi KS. In this case the image size is 17000 by 13300.  I was originally running on a 2014 iMac with a fairly limited GPU.

Russ

kswang1029:
@arusstaylor @veggiesaurus I have done some tests with a 20000x20000 image with gaussian noise only. I tried 20 levels to sample the gaussian distribution. This is an extreme use case I think but is a good performance/stress test I suppose. 
![Screenshot from 2022-11-28 14-45-56](https://user-images.githubusercontent.com/20819712/204212485-bbcfb9cf-d036-48f3-867d-ddf368c0bebd.png)

[...]

PR #1310, opened 2021-02-04
Angus/mip range icd change

veggiesaurus:
Companion PR of https://github.com/CARTAvis/carta-backend/pull/430

Allows loading of mip-mapped HDF5 files, in order to drastically reduce initial time to load. Below is an example of a 40K x 40K image loading off a normal hard drive

https://user-images.githubusercontent.com/592504/106865505-f46f4e80-66d3-11eb-9cce-fc0380370869.mp4

Requests cropped and mip-mapped spatial profiles while the cursor is not frozen. 

~~Note: smoothing currently seems to break things, I'll attempt to fix this soon.~~ Note: the entire image needs to be loaded in full-resolution when a contour is requested. #727 would help a lot with this.

HDF5 files with mipmapped datasets can be generated using the latest version of [fits2idia](https://github.com/idia-astro/fits2idia) (or ask @veggiesaurus for example files).


duidae (approved):

kswang1029:
With the frontend: origin/angus/mip_range_icd_change 91f1352 and the backend origin/confluence/hdf5-mipmaps 165db40 , I see the full resolution spatial profiles got updated only when I press F to freeze the cursor movement, not automatically updated when cursor stops moving for a short time. Is it the intended behavior? That is, we only see full resolution profile when the state of the cursor is frozen.

https://user-images.githubusercontent.com/20819712/107332414-1bbc8600-6aef-11eb-809c-8ca2d077286e.mp4

I see the behavior is the same in FITS. This will cause usability issues.

veggiesaurus:
> With the frontend: origin/angus/mip_range_icd_change [91f1352](https://github.com/CARTAvis/carta-frontend/commit/91f1352a4ffce2accff17fd67317a4034d361b8e) and the backend origin/confluence/hdf5-mipmaps 165db40 , I see the full resolution spatial profiles got updated only when I press F to freeze the cursor movement, not automatically updated when cursor stops moving for a short time. Is it the intended behavior? That is, we only see full resolution profile when the state of the cursor is frozen.

@kswang1029 this is fixed in `fb65497e` :slightly_smiling_face: Now it also shows the full profile when the cursor stops for 250 ms or more.

kswang1029:
> @kswang1029 this is fixed in `fb65497e` 🙂 Now it also shows the full profile when the cursor stops for 250 ms or more.

Confirmed! Nice

kswang1029:
@veggiesaurus do you think the following is an intended behavior? When the cursor is outside the image and moving, the spatial profiles are switched to the downsampled ones. Once the cursor stops moving, the profiles are switched back to the full resolution ones. In this case, we see identical profiles getting updated repeatedly along with the movement of the cursor in the empty area.
https://user-images.githubusercontent.com/20819712/116955227-82070e00-acc4-11eb-81b6-d7e74d285dbd.mp4

kswang1029:
with regular FITS image, changing channel does not work and we see the image does not update accordingly. HDF5 works fine.
Screen Shot 2021-05-04 at 10 52 34 AM

kswang1029:
I am also seeing the following warning from the console. 
Screen Shot 2021-05-04 at 11 00 00 AM
Screen Shot 2021-05-04 at 10 59 47 AM
Screen Shot 2021-05-04 at 10 59 27 AM

kswang1029:
@veggiesaurus when testing https://github.com/CARTAvis/carta-backend/pull/430 I see that when we export a spatial profile to tsv format, it is the decimated profile being exported, not the full resolution one (it should be). This is independent of the image type (FITS or HDF5).
 Could you confirm this is the case?

veggiesaurus:
> @veggiesaurus when testing [CARTAvis/carta-backend#430](https://github.com/CARTAvis/carta-backend/pull/430) I see that when we export a spatial profile to tsv format, it is the decimated profile being exported, not the full resolution one (it should be). This is independent of the image type (FITS or HDF5).
> Could you confirm this is the case?

@kswang1029 I can't reproduce this issue. Is it possible that the full resolution spatial profile is still being streamed?

confluence:
If the image is HDF5, the profile can only be decimated if the file doesn't have any mipmaps. Otherwise a low-res HDF5 profile should be downsampled. (Although it shouldn't be in this case.)

kswang1029:
@veggiesaurus could you try the s255 cube? Following is what I see from the tsv file when the frontend decimation is applied for visualization.

Screen Shot 2021-07-06 at 10 41 59 AM

If we maximize the spatial profiler and make a big enough browser window so that there is no frontend decimation, we will see the full resolution data in the tsv.

Screen Shot 2021-07-06 at 10 42 51 AM

PR #1315, opened 2021-02-08
Jason/issue 1127 4

### LINKED ISSUE #1127: [2.0 suggestion] AST refactoring: sort out wcsInfo, fullWCSInfor, wcsforTransformation...
so far we duplicate various AST objects such as wcsInfo, fullWCSInfor, wcsforConversion, so maybe we can sort out a way for clearer logic

- [x] init frame set with only at most 3D for fullWCSInfo & extracting spectral frame
- [x] extract 2D frame set from the frameset above
- [x] rearrange code in FrameStore

duidae:
closes #1127 

please help test the functions & values related wcs, thanks.


src/stores/FrameStore.ts
```diff
@@ -753,15 +727,33 @@ export class FrameStore {
...
-
+        // init WCS
         this.astFrameSet = this.initFrame();
         if (this.astFrameSet) {
-            this.spectralFrame = AST.getSpectralFrame(this.astFrameSet);
+            if (this.spectralAxis && this.spectralAxis.valid) {
+                this.spectralFrame = AST.getSpectralFrame(this.astFrameSet);
+            }
+
+            if (frameInfo.fileInfoExtended.depth > 1) { // 3D frame
```
veggiesaurus, line 737: I wonder... does this still work with 3D cubes with a single channel? (i.e. `NAXIS = 3` and `NAXIS3 = 1`)

duidae: frameInfo.fileInfoExtended.depth > 1 is an existing old logic and spreading in quite some places, if it is incorrect, we can correct it in this PR

veggiesaurus: `frameInfo.fileInfoExtended.depth > 1` is generally correct for checking whether a cube is 2D or 3D, but when it comes to AST, we might still need to check things more carefully.

duidae: the original logic of 3D cube initialization is:
`if frameInfo.fileInfoExtended.depth > 1 then initFullWCS()`
and initFullWCS() gets NAXIS1~3 for AST

and current logic is just the same except adding a check for the spectral/stokes axis, so not sure the difference & have no clue about the checking,

@veggiesaurus would u mind elaborate that directly? thanks

veggiesaurus: sorry, I think you're right, it shouldn't make a difference. I was just worried that situations where you've got an image with NAXIS = 3 and NAXIS3 = 1 (which for CARTA's purposes is really a 2D image, just with a degenerate 3rd axis), AST wouldn't let you use `astTran2` since the frame info would be 3D.

veggiesaurus (approved):
changes look good. Ready to merge once @kswang1029 has tested and approved

kswang1029:
@veggiesaurus @duidae will do by this week

kswang1029 (approved):
With some tests and comparisons with v1.4, I think the refactoring works. 👍

PR #1317, opened 2021-02-11
fixed issue 453 578

### LINKED ISSUE #453: Zooming in a profile in point style results in empty plot
The following is an example with 7860 channels. With much fewer channels (e.g 100), it is harder to trigger this issue.
Aug-07-2019 10-34-23

### LINKED ISSUE #578: missing grid lines when zooming the Q vs U scatter plot in the Stokes widget
frontend: ee77422

when zooming in in the Q vs U scatter plot in the Stokes widget, the grid lines may disappear even when zooming out. Double click will get then back.
Screen Shot 2019-10-28 at 14 51 15

panchyo0:
when user zoom too fast, min value will become larger than max
1. fixed #453 
2. fixed #578


veggiesaurus (changes requested):
changes look good. Please just remove the `.eslintcache` file from the commit. In fact, we should add it to the `.gitignore`

veggiesaurus (approved):
looks good. @kswang1029 can you confirm this fixes both issues?

kswang1029:
@panchyo0 @veggiesaurus good to merge 👍

kswang1029 (approved):

Issue #1175, opened 2020-08-31
support exporting a subimage (frontend)


kswang1029:
The feature is to be able to export a subimage from the original cube by specifying a rectangle region and a range of channels.

relavent backend issue: https://github.com/CARTAvis/carta-backend/issues/618

Requirement [doc](https://docs.google.com/document/d/1Jpju5UUzyFD98-iTgXomtmmHlfbWSHHokEVPCiA2t3I/edit#)

PR #1321, opened 2021-02-19
Hj/save sub image

zarda:
Support the feature of exporting subimage: (co-work with backend [implement](https://github.com/CARTAvis/carta-backend/issues/618)) in #1175.

Add functions:
- Export original image
- Export subimage in a spectral range of original image
- Export subimage in a combo of stokes of original image
- Export a region of chopped original image
- Export a shared region of chopped original image
- Drop degenerated axes

Add interface:
File -> Save Image

[Reference](https://docs.google.com/document/d/1Jpju5UUzyFD98-iTgXomtmmHlfbWSHHokEVPCiA2t3I/edit?pli=1#)

(Update Demo Clip)

https://user-images.githubusercontent.com/10767557/110899891-d634e800-833c-11eb-92ba-17bbc8ff1e25.mov

https://user-images.githubusercontent.com/10767557/110899900-d9c86f00-833c-11eb-81b3-8cee1170698c.mov


src/components/FileInfo/FileInfoComponent.tsx
```diff
@@ -1,15 +1,18 @@
...
+import { CARTA } from "carta-protobuf";
+import { TableComponent, TableComponentProps } from "components/Shared";
 import "./FileInfoComponent.scss";
+import { AppStore, FileBrowserStore } from "stores";
+import { SpectralSystem, SpectralType, SpectralUnit } from "models";
 
 export enum FileInfoType {
     IMAGE_FILE = "image-file",
     IMAGE_HEADER = "image-header",
+    CHOP_IMAGE = "chop-image",
```
veggiesaurus: I don't think this really belongs in the file info component, since it's more about settings for saving a file. I suggest we move this into a separate component. I also think we should use `SAVE_SUBIMAGE` instead of `CHOP_IMAGE`

src/services/BackendService.ts
```diff
@@ -1,9 +1,14 @@
-import {action, observable, makeObservable, runInAction} from "mobx";
-import {CARTA} from "carta-protobuf";
-import {Observable, Observer, Subject, throwError} from "rxjs";
-import {AppStore, PreferenceStore, RegionStore} from "stores";
-import {ApiService} from "./ApiService";
-import {mapToObject} from "utilities";
+import { action, observable, makeObservable, runInAction } from "mobx";
+
```
veggiesaurus: where do these new lines come from?

src/stores/AppStore.ts
```diff
@@ -1,11 +1,29 @@
...
+
 import * as Long from "long";
-import {Classes, Colors, IOptionProps, setHotkeysDialogProps} from "@blueprintjs/core";
-import {Utils} from "@blueprintjs/table";
+
 import * as AST from "ast_wrapper";
+
 import * as CARTACompute from "carta_computation";
-import {CARTA} from "carta-protobuf";
+
```
veggiesaurus: whitespace changes

src/stores/FileBrowserStore.ts
```diff
@@ -367,24 +423,40 @@ export class FileBrowserStore {
...
     }
 
-    constructor() {
-        makeObservable(this);
-        this.exportCoordinateType = CARTA.CoordinateType.WORLD;
-        this.exportFileType = CARTA.FileType.CRTF;
+    /// Transfer the request from stokes string
+    /// to [start, end, stride]
+    @computed get saveStokesRange(): number[] {
+        // [start, end, stride] for "ABCD"
```
veggiesaurus, line 473: if it's stokes, we should use IQUV instead of ABCD

src/stores/FileBrowserStore.ts
```diff
@@ -78,6 +105,9 @@ export class FileBrowserStore {
         if (AppStore.Instance.activeFrame && mode === BrowserMode.SaveFile) {
             this.saveFilename = AppStore.Instance.activeFrame.frameInfo.fileInfo.name;
         }
+        this.updateIniSaveSpectralRange();
```
veggiesaurus: could we not use a `@computed` value (`automaticSpectralRange` instead of this and an autorun?)
I don't like `IniSaveSpectralRange`. If `Ini` is short for `Initial`, rather use `Init` (or just use `Initial`)

src/stores/FileBrowserStore.ts
```diff
@@ -62,8 +69,28 @@ export class FileBrowserStore {
     @observable catalogFileInfo: CARTA.ICatalogFileInfo;
     @observable catalogHeaders: Array<CARTA.ICatalogHeader>;
 
+    // Save image
     @observable saveFilename: string = "";
     @observable saveFileType: CARTA.FileType = CARTA.FileType.CASA;
+    @observable saveSpectralRange: number[] = [0, 0];
+    @observable saveStokesOption: number;
+    @observable saveRegionId: number;
+    @observable isDropDegeneratedAxes: boolean;
```
veggiesaurus: rename to `shouldDropDegenerateAxes`

src/services/BackendService.ts
zarda: It comes from TSlint. Do we need this lint?

src/stores/FileBrowserStore.ts
zarda: It's only a semantic demo to quick explain the orders. Will use "IQUV" insteadly.

src/stores/FileBrowserStore.ts
zarda: Got it.

src/stores/FileBrowserStore.ts
zarda: There are several situations we need to initialize SaveSpectralRange: activeFrame changes, updateSpectralSystem, updateSpectralCoordinate, and start a new FileBrowserDialog. Suppose to use @reaction and @action to trigger the update of SaveSpectralRange.

zarda: Changed the naming

src/components/FileInfo/FileInfoComponent.tsx
zarda: Changed the naming

zarda: How about add a new component for image save, 'ImageSaveComponent' or 'FileSaveComponent'?

kswang1029:
Cannot edit value (eg, -20). The value persists at 0.

https://user-images.githubusercontent.com/20819712/108675147-5841a600-7521-11eb-85a2-f7bc8f2b22c4.mp4

kswang1029:
The value validation is triggered annoyingly too fast. 

https://user-images.githubusercontent.com/20819712/108676203-d3f02280-7522-11eb-86c9-66ba73a5c5ce.mp4

zarda:
> The value validation is triggered annoyingly too fast.
> 
>  Feb-22-2021.15-29-36.mp4

How about increasing debounce timeout to 2000 ms?

src/services/BackendService.ts
veggiesaurus: which tslint rule in particular? I don't mean the spaces between `{` and `action`, those are fine. I mean the blank lines between imports.

zarda: get the problem and correct it.

src/stores/AppStore.ts
zarda: Correct it

src/components/FileInfo/FileInfoComponent.tsx
zarda: May I have any hint about how to refactor it or any suggestion? I suppose to move out the codes related to 'save image' to a new component, 'components/SaveFile/ImageSaveComponent'.

veggiesaurus: I'd say `FileBrowser/ImageSaveComponent` would work

zarda: Got it.

kswang1029:
Value validation is still too aggressive as it modified the other field instead of the current field only. Maybe it is simpler to just validate values at once after clicking the Save button and use a popup to highlight errors.

https://user-images.githubusercontent.com/20819712/111099881-6b282300-8581-11eb-9f04-57e6f32010c2.mp4

kswang1029:
I just realized that the current layout may be misleading so that users think the output spectral definition is what have been chosen from the GUI. Could you change the UI to the following? That should be less ambiguous. 
Screen Shot 2021-03-15 at 11 34 43 AM

zarda:
@kswang1029 How about let the input box be in red while the input value is invalid? This means to change the `intent` of input box to 'danger' while it's invalid.
Screen Shot 2021-03-15 at 12 15 43

kswang1029:
> @kswang1029 How about let the input box be in red while the input value is invalid? This means to change the `intent` of input box to 'danger' while it's invalid.
> Screen Shot 2021-03-15 at 12 15 43

that might work.

kswang1029:
As saving an image may take some time (esp. for large cubes), we should disable the "Save" button once it is clicked for the first time, otherwise we can trigger multiple save requests.

kswang1029:
It appears that if we firstly save a Stokes IV cube from an IQUV cube, then we cannot save the IV cube to I or V only.

zarda:
> It appears that if we firstly save a Stokes IV cube from an IQUV cube, then we cannot save the IV cube to I or V only.

Could you have more details about how to reproduce this?

kswang1029:
> > It appears that if we firstly save a Stokes IV cube from an IQUV cube, then we cannot save the IV cube to I or V only.
> 
> Could you have more details about how to reproduce this?

Try the steps such as:
1) load an cube with IQUV
2) save IV as image_IV.fits
3) load or append image_IV.fits
4) try to save as a new image from image_IV.fits and the Stokes dropdown seems to have IV only. There should be IV, I and V three options.

zarda:
> > > It appears that if we firstly save a Stokes IV cube from an IQUV cube, then we cannot save the IV cube to I or V only.
> > 
> > 
> > Could you have more details about how to reproduce this?
> 
> Try the steps such as:
> 
> 1. load an cube with IQUV
> 2. save IV as image_IV.fits
> 3. load or append image_IV.fits
> 4. try to save as a new image from image_IV.fits and the Stokes dropdown seems to have IV only. There should be IV, I and V three options.

Would you mind checking the header to see if there are still 2 stokes?
In my testing, there are "IV", "I", and "V" in the options if the source image has two stokes.

kswang1029:
This is a test IV image generated from IQUV cube. Once loaded in CARTA we can see the Stokes parameter is correctly displayed in the cursor info bar. The header is defined properly. But when we try to save this image as I or V only, the dropdown displays IV only.
[test.fits.zip](https://github.com/CARTAvis/carta-frontend/files/6139414/test.fits.zip)

kswang1029:
in some cases, users might make the window shorter. Whee this happens, there is no way to access those UI elements not in the view. Should change the layout to be scrollable when needed. 

Screen Shot 2021-03-15 at 3 23 19 PM


In addition, I suggest to make the UI element "From-To" more responsive when width changes.

kswang1029:
When the width is not enough, we cannot access the full UI. Can the UI elements be responsive by reducing the width first then if the width limit is hit, changing into two lines.
Screen Shot 2021-03-25 at 10 37 02 AM

kswang1029:
The drop degenerate axes toggle should be only triggered by clicking either the text or the toggle itself. Now it can be triggered by clicking an empty area in the same row.

https://user-images.githubusercontent.com/20819712/112412005-c368dd00-8d58-11eb-8190-72c31d4e9869.mp4

kswang1029:
The save button should be disabled if red box presents. 

Screen Shot 2021-03-25 at 10 58 48 AM

kswang1029:
Due to the new save image tab, we need more space for the HDU dropdown.
https://user-images.githubusercontent.com/20819712/112413131-a208f080-8d5a-11eb-8ff9-a709e31610a1.mp4

kswang1029:
> Due to the new save image tab, we need more space for the HDU dropdown.
> https://user-images.githubusercontent.com/20819712/112413131-a208f080-8d5a-11eb-8ff9-a709e31610a1.mp4

is this fixed?

kswang1029:
> > Due to the new save image tab, we need more space for the HDU dropdown.
> > https://user-images.githubusercontent.com/20819712/112413131-a208f080-8d5a-11eb-8ff9-a709e31610a1.mp4
> 
> is this fixed?

using relative width breaks the layout when selecting a multi-HDU fits.
Screen Shot 2021-03-31 at 3 00 07 PM

Maybe you can try a fixed width other than 200px.

zarda:
Use unit em to base on the font-size in global preference. It maybe better while changing size of box.

kswang1029 (approved):
Feature-wise this PR is ready👍

veggiesaurus:
> Use unit em to base on the font-size in global preference. It maybe better while changing size of box.

Which global preference are you referring to?

zarda:
> > Use unit em to base on the font-size in global preference. It maybe better while changing size of box.
> 
> Which global preference are you referring to?

Sorry,
I misunderstand the effect range of ‘rem’ and ‘em’. Which one would be better in this case?

veggiesaurus:
> > > Use unit em to base on the font-size in global preference. It maybe better while changing size of box.
> > 
> > 
> > Which global preference are you referring to?
> 
> Sorry,
> I misunderstand the effect range of ‘rem’ and ‘em’. Which one would be better in this case?

I think it's fine to use `pix`, since we don't have any global font size. 

@zarda please fix the unnecessary whitespace changes mentioned above, and the merge conflicts. We should then be able to merge this in

src/components/FileInfo/FileInfoComponent.tsx
```diff
@@ -4,13 +4,15 @@ import {action, makeObservable, observable} from "mobx";
 import {Button, ButtonGroup, ControlGroup, Divider, FormGroup, HTMLSelect, InputGroup, IOptionProps, NonIdealState, Popover, PopoverInteractionKind, PopperModifiers, Position, Pre, Spinner, Tab, TabId, Tabs, Text} from "@blueprintjs/core";
 import {FixedSizeList as List} from "react-window";
 import AutoSizer from "react-virtualized-auto-sizer";
-import {CARTA} from "carta-protobuf";
+import {CARTA } from "carta-protobuf";
```
veggiesaurus: whitespace

src/components/FileInfo/FileInfoComponent.tsx
```diff
@@ -19,7 +21,7 @@ export enum FileInfoType {
 @observer
 export class FileInfoComponent extends React.Component<{
     infoTypes: FileInfoType[],
-    HDUOptions?: {HDUList: IOptionProps[], handleSelectedHDUChange: (hdu: string) => void;},
+    HDUOptions?: { HDUList: IOptionProps[], handleSelectedHDUChange: (hdu: string) => void; },
```
veggiesaurus: unnecessary whitespace change. We can change them in a separate issue (e.g. reformatting with prettier)

src/components/FileInfo/FileInfoComponent.tsx
```diff
@@ -34,13 +36,13 @@ export class FileInfoComponent extends React.Component<{
     @observable searchString: string = "";
     @observable matchedIter: number = 0;
     @observable isMouseEntered: boolean = false;
-    
+
     private isSearchOpened: boolean = false;
     private matchedTotal: number = 0;
-    private matchedIterLocation: {line: number, num: number} = {line: -1, num: -1};
+    private matchedIterLocation: { line: number, num: number } = { line: -1, num: -1 };
```
veggiesaurus: as above

veggiesaurus (commented):
@zarda can you fix the minor whitespace issues and the merge conflicts in your other PR (https://github.com/CARTAvis/carta-backend/pull/739) so we can merge this in?

src/components/FileInfo/FileInfoComponent.tsx
zarda: Retrieved

veggiesaurus (approved):

veggiesaurus:
@zarda I've made some changes to how you were handling the ranges, and fixed some spelling issues. @duidae can you have a look as well?

veggiesaurus (changes requested):
@zarda a few other issues I picked up that need fixing:
- ~~We should be able to save sub-images of 2D images (no channel selection required)~~ Sorry, this was due to my changes, don't worry about that :see_no_evil: 
- I think the default output format should be the same as the input format. For example, if the file is FITS, the default output should be FITS as well. CASA should only be default if the file itself is a CASA file.

zarda:
The spectral range could not be updated synchronously while we change the coordinate system or source image after 3bb3460

veggiesaurus:
@zarda it previously worked because you were breaking MobX strict mode rules. I have adjusted this now, it should work. @kswang1029 please check

veggiesaurus (approved):

kswang1029 (approved):
feature-wise it is good.

Issue #1171, opened 2020-08-27
add stokes dropdown in the histogram widget and the statistics widget


kswang1029:
Stokes dropdown needs to be added to the histogram widget and the statistics widget. This allows users to view Stokes analytics independently without using the animator stokes slider.

markccchiang:
Is this feature for version 1.4? It seems not so straightforward to extract the stokes as an independent input parameter for histogram and statistics calculations. Not sure it can be done by this week.

veggiesaurus:
I suggest we leave it out, it's too risky

kswang1029:
ok

kswang1029:
then there is no need to backport the frontend part.

kswang1029:
remove the backport tag and add to v1.5 release.

markccchiang:
The backend branch [mark/stokes_for_stat_and_hist_calc](https://github.com/CARTAvis/carta-backend/tree/mark/stokes_for_stat_and_hist_calc) and protobuf branch [mark/stokes_for_stat_and_hist_calc](https://github.com/CARTAvis/carta-protobuf/tree/mark/stokes_for_stat_and_hist_calc) are ready to test with the frontend.

PR #1323, opened 2021-02-20
Yuhsuan/issue 1262 stats output

### LINKED ISSUE #1262: statistics widget: export as human and machine readable format
Users request to be able to save the content in the statistics widget. The format should be human and machine readable.

YuHsuan-Hwang:
This PR adds feature requested in https://github.com/cartavis/carta-frontend/issues/1262.
The output tsv file follows the same format with those in the x/y/z profile dialogs.
To let the file be more human readable, the x column strings are padded with spaces at the end, and y column values are all shown in scientific notation.


kswang1029:
although it is not related to this PR I suggest we make the default height of the statistics widget (when as a floating widget) taller so that all statistic quantities are visible in the view.

kswang1029:
Few suggestions:
1. in the tsv file, we remove `# xLabel: Statistic`, `# yLabel: Value` and `# zLabel: Unit` from header and replace `# x y z` by `# statistic value unit`.
2. add channel index and Stokes info to the header when possible (for a 3d or 4d cube).
3. at the moment when a point region or no region is selected the statistics refer to the full image in the view. So in the header of the tsv file we should mention it like `# full image` in addition to item 2 (when possible).

YuHsuan-Hwang:
@kswang1029 The widget height and the header format are adjusted.

src/components/Stats/StatsComponent.tsx
```diff
@@ -115,43 +125,109 @@ export class StatsComponent extends React.Component<WidgetProps> {
...
 
+    onMouseEnter = () => {
+        this.showMouseEnterWidget();
+    };
+
+    onMouseLeave = () => {
+        this.hideMouseEnterWidget();
+    };
+
+    private getTableValue = (index, type) => {
```
duidae: add type check: index: numeric, type: CARTA.StatsType

src/components/Stats/StatsComponent.tsx
```diff
@@ -115,43 +125,109 @@ export class StatsComponent extends React.Component<WidgetProps> {
...
+        let numString = "";
+        let unitString = "";
+        
+        if (isFinite(index) || type) {
```
duidae: this OR logic is incorrect, should be if (this.statsData && isFinite(index) && index >= 0 && index < this.statsData.statistics?.length)

src/components/Stats/StatsComponent.tsx
```diff
@@ -115,43 +125,109 @@ export class StatsComponent extends React.Component<WidgetProps> {
...
+                } else if (type === CARTA.StatsType.FluxDensity) {
+                    if (unit === "Jy/beam") {
+                        unitString = "Jy";
+                    }
+                } else {
+                    unitString = unit;
+                }
+            }
+            
+            const value =  this.statsData?.statistics[index].value;
```
duidae: optional chaining(?.) is not necessary if the above if() protection was added

src/components/Stats/StatsComponent.tsx
```diff
@@ -115,43 +125,109 @@ export class StatsComponent extends React.Component<WidgetProps> {
...
         let formContent;
+        let exportDataComponent = null;
         if (this.statsData) {
             // stretch value column to cover width
             const valueWidth = Math.max(0, this.width - StatsComponent.NAME_COLUMN_WIDTH);
 
             let rows = [];
             StatsComponent.STATS_NAME_MAP.forEach((name, type) => {
-                const index = this.statsData.statistics.findIndex(s => s.statsType === type);
+                const index = this.statsData?.statistics.findIndex(s => s.statsType === type);
```
duidae: optional chaining(?.) is not necessary since the above if() protection

duidae (changes requested):
please help to update to lastest dev before review, thanks.

kswang1029:
When there are spectrally matched images, the channel index in the header should be the matched index of the image selected by the image dropdown. In the following example, the current image in the view is `HD163296_13CO_2-1.fits` and the displayed channel is `54`. If we create two statistics widgets and let each display the statistics of the two matched images by configuring the image dropdown menu then request tsv of `HD163296_CO_2-1.fits` which is currently not in the image view we will see the channel index is `54` not `59`. Stokes index has the same issue. (note there is an ongoing issue to add Stokes dropdown to the statistics widget (https://github.com/CARTAvis/carta-frontend/issues/1171). At the moment we cannot select Stokes value and have to rely on the animator)

Screen Shot 2021-02-22 at 11 08 39 AM

src/components/Stats/StatsComponent.tsx
```diff
@@ -115,43 +125,109 @@ export class StatsComponent extends React.Component<WidgetProps> {
...
+                    }
+                }
+            } else {
+                regionInfo += "# full image\n"
+            }
+
+            let channelInfo = ""
+            let stokesInfo = ""
+            const activeFrame = AppStore.Instance.activeFrame;
+            if (activeFrame) {
```
duidae: should be effective frame?

YuHsuan-Hwang:
@duidae @kswang1029 The issues are fixed, thanks.

kswang1029:
> When there are spectrally matched images, the channel index in the header should be the matched index of the image selected by the image dropdown. In the following example, the current image in the view is `HD163296_13CO_2-1.fits` and the displayed channel is `54`. If we create two statistics widgets and let each display the statistics of the two matched images by configuring the image dropdown menu then request tsv of `HD163296_CO_2-1.fits` which is currently not in the image view we will see the channel index is `54` not `59`. Stokes index has the same issue. (note there is an ongoing issue to add Stokes dropdown to the statistics widget (#1171). At the moment we cannot select Stokes value and have to rely on the animator)
> 
> Screen Shot 2021-02-22 at 11 08 39 AM

Issue fixed. 👍

duidae (approved):

src/utilities/export.ts
```diff
@@ -0,0 +1,17 @@
+export function getTimestamp() {
```
veggiesaurus: I suggest we use `moment` here. It's already a dependency. How about something like this:
`export function getTimestamp(format: string = "YYYY-MM-DD-HH-mm-SS")` (just check the momentjs docs to ensure I've got the right values there)

veggiesaurus (changes requested):

src/utilities/export.ts
YuHsuan-Hwang: @veggiesaurus This part is modified.

src/utilities/export.ts
```diff
@@ -0,0 +1,18 @@
+import moment from "moment";
+
+export function getTimestamp(format: string = "YYYY-M-D-H-m-s") {
```
veggiesaurus: using `M` instead of `MM` is a bit problematic, because it means things are no longer easily sortable. For example, `2021-12-1` will be above `2012-8-1`. Rather use the zero-padded alternatives`MM`, `DD`, etc: https://momentjs.com/docs/#/displaying/format/

src/utilities/export.ts
```diff
@@ -0,0 +1,18 @@
...
+    const tsvData = `data:text/tab-separated-values;charset=utf-8,${content}\n`;
+    const dataURL = encodeURI(tsvData).replace(/#/g, "%23");
+
+    const a = document.createElement("a") as HTMLAnchorElement;
+    a.href = dataURL;
+
+    a.download = `${imageName}-${plotName.replace(" ", "-")}-${getTimestamp()}.tsv`;
+    a.dispatchEvent(new MouseEvent("click"));
+
+    return null
```
veggiesaurus: semicolon

src/utilities/export.ts
YuHsuan-Hwang: I agree this is better for sorting files. This part is changed.

src/utilities/export.ts
YuHsuan-Hwang: This is fixed.

veggiesaurus (approved):

PR #1326, opened 2021-02-22
force common spectral units when matching

### LINKED ISSUE #1283: spectral matching with different conventions in the headers
When exporting an image to a FITS file in CASA, there is an option to set the spectral convention as velocity in the header. So in principle we should see the original image and the exported image are perfectly matched with each others. However, we see an offset in the matched channel index. 
Screen Shot 2020-12-23 at 1 44 32 PM

[test image](https://drive.google.com/file/d/1yteh2a7EXNLMzcStOmmWxg3BUYu8hufY/view?usp=sharing)

veggiesaurus:
Closes #1283 

Makes sure the units are the same when doing spectral matching.

### DIFF
--- a/src/utilities/wcs.ts
+++ b/src/utilities/wcs.ts
@@ -1,6 +1,6 @@
 import {CARTA} from "carta-protobuf";
 import * as AST from "ast_wrapper";
-import {Point2D, WCSPoint2D, SpectralType} from "models";
+import {Point2D, WCSPoint2D, SpectralType, SPECTRAL_DEFAULT_UNIT} from "models";
 import {NumberFormatType} from "stores";
 import {add2D, polygonPerimeter, rotate2D, scale2D, subtract2D, magDir2D} from "./math2d";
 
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
Confirmed the fix! 👍

veggiesaurus:
@kswang1029 can you make sure it doesn't break other things (cursor, spectral profile output values)?

kswang1029:
@veggiesaurus yes they all appear fine.

kswang1029:
@duidae for spectral line labels do we also consider units when converting line frequency to whatever users choose for x-axis display in the spectral profiler widget?

duidae (approved):

duidae:
> @duidae for spectral line labels do we also consider units when converting line frequency to whatever users choose for x-axis display in the spectral profiler widget?

image
we have different units in the dropdown?

kswang1029:
@duidae I just checked line label seems to work fine. So I guess unit here is considered in the calculation. :)

Issue #1329, opened 2021-02-24
animation playback not working for the image iterator in the animator


kswang1029:
To reproduce, load two images, select the image radio button in the animator and click the play button. 

The frontend does not send out START_ANIMATION message to the backend. It does send out the message if it is channel or stokes.

veggiesaurus:
Image animation is front-end only, so it shouldn't be sending anything. any messages in the dev console?

kswang1029:
when the button is clicked, no new message is seen in the console.

kswang1029:
` ksw/test_image_playback ` provides a one-liner fix to the issue. @veggiesaurus would you have a look to see if that makes sense when you have time?

UPDATE: same branch might fix https://github.com/CARTAvis/carta-frontend/issues/1270 too.

Issue #1270, opened 2020-11-29
Changing order of images does not change order in animator


keflavich:
I can change the order of images using the Image List tool.  In the animator, if I drag the tracker box left to right, it does the expected thing and obeys the order given in Image List.  However, if I use the Next/Previous tool, it retains the order in which the files were loaded instead of the modified order in Image List.

kswang1029:
This issue is blocked by https://github.com/CARTAvis/carta-frontend/issues/1329 where the play button is not functional for the image iterator.

kswang1029:
A set of test images for this issue is available in `set_dice` where there are six images as the "dice".

PR #1332, opened 2021-03-01
fix play button for image and image animation order

kswang1029:
This PR fixes 
1. https://github.com/CARTAvis/carta-frontend/issues/1329 the animation state when animation playback is triggered is not set correctly.
2. https://github.com/CARTAvis/carta-frontend/issues/1270 the frame order is managed by `handleFileReordered` in LayerListComponent so there is no need to sort the frame list first when determining which frame to show next in AppStore.

### DIFF
--- a/src/components/Menu/RootMenuComponent.tsx
+++ b/src/components/Menu/RootMenuComponent.tsx
@@ -168,7 +168,7 @@ export class RootMenuComponent extends React.Component {
             </Menu>
         );
 
-        let layerItems = appStore.frames.slice().sort((a, b) => a.frameInfo.fileId <= b.frameInfo.fileId ? -1 : 1).map(frame => {
+        let layerItems = appStore.frames.map(frame => {
             return (
                 <Menu.Item
                     text={frame.filename}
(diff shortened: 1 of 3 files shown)


kswang1029:
@acdo2002 please help to perform some tests to see if the two issues are fixed properly.

acdo2002:
1. The play button works once I would like to iterate the images (set frameRate as 1).
2. minor improvement suggestion, since the menu bar>View>Images, it shows a different order to Image List widget if I change the image list order on the widget. 
Image-List inconsistent
Could you change `carta-frontend/src/components/Menu/RootMenuComponent.tsx` line 171? As you have done before, remove `.sort((a, b) => a.frameInfo.fileId <= b.frameInfo.fileId ? -1 : 1)`, it may fix the inconsistent between menu bar and widget.

kswang1029:
@acdo2002 updated with c3fde1f, please help to re-test.

acdo2002 (approved):
UI works good!!!

veggiesaurus:
can you test with > 10 images, just to make sure there isn't any weird alphabetic sorting at play

kswang1029:
uh.... that might break things 🤔

kswang1029:
Just tried 12 images and it works

acdo2002:
I have tried 10-13 images several times, so far they work as what I expect (include Previous image/Next image function).

veggiesaurus (approved):

PR #1336, opened 2021-03-03
Angus/fix 932

### LINKED ISSUE #932: File browser opens behind contour widget
When using the contour configuration widget, and the file browser is opened (e.g. to append an image), the contour widget stays in front of it. It should open on top of all the widgets. It would actually be preferable to be able to pin the contour widget like the others. I don't see why it's different.

This should come under Epic #692

veggiesaurus:
Rearranges dialog ordering to fix #932


veggiesaurus:
@Jordatious please review commit `4a931a6` on mix n match

kswang1029 (approved):
Looks fine 👍

Jordatious:
This works for me too. But in general, I think this represents a larger issue, which is that the user cannot click on a window to bring it in front of another. So if I now open the header, and then the contour config, the latter opens behind the former. I think it's useful to have the file browser open on top of everything, but generally, I think the user should be able to choose which window is at the top simply by clicking on it. Obviously it's best to pin widgets, but some users won't want that, particularly for temporary windows like the header, file browser, and perhaps the contour config. Should I open a new ticket for this?

I'd also like to be able to pin the contour config. Is there reason to not do that? If not, should I reopen #932?

veggiesaurus:
>  I think the user should be able to choose which window is at the top simply by clicking on it.

Yes, perhaps that's a better approach. Could you make a _new_ issue for this? Since its more general than #932 (and also trickier to solve appropriately, I suspect)

> I'd also like to be able to pin the contour config. Is there reason to not do that? If not, should I reopen #932?

I think you can open a separate issue for that, rather. The two issues are separate.

PR #1338, opened 2021-03-04
Thchang/adding transparent image background option in preference

### LINKED ISSUE #923: Invisible labels of exported images during dark mode
When exporting images (including plots) during dark mode, the labels in the output figure are invisible, or otherwise difficult to see, since they're white, over a gray background, which seems to be semi-transparent. Using black would be much more useful, as when exporting images during light mode. I can't find a preference to configure the label colours, but perhaps there is, in which case the default should be updated.

TienHao:
closes #923 

Adding a controller in preference to export images with transparent or filled background.(default: filled) The color of filled background depends on the current theme.

Transparent background(light theme):
HD163296_CO_2-1_bin3 image-X-profile-2021-03-04-12-03-07

Transparent background(dark theme):
HD163296_CO_2-1_bin3 image-X-profile-2021-03-04-12-02-08

Filled background(light theme):
HD163296_CO_2-1_bin3 image-X-profile-2021-03-04-12-03-01

Filled background(dark theme):
HD163296_CO_2-1_bin3 image-X-profile-2021-03-04-12-02-34

### DIFF
--- a/src/stores/PreferenceStore.ts
+++ b/src/stores/PreferenceStore.ts
@@ -19,7 +19,7 @@ export enum PreferenceKeys {
     GLOBAL_DRAG_PANNING = "dragPanning",
     GLOBAL_SPECTRAL_MATCHING_TYPE = "spectralMatchingType",
     GLOBAL_AUTO_WCS_MATCHING = "autoWCSMatching",
-    GLOBAL_TRANSPARENT_EXPORTED_IMAGE = "transparentExportedImage",
+    GLOBAL_TRANSPARENT_IMAGE_BACKGROUND = "transparentImageBackground",
 
     RENDER_CONFIG_SCALING = "scaling",
     RENDER_CONFIG_COLORMAP = "colormap",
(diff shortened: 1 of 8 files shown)


src/stores/PreferenceStore.ts
```diff
@@ -86,6 +87,7 @@ const DEFAULTS = {
         dragPanning: true,
         spectralMatchingType: SpectralType.VRAD,
         autoWCSMatching: WCSMatchingType.NONE,
+        transparentImageBackground: true,
```
veggiesaurus: I think the default should be `false`, especially since the older schema doesn't have this field

veggiesaurus (changes requested):

src/stores/PreferenceStore.ts
TienHao: fixed

veggiesaurus (approved):

TienHao:
I suggest that we can add tooltips to the export image buttons, show different text depending on `transparentImageBackground`  to remind user.(Maybe like the images below?)

`transparentImageBackground` is diabled 
transparent image background2

`transparentImageBackground` is enabled 
transparent image background1

kswang1029 (approved):
works well 👍

PR #1340, opened 2021-03-05
Yuhsuan/issue 1218 file browser date

YuHsuan-Hwang:
This fixed issue https://github.com/CARTAvis/carta-frontend/issues/1218.
`moment(0, "HH").diff(t, "days")` returns zero when the difference is shorter than 1 day. Changed to `moment(0, "HH").diff(t)`, which returns negative value when the created time of the file is after today 0 a.m.


veggiesaurus (approved):

duidae (approved):

kswang1029:
I will report my test result TOMORROW

kswang1029 (approved):
Tested and it works~ 👍

PR #1342, opened 2021-03-05
Yuhsuan/issue 48 header search

### LINKED ISSUE #1230: cannot select text from file info
potentially regression of https://github.com/CARTAvis/carta-frontend/pull/1222

The current dev does not allow users to select texts from the file info and file header text boxes, hence does not allow copy-paste action.

YuHsuan-Hwang:
This PR is for new feature requested in https://github.com/CARTAvis/carta/issues/48.
Some details:
- The button becomes visible when hovering on the information panel.
- The search panel can be easily closed by clicking anywhere outside.
- Pressing enter moves to the next matched string.
- The style of highlighting and scrolling is similar to the text search in Chrome browser, but other suggestions are welcome.

### DIFF
--- a/src/components/FileInfo/FileInfoComponent.tsx
+++ b/src/components/FileInfo/FileInfoComponent.tsx
@@ -61,7 +61,7 @@ export class FileInfoComponent extends React.Component<{
     };
 
     @action setSearchString = (inputSearchString: string) => {
-        this.searchString = inputSearchString.replace("\b","");
+        this.searchString = inputSearchString.replace("\b", "");
     };
 
     @action resetMatchedNums = () => {
@@ -121,7 +121,7 @@ export class FileInfoComponent extends React.Component<{
         this.setSearchString(ev.target.value);
         this.resetMatchedNums();
         
-        if (this.searchString !== "" && this.searchString !== "\b") {
+        if (this.searchString !== "") {
             // RegExp ignores the difference of lettercase; use special characters as normal strings by putting \ in the front
             const searchStringRegExp = new RegExp(this.searchString.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'), 'i');
             this.splitLengthArray = [];
--- a/src/components/FileInfo/FileInfoComponent.tsx
+++ b/src/components/FileInfo/FileInfoComponent.tsx
@@ -61,7 +61,7 @@ export class FileInfoComponent extends React.Component<{
     };
 
     @action setSearchString = (inputSearchString: string) => {
-        this.searchString = inputSearchString;
+        this.searchString = inputSearchString.replace("\b","");
     };
 
     @action resetMatchedNums = () => {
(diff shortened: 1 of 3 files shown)


kswang1029 (approved):
This works nicely 👍  

this issue https://github.com/CARTAvis/carta-frontend/issues/1230 (cannot select by mouse) seems fixed too (tested with my local dev build 🤔)

veggiesaurus (changes requested):
this is _really_ nice! I only have one minor request: The search icon looks a bit weird when you have a header that requires a scroll tab

Firefox: 
image

Chrome:
Screenshot from 2021-03-05 16-18-52

Wonder if we should move it in a bit?

veggiesaurus:
It's pretty quick when handling large headers as well! :+1: 

https://user-images.githubusercontent.com/592504/110128593-a67f6080-7dcf-11eb-8955-4e8f707e1a8d.mp4

One thing that we might want to fix is the search bar disappearing when you scroll, because you can't actually see all the highlighted results.

kswang1029:
The location of the search button is tricky as different browsers seem to have different scroll bar widths and may expand when hovering. If the button is too far away from the bottom right corner, it is weird too🙈

YuHsuan-Hwang:
Different Chrome version also seems to have different scroll bar width

kswang1029:
@YuHsuan-Hwang could you also implement the same behavior as “enter” for the “previous” and “next” button? That will make the UX even better I guess.

YuHsuan-Hwang:
@veggiesaurus @kswang1029  Some changes:
1. Fixed #1230. Please help checking.
2. The search panel can be closed only by clicking the button. This avoids closing the panel when the users scroll.
3. Keep the buttons pressed over 0.5 sec will continuously scroll through the matched strings. Pressing enter after clicking the buttons will move forward/backward (the same as clicking the buttons). Pressing enter after clicking the text input will move forward.
4. The position of the button is not changed. It seems to be hard to fit all kinds of scroll bar.

duidae (approved):

veggiesaurus:
@YuHsuan-Hwang this is looking good. One issue: if you search for something, and then select another file, the search results persist and are buggy:

image

veggiesaurus:
I also encountered a hard-crash: 

```
Uncaught TypeError: Cannot read property 'forEach' of undefined
    at FileInfoComponent.highlightString
```

src/components/FileInfo/FileInfoComponent.tsx
```diff
@@ -97,22 +253,94 @@ export class FileInfoComponent extends React.Component<{
...
+        
+        const splitLength = this.splitLengthArray[index]
+        const nameValueLength = name.length + 3 + value.length;
+        let highlightedString = [];
+        let keyIter = 0; // add unique keys to span to avoid warning
+        let highlighClassName = "";
+        let typeClassName = "header-name";
+        let usedLength = 0; 
+
+        function addHighlightedString(sliceStart: number, sliceEnd: number) {
```
veggiesaurus: please make this an arrow function instead,.

src/components/FileInfo/FileInfoComponent.tsx
```diff
@@ -97,22 +253,94 @@ export class FileInfoComponent extends React.Component<{
...
+                return;
+            }
+            highlightedString.push(<span className={typeClassName + highlighClassName} key={keyIter.toString()}>
+                {((name !== "END") ? `${name} = ${value}${comment && " / " + comment}` : name).slice(sliceStart, sliceEnd)}</span>);
+            keyIter += 1;
+            return;
+        }
+
+        splitLength.forEach((arrayValue, arrayIndex) => {
+            highlighClassName = "";
```
veggiesaurus: typo

YuHsuan-Hwang:
@veggiesaurus Fixed the issues. Thanks for checking.

veggiesaurus (approved):

PR #1344, opened 2021-03-08
Jason/issue 1150 pv

### LINKED ISSUE #1150: display xy labels and ticks properly for a position-velocity image
The x and y labels and ticks should be displayed properly based on the image header. (AST support)

duidae:
closes #1150 

need backend support of PV


kswang1029:
@duidae This is very nice! Only few minor changes:
- [x] 1. When selecting Coordinate="air wavelength", change the label "Wavelength in air" in the image viewer to "Air wavelength" to be consistent. Same for "Vacuum wavelength". Add system label too (same as the spectral profiler widget)
Eg "LSRK, Air wavelength (mm)"
- [ ] 2. Can the coordinate dropdown include "Channel" (meaning channel index) too? 
- [x] 3. Apply "Radio velocity (km/s)" if possible by default to fit general use cases better (P"V"). Otherwise select the one defined in header.
- [x] 4. change tab title from "Spectral" to "Conversion"
- [x] 5. Add a text label "For spatial-spectral image" as the first line. This would be more generic when we support permuted cubes (eg RA-Frequency-Dec). Then add the label "Spectral axis" as the 2nd line (then Coordinate and System as usual).
- [ ] 6. Is it possible to change the unit for OFFSET? We will need to support: "milli-arcsec", "arcsec", "arcmin", and "deg". If possible, the layout is a text label "Spatial" followed by "Coordinate" as a dropdown with just "OFFSET" (whatever it is called in the header) for now and "Unit" as a dropdown including "milli-arcsec", "arcsec", "arcmin", and "deg". 
- [ ] 7. show cursor "world" values in the cursor info bar too.

veggiesaurus:
> @duidae This is very nice! Only few minor changes:
> - [ ] When selecting Coordinate="air wavelength", change the label "Wavelength in air" in the image viewer to "Air wavelength" to be consistent. Same for "Vacuum wavelength". Add system label too (same as the spectral profiler widget)
> Eg "LSRK, Air wavelength (mm)"

I think this is all handled by AST, so we'll need to specify a custom axis label in our AST plot call

duidae:
> @duidae This is very nice! Only few minor changes:
> 
> * [x]  1. When selecting Coordinate="air wavelength", change the label "Wavelength in air" in the image viewer to "Air wavelength" to be consistent. Same for "Vacuum wavelength". Add system label too (same as the spectral profiler widget)
>   Eg "LSRK, Air wavelength (mm)"
> * [ ]  2. Can the coordinate dropdown include "Channel" (meaning channel index) too?
> * [x]  3. Apply "Radio velocity (km/s)" if possible by default to fit general use cases better (P"V"). Otherwise select the one defined in header.
> * [x]  4. change tab title from "Spectral" to "Conversion"
> * [x]  5. Add a text label "For spatial-spectral image" as the first line. This would be more generic when we support permuted cubes (eg RA-Frequency-Dec). Then add the label "Spectral axis" as the 2nd line (then Coordinate and System as usual).
> * [ ]  6. Is it possible to change the unit for OFFSET? We will need to support: "milli-arcsec", "arcsec", "arcmin", and "deg". If possible, the layout is a text label "Spatial" followed by "Coordinate" as a dropdown with just "OFFSET" (whatever it is called in the header) for now and "Unit" as a dropdown including "milli-arcsec", "arcsec", "arcmin", and "deg".
> * [ ]  7. show cursor "world" values in the cursor info bar too.

@kswang1029 
(2)
image
AST does not have option like "UNKNOWN" for spectral domain, so showing simply channel index seems not trivial to do.

(5) no sure what this means, please elaborate more.

(6)
tried setting the unit of axis of OFFSET to other unit (arcmin/deg/milli-arcsec) but the values of the grid are not changing, apparently AST doesnt not trigger unit conversion of 'OFFSET' axis, probably it does not recognize 'OFFSET' as a spatial domain axis, also not trivial to do.

(7)
this seems not inside the scope of issue 1150, will go back to this if have time left, need to handle other issue first.

kswang1029:
for (5) I was considering
Screen Shot 2021-03-16 at 2 55 53 PM
The red box part can be ignored for now due to (6).

duidae:
image

(5) fixed,
(2)(6)(7) may not trivial to do.

@kswang1029 please have a look, thanks.

veggiesaurus:
> tried setting the unit of axis of OFFSET to other unit (arcmin/deg/milli-arcsec) but the values of the grid are not changing, apparently AST doesnt not trigger unit conversion of 'OFFSET' axis, probably it does not recognize 'OFFSET' as a spatial domain axis, also not trivial to do.

We should just handle the conversions here manually, if they are definitely needed.

veggiesaurus:
> AST does not have option like "UNKNOWN" for spectral domain, so showing simply channel index seems not trivial to do.

For situations where we're not using a spectral coordinate, we should just use image pixel coordinates for the axis.

kswang1029:
> > tried setting the unit of axis of OFFSET to other unit (arcmin/deg/milli-arcsec) but the values of the grid are not changing, apparently AST doesnt not trigger unit conversion of 'OFFSET' axis, probably it does not recognize 'OFFSET' as a spatial domain axis, also not trivial to do.
> 
> We should just handle the conversions here manually, if they are definitely needed.

ok to just support the unit of the offset axis defined in header for v2.0.

kswang1029:
@duidae @veggiesaurus if time does not allow I suggest we get (7) implemented and postpone (2) and (6) to a future release. The impact of not having (7) is more critical than the others.

duidae:
> @duidae @veggiesaurus if time does not allow I suggest we get (7) implemented and postpone (2) and (6) to a future release. The impact of not having (7) is more critical than the others.

@kswang1029 I am quite heavily loaded for sharing TienHao's loading of multiple profile, I think we just file an new issue for (7) & can @TienHao help for that?

kswang1029:
> > @duidae @veggiesaurus if time does not allow I suggest we get (7) implemented and postpone (2) and (6) to a future release. The impact of not having (7) is more critical than the others.
> 
> @kswang1029 I am quite heavily loaded for sharing TienHao's loading of multiple profile, I think we just file an new issue for (7) & can @TienHao help for that?

ok

kswang1029 (approved):
This is looking good. Please change the font style to regular instead of bold face and have the font size a bit smaller so that the layout is more compact (eg using the style same as "Coordinate"). Once this is done the PR is ready to merge in. 👍 
Screen Shot 2021-03-23 at 11 08 19 AM

TienHao:
> 
> @kswang1029 I am quite heavily loaded for sharing TienHao's loading of multiple profile, I think we just file an new issue for (7) & can @TienHao help for that?

yes, I can help for that.

veggiesaurus (approved):

duidae:
@veggiesaurus seems current dev backend has supported opening pv image, is this ok to merge now? thanks.

PR #1345, opened 2021-03-08
Jason/issue 1335

### LINKED ISSUE #1335: validate preferences.json and layout.json then apply error handling when needed
PR https://github.com/CARTAvis/carta-backend/pull/746 introduces saving config files preferences.json and layout .json to the .carta folder. (Note the same issue already exists if users edit localStorge from browser)

These json files might be edited by users and might be edited so that some attributes are not understandable by the frontend resulting that frontend is initialized in an unexpected state (eg empty layout). We will need a way to catch exceptions and apply a default value. For example, in the preferences.json I set a wrong colormap to "jettt" then the frontend should catch it and apply a default colormap such as inferno.

duidae:
closes #1335


veggiesaurus (approved):

kswang1029 (approved):
This works nicely. 👍

PR #1348, opened 2021-03-09
pass header string as ptr instead of stack variable

### LINKED ISSUE #1347: Out of bounds error when calling initFrame with a very large header
Spin-off issue of https://github.com/CARTAvis/carta-backend/issues/768

The AST wrapper's `initFrame` function is crashing when passing it a very long header string.

veggiesaurus:
Closes #1347

We were simply wrapping the `initFrame` function using `cwrap` with a string argument. This meant that the string was sent to WebAssembly as a stack variable. The test image in CARTAvis/carta-backend#768 has a 4MB header string, which naturally causes a problem. 

The solution is to `malloc` a string in the WebAssembly heap first, copy the string into that memory and then pass the string as a pointer.

### DIFF
--- a/wasm_src/ast_wrapper/post.ts
+++ b/wasm_src/ast_wrapper/post.ts
@@ -92,7 +92,6 @@ Module.setCanvas = function (canvas) {
 };
 
 Module.plot = Module.cwrap("plotGrid", "number", ["number", "number", "number", "number", "number", "number", "number", "number", "number", "number", "number", "string"]);
-Module.initFrame = Module.cwrap("initFrame", "number", ["string"]);
 Module.getSpectralFrame = Module.cwrap("getSpectralFrame", "number", ["number"]);
 Module.getSkyFrameSet = Module.cwrap("getSkyFrameSet", "number", ["number"]);
 Module.initDummyFrame = Module.cwrap("initDummyFrame", "number", []);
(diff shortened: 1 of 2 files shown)


duidae (approved):

duidae:
are comments causing the problem? if so, probably we can also let backend ignore those comments to reduce redundant info for AST.

kswang1029:
for users it would be good to see everything defined in the header. However, for feeding AST, we can remove COMMENT, HISTORY etc as they are not doing anything useful to AST.

duidae:
> for users it would be good to see everything defined in the header. However, for feeding AST, we can remove COMMENT, HISTORY etc as they are not doing anything useful to AST.

hmm, I see

veggiesaurus:
> are comments causing the problem? if so, probably we can also let backend ignore those comments to reduce redundant info for AST.

In this particular file, it's not that. It's just that there are _loads_ of header entries.

I'm wondering if we shouldn't rather use `astPutFits` to do the construction line-by-line. I'll give it a try in a separate branch. I suspect concatenating a massive string up to 4 MB is really bogging things down!

kswang1029:
Yes for this particular FITS file (made by me for testing purpose) the header is just long even without COMMENT/HISTORY. In some optical FITS images, the header might be like that unfortunately (but may not be _that_ long).

PR #1349, opened 2021-03-10
populate FITS header cards individually

### LINKED ISSUE #1347: Out of bounds error when calling initFrame with a very large header

veggiesaurus:
Closes #1347

This is a different approach to that of #1348. Instead of passing the header string as a pointer, we use `astPutFits` to populate individual FITS cards. This has two advantages:
- We can safely use the stack, since the individual cards are 80 characters or less
- We don't have to build up a giant string

This seems to be quite a bit faster than #1348 for situations where you have lots of header entries

### DIFF
--- a/src/stores/FrameStore.ts
+++ b/src/stores/FrameStore.ts
@@ -829,7 +829,8 @@ export class FrameStore {
     private initFrame = (): number => {
         const dimension = this.frameInfo.fileInfoExtended.depth > 1 ? "3" : "2";
 
-        let headerString = "";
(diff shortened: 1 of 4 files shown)


duidae (approved):

kswang1029 (approved):
tested with some usual use cases and I don't see odd things so far. good to go

Issue #1351, opened 2021-03-16
log and power scaling function mapping issue


kswang1029:
With some discussions with @YuHsuan-Hwang when implementing a colorbar (https://github.com/CARTAvis/carta-frontend/issues/601) to the image viewer, we think the current log and power scaling functions do not map value [0, 1] to [0, 1] exactly. Rather, they map the value [0, 1] to [0, ~<1] when the control parameter "alpha" is large enough. However, if users define a small alpha such as 10, the upper bound will be significantly less than unity (e.g. if alpha=10, log scaling function gives 0.954). This results in color saturation for those values mapped above that upper bound.

To retain consistency, the log and power functions should be modified to
<img width="146" alt="Screen Shot 2021-03-16 at 1 40 53 PM" src="https://user-images.githubusercontent.com/20819712/111261319-430df200-865d-11eb-9b25-3b9158db4318.png">

The impact to the current users should be negligible as the default alpha is large enough. Note that after the change, the log and power scaling functions will be slightly different from the ones from ds9 where original we quote the formulae there.

kswang1029:
@veggiesaurus do you think we can make the change here? The shader code needs some tiny changes too. eg https://github.com/CARTAvis/carta-frontend/blob/4b7a4ca26970eccb93ad1e8211875585193628f5/src/services/GLSL/pixel_shader_raster.glsl#L70
and
https://github.com/CARTAvis/carta-frontend/blob/4b7a4ca26970eccb93ad1e8211875585193628f5/src/services/GLSL/pixel_shader_raster.glsl#L73

veggiesaurus:
yes please :+1: Assign to @YuHsuan-Hwang?

PR #1353, opened 2021-03-16
change log and power scaling

YuHsuan-Hwang:
This PR fixed issue https://github.com/CARTAvis/carta-frontend/issues/1351.
Adjusted the scaling formulas of Log and Power.


veggiesaurus (changes requested):
We also need to change the scaling functions in `math.ts` (these are used in the contour generator) and update the log and power PNGs in `static/equations` to make these changes consistent.

YuHsuan-Hwang:
@veggiesaurus I changed the icons and the scaling functions for the contour generator. The impact to the contour generator is also negligible since alpha is always set to be 1000.

veggiesaurus (approved):

kswang1029 (approved):

PR #1364, opened 2021-03-25
ksw/revise axes label of the ellipse region

kswang1029:
trivial improvement of the axes label of the ellipse region in the region dialog.

Label "Axes" is changed to "Semi-axes".

Screen Shot 2021-03-25 at 10 23 17 AM

@Jordatious please check if this fixes the issue https://github.com/CARTAvis/carta/issues/70

### DIFF
--- a/src/components/Dialogs/RegionDialog/EllipticalRegionForm/EllipticalRegionForm.tsx
+++ b/src/components/Dialogs/RegionDialog/EllipticalRegionForm/EllipticalRegionForm.tsx
@@ -310,7 +310,7 @@ export class EllipticalRegionForm extends React.Component<{ region: RegionStore,
                             <td><span className="info-string">{infoString}</span></td>
                         </tr>
                         <tr>
-                            <td>Axes {pxUnitSpan}</td>
+                            <td>Semi-axes {pxUnitSpan}</td>
                             <td>{sizeWidthInput}</td>
                             <td>{sizeHeightInput}</td>
                             <td><span className="info-string">{sizeInfoString}</span></td>


Jordatious:
I like the grey "Semi-minor" text (and presumably "Semi-major" for the other field). But is "Semi-axes" a real term?

kswang1029:
> I like the grey "Semi-minor" text (and presumably "Semi-major" for the other field). But is "Semi-axes" a real term?

Would you recommend a sensible term instead?

Jordatious:
Well I'm not sure. Maybe it's a term, but I haven't heard it. Is radius/radii specific to circles rather than ellipses? I think if you have semi-major and semi-minor as the grey text you could maybe have "axes". But then the grey labels would be blank if when the region is already drawn. Could you also have it in the pop-up from when you hover over it, along with the format? Or is that too busy? Just basically thinking out loud here!

Jordatious:
Ok my mistake, I didn't realise "semi-major" and "semi-minor" were already there for v1.4. I think I've been immediately overwriting the values so never saw it. So perhaps that suggests we need something else, like adding it to the pop-up when you hover over it.

kswang1029:
https://www.merriam-webster.com/dictionary/semiaxis

Looks like “Semi-axis” is a term. 🤔

arusstaylor:
Semi-axis is actually a mathematic term for ellipses.  It degenerates to semi-diameter = radius for circles.

Russ

kswang1029:
> Semi-axis is actually a mathematic term for ellipses. It degenerates to semi-diameter = radius for circles. Russ
> […](#)

So I think we can use “Semi-axis” as the label then users should know the meaning of following boxes.

Jordatious:
Sounds good.

Jordatious (approved):

Issue #601, opened 2019-11-11
support colorbar in the image viewer


kswang1029:
For publication purpose, a colorbar is needed along the the raster image plot to indicate intensity color mapping.

Colorbar requirements:
- [ ] vertical at the right of the image plot (default)
- [ ] horizontal at the top of the image plot
- [ ] colorbar with value ticks and label (e.g., "Intensity (Jy/beam)")
- [ ] options to show (default) or not show a colorbar

veggiesaurus:
@lmarIT See, it's already an issue 😆 Scheduled for next release

veggiesaurus:
@TienHao could we use a chartjs chart for the ticks and numbers? I can write a WebGL shader to dynamically display the correct colours in real time

PR #1366, opened 2021-03-26
Yuhsuan/issue 601 colorbar

YuHsuan-Hwang:
This PR is for https://github.com/CARTAvis/carta-frontend/issues/601 adding colorbar in the image viewer.
Features:
* Added colorbar at the right side of the main image. Customize the style (ex: width, offset, color, …) from Image View Settings -> Colorbar.
* When moving the cursor on the colorbar, a bar is shown at the precise position of the cursor, and the exact value is shown at the right bottom corner. Turn on/off this feature from Image View Settings -> Colorbar -> Cursor info.
* The number of ticks changes with the size of the window.

Some details:
* Changed the range of gamma and alpha to [0.01, 2] and [0.01, 1000000].
* Force OverlayStore.renderWidth/Height to be >= 1 to avoid crashing.
* The setting of min=0 was not working in safeInputNumber. Fixed the incorrect logic.


veggiesaurus:
this is looking great. One thing: If there are no units defined, the label should be hidden automatically, otherwise there's just empty space

m51_xray fits-image-2021-03-26-12-22-25

YuHsuan-Hwang:
> this is looking great. One thing: If there are no units defined, the label should be hidden automatically, otherwise there's just empty space
> 
> m51_xray fits-image-2021-03-26-12-22-25

Ok. I will try to hide it.

YuHsuan-Hwang:
@veggiesaurus I make the space of the label hidden if the text is empty. It will also be hidden if users switch on (Image View Settings -> Colorbar ->) Label custom text but leave the label text column empty. This will ensure that there is no useless space in the window. Let me know if you prefer other design.

Jordatious:
Looking good! I get quotes for the beam unit / colorbar label for some images.

veggiesaurus:
My personal preference would be to keep the label hidden by default, and then just turn it on for publication plots, if required. @kswang1029 what do you think?

kswang1029:
> My personal preference would be to keep the label hidden by default, and then just turn it on for publication plots, if required. @kswang1029 what do you think?

I don't have strong opinions on this. There are pros and cons of having the label on/off by default. 
On: (pros) straightforward to have a presentation-ready plot for saving as png (cons) space issue
Off: (pros) more compact layout (cons) _may_ (not every user needs the label) need extra effort to enable the label for a presentation-ready plot

So given the above arguments, having it off by default seems okay (this sounds a good candidate as a preference). @Jordatious ?

YuHsuan-Hwang:
@Jordatious The quotes are also shown in other panels. This could be another issue.

YuHsuan-Hwang:
@veggiesaurus @kswang1029 I think setting the label hidden by default is Ok because the unit is shown when hovering on both the image and the colorbar.
I think users may be confused if they see nothing happens after turning on the label. My new idea is that we show the label text as "undefined" when the unit is _undefined_ to prevent this. We don't need to hide any extra space in this case.

kswang1029:
> @veggiesaurus @kswang1029 I think setting the label hidden by default is Ok because the unit is shown when hovering on both the image and the colorbar.
> I think users may be confused if they see nothing happens after turning on the label. My new idea is that we show the label text as "undefined" when the unit is _undefined_ to prevent this. We don't need to hide any extra space in this case.

good idea

veggiesaurus:
> @veggiesaurus @kswang1029 I think setting the label hidden by default is Ok because the unit is shown when hovering on both the image and the colorbar.
> I think users may be confused if they see nothing happens after turning on the label. My new idea is that we show the label text as "undefined" when the unit is _undefined_ to prevent this. We don't need to hide any extra space in this case.

I think "undefined" looks a bit clumsy, and doesn't really add any information. What about the more standard "arbitrary units"?

YuHsuan-Hwang:
@veggiesaurus Sure. We can change to that.

YuHsuan-Hwang:
@veggiesaurus @kswang1029 Changed to new design for undefined units. Also changed the order in the setting panel so that important params like ticks density and label visible can be easily found.

If there's no other big issue about the features, I'm changing this draft PR to normal PR.

kswang1029:
> @Jordatious The quotes are also shown in other panels. This could be another issue.

At the moment, CARTA adopts BUNIT string as the unit. If the string contains " then we will see it. We could improve this by pre-processing the BUNIT string if needed.

Jordatious:
I'd probably choose to have the label there by default, so sounds like a good switch to add to the preferences. I also prefer "arbitrary units". But we should add support for the user to change the label with custom text, which is pre-filled with the current label. This would be good for basically all the labels. I believe other visualisation software also allows for this.

kswang1029:
> I'd probably choose to have the label there by default, so sounds like a good switch to add to the preferences. I also prefer "arbitrary units". But we should add support for the user to change the label with custom text, which is pre-filled with the current label. This would be good for basically all the labels. I believe other visualisation software also allows for this.

Ok let’s try to have colorbar preferences later on. Custom label is implemented.

kswang1029 (approved):
Nice 👍

Jordatious:
How do I do a custom label? Is that also present for the other labels? I don't think I knew about this functionality.

kswang1029:
> How do I do a custom label? Is that also present for the other labels? I don't think I knew about this functionality.

A new tab "colorbar" is added to the image viewer settings. Via that we can enable colorbar label with an option to make a custom one. 
Screen Shot 2021-03-31 at 3 45 51 PM


ps. even the label in RA and Dec axes can be customized too via the "labels" tab.

Jordatious:
Ahhh yes! Very cool! Thanks! I couldn't find this menu for a bit. I think it has been moved from where it used to be.

It would be useful to even support latex format for the custom labels.

kswang1029:
> Ahhh yes! Very cool! Thanks! I couldn't find this menu for a bit. I think it has been moved from where it used to be.
> 
> It would be useful to even support latex format for the custom labels.

Would you like to file an issue in frontend repo to support latex label?

Jordatious:
Sure - https://github.com/CARTAvis/carta-frontend/issues/1385

veggiesaurus (approved):

PR #1369, opened 2021-03-29
add certain className for popover width adjustment

YuHsuan-Hwang:
This PR fixed https://github.com/CARTAvis/carta-frontend/issues/1368.
Used a certain className `header-search-input` instead of `bp3-popover` to avoid changing all the popover widths.

PR #1371, opened 2021-03-30
Ksw/scaling eqn png

YuHsuan-Hwang:
This PR is for updating the equation images for both light and dark themes.
image

The incorrect the popover width is fixed in another PR: https://github.com/CARTAvis/carta-frontend/issues/1369.


kswang1029 (approved):
👍

veggiesaurus:
This looks fine, but I'm wondering if there isn't a more sustainable solution here. What about using the original PNG and just inverting it with CSS?

YuHsuan-Hwang:
@veggiesaurus Ok. I will change it.

veggiesaurus:
> @veggiesaurus Ok. I will change it.

Well, give it a try and see if its straightforward 😊 I know CSS has some useful filters. I guess it depends whether we can keep the transparency layer uninverted

YuHsuan-Hwang:
@veggiesaurus Changed to CSS style solution. The images look the same as the above.
image

veggiesaurus:
> @veggiesaurus Changed to CSS style solution. The images look the same as the above.
> image

@YuHsuan-Hwang I know this is now a little bit of feature creep, but do you think we could try to also highlight the equations in a similar fashion to that of the text?

YuHsuan-Hwang:
@veggiesaurus Sure, I also thought about this. Let me try if `brightness` works.

YuHsuan-Hwang:
I have changed the color to white, but it's not significant. We may have to change the characters into bold.
image

kswang1029:
> I have changed the color to white, but it's not significant. We may have to change the characters into bold.
> image

Try drop-shadow?

YuHsuan-Hwang:
@kswang1029 I think drop-shadow works. Thanks.
image

veggiesaurus (approved):

PR #1384, opened 2021-03-31
add active style

YuHsuan-Hwang:
This PR extends PR https://github.com/CARTAvis/carta-frontend/pull/1371.
Added a style for the selected scaling for both light and dark mode:
image
image


kswang1029 (approved):

YuHsuan-Hwang:
Found that the characters may look blurry on some screens with drop-shadow. Also found that if we only use `brightness(2)`, the difference is significant on some screens. Removed drop-shadow to avoid blurry characters.

Issue #1356, opened 2021-03-18
Gamma spinbox is not rendered properly


kswang1029:
This issue seems hiding there for a while

![Screen Shot 2021-03-18 at 9 39 30 AM](https://user-images.githubusercontent.com/20819712/111560667-eed64f80-87cd-11eb-9232-493cb6e057c1.png)

Issue #597, opened 2019-11-06
support contrast and bias 2D widget in render config


low-sky:
[edit by KSW]
This is a feature request to support a 2D widget so that users can use it to adjust bias and contrast of the image easily. The selected scaling function with bias and contrast should be displayed on top of the histogram plot. 

The widget is something like
<img width="185" alt="Screen Shot 2021-03-30 at 11 41 19 AM" src="https://user-images.githubusercontent.com/20819712/112930449-e32f4500-914c-11eb-935e-adaf1ca09be5.png">
where by dragging the circle in the box, bias (in x direction) and contrast (in y direction) can be defined.

Mockup
![carta_renderConfig_biasContrast_mockup](https://user-images.githubusercontent.com/20819712/112931118-1c1be980-914e-11eb-97ef-7b8a7bd307c6.png)

[early messages]
Jayanne English (U. Manitoba) has requested a new transform / colour table reminiscent of the greyscale3 option in the kvis package.

[Here](https://nbviewer.jupyter.org/urls/dl.dropbox.com/s/goixi4li7oe0fot/Greyscale3.ipynb) is an implementation of the transform function.

The color map itself is just a linear greyscale (i.e., already implemented), but this feature request would presumably create a new transform called logistic with a two parameter function.  Jayanne specifically requests a 2D adjustment tool for controlling the parameters rather that 2 sliders or 2 text entries. 

There is an extra request for a color bar, which I'll open as a separate issue.

veggiesaurus:
@jott3077 has requested something similar in terms of a 2D control for adjusting contrast/bias. I believe the required control can be reused for "greyscale3" as well.

veggiesaurus:
I think we can put this control next to the histogram, and simply hide this control when the widget doesn't have space for it. That would prevent it from cluttering up the interface when users don't need it

kswang1029:
@veggiesaurus @low-sky So this issue is requesting the greyscale3 colormap, the entire render config of kvis, or just a new scaling (transfer) function that can be apply to all other colormaps?

greyscale3
![Screen Shot 2021-03-27 at 6 46 53 PM](https://user-images.githubusercontent.com/20819712/112718265-d8966500-8f2c-11eb-84fb-fbf98c2e3206.png)
kvis render config
[...]

PR #1390, opened 2021-04-07
Yuhsuan/issue 597 contrast bias

YuHsuan-Hwang:
This PR is for feature https://github.com/CARTAvis/carta-frontend/issues/597. Also fixed issue https://github.com/CARTAvis/carta-frontend/issues/1356.
The color scaling line on the histogram can be turned on/off in the setting panel so that users can export images with or without the line.


YuHsuan-Hwang:
Implementing a new mode for smoothed bias and contrast by using Glynn Rogers function. Will reopen the PR after this part is completed.

YuHsuan-Hwang:
@veggiesaurus @kswang1029 @duidae 
A new mode for smoothed bias and contrast is added and set as default.

Some details:
* Switch the mode in the render configuration settings panels -> Colormap Scaling.
* Adjust bias and contrast by dragging the point or clicking in the rectangle.

Please check if everything works well in both two modes. Thanks.

src/utilities/math.ts
```diff
@@ -76,37 +76,81 @@ export function getPercentiles(histogram: CARTA.IHistogram, ranks: number[]): nu
...
 }
 
-export function scaleValue(x: number, scaling: FrameScaling, alpha: number = 1000, gamma: number = 1.5) {
+function errorFunction(x: number, c: number, x0: number) {
+    const y = Math.exp(c * (x - x0));
+    return y / (y + 1);
+}
+
+export function scaleValue(x: number, scaling: FrameScaling, alpha: number = 1000, gamma: number = 1.5, bias: number = 0, contrast: number = 1,
+    biasContrastMode: boolean = false, smoothedBias: number = 0, smoothedContrast: number = 0) {
```
veggiesaurus: can we not just use the same arguments (`bias` and `contrast` for both modes, but use `biasContrastMode` to determine which function to use?

src/utilities/math.ts
```diff
@@ -76,37 +76,81 @@ export function getPercentiles(histogram: CARTA.IHistogram, ranks: number[]): nu
...
+            break;
         case FrameScaling.GAMMA:
-            return Math.pow(x, gamma);
+            scaleValue = Math.pow(x, gamma);
+            break;
         default:
-            return x;
+            scaleValue = x;
+    }
+    if (biasContrastMode) {
```
veggiesaurus: since `biasContrastMode` is a boolean, I think it just be something like `useSmoothBiasContrast`

src/utilities/math.ts
```diff
@@ -76,37 +76,81 @@ export function getPercentiles(histogram: CARTA.IHistogram, ranks: number[]): nu
...
+        const offset = errorFunction(0, c, smoothedBias);
+        let denominator = errorFunction(1, c, smoothedBias) - offset;
+        if (denominator <= 0) {
+            denominator = 0.1;
+        }
+
+        scaleValue = (errorFunction(scaleValue, c, smoothedBias) - offset) / denominator;
+    } else {
+        scaleValue = clamp(scaleValue - bias, 0, 1);
+        scaleValue = clamp((scaleValue - 0.5) * contrast + 0.5, 0, 1);
```
veggiesaurus, line 133: if we use `(contrast + 1)`, wouldn't we be able to use the same range and defaults for both bias/contrast modes?

veggiesaurus: (or perhaps we could change the other definition to match this mode? :thinking: )

src/stores/RenderConfigStore.ts
```diff
@@ -277,6 +304,42 @@ export class RenderConfigStore {
...
+
+    @action setContrast = (contrast: number) => {
+        this.contrast = contrast;
+    };
+
+    @action resetContrast = () => {
+        this.contrast = 1;
+    };
+
+    @action setSmoothedBias = (smoothedBias: number) => {
```
veggiesaurus: again, I think we should just use the same bias/contrast variables here. When we switch modes, we can automatically reset them to the correct defaults

src/stores/RenderConfigStore.ts
```diff
@@ -56,11 +56,22 @@ export class RenderConfigStore {
     static readonly GAMMA_MAX = 2;
     static readonly ALPHA_MIN = 0.1;
     static readonly ALPHA_MAX = 1000000;
+    static readonly BIAS_MIN = -1;
```
veggiesaurus, line 60: it would be nice if we could just adjust our bias/contrast equations to a form such that all of these variables just have the same range of `[0, 1]`.

kswang1029:
@YuHsuan-Hwang would be good to add "double-click" (in the 2D box) as reset bias and contrast.

src/stores/RenderConfigStore.ts
kswang1029: For the 2D box, wonder if we can apply a transformation internally so that the default position of the circle is at the center of the box?

YuHsuan-Hwang: @veggiesaurus Sure, we can modify the functions for normal bias/contrast and combine the parameters. We may need to consider that the default value for normal bias and contrast will become 0.5 and 0.5, and this is not as straightforward as 0 and 1.

src/utilities/math.ts
YuHsuan-Hwang: Sure. Will change the naming.

src/stores/RenderConfigStore.ts
YuHsuan-Hwang: @kswang1029 not sure how we can transform the position to the center

YuHsuan-Hwang:
> @YuHsuan-Hwang would be good to add "double-click" (in the 2D box) as reset bias and contrast.

Sure, I will try to add the event.

YuHsuan-Hwang:
@veggiesaurus @kswang1029 A solution for combining the parameters and placing the default circle in the center of the box:
 - Set the ranges and default values of smoothed bias/contrast the same as normal bias/contrast.
 - Map smoothed bias from [-1, 1] to [0, 1] and apply Glynn Rogers function.
 - Separate smoothed contrast into two range: [0, 1] and [1, 2]. For range [1, 2], map the value from [1, 2] to [0, 1] and apply Glynn Rogers function. For range [0, 1], apply function `clamp((value - (1.0 - smoothedBias)) * contrast + (1.0 - smoothedBias), 0, 1)`.

Cons: When contrast is [0, 1], the function is not so straightforward.

YuHsuan-Hwang:
> @veggiesaurus @kswang1029 A solution for combining the parameters and placing the default circle in the center of the box:
> 
> * Set the ranges and default values of smoothed bias/contrast the same as normal bias/contrast.
> * Map smoothed bias from [-1, 1] to [0, 1] and apply Glynn Rogers function.
> * Separate smoothed contrast into two range: [0, 1] and [1, 2]. For range [1, 2], map the value from [1, 2] to [0, 1] and apply Glynn Rogers function. For range [0, 1], apply function `clamp((value - (1.0 - smoothedBias)) * contrast + (1.0 - smoothedBias), 0, 1)`.
> 
> Cons: When contrast is [0, 1], the function is not so straightforward.

Applied the new transformations to bias/contrast and added double click event in the 2D box.

src/utilities/math.ts
YuHsuan-Hwang: Combined the parameters.

src/utilities/math.ts
YuHsuan-Hwang: Combined the parameters.

src/stores/RenderConfigStore.ts
YuHsuan-Hwang: Combined the parameters.

veggiesaurus:
@kswang1029 don't you think the "smooth bias/contrast" setting should be in preferences, rather than the widget config? It's related to the render configuration, not the widget settings.

One minor issue: when switching the "show color scaling" setting on and off, it changes the graph zoom. This is rather confusing, because it only occurs when the graph's bounds are automated. I suspect this may confuse users

veggiesaurus:
@YuHsuan-Hwang the contrast/bias widget works nicely if you click and drag on the _middle_ of the circle, but it's jumpy if I click and drag on the edge of the point: 


https://user-images.githubusercontent.com/592504/115233271-f7e17600-a117-11eb-8346-df338ecb1d88.mp4


I think you just need to take the initial offset into account when moving the point.

YuHsuan-Hwang:
> @YuHsuan-Hwang the contrast/bias widget works nicely if you click and drag on the _middle_ of the circle, but it's jumpy if I click and drag on the edge of the point:
> 
>  Peek.2021-04-19.14-01.mp4 
> I think you just need to take the initial offset into account when moving the point.

I see. Will change this part.

YuHsuan-Hwang:
@kswang1029 @veggiesaurus @duidae Please check the new changes:
* Moved the smoothed bias/contrast setting to the preferences dialog. Removed the show scaling function setting.
* Show the full range of the histogram instead of zooming to Min/Max.
* Set the opacity of the scaling function line to 0.5.
* Fixed the issue of dragging on the edge of the point.

kswang1029 (approved):
Feature-wise this looks good. 👍

veggiesaurus:
@YuHsuan-Hwang we will need to adjust the preferences schema in order to accommodate your new setting, won't we?

veggiesaurus (approved):

PR #1391, opened 2021-04-07
Jason/issue 646 multi profile separate buttons

### LINKED ISSUE #646: ability to view multiple profiles in the spectral profile widget
sub task for epic #643 

When image overlay is implemented, we will need to be able to view multiple spectral profiles from different images in a single spectral profile widget _with units in x and y properly converted if they are not consistent._ 

Suggest to have a dropdown with a list of checkboxes for users to select which images are used for generating spectral profiles. 

related discussions from the science team https://docs.google.com/document/d/1lPG_PxX4LnmZ5a6w8CUxKgryvPad4J9NkA6Ud30j7CA/edit#heading=h.6a98bch91ewh

### LINKED ISSUE #1398: z profiler profile settings are not cleaned after closing a image in 1.4
bug 1 reproduce:
1. open a image with stokes, select stokes to Q
2. create a region, select stat to RMS
3. close this image 
4. reopen this image or other image, the previous settings (stat/stoke) remains
 
juvv9-3ke63

bug 2 reproduce:
1. open a image
2. select to the image instead of active, should be highlighted
3. open another image
4. the image field stay highlighted for Active option, which is incorrect
 
kvlk5-13paf (1)

duidae:
closes #646 & closes #1398


kswang1029:
Do you think multiple profiles from different images need backend changes or even ICD changes? Or maybe not as in v1.4 we can have multiple widgets with each one showing a profile from a different image cube.

duidae:
> Do you think multiple profiles from different images need backend changes or even ICD changes? Or maybe not as in v1.4 we can have multiple widgets with each one showing a profile from a different image cube.

I am not sure, will figure out soon.

duidae:
> Do you think multiple profiles from different images need backend changes or even ICD changes? Or maybe not as in v1.4 we can have multiple widgets with each one showing a profile from a different image cube.

I am not sure, will figure out soon.

duidae:
@kswang1029  can u please help test the behavior of spectral profiler in single line mode (i.e. Image/Region/Statistic/Stokes buttons are **not** pressed) are the same as before from commit 833f8f2? thanks.

kswang1029:
> @kswang1029 can u please help test the behavior of spectral profiler in single line mode (i.e. Image/Region/Statistic/Stokes buttons are **not** pressed) are the same as before from commit [833f8f2](https://github.com/CARTAvis/carta-frontend/commit/833f8f2a2d0961b2abf26fa1e2c38d2c6493861e)? thanks.

Overall it behaves correctly. The dropdown menus of image and region become unselected (no label displayed except a triangle) when the image or region is closed or deleted. The displayed profile is still correct (switch to "Active") but it is just the dropdown does not show the label. For Stokes, it was "Current" and now it is "z". I suggest we use "Current" or "Active" instead.

duidae:
> > @kswang1029 can u please help test the behavior of spectral profiler in single line mode (i.e. Image/Region/Statistic/Stokes buttons are **not** pressed) are the same as before from commit [833f8f2](https://github.com/CARTAvis/carta-frontend/commit/833f8f2a2d0961b2abf26fa1e2c38d2c6493861e)? thanks.
> 
> Overall it behaves correctly. The dropdown menus of image and region become unselected (no label displayed except a triangle) when the image or region is closed or deleted. The displayed profile is still correct (switch to "Active") but it is just the dropdown does not show the label. For Stokes, it was "Current" and now it is "z". I suggest we use "Current" or "Active" instead.

thanks.

duidae:
@kswang1029 want to confirm: when showing multiple profiles of **matched** images, the region should fix to **Cursor** since Region 1,2,3... are belonging to a certain frame and not available for other frames, right?

image

kswang1029:
> @kswang1029 want to confirm: when showing multiple profiles of **matched** images, the region should fix to **Cursor** since Region 1,2,3... are belonging to a certain frame and not available for other frames, right?
> 
> image

No, when there are matched images, regions are _shared_ among all the matched images. For example, image A and image B are matched with A as the reference. Then when we create a region C, it is shared in image A and B and the region reference image is A.

duidae:
> > @kswang1029 want to confirm: when showing multiple profiles of **matched** images, the region should fix to **Cursor** since Region 1,2,3... are belonging to a certain frame and not available for other frames, right?
> > image
> 
> No, when there are matched images, regions are _shared_ among all the matched images. For example, image A and image B are matched with A as the reference. Then when we create a region C, it is shared in image A and B and the region reference image is A.

ok, thanks.
BTW, @kswang1029 how to handle stokes for matched images?

kswang1029:
> > > @kswang1029 want to confirm: when showing multiple profiles of **matched** images, the region should fix to **Cursor** since Region 1,2,3... are belonging to a certain frame and not available for other frames, right?
> > > image
> > 
> > 
> > No, when there are matched images, regions are _shared_ among all the matched images. For example, image A and image B are matched with A as the reference. Then when we create a region C, it is shared in image A and B and the region reference image is A.
> 
> ok, thanks.
> BTW, @kswang1029 how to handle stokes for matched images?

> > > @kswang1029 want to confirm: when showing multiple profiles of **matched** images, the region should fix to **Cursor** since Region 1,2,3... are belonging to a certain frame and not available for other frames, right?
> > > image
> > 
> > 
> > No, when there are matched images, regions are _shared_ among all the matched images. For example, image A and image B are matched with A as the reference. Then when we create a region C, it is shared in image A and B and the region reference image is A.
> 
> ok, thanks.
> BTW, @kswang1029 how to handle stokes for matched images?

I suggest the following:
When it is in single-line mode, the content of the stokes dropdown depends on the image (same as before). 
When it is in multi-line mode, we show z, I, Q U, V regardless the matched images. The frontend just requests what’s been selected and let the backend to handle exceptions (eg requesting Qz from a cube without Stokes axis).

duidae:
@kswang1029 can u provide some test images for multiple profile of matched images? thanks.

duidae:
@veggiesaurus @kswang1029  multi profile should be ready, please have a look, thanks.

kswang1029:
@duidae Very well done! This new feature works nicely. Thank you for your hard work on this (and apologies for the late notification of the issue re-assignment). After a quick test I found the following for improvements:

- [x] 1. when the dropdown menu of the stokes is grey out, it is still clickable. Same for statistic

- [x] 2. in single profile mode, the "Overlay" toggle in the Smoothing tab of the settings dialogue has no function. 

- [x] 3. try to convert the x-axis to radio velocity when enable multi-profile from "Image". This fits the common use case better. 

- [x] 4. could you briefly explain how colors are assigned? With three images matched, when we switch to different images, we see the three profiles change colors with image="active". When showing multiple Stokes profiles, the situation is more critical as the colors are very similar thus harder to read (eg plotting Q U and V). The same issue exist for multiple regions too.
Screen Shot 2021-04-20 at 11 10 35 AM


- [x] 5. could you make the layout responsive? That is, if I resize my widget vertically, the profile plot also gets resized too. Maybe set the vertical length to relative (to the size of the widget)?

kswang1029:
@veggiesaurus @duidae As potentially what @TienHao is working on on profile fitting may conflict with this PR, let's try to get this merged first if possible.

duidae:
> * [ ]  4. could you briefly explain how colors are assigned? With three images matched, when we switch to different images, we see the three profiles change colors with image="active". When showing multiple Stokes profiles, the situation is more critical as the colors are very similar thus harder to read (eg plotting Q U and V). The same issue exist for multiple regions too.
> 
> Screen Shot 2021-04-20 at 11 10 35 AM

the logic of color assignment for multiple profiles is:
1. assign primary color to the first line: this is in order to be compatible with the primary color function in the saved layout, i.e. we allow users to save their preferred color in the layout, so once spectral widget only shows 1 line, it will choose the primary color.
2. assign colors for 2nd, 3rd... lines by their item index in the drop down, i.e.
image

kswang1029:
> > * [ ]   4. could you briefly explain how colors are assigned? With three images matched, when we switch to different images, we see the three profiles change colors with image="active". When showing multiple Stokes profiles, the situation is more critical as the colors are very similar thus harder to read (eg plotting Q U and V). The same issue exist for multiple regions too.
> > 
> > Screen Shot 2021-04-20 at 11 10 35 AM
> 
> the logic of color assignment for multiple profiles is:
> 
> 1. assign primary color to the first line: this is in order to be compatible with the primary color function in the saved layout, i.e. we allow users to save their preferred color in the layout, so once spectral widget only shows 1 line, it will choose the primary color.
> 2. assign colors for 2nd, 3rd... lines by their item index in the drop down, i.e.
>    image

can we follow this so that adjacent colors are easy to read and distinguishable?
Screen Shot 2021-04-20 at 1 48 25 PM

Alternatively, maybe we can shuffle the colors a bit based on the example from matplotlib to make adjacent colors distinguishable

duidae:
> Alternatively, maybe we can shuffle the colors a bit based on the example from matplotlib to make adjacent colors distinguishable

@kswang1029 cuz we are using a shared color list now, will the shuffled colors apply to other color picker(image viewer, spatial profiler...etc) or just for spectral profiler(create a new color list)?

kswang1029:
> > Alternatively, maybe we can shuffle the colors a bit based on the example from matplotlib to make adjacent colors distinguishable
> 
> @kswang1029 cuz we are using a shared color list now, will the shuffled colors apply to other color picker(image viewer, spatial profiler...etc) or just for spectral profiler(create a new color list)?

Let's make the shuffled color list shared to other components. Before this multi-profile feature, the order in the current color list is meaningless. So making the new shuffled list global should be safe and fine.

duidae:
@kswang1029  1~5 fixed, please test, thanks.

kswang1029:
@duidae regarding the color issue, I did some experiments and compared any two of the colors in order to maximize the contrast. Could you update the color list to the following accordingly? Basically, some colors are removed and the order is adjusted a little bit. Once the color list is mapped to the selection order, not the dropdown list order, this new color list should work properly in general cases. Please let me know if you see other problems due to the change. 

```
export const SWATCH_COLORS = [
    Colors.BLUE3,
    Colors.ORANGE3,
    Colors.GREEN3,
    Colors.RED3,
    Colors.VIOLET3,
    Colors.SEPIA3,
    Colors.INDIGO3,
    //Colors.ROSE3,
    Colors.GRAY3,
    Colors.LIME3,
    Colors.TURQUOISE3,
    //Colors.VERMILION3,
    Colors.FOREST3,
    //Colors.INDIGO3,
    Colors.GOLD3,
    Colors.COBALT3,
    Colors.LIGHT_GRAY3,
    Colors.DARK_GRAY3,
    Colors.WHITE,
    Colors.BLACK
];
```


What they see against the theme background is like (left: light, right: dark)
Screen Shot 2021-04-21 at 12 10 42 PM

kswang1029:
> @kswang1029 1~5 fixed, please test, thanks.

@duidae Thanks the fix looks good. 👍

One thing I spotted is if we append _one more_ image and match it, this new image is still plotted in frequency (as defined in header), not being converted to radio velocity first.

https://user-images.githubusercontent.com/20819712/115496207-bf6aa500-a29b-11eb-8ed4-5c1655a7e41d.mp4

 
Do you think we can globally set "radio velocity" as default for the spectral profiler, if this way is easier to fix the issue? if the conversion is not possible then we fall back to what is defined in header (eg frequency).

duidae:
@kswang1029  the color picking part should be ok, please check.

kswang1029:
> @kswang1029 the color picking part should be ok, please check.

Hmm I see repetition of colors when deselete and reselect. Do we re-use colors in round-robin way?

Screen Shot 2021-04-21 at 3 58 37 PM

duidae:
> > @kswang1029 the color picking part should be ok, please check.
> 
> Hmm I see repetition of colors when deselete and reselect. Do we re-use colors in round-robin way?
> 
> Screen Shot 2021-04-21 at 3 58 37 PM

yes, it's impossible to avoid repetition, we just have 17 colors.

kswang1029:
> > > @kswang1029 the color picking part should be ok, please check.
> > 
> > 
> > Hmm I see repetition of colors when deselete and reselect. Do we re-use colors in round-robin way?
> > Screen Shot 2021-04-21 at 3 58 37 PM
> 
> yes, it's impossible to avoid repetition, we just have 17 colors.

Maybe pre-mapping the color to the list order still works fine using the new set of color? I see the complexity now...... Premapping is not a good idea. If we map the selection order to 10 colors and apply round-robin would this work?

duidae:
> > > > @kswang1029 the color picking part should be ok, please check.
> > > 
> > > 
> > > Hmm I see repetition of colors when deselete and reselect. Do we re-use colors in round-robin way?
> > > Screen Shot 2021-04-21 at 3 58 37 PM
> > 
> > 
> > yes, it's impossible to avoid repetition, we just have 17 colors.
> 
> Maybe pre-mapping the color to the list order still works fine using the new set of color?

duidae:
> > > > @kswang1029 the color picking part should be ok, please check.
> > > 
> > > 
> > > Hmm I see repetition of colors when deselete and reselect. Do we re-use colors in round-robin way?
> > > Screen Shot 2021-04-21 at 3 58 37 PM
> > 
> > 
> > yes, it's impossible to avoid repetition, we just have 17 colors.
> 
> Maybe pre-mapping the color to the list order still works fine using the new set of color? I see the complexity now...... Premapping is not a good idea. If we map the selection order to 10 colors and apply round-robin would this work?

dont understand, please elaborate

duidae:
> > @kswang1029 1~5 fixed, please test, thanks.
> 
> @duidae Thanks the fix looks good. 
> 
> One thing I spotted is if we append _one more_ image and match it, this new image is still plotted in frequency (as defined in header), not being converted to radio velocity first.
> 
>  Apr-21-2021.12-18-27.mp4 
> Do you think we can globally set "radio velocity" as default for the spectral profiler, if this way is easier to fix the issue? if the conversion is not possible then we fall back to what is defined in header (eg frequency).



> > @kswang1029 1~5 fixed, please test, thanks.
> 
> @duidae Thanks the fix looks good. 
> 
> One thing I spotted is if we append _one more_ image and match it, this new image is still plotted in frequency (as defined in header), not being converted to radio velocity first.
> 
>  Apr-21-2021.12-18-27.mp4 
> Do you think we can globally set "radio velocity" as default for the spectral profiler, if this way is easier to fix the issue? if the conversion is not possible then we fall back to what is defined in header (eg frequency).

fixed, sync spectral settings of newly matched image to spectral reference.

kswang1029:
> > > > > @kswang1029 the color picking part should be ok, please check.
> > > > 
> > > > 
> > > > Hmm I see repetition of colors when deselete and reselect. Do we re-use colors in round-robin way?
> > > > Screen Shot 2021-04-21 at 3 58 37 PM
> > > 
> > > 
> > > yes, it's impossible to avoid repetition, we just have 17 colors.
> > 
> > 
> > Maybe pre-mapping the color to the list order still works fine using the new set of color? I see the complexity now...... Premapping is not a good idea. If we map the selection order to 10 colors and apply round-robin would this work?
> 
> dont understand, please elaborate

The idea is to form a queue with 10 colors. When we plot a profile, a color is used and removed from the queue. When a profile is removed, the assigned color is put back to the head of the queue. Would this work? (Sorry this seems not round-robin in the original idea)

duidae:
> > > > > > @kswang1029 the color picking part should be ok, please check.
> > > > > 
> > > > > 
> > > > > Hmm I see repetition of colors when deselete and reselect. Do we re-use colors in round-robin way?
> > > > > Screen Shot 2021-04-21 at 3 58 37 PM
> > > > 
> > > > 
> > > > yes, it's impossible to avoid repetition, we just have 17 colors.
> > > 
> > > 
> > > Maybe pre-mapping the color to the list order still works fine using the new set of color? I see the complexity now...... Premapping is not a good idea. If we map the selection order to 10 colors and apply round-robin would this work?
> > 
> > 
> > dont understand, please elaborate
> 
> The idea is to form a queue with 10 colors. When we plot a profile, a color is used and removed from the queue. When a profile is removed, the assigned color is put back to the head of the queue. Would this work? (Sorry this seems not round-robin in the original idea)

I do think ensuring the function correctness should be the most crucial issue at this moment, if we really want some fancy things, it's better planned ahead or file an improvement issue in the future.

kswang1029:
> > > > > > > @kswang1029 the color picking part should be ok, please check.
> > > > > > 
> > > > > > 
> > > > > > Hmm I see repetition of colors when deselete and reselect. Do we re-use colors in round-robin way?
> > > > > > Screen Shot 2021-04-21 at 3 58 37 PM
> > > > > 
> > > > > 
> > > > > yes, it's impossible to avoid repetition, we just have 17 colors.
> > > > 
> > > > 
> > > > Maybe pre-mapping the color to the list order still works fine using the new set of color? I see the complexity now...... Premapping is not a good idea. If we map the selection order to 10 colors and apply round-robin would this work?
> > > 
> > > 
> > > dont understand, please elaborate
> > 
> > 
> > The idea is to form a queue with 10 colors. When we plot a profile, a color is used and removed from the queue. When a profile is removed, the assigned color is put back to the head of the queue. Would this work? (Sorry this seems not round-robin in the original idea)
> 
> I do think ensuring the function correctness should be the most crucial issue at this moment, if we really want some fancy things, it's better planned ahead or file an improvement issue in the future.

This is what I see when color repetition happens. At least the current logic should be refined to avoid color repetition. 
https://user-images.githubusercontent.com/20819712/115535319-366b6200-a2cb-11eb-80ea-5288f6e403b5.mp4

kswang1029:
> > > @kswang1029 1~5 fixed, please test, thanks.
> > 
> > 
> > @duidae Thanks the fix looks good.
> > One thing I spotted is if we append _one more_ image and match it, this new image is still plotted in frequency (as defined in header), not being converted to radio velocity first.
> > Apr-21-2021.12-18-27.mp4
> > Do you think we can globally set "radio velocity" as default for the spectral profiler, if this way is easier to fix the issue? if the conversion is not possible then we fall back to what is defined in header (eg frequency).
> 
> > > @kswang1029 1~5 fixed, please test, thanks.
> > 
> > 
> > @duidae Thanks the fix looks good.
> > One thing I spotted is if we append _one more_ image and match it, this new image is still plotted in frequency (as defined in header), not being converted to radio velocity first.
> > Apr-21-2021.12-18-27.mp4
> > Do you think we can globally set "radio velocity" as default for the spectral profiler, if this way is easier to fix the issue? if the conversion is not possible then we fall back to what is defined in header (eg frequency).
> 
> fixed, sync spectral settings of newly matched image to spectral reference.

👍 thanks.

duidae:
> This is what I see when color repetition happens. At least the current logic should be refined to avoid color repetition.
> https://user-images.githubusercontent.com/20819712/115535319-366b6200-a2cb-11eb-80ea-5288f6e403b5.mp4

changed back to assign color using drop down item index

kswang1029:
The concatenation of profile stream progress needs to be refined. In the following example, we expect to have two profiles being updated. However, initially the progress is >100%. Once it updates to 200%, there will be another progress showing up for the second profile. 

https://user-images.githubusercontent.com/20819712/116355761-e5a8bb80-a82c-11eb-91cf-f7daa666e135.mp4

kswang1029:
> The concatenation of profile stream progress needs to be refined. In the following example, we expect to have two profiles being updated. However, initially the progress is >100%. Once it updates to 200%, there will be another progress showing up for the second profile.
> 
>  Apr-28-2021.14-17-08.mp4

The concatenated progress does work as expected for the first time. The subsequent ones appear not. 

https://user-images.githubusercontent.com/20819712/116357306-ee9a8c80-a82e-11eb-940c-b3eb42242b43.mp4

duidae:
fixed to clean previous profiles.

duidae:
> fixed to clean previous profiles.


https://user-images.githubusercontent.com/25971696/116524178-ab611c00-a909-11eb-9041-478b437a41c8.mp4

yesterday's fix tried to clean up the profile data when region is updated and creates flashy effect, thanks to @YuHsuan-Hwang for spotting this issue, 

and the root cause is that we used to not clean up profile data and backend sends different spectral profile data sequentially:

profile A: progress 0, 0.1, 0.2, .... 1 -> profile B: progress 0, 0.1, 0.2, .... 1

rather than in parallel after the first time(first time they are sent in parallel), so wondering if we can send the profile in the sequence of:

profile A: progress 0 
profile B: progress 0
profile A: progress 0.1, 0.2, .... 1
profile B: progress 0.1, 0.2, .... 1

so that we can update the profile data to empty in the beginning to avoid showing old data when showing multiple profiles,
@veggiesaurus @markccchiang, do you think it's feasible? thanks.

kswang1029:
if the cursor movement is less than one pixel, we will see an empty plot. 
https://user-images.githubusercontent.com/20819712/116533554-17e11880-a914-11eb-97e9-96c4a1faba6c.mp4

(can consider to revert the commit and mark this as a known minor issue for beta/release)

kswang1029:
> > fixed to clean previous profiles.
> 
>  Kazam_screencast_00002.mp4 
> 
> yesterday's fix tried to clean up the profile data when region is updated and creates flashy effect, thanks to @YuHsuan-Hwang for spotting this issue,
> 
> and the root cause is that we used to not clean up profile data and backend sends different spectral profile data sequentially:
> 
> profile A: progress 0, 0.1, 0.2, .... 1 -> profile B: progress 0, 0.1, 0.2, .... 1
> 
> rather than in parallel after the first time(first time they are sent in parallel), so wondering if we can send the profile in the sequence of:
> 
> profile A: progress 0
> profile B: progress 0
> profile A: progress 0.1, 0.2, .... 1
> profile B: progress 0.1, 0.2, .... 1
> 
> so that we can update the profile data to empty in the beginning to avoid showing old data when showing multiple profiles,
> @veggiesaurus @markccchiang, do you think it's feasible? thanks.

the overhead might be larger? and the empty plot could still happen I guess.

duidae:
> > > fixed to clean previous profiles.
> > 
> > 
> > Kazam_screencast_00002.mp4
> > yesterday's fix tried to clean up the profile data when region is updated and creates flashy effect, thanks to @YuHsuan-Hwang for spotting this issue,
> > and the root cause is that we used to not clean up profile data and backend sends different spectral profile data sequentially:
> > profile A: progress 0, 0.1, 0.2, .... 1 -> profile B: progress 0, 0.1, 0.2, .... 1
> > rather than in parallel after the first time(first time they are sent in parallel), so wondering if we can send the profile in the sequence of:
> > profile A: progress 0
> > profile B: progress 0
> > profile A: progress 0.1, 0.2, .... 1
> > profile B: progress 0.1, 0.2, .... 1
> > so that we can update the profile data to empty in the beginning to avoid showing old data when showing multiple profiles,
> > @veggiesaurus @markccchiang, do you think it's feasible? thanks.
> 
> the overhead might be larger? and the empty plot could still happen I guess.

cleaning makes the largest overhead I think, and empty slot also comes from cleaning.

kswang1029:
> > > > fixed to clean previous profiles.
> > > 
> > > 
> > > Kazam_screencast_00002.mp4
> > > yesterday's fix tried to clean up the profile data when region is updated and creates flashy effect, thanks to @YuHsuan-Hwang for spotting this issue,
> > > and the root cause is that we used to not clean up profile data and backend sends different spectral profile data sequentially:
> > > profile A: progress 0, 0.1, 0.2, .... 1 -> profile B: progress 0, 0.1, 0.2, .... 1
> > > rather than in parallel after the first time(first time they are sent in parallel), so wondering if we can send the profile in the sequence of:
> > > profile A: progress 0
> > > profile B: progress 0
> > > profile A: progress 0.1, 0.2, .... 1
> > > profile B: progress 0.1, 0.2, .... 1
> > > so that we can update the profile data to empty in the beginning to avoid showing old data when showing multiple profiles,
> > > @veggiesaurus @markccchiang, do you think it's feasible? thanks.
> > 
> > 
> > the overhead might be larger? and the empty plot could still happen I guess.
> 
> cleaning makes the largest overhead I think, and empty slot also comes from cleaning.

Ok to just mark this as a known issue. This has very little impact to users so we can have more time to work this out.

veggiesaurus:
> so that we can update the profile data to empty in the beginning to avoid showing old data when showing multiple profiles,
> @veggiesaurus @markccchiang, do you think it's feasible? thanks.

I think we'd need to make quite a few changes in order to do that. Spectral profile requests are handled sequentially by design.

I'm wondering why they're sent in parallel at first though?

duidae:
> > so that we can update the profile data to empty in the beginning to avoid showing old data when showing multiple profiles,
> > @veggiesaurus @markccchiang, do you think it's feasible? thanks.
> 
> I think we'd need to make quite a few changes in order to do that. Spectral profile requests are handled sequentially by design.
> 
> I'm wondering why they're sent in parallel at first though?

guess it's because we are requesting the spectral profiles of different files, so that can be processed in parallel, but somehow the mechanism changes for later requests.
ok thanks for the explanation, will change back to previous implementation.

markccchiang:
> guess it's because we are requesting the spectral profiles of different files, so that can be processed in parallel, but somehow the mechanism changes for later requests.
> ok thanks for the explanation, will change back to previous implementation.

Looks like the frontend sends two `SET_SPECTRAL_REQUIREMENTS` commands for two different images the first time, so the backend processes in parallel. After that when a region states changed, the backend processes the region spectral profiles in sequential.

kswang1029:
Please disable "show mean/rms" when multi-profile mode is enabled. Would be good to add a tooltip to explain why. 
Screen Shot 2021-05-03 at 1 49 24 PM

duidae:
> Please disable "show mean/rms" when multi-profile mode is enabled. Would be good to add a tooltip to explain why.
> Screen Shot 2021-05-03 at 1 49 24 PM

when in multiple profile mode, there could still be only one profile showed, so mean/rms is still preserved & visible in this situation, wouldn't this be better?

kswang1029:
> > Please disable "show mean/rms" when multi-profile mode is enabled. Would be good to add a tooltip to explain why.
> > Screen Shot 2021-05-03 at 1 49 24 PM
> 
> when in multiple profile mode, there could still be only one profile showed, so mean/rms is still preserved & visible in this situation, wouldn't this be better?

good point. Could the condition be "if n_profile >1, disable show mean/rms"?

duidae:
> > > Please disable "show mean/rms" when multi-profile mode is enabled. Would be good to add a tooltip to explain why.
> > > Screen Shot 2021-05-03 at 1 49 24 PM
> > 
> > 
> > when in multiple profile mode, there could still be only one profile showed, so mean/rms is still preserved & visible in this situation, wouldn't this be better?
> 
> good point. Could the condition be "if n_profile >1, disable show mean/rms"?

it's already implemented, when profiles > 1, spectral profiler just wont show mean/rms

kswang1029:
> > > > Please disable "show mean/rms" when multi-profile mode is enabled. Would be good to add a tooltip to explain why.
> > > > Screen Shot 2021-05-03 at 1 49 24 PM
> > > 
> > > 
> > > when in multiple profile mode, there could still be only one profile showed, so mean/rms is still preserved & visible in this situation, wouldn't this be better?
> > 
> > 
> > good point. Could the condition be "if n_profile >1, disable show mean/rms"?
> 
> it's already implemented, when profiles > 1, spectral profiler just wont show mean/rms

uh, what I meant is the toggle better be disabled when n_profile > 1.

duidae:
> > > > > Please disable "show mean/rms" when multi-profile mode is enabled. Would be good to add a tooltip to explain why.
> > > > > Screen Shot 2021-05-03 at 1 49 24 PM
> > > > 
> > > > 
> > > > when in multiple profile mode, there could still be only one profile showed, so mean/rms is still preserved & visible in this situation, wouldn't this be better?
> > > 
> > > 
> > > good point. Could the condition be "if n_profile >1, disable show mean/rms"?
> > 
> > 
> > it's already implemented, when profiles > 1, spectral profiler just wont show mean/rms
> 
> uh, what I meant is the toggle better be disabled when n_profile > 1.

consider user is checking single profile with mean/rms on, and switch to multiple profiles and want to turn off mean/rms but it's disabled and we dont let him turn that off, seems a little bit redundant?

veggiesaurus:
> consider user is checking single profile with mean/rms on, and switch to multiple profiles and want to turn off mean/rms but it's disabled and we dont let him turn that off, seems a little bit redundant?

Yeah, I agree. The current behaviour seems good to me.

duidae:
> > consider user is checking single profile with mean/rms on, and switch to multiple profiles and want to turn off mean/rms but it's disabled and we dont let him turn that off, seems a little bit redundant?
> 
> Yeah, I agree. The current behaviour seems good to me.

I think the ambiguity comes from users may not know when mean/rms will be visible for the first time, added helper text to that for clarity.

image

kswang1029 (approved):
This PR is one big milestone on the way to replace casaviewer. Excellent work. 👍 

Minor issues will be filed as new issues for post 2.0.

veggiesaurus (approved):

Issue #1363, opened 2021-03-23
displaying world coordinate of a pv image in the cursor info bar


kswang1029:
After https://github.com/CARTAvis/carta-frontend/pull/1344 is merged, we will need to support displaying "world" coordinates of a pv image in the cursor info bar.

Issue #1360, opened 2021-03-20
Support labeling and cursor info for uv image 


kswang1029:
the relevant backend issue is https://github.com/CARTAvis/carta-backend/issues/663. 

The x and y labels should just display what CTYPE1 and CTYPE2 are. The coordinate grid should just be a normal Cartesian coordinate centered at 0,0. Cursor info should display (u, v) coordinate too along with the pixel value and pixel coordinates.

From the image header we expect to see the following CTYPE1/2:
- UU, VV
- uu, vv

CUNIT1/2 are expected to be:
- lambda (ie wavelength)
- klambda (ie 1000x wavelength)
- m (ie meter)

The xy labels for the UV image should be displayed like: 
- UU (lambda), VV (lambda), ie CTYPE1 (CUNIT1), CTYPE2 (CUNIT2)

PR #1362, opened 2021-03-22
Angus/1151 rectangular pixels

veggiesaurus:
Closes #1151.

This allows users to view images that have a fixed size rectangular pixel (e.g. `CUNIT1 == CUNIT2` but `CDELT1 != CDELT2`), or an arbitrary pixel size (e.g. `CUINT1 != CUNIT2`), in which case the pixel size is determined by the size of the image and the size of the image view component.

Also closes #1400 and fixes the behaviour of `AST.onReady`'s promises.

PR #1394, opened 2021-04-08
Yuhsuan/issue 1360 uv label

YuHsuan-Hwang:
This PR is for issue https://github.com/CARTAvis/carta-frontend/issues/1360 and https://github.com/CARTAvis/carta-frontend/issues/1363. Added labels for uv images and also cursor coordinate info for both uv images and pv images.

### DIFF
--- a/src/stores/FrameStore.ts
+++ b/src/stores/FrameStore.ts
@@ -435,7 +435,17 @@ export class FrameStore {
     }
 
     @computed get positionAxis(): number {
-        return this.specialAxis(/offset|position|offset position/i);
+        if (this.frameInfo?.fileInfoExtended?.headerEntries) {
+            const entries = this.frameInfo.fileInfoExtended.headerEntries;
+            const axis1 = entries.find(entry => entry.name.includes("CTYPE1"));
+            const axis2 = entries.find(entry => entry.name.includes("CTYPE2"))
+            if (axis1?.value?.match(/offset|position|offset position/i)) {
+                return 1;
+            } else if (axis2?.value?.match(/offset|position|offset position/i)) {
+                return 2;
+            }
+        }
+        return undefined;
     }
 
     @computed get isUVImage(): boolean {
@@ -443,20 +453,18 @@ export class FrameStore {
     }
 
     @computed get uvAxis(): number {
-        return this.specialAxis(/uu/i);
-    }
-
-    private specialAxis = (regex): number => {
         if (this.frameInfo?.fileInfoExtended?.headerEntries) {
             const entries = this.frameInfo.fileInfoExtended.headerEntries;
             const axis1 = entries.find(entry => entry.name.includes("CTYPE1"));
             const axis2 = entries.find(entry => entry.name.includes("CTYPE2"))
-            if (axis1?.value?.match(regex) || axis2?.value?.match(regex)) {
-                return axis1?.value?.match(regex) ? 1 : 2;
+            if (axis1?.value?.match(/uu/i)) {
+                return 1;
+            } else if (axis2?.value?.match(/uu/i)) {
+                return 2;
             }
         }
         return undefined;
-    };
+    }
 
     @computed get spectralAxis(): { valid: boolean; dimension: number, type: SpectralTypeSet, specsys: string } {
         if (this.frameInfo?.fileInfoExtended?.headerEntries) {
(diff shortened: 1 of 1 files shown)


src/stores/FrameStore.ts
```diff
@@ -759,11 +770,24 @@ export class FrameStore {
...
 
         if (this.isPVImage) {
-            this.astFrameSet = this.initPVImage();
+            this.astFrameSet = this.initFrame2D();
             if (this.astFrameSet) {
                 this.spectralFrame = AST.getSpectralFrame(this.astFrameSet);
                 this.wcsInfo = AST.copy(this.astFrameSet);
             }
+        } else if (this.isUVImage) {
+            this.astFrameSet = this.initFrame();
```
duidae: need to remind that this is not an elegant solution because there should be some redundancy between astFrameSet/astFrameSet2D, please add some comments to specify that this should be refactored in the future.

YuHsuan-Hwang: @duidae Added a comment for this part.

duidae: 3Q

src/stores/FrameStore.ts
```diff
@@ -435,17 +435,28 @@ export class FrameStore {
...
 
     @computed get positionAxis(): number {
+        return this.specialAxis(/offset|position|offset position/i);
+    }
+
+    @computed get isUVImage(): boolean {
+        return this.uvAxis !== undefined;
+    }
+
+    @computed get uvAxis(): number {
```
veggiesaurus, line 455: I don't think this should be a `@computed`, should it? since nothing here is an observable.

Same goes for the above two functions

src/stores/FrameStore.ts
```diff
@@ -435,17 +435,28 @@ export class FrameStore {
...
+
+    private specialAxis = (regex): number => {
         if (this.frameInfo?.fileInfoExtended?.headerEntries) {
             const entries = this.frameInfo.fileInfoExtended.headerEntries;
-            const positionAxisRegex = /offset|position|offset position/i;
             const axis1 = entries.find(entry => entry.name.includes("CTYPE1"));
             const axis2 = entries.find(entry => entry.name.includes("CTYPE2"))
-            if (axis1?.value?.match(positionAxisRegex) || axis2?.value?.match(positionAxisRegex)) {
-                return axis1?.value?.match(positionAxisRegex) ? 1 : 2;
+            if (axis1?.value?.match(regex) || axis2?.value?.match(regex)) {
```
veggiesaurus: personally I think this would be easier to understand as an `if (...) else if (...)` block.

src/stores/FrameStore.ts
```diff
@@ -759,11 +770,25 @@ export class FrameStore {
         this.animationChannelRange = [0, frameInfo.fileInfoExtended.depth - 1];
 
         if (this.isPVImage) {
-            this.astFrameSet = this.initPVImage();
+            this.astFrameSet = this.initFrame2D();
```
veggiesaurus, line 781: we don't use `this.astFrameSet` anywhere outside of the constructor. I'd previously refactored this and turned it into a local variable, but I think it got lost in a merge commit or something :thinking: @duidae do you see any reason to have `this.astFrameSet` and `this.wcsInfo`?

veggiesaurus: oh wait, ignore me... that's part of my PR  (#1362) :facepalm:

veggiesaurus (changes requested):

src/stores/FrameStore.ts
YuHsuan-Hwang: Refactored this part so that the values will be updated correctly.

src/stores/FrameStore.ts
YuHsuan-Hwang: Changed to an `if ... else if` block.

src/stores/FrameStore.ts
YuHsuan-Hwang: Also refactored this part to shorten the code.

src/stores/FrameStore.ts
veggiesaurus: Perhaps you misunderstood me. Having it as a get function is fine, but the '@computed' decorator is only useful if your get function reads observables that change

YuHsuan-Hwang: I thought the @computed decorator detects observables in functions, but then I realized that it does not and refactored this part. I think the @computed decorator is needed here.

veggiesaurus: to be dealt with in #1402

veggiesaurus (approved):
code looks fine to me. Will merge after @kswang1029's review

kswang1029 (approved):
Mostly it works well. One minor issue that can be dealt with LATER is the region dialogue where we are not able to define a region with "world coordinates" (uv and pv images).

veggiesaurus:
> Mostly it works well. One minor issue that can be dealt with LATER is the region dialogue where we are not able to define a region with "world coordinates" (uv and pv images).

@kswang1029 is there an existing issue for this?

Issue #1237, opened 2020-10-26
Catalog tool: Enhanced marker-based visualization


low-sky:
Allow for more catalog columns to be mapped to marker properties:

- Size of plotting symbol
- Colour of plotting symbol
- Orientation of plotting symbol

Build user interface to set transform between marker properties and catalog entries.

kswang1029:
@low-sky A question:
for size/color/orientation mapping, I suppose we will only accept numeric columns (not string columns) as input, right? Then we can set up normalization via the GUI. Do we expect to support mapping with string columns?

low-sky:
Yes, just numerical columns.  Sometimes you do want to do text annotation (e.g., source number) but just numerical initially.

veggiesaurus:
Size and color are easy enough, but orientation will require a switch from plotly to our own custom built WebGL plotting library

kswang1029:
if orientation is really required, the storypoint should be increased (currently 5).

low-sky:
As noted in discussion with @panchyo0, we will need some controls for mapping numerical values to ranges.  A good example of this is for colour where the already have some tools in the render configuration for the image browser.  We'll want some similar functionality for colour, namely ability to set min and max of ranges considered, colour table for the points, and ideally the transform between numerical value and the output numerical value (i.e., log, sqrt, linear, etc.).

For size, we'll want to be able to do similar mapping (min and max of range considered) but also some tools on the control of the transform (i.e., does the value map to radius or area of circle; log transform of data, etc.).

Jordatious:
It would be nice to increase this functionality and plot log axes, for example. Not sure why we need orientation of markers. I would have thought that's much less a priority than other basic plotting functionality.

low-sky:
Orientation of markers is useful for plotting things like source catalogues for objects with major, minor axes and position angles.

Jordatious:
Wouldn't that be less markers and more ellipses? Many source finders that write those properties to catalogues also write ds9 region files, which CARTA could import. I agree it'd be useful, but I'd prefer to see more basic plotting functionality added first.

panchyo0:
closed with PR #1401

Issue #814, opened 2020-03-04
WebGL rendering of catalog points


veggiesaurus:
Rather than using Konva, we should use WebGL to render catalog points.

We can do this by generating shapes and lines from the individual catalog points, and storing them in a vertex buffer, in a similar process to the contour generation.

More details are needed on what types of shapes need to be rendered, and whether the line thickness needs to be adjusted, or whether single-pixel lines can be used.

veggiesaurus:
My suggestion is to provide a simple list of symbol shapes, similar to that of DS9:
- Box
- Circle
- Cross
- X
- Hexagon
- Triangle

I think it's important to be able to edit the line width, especially for HiDPI screens. I also think it's important to be able to assign a different symbol shape, color and size to each point, based on a column or criteria

kswang1029:
we probably can apply some rotations to these symbol shapes like the attached image. For closed shape, we can set it as filled or not.
<img width="889" alt="Screen Shot 2020-03-04 at 6 28 10 PM" src="https://user-images.githubusercontent.com/20819712/75870663-42396200-5e46-11ea-96bf-d2a7ccf32280.png">

arusstaylor:
This seems a good comprehensive set of symbols.  I can't imaging a use case where we want use filled versions.   Colour and size and line thickness I think would be the attributes that we want to be able to set.   There should be a "highlight" mode for each when for example selecting an object in the catalogue table highlights the symbol for that object on the image.   Should probably be an action that draws the eye like blinking.

veggiesaurus:
> Should probably be an action that draws the eye like blinking.
Yes, this is a very useful feature of DS9, along with automatically centering on a point. 

I don't think we should aim for the reverse (clicking on a point on the image view to highlight a point in the catalog), because the catalog overlay could be tens of thousands of points, and interacting with that many points is tricky.

arusstaylor:
So is the issue capturing a click associated with a symbol in the image when the density of symbols is high?  I have been thinking that it would be good have something like a little popup that displayed metadata from the table for a symbol when selected.

veggiesaurus:
> So is the issue capturing a click associated with a symbol in the image when the density of symbols is high? I have been thinking that it would be good have something like a little popup that displayed metadata from the table for a symbol when selected.

There are a few issues:

1. This would require us to also render a hit graph for the catalogs. Not too hard to do, but will make the render process a bit more expensive
[...]

PR #1401, opened 2021-04-15
Qi/catalog 1237

panchyo0:
related issues #1237,  #814. performance highly depend on shape size
1. switch catalog overlay from plotly to webgl2, Due to the performance issue with 1M data points with different colours
2. add all shapes required in #814,  line segments and ellipse with blend and depth test. Apply shape outlines at fragment shader.
3. add size (major/ minor), color, orientation maps

update
4. enable webgl2.0 popup for Safari


src/components/CatalogOverlay/CatalogOverlayPlotSettingsPanelComponent/CatalogOverlayPlotSettingsPanelComponent.tsx
```diff
@@ -31,28 +33,50 @@ const IconWrapper = (path: React.SVGProps<SVGPathElement>, color: string, fill:
...
+        CatalogOverlayShape.CircleLined,
+        CatalogOverlayShape.CrossFilled,
+        CatalogOverlayShape.EllipseLined,
+        CatalogOverlayShape.HexagonLined,
+        CatalogOverlayShape.HexagonLined2,
+        CatalogOverlayShape.RhombLined,
+        CatalogOverlayShape.TriangleDownLined,
+        CatalogOverlayShape.TriangleUpLined,
+        CatalogOverlayShape.XFilled,
+        CatalogOverlayShape.LineSegmentFILLED
```
veggiesaurus: Why is this in caps?

src/services/CatalogWebGLService.ts
```diff
@@ -0,0 +1,194 @@
+import {createTextureFromArray, getShaderFromString, initWebGL2, loadImageTexture} from "utilities";
+import allMaps from "../static/allmaps.png";
+import vertexShaderLine from "!raw-loader!./GLSL/vertex_shader_catalog.glsl";
+import pixelShaderDashed from "!raw-loader!./GLSL/pixel_shader_catalog.glsl";
```
veggiesaurus: these names should be changed to something more appropriate

src/services/CatalogWebGLService.ts
```diff
@@ -0,0 +1,194 @@
...
+    PositionTexture: WebGLUniformLocation,
+    OrientationTexture: WebGLUniformLocation,
+    ColorTexture: WebGLUniformLocation,
+    SizeTexture: WebGLUniformLocation,
+    SelectedSourceTexture: WebGLUniformLocation,
+    SizeMinorTexture: WebGLUniformLocation,
+    // size
+    SmapEnabled: WebGLUniformLocation,
+    AreaMode: WebGLUniformLocation,
+    SminorMapEnabled: WebGLUniformLocation,
```
veggiesaurus: please rename to `SizeMinorMapEnabled`

src/services/GLSL/pixel_shader_catalog.glsl
```diff
@@ -0,0 +1,479 @@
...
+    float alpha = getAlphaValue(posPixelSpace, rMin, rMax);
+
+    // Blending
+    if (uCmapEnabled && !isNaN(v_colour)) {
+        float x = clamp(v_colour, 0.0, 1.0);
+        float cmapYVal = (float(uCmapIndex) + 0.5) / float(uNumCmaps);
+        vec2 cmapCoords = vec2(x, cmapYVal);
+        outColor = (1.0 - outline) * vec4(texture(uCmapTexture, cmapCoords).xyz, alpha) + outline * vec4(uSelectedSourceColor, alpha2);
+    } else {
+        // outColor = vec4((1.0 - outline) * uPointColor + outline * uSelectedSourceColor, alpha);
```
veggiesaurus: remove comment

src/services/GLSL/pixel_shader_catalog.glsl
```diff
@@ -0,0 +1,479 @@
...
+            gl_FragDepth = 0.0;
+        } 
+    }
+    float alpha = getAlphaValue(posPixelSpace, rMin, rMax);
+
+    // Blending
+    if (uCmapEnabled && !isNaN(v_colour)) {
+        float x = clamp(v_colour, 0.0, 1.0);
+        float cmapYVal = (float(uCmapIndex) + 0.5) / float(uNumCmaps);
+        vec2 cmapCoords = vec2(x, cmapYVal);
```
veggiesaurus: each point is the same color, so we should move the colormap texture lookup from the fragment shader to the vertex shader, and then pass it from the vertex to fragment shader as a `varying`. The end result of this should be a reduction in the number of texture lookups we need to do. I think we can also do a direct texel lookup instead of a `texture` sampling, in order to read out a specific pixel, rather than using texture coordinates to sample a pixel.

Note: We should probably do this with the contours as well. Sorry, I should have thought of that when I wrote that set of shaders!

veggiesaurus: I do wonder if it will actually make any appreciable performance difference, but in theory, we'd be doing far fewer lookups (assuming a point size of 5x5 pix^2, that would be 1 lookup per point, instead of 25), even if most of the lookups are cached.

src/services/GLSL/vertex_shader_catalog.glsl
```diff
@@ -0,0 +1,177 @@
...
+
+    v_colour = -1.0;
+    v_orientation = -1.0;
+    v_minorSize = -1.0;
+    v_selected = selectedSource.x;
+    v_pointSize = uPointSize;
+
+    if (uCmapEnabled) {
+        vec4 color = getValueByIndexFromTexture(uColorTexture, gl_VertexID);
+        v_colour = color.x;
```
veggiesaurus: rather calculate the actual color here, instead of just passing through the color _index_. If the color map is not enabled, you can just pass through the uniform color `uPointColor` instead.

src/services/GLSL/vertex_shader_catalog.glsl
```diff
@@ -0,0 +1,177 @@
...
+uniform bool uShowSelectedSource;
+uniform bool uSminorMapEnabled;
+uniform bool uAreaModeMinor;
+uniform bool uCmapEnabled;
+uniform bool uOmapEnabled;
+
+out float v_colour;
+out float v_pointSize;
+out float v_orientation;
+out float v_selected;
```
veggiesaurus, line 65: can't we actually just do the calculations in the vertex shader for this as well? you can just adjust `v_pointSize` depending on whether the point is selected, surely?

src/stores/CatalogStore.ts
```diff
@@ -49,86 +46,55 @@ export class CatalogStore {
...
-            this.catalogData.set(fileId,
-                {
-                    xImageCoords: catalogDataInfo.xImageCoords,
-                    yImageCoords: catalogDataInfo.yImageCoords,
-                    xSelectedCoords: xSelectedCoords,
-                    ySelectedCoords: ySelectedCoords,
-                    showSelectedData: catalogDataInfo.showSelectedData,
-                    displayed: catalogDataInfo.displayed
-                });
+            catalog.dataPoints = dataPoints;
```
veggiesaurus: if the `dataPoints` array is actually just for X and Y, then I think we should rather just call it `posPoints` or `positionPoints`

src/components/ImageView/CatalogView/CatalogViewGLComponent.tsx
```diff
@@ -0,0 +1,300 @@
...
+            const orientationTexture = this.catalogWebGLService.getDataTexture(fileId, CatalogTextureType.Orientation);
+            if (!catalogWidgetStore.disableOrientationMap && orientationTexture) {
+                this.gl.uniform1i(shaderUniforms.OmapEnabled, 1);
+                this.gl.activeTexture(GL2.TEXTURE4);
+                this.gl.bindTexture(GL2.TEXTURE_2D, orientationTexture);
+                this.gl.uniform1i(shaderUniforms.OrientationTexture, 4);
+            }
+
+            // selected source
+            const selectedSource = this.catalogWebGLService.getDataTexture(fileId, CatalogTextureType.SelectedSource);
```
veggiesaurus, line 213: I don't think we should be using a 32-bit float for this. We could just use an 8-bit texture for this instead, and that would make things much smaller.

src/stores/CatalogProfileStore.ts
```diff
@@ -352,25 +353,23 @@ export class CatalogProfileStore {
...
-        const coords = CatalogStore.Instance.catalogData.get(this.catalogFileId);
-        if (coords?.xImageCoords?.length) {
+        const coordsArray = CatalogStore.Instance.catalogGLData.get(this.catalogFileId);
+        if (coordsArray?.dataPoints?.length) {
             let selectedX = [];
             let selectedY = [];
-            let xArray = coords.xImageCoords;
-            let yArray = coords.yImageCoords;
-
+            const selectedData = new Float32Array(coordsArray.dataPoints.length / 2);
```
veggiesaurus: can use a `UInt8Array` for a large memory saving

src/stores/CatalogStore.ts
```diff
@@ -49,86 +46,55 @@ export class CatalogStore {
...
+        if (catalog) {
+            const dataSize = catalog.dataPoints.length;
+            let dataPoints = new Float32Array(dataSize + xData.length * 2);
+            dataPoints.set(catalog.dataPoints);
             switch (catalogFrame) {
                 case CatalogSystemType.Pixel0:
                     for (let i = 0; i < xData.length; i++) {
-                        catalogDataInfo.xImageCoords.push(xData[i]);
-                        catalogDataInfo.yImageCoords.push(yData[i]);
+                        dataPoints[dataSize + i * 2] = xData[i];
```
veggiesaurus: I wonder if it would make more sense to just have separate textures for the X and Y columns? Then we don't need to duplicate all this data.

src/utilities/templates.tsx
```diff
@@ -10,5 +10,15 @@ export const TemplateNodes = {
                 <small> If you are using CARTA through a browser, please ensure your browser supports WebGL. If you are using the CARTA standalone app, please ensure you are not running in a remote desktop. </small>
             </p>
         </React.Fragment>
+    ),
+    WebGL2ErrorMessage: (
+        <div>
+            <div>
+                <p>Could not load WebGL2.0. Catalog will not be displayed properly.</p>
```
veggiesaurus: "Catalog will not be displayed properly." -> "The catalog overlay feature will not be available".

wasm_src/carta_computation/carta_computation.cc
```diff
@@ -222,4 +235,83 @@ void generateVertexData(void* dst, size_t dstCapacity, float* srcVertices, int n
...
+
+void calculateCatalogSizeDiameter(float* data, size_t N, float min, float max, int sizeMin, int sizeMax, int scaling, int devicePixelRatio, float alpha, float gamma) {
+    float columnMin = scaleValue(min, scaling, alpha, gamma);
+    float columnMax = scaleValue(max, scaling, alpha, gamma);
+    float range = columnMax - columnMin;
+    for (size_t i = 0; i < N; i++)
+    {
+        float v = clamp(*(data + i), min, max);
+        float value = scaleValue(v, scaling, alpha, gamma);
+        *(data + i) = ((value - columnMin) / range * (sizeMax - sizeMin) + sizeMin) * devicePixelRatio;
```
veggiesaurus: `data[i]` would be neater

wasm_src/carta_computation/carta_computation.cc
```diff
@@ -222,4 +235,83 @@ void generateVertexData(void* dst, size_t dstCapacity, float* srcVertices, int n
...
+
+void calculateCatalogSizeArea(float* data, size_t N, float min, float max, int sizeMin, int sizeMax, int scaling, int devicePixelRatio, float alpha, float gamma) {
+    float columnMin = scaleValue(min, scaling, alpha, gamma);
+    float columnMax = scaleValue(max, scaling, alpha, gamma);
+    float range = columnMax - columnMin;
+    for (size_t i = 0; i < N; i++)
+    {
+        float v = clamp(*(data + i), min, max);
+        float value = scaleValue(v, scaling, alpha, gamma);
+        *(data + i) = (sqrt((value - columnMin) / range) * (sizeMax - sizeMin) + sizeMin) * devicePixelRatio;
```
veggiesaurus: same here: `data[i]`. No need for old-school pointer arithmetic in this case, is there?

wasm_src/carta_computation/carta_computation.cc
```diff
@@ -222,4 +235,83 @@ void generateVertexData(void* dst, size_t dstCapacity, float* srcVertices, int n
...
+        float value = (scaleValue(v, scaling, alpha, gamma) - columnMin) / range;
+        if (invert)
+        {
+            value = 1 - value;
+        }
+        *(data + i) = value;
+    }
+}
+
+void calculateCatalogOrientationMap(float* data, size_t N, float min, float max, float angleMin, float angleMax, int scaling, float alpha, float gamma) {
```
veggiesaurus: there's a lot of repetition in these functions. Can't we reuse one function?

wasm_src/carta_computation/carta_computation.cc
```diff
@@ -222,4 +235,83 @@ void generateVertexData(void* dst, size_t dstCapacity, float* srcVertices, int n
...
+    float range = columnMax - columnMin;
+    for (size_t i = 0; i < N; i++)
+    {
+        float v = clamp(*(data + i), min, max);
+        float value = scaleValue(v, scaling, alpha, gamma);
+        *(data + i) = (sqrt((value - columnMin) / range) * (sizeMax - sizeMin) + sizeMin) * devicePixelRatio;
+    }
+}
+
+void calculateCatalogColorMap(float* data, size_t dataSize, bool invert, float min, float max, int scaling, float alpha, float gamma) {
```
veggiesaurus: I'd rather we used `dataMin` and `dataMax`, to avoid confusion with `std::min/max`

wasm_src/carta_computation/post.ts
```diff
@@ -103,4 +107,48 @@ Module.GenerateVertexData = (sourceVertices: Float32Array, indexOffsets: Int32Ar
...
+    Module.HEAPF32.set(data, dataOnWasmHeap / bytes_per_element);
+
+    if (area) {
+        calculateCatalogSizeArea(dataOnWasmHeap, N, min, max, sizeMin, sizeMax, scaling, devicePixelRatio, alpha, gamma);
+    } else {
+        calculateCatalogSizeDiameter(dataOnWasmHeap, N, min, max, sizeMin, sizeMax, scaling, devicePixelRatio, alpha, gamma);
+    }
+
+    const float32 = new Float32Array(Module.HEAPF32.buffer, dataOnWasmHeap, N);
+    Module._free(dataOnWasmHeap);
```
veggiesaurus, line 120: technically, if we call `Module._free` on that pointer, then the `float32` array is now pointing to memory that can be freely used/overwritten

src/stores/widgets/CatalogWidgetStore.ts
```diff
@@ -64,22 +81,432 @@ export class CatalogWidgetStore {
...
+        this.sizeMinorColumnMinLocked = false;
+        this.sizeMinorColumnMaxLocked = false;
+
+        reaction(() => this.sizeMapData, (column) => {
+            const result = minMaxArray(column);
+            this.setSizeColumnMin(isFinite(result.minVal)? result.minVal : 0, "default");
+            this.setSizeColumnMax(isFinite(result.maxVal)? result.maxVal : 0, "default");
+        });
+
+        reaction(() => this.sizeArray(), (size) => {
```
veggiesaurus, line 172: I'm confused. How you can you react to a simple function that returns a `Float32Array`. Should `sizeArray` be a `@computed` function?

src/stores/widgets/CatalogWidgetStore.ts
```diff
@@ -64,22 +81,432 @@ export class CatalogWidgetStore {
...
+
+        reaction(() => this.sizeMinorMapData, (column) => {
+            const result = minMaxArray(column);
+            this.setSizeMinorColumnMin(isFinite(result.minVal)? result.minVal : 0, "default");
+            this.setSizeMinorColumnMax(isFinite(result.maxVal)? result.maxVal : 0, "default");
+        });
+
+        reaction(() => this.sizeMinorArray(), (size) => {
+            if (size.length) {
+                CatalogWebGLService.Instance.updateDataTexture(this.catalogFileId, size, CatalogTextureType.SizeMinor);
```
veggiesaurus, line 198: would it not be simpler to just copy the column data itself, and then do the scaling/clamping on the GPU in the vertex shader? that would mean changing any of the parameters would be as simple as just changing a `uniform` parameter, instead of an expensive re-calculation and re-upload

veggiesaurus (changes requested):
this is a really big PR, and the results are really excellent :rocket: , but I do think it needs some ironing out and optimisation.

src/stores/widgets/CatalogWidgetStore.ts
panchyo0: I think that because we have a `@computed` function inside `sizeArray`. with `@computed sizeArray`, we get a render bug when user pre set size map.

kswang1029:
When ellipse marker is selected, we can map major and minor axes to columns. Usually they will be mapped to size estimate from Gaussian fit so the two columns are related and have same scales and unit. So when we lock one clip max or min from either major or minor axis, we should lock _all_ parameters including clip min, clip max from major axis and clip min, clip max from minor axis at the same time so that we map column values to maker sizes with _same_ scaling. @low-sky would this make sense to you? Or do you prefer to have the freedom that users can lock the clip values independently? 

https://user-images.githubusercontent.com/20819712/116181932-0f3ee580-a74e-11eb-80a8-7e511ee81a18.mp4

Also when minor axis mapping is disabled, we should unlock clip min/max of the major axis if they were locked by the minor axis.

kswang1029:
It appears that when the minor axis of an ellipse is to small, we see blurry markers. This happens quite easily when we map major axis and minor axis of the ellipse marker to data columns. 

Screen Shot 2021-04-27 at 12 00 05 PM

veggiesaurus:
> It appears that when the minor axis of an ellipse is to small, we see blurry markers. This happens quite easily when we map major axis and minor axis of the ellipse marker to data columns. 

It looks like the line feathering is getting stretched out as well as the ellipse

low-sky:
Replying to above @kswang1029  -- I agree that if you end up locking one set of clips (e.g., major axis) between Major / Minor, the all the parameters should be locked to the same clip.   In fact, the default can be "locked."  However, I think it's useful to the corner cases of using orientation and ellipse parameters to represent non-spatial variables.

src/services/GLSL/vertex_shader_catalog.glsl
panchyo0: related issue #1426

src/stores/CatalogProfileStore.ts
panchyo0: the range for UInt8Array 0 to 255, should we use `Unit32Array`?

veggiesaurus: Am I missing something? Aren't you just storing a 0 or a 1 depending on whether it's selected?

panchyo0: yes, I just confuse myself, thanks

src/stores/widgets/CatalogWidgetStore.ts
panchyo0: related issue #1426

wasm_src/carta_computation/post.ts
veggiesaurus: this is ok for now, but please keep this in mind down the line. If we intersperse any calls to `CartaComputation` and our use of these results, undefined behaviour would ensue. The safe way to do this is to `slice()` the array, in order to move it from WASM back to javascript

panchyo0: ah, Thanks for point it out.

kswang1029:
Here is a minor layout issue ("clip max")
Screen Shot 2021-05-03 at 11 39 53 AM

kswang1029:
Another minor UI issue: when we switch to a marker other than "ellipse", the minor tab is still activated.
https://user-images.githubusercontent.com/20819712/116839418-ec587980-ac04-11eb-8342-eb4374626c58.mp4

kswang1029:
UI issue about the file dropdown when the file name of the catalog is long.

Screen Shot 2021-05-03 at 11 56 15 AM

kswang1029:
UX improvement: Size Min/Size Max as spinbox (cf Size or Thickness)
Screen Shot 2021-05-03 at 12 18 49 PM

kswang1029:
Minor issue: inconsistent layer order as seen from the viewer (left) and png export (right).
Screen Shot 2021-05-03 at 12 28 11 PM

in the viewer, catalog layer is _above_ the grid layer. 
Screen Shot 2021-05-03 at 12 28 22 PM

Either way is fine I think. We will just need to make sure they are consistent.

veggiesaurus:
> Either way is fine I think. We will just need to make sure they are consistent.

I suggest they go above the grid and below the regions

veggiesaurus (approved):

panchyo0:
@kswang1029  I fixed issue #1427 and UI minor issues. please check again. for UI improvement, we should do it in a separate PR and trade it carefully with performance test, since the button behave is a little different. for example, line thickness changed right away as the value changed and size min/max doesn't.

kswang1029 (approved):
Nice PR with powerful features and good performance. 👍

Issue #970, opened 2020-06-30
safari specific: spatial profiler not resized properly


kswang1029:
This issue appears to happen ONLY with Safari browser (Version 13.1.1 (15609.2.9.1.2)), not with Chrome or Firefox... 

![Jun-30-2020 11-02-10](https://user-images.githubusercontent.com/20819712/86078703-3e9aac80-bac1-11ea-8c5b-b993544c35aa.gif)

kswang1029:
Tested with Version 14.0.2 (15610.3.7.1.10, 15610) and the issue remains.

kswang1029:
tested with Version 14.0.3 (15610.4.3.1.7, 15610) and the issue remains (as of 20210421 dev)

kswang1029:
Version 15.3 (17612.4.9.1.8) is free of this issue. Close for now.

Issue #1404, opened 2021-04-16
frontend broken with Safari 13


panchyo0:
<img width="1431" alt="Screen Shot 2021-04-15 at 8 34 57 PM" src="https://user-images.githubusercontent.com/22226157/114963457-0e42b580-9e2a-11eb-8afb-4be36f376878.png">

related issue https://github.com/maslianok/react-resize-detector/issues/141

veggiesaurus:
@panchyo0 can you see if just updating that module fixes this?

veggiesaurus:
workaround: upgrade to safari 13.1

PR #1409, opened 2021-04-19
issue 1404

panchyo0:
related issue #1404
`Starting from version 6 the library uses ResizeObserver without any polyfills since it's supported by all major browsers.` 

we need to replace `import ReactResizeDetector from "react-resize-detector"` by `import ResizeObserver from "react-resize-detector/build/withPolyfill"`

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 
 import {observer} from "mobx-react";
-import ReactResizeDetector from "react-resize-detector";
+import ResizeObserver from "react-resize-detector/build/withPolyfill";
 import {Alert, Intent} from "@blueprintjs/core";
 import {UIControllerComponent, FloatingWidgetManagerComponent} from "./components";
 import {TaskProgressDialogComponent} from "./components/Dialogs";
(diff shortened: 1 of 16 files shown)


veggiesaurus:
safari 13.1 [supports](https://caniuse.com/resizeobserver) resize observer. Is there any reason to continue supporting 13.0 and below?

panchyo0:
it depends on users computer. I have macOS 10.15.3 with Safari 13.0, it's broken. Or we can make it clear to user, it they reach this issue, just update the safari version.

veggiesaurus:
And you can't update? I'm not sure how far back we should go. 13.1 has been out for over a year

panchyo0:
good point!

veggiesaurus:
@kswang1029 what do you think? should we continue to support Safari 13.0?

kswang1029:
> @kswang1029 what do you think? should we continue to support Safari 13.0?

I suggest we move forward so no

veggiesaurus:
> > @kswang1029 what do you think? should we continue to support Safari 13.0?
> 
> I suggest we move forward so no

@kswang1029 I don't suppose this somehow magically fixes #970 ?

kswang1029:
> > > @kswang1029 what do you think? should we continue to support Safari 13.0?
> > 
> > 
> > I suggest we move forward so no
> 
> @kswang1029 I don't suppose this somehow magically fixes #970 ?

tested with Safari and no luck there.

PR #1414, opened 2021-04-21
Angus/1214 manual reconnection

### LINKED ISSUE #1214: Google sign in non-authorized user getting past the dashboard to continuously fail authenticating at the frontend
When using Google Sign in, a user successfully authenticates with their Institutional Google account, but they are not listed in the `usertable.txt`. They are probably unaware that they are not authorized to access it because it accepted the Google login and began to load the frontend. They might assume something is wrong with the service, and perhaps refreshed their page(?, but ultimately left their browser tab open. Now the node.js log is continuously reporting a message in the following form, every 5 seconds:

> carta-node-server > WS upgrade request from 333.33.3.333 for authenticated user xxx
> carta-node-server > Could not find username xxx in the user map 

That will be a lot of unnecessary messages.

Expected behavior would be for the user to immediately get redirected to the dashboard, with no way to access the frontend, so the messages stop.

veggiesaurus:
Closes #1214 by implementing a manual reconnection strategy: 

- Initially, the frontend will attempt to connect to the backend up to 15 times (with a one second delay between connection attempts).
- If this fails, the user will be required to manually click the "Retry" button to reconnect
- If the connection drops later on, the frontend will not attempt to automatically reconnect. The user will have to manually click "Retry".


kswang1029:
@veggiesaurus tested with a local build and it works properly. I don't have the environment to test the case with an auth system (eg controller).🤔

veggiesaurus:
I've tested it with the latest controller package, seems to work ok. Perhaps @ajm-asiaa could test it though?

ajm-ska:
I put this frontend on our "cartaviewer" server. I think it works. When I manually killed the backend process in a terminal I see the "Retry" notification window.
Screenshot 2021-04-23 at 10 53 44

kswang1029:
> I put this frontend on our "cartaviewer" server. I think it works. When I manually killed the backend process in a terminal I see the "Retry" notification window.
> Screenshot 2021-04-23 at 10 53 44

what's the behavior with auth? When trying to reconnect, no auth is required as long as it is not expired?

ajm-ska:
> > I put this frontend on our "cartaviewer" server. I think it works. When I manually killed the backend process in a terminal I see the "Retry" notification window.
> > Screenshot 2021-04-23 at 10 53 44
> 
> what's the behavior with auth? When trying to reconnect, no auth is required as long as it is not expired?

How do you test that? I didn't need to log in again if that is what you mean? But I just clicked Retry straight away. How long does it take for the auth to expire?

kswang1029:
> > > I put this frontend on our "cartaviewer" server. I think it works. When I manually killed the backend process in a terminal I see the "Retry" notification window.
> > > Screenshot 2021-04-23 at 10 53 44
> > 
> > 
> > what's the behavior with auth? When trying to reconnect, no auth is required as long as it is not expired?
> 
> How do you test that? I didn't need to log in again if that is what you mean? But I just clicked Retry straight away. How long does it take for the auth to expire?

My question is, if the previous authentication expired (eg google or LDAP) and the user hits the retry button, what would happen? I am not sure how to test this... Maybe you need another CARTA tab and logout manually via the dashboard. Then go back to the original tab and hit retry.

veggiesaurus:
> My question is, if the previous authentication expired (eg google or LDAP) and the user hits the retry button, what would happen? I am not sure how to test this... Maybe you need another CARTA tab and logout manually via the dashboard. Then go back to the original tab and hit retry.

the reconnection will fail if they are no longer authenticated. If they re-authenticate, it will succeed.

src/App.tsx
```diff
@@ -21,6 +21,53 @@ export class App extends React.Component {
         }
     };
 
+    renderAlertComponent = (alertStore: AlertStore, darkTheme: boolean) => {
```
duidae: private

src/App.tsx
```diff
@@ -21,6 +21,53 @@ export class App extends React.Component {
...
+                        onClose={alertStore.handleInteractiveAlertClosed}
+                        canEscapeKeyCancel={false}
+                    >
+                        <p>{alertStore.interactiveAlertText}</p>
+                    </Alert>
+                );
+            default:
+                return null;
+        }
+    }
```
duidae: semi-colon

src/App.tsx
```diff
@@ -30,24 +77,12 @@ export class App extends React.Component {
             glClassName += " dark-theme";
         }
 
+        let alertComponent = this.renderAlertComponent(appStore.alertStore, appStore.darkTheme);
```
duidae: const

src/stores/AppStore.ts
```diff
@@ -1507,7 +1514,7 @@ export class AppStore {
             console.error(err);
             this.alertStore.showAlert("Error resuming session");
         });
-    };
+    }
```
duidae: semi-colon

duidae (changes requested):

src/stores/AlertStore.ts
```diff
@@ -14,33 +20,40 @@ export class AlertStore {
...
-    @action("dismiss_alert") dismissInteractiveAlert = () => {
-        this.interactiveAlertVisible = false;
-    };
+    @action showRetryAlert = (text: string | React.ReactNode, onClose: (confirmed: boolean) => void, icon?: any) => {
+        this.interactiveAlertText = text;
+        this.alertIcon = icon;
+        this.alertType = AlertType.Retry;
+        this.alertVisible = true;
+        this.interactiveAlertCallback = onClose;
+    }
```
duidae: semi-colon

duidae (changes requested):

veggiesaurus:
thanks @duidae, have fixed the minor issues

TienHao (approved):

YuHsuan-Hwang (approved):

Issue #1324, opened 2021-02-20
greyed out shortcuts are not really disabled 


kswang1029:
When CARTA is just initialized with a file browser launched, the file menu looks like
![Screen Shot 2021-02-20 at 9 23 07 PM](https://user-images.githubusercontent.com/20819712/108596732-dab35400-73c1-11eb-89f3-f18b952cd96a.png)
However, if we use shortcut to trigger
- append image (alt + L)
- save image (alt + S)
- import catalog (alt + G)
the frontend accepts the command even though in the menu they are greyed out.

This introduces a more critical issue when triggering "import catalog" as demonstrated in the video. The breadcrumb shows $BASE and if we click the "computer" icon then try again to load an image, the folder image (eg CASA and MIRIAD) is not recognised as images anymore. FITS image cannot be loaded either.  

https://user-images.githubusercontent.com/20819712/108597241-fa974780-73c2-11eb-8608-3d5b54f255df.mp4

The messages are
![Screen Shot 2021-02-20 at 9 20 02 PM](https://user-images.githubusercontent.com/20819712/108596814-61683100-73c2-11eb-8abd-af1b6d6e749b.png)

Issue #1406, opened 2021-04-16
image archive use case: disable write request


kswang1029:
pairing to https://github.com/CARTAvis/carta-backend/issues/798 

Once the frontend sees the flag in the message, save related options should be strictly disabled.

PR #1416, opened 2021-04-22
Yuhsuan/issue 1406 read only mode

YuHsuan-Hwang:
This PR is for issue https://github.com/CARTAvis/carta-frontend/issues/1406: grey out Save image and Export regions in the file menu in read-only mode.
Also fixed issue https://github.com/CARTAvis/carta-frontend/issues/1324: avoid opening the file browser dialog when users use greyed out shortcuts.


kswang1029:
@YuHsuan-Hwang could you add tooltip to those disabled options to let user know why? (only when the frontend sees the read_only flag)

YuHsuan-Hwang:
> @YuHsuan-Hwang could you add tooltip to those disabled options to let user know why? (only when the frontend sees the read_only flag)

Ok. Will add tooltips to the menu items.

veggiesaurus (approved):

kswang1029:
@YuHsuan-Hwang tooltip works nicely and intelligently. 👍

kswang1029 (approved):

Issue #1415, opened 2021-04-21
colorbar not rendered correctly when colormap is inverted


veggiesaurus:
The colorbar looks fine until I invert the colormap. 

Normal:
![ds91 fits-image-2021-04-21-16-54-08](https://user-images.githubusercontent.com/592504/115574821-42065b00-a2c2-11eb-946a-9bd3c93187ab.png)

Inverted: 
![ds91 fits-image-2021-04-21-16-54-32](https://user-images.githubusercontent.com/592504/115574869-4e8ab380-a2c2-11eb-8e3c-2b64fde7d90e.png)

kswang1029:
In dev with horizontal colorbar merged?

kswang1029:
Ok I see this in bias/contrast branch, not in dev (before merging bias/contrast) and I missed it🙈

veggiesaurus:
> Ok I see this in bias/contrast branch, not in dev (before merging bias/contrast)

oops, I thought it was in `dev` 🙈

PR #1419, opened 2021-04-23
Yuhsuan/issue 1415 invert colorbar

YuHsuan-Hwang:
This PR is for https://github.com/CARTAvis/carta-frontend/issues/1415. Modified the functions so that the colorbar works fine with inverted color and also with extreme cases (single color).


kswang1029 (approved):
confirmed 👍

veggiesaurus (approved):

PR #1222, opened 2020-10-07
Angus/build updates

veggiesaurus:
Closes #321 and closes #706 

This is a rather large change to the build, because it now uses `create-react-app@3.4.3`, which comes with new build scripts. Some other changes:

- allows us to import SCSS files directly, no need to run all the CSS-watching stuff
- uses eslint (with typescript plugins) instead of tslint
- uses browserlist to ensure that we compile to a sensible target, rather than supporting ancient browsers

There are a few stricter eslint rules by default, and this has required a number of minor changes to the code. Hopefully this doesn't break anything. 

@kswang1029 can you do some testing of the frontend in general?
@ajm-asiaa can you try build electron apps using this branch?
@duidae @panchyo0 @TienHao can you check through the changes to the code itself? Most were suggested directly by eslint.

note: this seems to break the breadcrumbs fix by @ajm-asiaa, I'll look into solving this during the review process

PR #1423, opened 2021-04-26
readme changes

veggiesaurus:
Fixes weird repetitions in the readme, introduced in #1222.


confluence (approved):
This looks OK to me, but I think we should maybe also add a section about packages (Ubuntu and npm).

veggiesaurus:
> This looks OK to me, but I think we should maybe also add a section about packages (Ubuntu and npm).

ok. I'll file an issue for this

Issue #1429, opened 2021-05-03
progress indicator of the spectral profiler 


kswang1029:
POST v2.0 issue

PR https://github.com/CARTAvis/carta-frontend/pull/1391 enables the support of plotting multiple profiles in one spectral profiler widget.

The profile streaming progress indicator has two issues:
- [x] if the image is 2D, the progress stays as [0%/100% complete]
- [ ] if there are multiple profile streams, the progress indicator will appear and disappear repeatedly until all profiles are streamed. The true accumulated progress needs to be calculated properly so that the progress indicator shows up only once when profiles are streaming.

PR #1431, opened 2021-05-04
Jason/issue 1429 and tooltip

duidae:
this PR fixed:
1. first part of #1429 (if the image is 2D, the progress stays as [0%/100% complete])
2. add more tooltips to multi profile selection buttons of spectral profiler.


kswang1029:
This might be related to this PR (let me know if it is not then I will file a new issue).

The image dropdown will run into a grey-out state persistently if we select an image without channels. 

https://user-images.githubusercontent.com/20819712/117103001-1304e480-adac-11eb-9c0c-248d3ea7aa80.mp4

kswang1029:
please revise "stokes coordinate" to "Stokes parameter" (official term) in tooltips.

kswang1029:
Would the following tooltips be sensible?

> Image
When enabled, Spectral Profiler will show multiple profiles from different images which are matched both spatially and spectrally via the Image List widget.
Region
When enabled, Spectral Profiler will show multiple profiles from different selected regions.
Statistic
When enabled, Spectral Profiler will show multiple profiles with different selected statistic quantities.
Stokes
When enabled, Spectral Profiler will show multiple profiles with different selected Stokes.

duidae:
> This might be related to this PR (let me know if it is not then I will file a new issue).
> 
> The image dropdown will run into a grey-out state persistently if we select an image without channels.
> 
>  May-05-2021.14-12-56.mp4

tooltip wording refined. please have a look.
And the deliberate grey-out is to guide users not to click the button since enabling multiple profile mode of the category with disabled dropdown is kind of meaningless.

For example, when an image has no stokes, the dropdown is disabled, and the button is still in bold font looking like functional , users may still want to click them but it is just in vain (nothing happen) after clicking:
image

The reason we still allow users to click is that when switching to other image having multiple stokes, the clicked button would still work.

duidae:
> Would the following tooltips be sensible?
> 
> > Image
> > When enabled, Spectral Profiler will show multiple profiles from different images which are matched both spatially and spectrally via the Image List widget.
> > Region
> > When enabled, Spectral Profiler will show multiple profiles from different selected regions.
> > Statistic
> > When enabled, Spectral Profiler will show multiple profiles with different selected statistic quantities.
> > Stokes
> > When enabled, Spectral Profiler will show multiple profiles with different selected Stokes.

will change to above texts, but will also keeps the XY/Z in Image part since there seems no other tutorial in spectral profiler to teach users how to create spatial/spectral matched images, so may need to provide clear instructions.

kswang1029:
> > Would the following tooltips be sensible?
> > > Image
> > > When enabled, Spectral Profiler will show multiple profiles from different images which are matched both spatially and spectrally via the Image List widget.
> > > Region
> > > When enabled, Spectral Profiler will show multiple profiles from different selected regions.
> > > Statistic
> > > When enabled, Spectral Profiler will show multiple profiles with different selected statistic quantities.
> > > Stokes
> > > When enabled, Spectral Profiler will show multiple profiles with different selected Stokes.
> 
> will change to above texts, but will also keeps the XY/Z in Image part since there seems no other tutorial to teach users how to create spatial/spectral matched images, so may need to provide clear instructions.

ok. 

stokes -> Stokes (capital S as it is a name who invented the Stokes parameters :) )

duidae:
@kswang1029 tooltip updated, please have a look, thanks.

kswang1029:
> @kswang1029 tooltip updated, please have a look, thanks.

Thanks. The tooltips are fine. Any idea on the forever-grey out image dropdown? One solution I can think of is to grey out 2D image in the list so users won't have a change to trigger the issue.

duidae:
Grey out 2D image options. please have a look.

kswang1029 (approved):
looking good 👍

veggiesaurus (approved):

PR #1440, opened 2021-05-12
Finish exporting header as txt file

### LINKED ISSUE #1357: export header as a text file
as requested from #1355, it would be handy to have an export button so that the header file can be saved as a text file.

acdo2002:
Try to fix #1357 
I follow the LinePlotComponent.tsx (like channel-N-histogram.tsv to export data) & FileInfoComponent.tsx (like search-text button) structure to solve this issue. 
Thanks Yu-Hsuan's help!

Any suggestion is welcome.

### DIFF
--- a/src/components/FileInfo/FileInfoComponent.scss
+++ b/src/components/FileInfo/FileInfoComponent.scss
@@ -83,6 +83,16 @@
             user-select: none;
         }
     }
+    
+    .header-export{
+        position: absolute;
+        bottom: 0px;
+        right: 0;
+        margin: 0 10px 10px 0;
+        transition: opacity 400ms;
+    }
+
+
 }
 
 .header-search-input {
--- a/src/components/FileInfo/FileInfoComponent.scss
+++ b/src/components/FileInfo/FileInfoComponent.scss
@@ -91,8 +91,6 @@
         margin: 0 10px 10px 0;
         transition: opacity 400ms;
     }
-
-
 }
 
 .header-search-input {
--- a/src/components/FileInfo/FileInfoComponent.scss
+++ b/src/components/FileInfo/FileInfoComponent.scss
@@ -77,7 +77,7 @@
         bottom: -40px;
         right: 0;
         margin: 0 10px 10px 0;
-        transition: 400ms;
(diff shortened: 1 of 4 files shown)


kswang1029:
@acdo2002 I suggest we add filename as the first line of the text file starting with #
Also, we should try to have a button group with the download button on the right hand side and the search button on the left hand side.

veggiesaurus:
Should we not be trying to export this as someone that can be used as a fits header?

acdo2002:
@kswang1029 
Does this format match to your thinking? (using button group, search button is on the left-hand-side, while export txt button is on right-hand-side)
Screenshot from 2021-05-12 21-39-09-1
I can add filename as the first line of the text file starting with # later.

@veggiesaurus 
Sorry I am a bit confused, do you mean the exported txt file may be used by other? 
At least my experience is that the header will generate together with the data calibration/creation rather than I import the header...

veggiesaurus:
> @veggiesaurus
> Sorry I am a bit confused, do you mean the exported txt file may be used by other?
> At least my experience is that the header will generate together with the data calibration/creation rather than I import the header...

I was just thinking that it makes sense to make them look as similar as possible to a FITS header, but perhaps that doesn't matter

acdo2002:
@kswang1029 I have followed your suggestion to improve the function, could you check it?

kswang1029 (approved):
this looks fine. 

To improve the readability, we probably want to align "=" (cf. ds9). This could happen as a different issue.

acdo2002:
@TienHao Because I put export txt as a utilities function and change small part of your code, could you check whether it is OK? I just remove the duplicate parts.

YuHsuan-Hwang (approved):

acdo2002:
@kswang1029 the minor code refactoring has been done, could you look at it again? 
there are two parts used the exportTxt function (because we wrote in the same way): One is header export, another one is spectral-line fitting results saved as log.

TienHao (approved):

kswang1029 (approved):

veggiesaurus (approved):

PR #1441, opened 2021-05-12
Thchang/issue 961

### LINKED ISSUE #961: catalog: region creation mode is not activated immediately by click when catalog selection mode is enabled
When the catalog selection mode is enabled, we need TWO clicks to enable region creation mode instead of the expected one click.

Jun-29-2020 10-07-38

TienHao:
closes #961. Fixed region create button, which was not activated immediately by click when catalog selection mode was enabled.

### DIFF
--- a/src/components/ImageView/Toolbar/ToolbarComponent.tsx
+++ b/src/components/ImageView/Toolbar/ToolbarComponent.tsx
@@ -64,10 +64,10 @@ export class ToolbarComponent extends React.Component<ToolbarComponentProps> {
 
     private handelActiveLayerClicked = (layer: ImageViewLayer) => {
         this.props.onActiveLayerChange(layer);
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):

src/components/ImageView/Toolbar/ToolbarComponent.tsx
```diff
@@ -64,7 +64,9 @@ export class ToolbarComponent extends React.Component<ToolbarComponentProps> {
 
     private handelActiveLayerClicked = (layer: ImageViewLayer) => {
```
veggiesaurus: can you fix this typo at the same time? `handel` -> `handle`

veggiesaurus (changes requested):

veggiesaurus (approved):

Issue #1201, opened 2020-09-11
angular distance measuring tool


kswang1029:
as one of the NRAO priorities
@jott3077 please help to add descriptions. 

req. doc:
https://docs.google.com/document/d/17S6JP5s0_6OZGKLYMrDuY9rV86-TR1Vqgcpra4Ocj78/edit

remaining features:
- [x] add a distance measuring dialog and support pixel position input
- [x] support wcs position input
- [ ] support style changes: add option for measuring distance on all three sides
- [ ] support style changes: distance units, color, line width

kswang1029:
The initial implementation with the basic functions is merged. The rest features will be scheduled later for the v3.0 release.

kswang1029:
@YuHsuan-Hwang shall we close this one since now we have ruler annotation where we can have a custom styling.

PR #1446, opened 2021-05-18
Yuhsuan/issue 1201 distance measure

YuHsuan-Hwang:
This PR is for the basic feature requested in https://github.com/CARTAvis/carta-frontend/issues/1201: plot geodesic triangles and show distance in the image.

Other parts of the feature, including adding a distance measuring dialog (support position input and unit conversion) and a setting panel (support style changes) will be implemented in different PRs.

Will wait for PR https://github.com/CARTAvis/carta-frontend/pull/1441 to be merged (need to resolve conflicts).


kswang1029:
@YuHsuan-Hwang overall this looks nice. Few suggestions:
1) if the first click does not result in a valid world coordinate, we should ignore it. In the following video, the first click is outside the WCS grid and the second is inside.

https://user-images.githubusercontent.com/20819712/118593778-d63ce280-b7da-11eb-9b88-1e5a3b3ec53c.mp4

ps. the projection effect is cool~

2) add the distance to the other two sides?
3) change the cursor style to something like a pen or a + as an indicator of the distance measure mode?
4) when the image does not have a valid WCS, we use pixel coordinate. So the distance should also show the unit as "pix"
LMCq fits-image-2021-05-18-13-10-34
5) the icon svg looks a bit blurred in one side due to dashed line?.
Screen Shot 2021-05-18 at 1 17 50 PM

kswang1029:
@YuHsuan-Hwang performance-wise, it seems the distance triangle is re-rendered a bit slowly?

https://user-images.githubusercontent.com/20819712/118595339-772c9d00-b7dd-11eb-89ca-2de6b41794c2.mp4

YuHsuan-Hwang:
> @YuHsuan-Hwang overall this looks nice. Few suggestions:
> 
> 1. if the first click does not result in a valid world coordinate, we should ignore it. In the following video, the first click is outside the WCS grid and the second is inside.
> 
>  May-18-2021.13-09-35.mp4 
> ps. the projection effect is cool~
> 
> 2. add the distance to the other two sides?
> 3. change the cursor style to something like a pen or a + as an indicator of the distance measure mode?
> 4. when the image does not have a valid WCS, we use pixel coordinate. So the distance should also show the unit as "pix"
>    LMCq fits-image-2021-05-18-13-10-34
> 5. the icon svg looks a bit blurred in one side due to dashed line?.
>    Screen Shot 2021-05-18 at 1 17 50 PM

1.3.4. Will fix the issues
2. I will set this as a style setting in another PR. I think it is more clear to show one number as default.
5. The icon is a bit blurred due to the design. It's also similar to region icons. We may change to another icon. So far two options:
imageimage
Any suggestions about the icon?

YuHsuan-Hwang:
> @YuHsuan-Hwang performance-wise, it seems the distance triangle is re-rendered a bit slowly?
> 
>  May-18-2021.13-31-07.mp4

It's about AST plot rendering. Not sure how to avoid this.

kswang1029:
> > @YuHsuan-Hwang performance-wise, it seems the distance triangle is re-rendered a bit slowly?
> > May-18-2021.13-31-07.mp4
> 
> It's about AST plot rendering. Not sure how to avoid this.

The gridline rendering has same issue and root cause I think. We can leave it for now.

veggiesaurus:
> > > @YuHsuan-Hwang performance-wise, it seems the distance triangle is re-rendered a bit slowly?
> > > May-18-2021.13-31-07.mp4
> > 
> > 
> > It's about AST plot rendering. Not sure how to avoid this.
> 
> The gridline rendering has same issue and root cause I think. We can leave it for now.

We're being cautious in terms of how often we update the AST grid, because in some scenarios it will take longer than 16 ms to update. I'd like to fix this, though. It would make sense to introduce a dynamic frame rate, depending on how long it actually takes to plot.

I agree we should leave it as a separate issue though

kswang1029:
> > > > @YuHsuan-Hwang performance-wise, it seems the distance triangle is re-rendered a bit slowly?
> > > > May-18-2021.13-31-07.mp4
> > > 
> > > 
> > > It's about AST plot rendering. Not sure how to avoid this.
> > 
> > 
> > The gridline rendering has same issue and root cause I think. We can leave it for now.
> 
> We're being cautious in terms of how often we update the AST grid, because in some scenarios it will take longer than 16 ms to update. I'd like to fix this, though. It would make sense to introduce a dynamic frame rate, depending on how long it actually takes to plot.
> 
> I agree we should leave it as a separate issue though

The current solution for gridline is we apply a higher tolerance (a control parameter of line smoothness from ast) _during_ panning/zooming.

YuHsuan-Hwang:
@kswang1029 Fixed 1.3.4.

YuHsuan-Hwang:
@kswang1029 @veggiesaurus @duidae Changed to formal PR.

src/components/ImageView/Toolbar/ToolbarComponent.tsx
```diff
@@ -190,18 +192,20 @@ export class ToolbarComponent extends React.Component<ToolbarComponentProps> {
...
+                </Tooltip>
                 <Tooltip position={tooltipPosition} content={<span>Catalog selection<br/><i><small>Click to select single catalog source</small></i></span>}>
-                    <AnchorButton icon={"locate"} active={catalogOverlayEnabled} onClick={() => this.handleActiveLayerClicked(ImageViewLayer.Catalog)} disabled={catalogSelectionDisabled}/>
+                    <AnchorButton icon={"locate"} active={appStore.activeLayer === ImageViewLayer.Catalog} onClick={() => this.handleActiveLayerClicked(ImageViewLayer.Catalog)} disabled={catalogSelectionDisabled}/>
                 </Tooltip>
                 {frame.regionSet.mode === RegionMode.CREATING &&
                 <Tooltip position={tooltipPosition} content={<span>Create region<br/><i><small>Click to select region type</small></i></span>}>
                     <Popover content={regionMenu} position={Position.TOP} minimal={true}>
-                        <AnchorButton icon={regionIcon} active={true}/>
+                        <AnchorButton icon={regionIcon} active={appStore.activeLayer === ImageViewLayer.RegionCreating} onClick={() => this.handleActiveLayerClicked(ImageViewLayer.RegionCreating)}/>
```
YuHsuan-Hwang, line 210: @TienHao I resolved conflicted with PR https://github.com/CARTAvis/carta-frontend/pull/1441. Please check if this part is ok, thanks.

kswang1029 (approved):
looks nice

TienHao (approved):

veggiesaurus (changes requested):
@YuHsuan-Hwang this is really nice. However, one bug I spotted: if you change coordinates (e.g. from FK5 to GAL) you have to click a few times before it works, or you have to pan the image a little bit.

veggiesaurus:
@kswang1029 @YuHsuan-Hwang am I misunderstanding something here? This is a GALFACTS image with FK5 and `--CAR` projection. Shouldn't the individual components of the distance vector align with the grid lines here?

image

kswang1029:
> @kswang1029 @YuHsuan-Hwang am I misunderstanding something here? This is a GALFACTS image with FK5 and `--CAR` projection. Shouldn't the individual components of the distance vector align with the grid lines here?
> 
> image

in dec axis, it is a straight line as expected. The other two sides should not be straight lines. Straight lines parallel to the ra axis are not the shortest distance I think.

YuHsuan-Hwang:
> @YuHsuan-Hwang this is really nice. However, one bug I spotted: if you change coordinates (e.g. from FK5 to GAL) you have to click a few times before it works, or you have to pan the image a little bit.

Sorry, I am not sure what you mean. Could you show how to reproduce it?

YuHsuan-Hwang:
> > @kswang1029 @YuHsuan-Hwang am I misunderstanding something here? This is a GALFACTS image with FK5 and `--CAR` projection. Shouldn't the individual components of the distance vector align with the grid lines here?
> > image
> 
> in dec axis, it is a straight line as expected. The other two sides should not be straight lines. Straight lines parallel to the ra axis are not the shortest distance I think.

Yes, the lines are geodesic, so they are parallel to only dec axis.

veggiesaurus:
> > @YuHsuan-Hwang this is really nice. However, one bug I spotted: if you change coordinates (e.g. from FK5 to GAL) you have to click a few times before it works, or you have to pan the image a little bit.
> 
> Sorry, I am not sure what you mean. Could you show how to reproduce it?

@YuHsuan-Hwang using `C.fits` and switching from WCS to GAL. It doesn't work properly until I pan the image at least one. And after that, it seems to pick up the totally incorrect starting coordinate.


https://user-images.githubusercontent.com/592504/121645308-554cd000-ca94-11eb-9ddf-8d432fa9dc06.mp4

YuHsuan-Hwang:
> > > @YuHsuan-Hwang this is really nice. However, one bug I spotted: if you change coordinates (e.g. from FK5 to GAL) you have to click a few times before it works, or you have to pan the image a little bit.
> > 
> > 
> > Sorry, I am not sure what you mean. Could you show how to reproduce it?
> 
> @YuHsuan-Hwang using `C.fits` and switching from WCS to GAL. It doesn't work properly until I pan the image at least one. And after that, it seems to pick up the totally incorrect starting coordinate.
> 
>  2021-06-11.09-00-27.mp4

I see, I will try to let it responsive to coordinate changes. However, I can create new lines without extra clicks or panning, so I am not sure about this part. Does this happen to every image?

veggiesaurus:
> I see, I will try to let it responsive to coordinate changes. However, I can create new lines without extra clicks or panning, so I am not sure about this part. Does this happen to every image?

Yes. Happens for me on both Chrome and Firefox (Ubuntu 20.04).

Its far less noticable if you go from FK5 to FK4. If you go from FK5 to GAL it's a big jump

kswang1029:
@veggiesaurus @YuHsuan-Hwang I suspect the jump is due to a missing step of coordinate transformation from the original reference system (eg GAL) to a target reference system (eg FK5). The same world coordinate such as (100deg, -20deg) points to different positions in different reference systems.

kswang1029:
> > I see, I will try to let it responsive to coordinate changes. However, I can create new lines without extra clicks or panning, so I am not sure about this part. Does this happen to every image?
> 
> Yes. Happens for me on both Chrome and Firefox (Ubuntu 20.04).
> 
> Its far less noticable if you go from FK5 to FK4. If you go from FK5 to GAL it's a big jump

@YuHsuan-Hwang @veggiesaurus regarding the extra-clicks, if we use "double-click" to switch to a different coordinate system, the issue can be reproduced.

veggiesaurus (approved):

PR #1452, opened 2021-05-21
Jason/issue 1444

### LINKED ISSUE #1444: label in region dropdown menus of the spectral profiler and moment generator out of sync
When we change region in the moment generator, the spectral profile data get updated in the spectral profiler as expected. However the label does not get updated. @duidae could you help to have a look for v2.0 final?

https://user-images.githubusercontent.com/20819712/118429151-97882900-b703-11eb-984c-e1d6f60c8c9a.mp4

duidae:
closes #1444 

P.S. image selection(image drop down in moment tab) still needs to trigger frame change in spectral profiler because range values of moment are calculated from the selected image, or users will get weird range values from the unmatched frame during range selection.

### DIFF
--- a/src/components/Shared/RegionSelector/RegionSelectorComponent.tsx
+++ b/src/components/Shared/RegionSelector/RegionSelectorComponent.tsx
@@ -7,8 +7,7 @@ import {RegionWidgetStore, RegionsType, RegionId, ACTIVE_FILE_ID} from "stores/w
 import "./RegionSelectorComponent.scss";
 
 @observer
-export class RegionSelectorComponent extends React.Component<{ widgetStore: RegionWidgetStore, onFrameChanged?: (newFrame: FrameStore) => void, nonClosedDisabled?: boolean }> {
-
+export class RegionSelectorComponent extends React.Component<{ widgetStore: RegionWidgetStore, onFrameChanged?: (newFrame: FrameStore) => void }> {
     private handleFrameChanged = (changeEvent: React.ChangeEvent<HTMLSelectElement>) => {
         const appStore = AppStore.Instance;
         const widgetStore = this.props.widgetStore;
@@ -38,10 +37,7 @@ export class RegionSelectorComponent extends React.Component<{ widgetStore: Regi
 
         let enableFrameselect = false;
         let selectedFrameValue: number = ACTIVE_FILE_ID;
-        let frameOptions: IOptionProps[] = [{value: ACTIVE_FILE_ID, label: "Active"}];
-
         if (appStore.activeFrame) {
-            frameOptions = frameOptions.concat(appStore.frameNames);
             selectedFrameValue = widgetStore.fileId;
             enableFrameselect = true;
         }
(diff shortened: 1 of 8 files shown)


kswang1029 (approved):
👍

veggiesaurus:
The one issue I have is that its not possible to generate a moment map for the entire image if you've got a region selected, even by using the dropdown. Shouldn't there be an "image" entry as well? I guess this is out of the scope of the PR though, and a separate issue

veggiesaurus:
image

The question mark icon as a home for a tooltip seems out of place to me. Do we use this approach anywhere else in the UI?

duidae:
> image
> 
> The question mark icon as a home for a tooltip seems out of place to me. Do we use this approach anywhere else in the UI?

yeah I got confused with previous implementation (dropdown did not include `image` option) as well, the question mark is to clarify how to generate a moment map for the entire image, which was implicit previously.

veggiesaurus:
> > image
> > The question mark icon as a home for a tooltip seems out of place to me. Do we use this approach anywhere else in the UI?
> 
> yeah I got confused with previous implementation (dropdown did not include `image` option) as well, the question mark is to clarify how to generate a moment map for the entire image, which was implicit previously.

But why not put the tooltip on the dropdown instead? Or only show it when the selected region is "active". It's not an approach we've used elsewhere. 

If we do stick with the icon, then we should at least make its intent "info" instead of "warning"

duidae:
> The one issue I have is that its not possible to generate a moment map for the entire image if you've got a region selected, even by using the dropdown. Shouldn't there be an "image" entry as well? I guess this is out of the scope of the PR though, and a separate issue

the major ambiguity is from:
1. if current region is 'cursor', generate the moment map of whole image, but point regions are not selectable in the dropdown in moment tab, previous implementation solved this by using the 'Active' option.
2. the 'Active' option is to provide a convenient way for region selection.
3. and kind of weird to append 'Image' option because 'Active' option and 'Image' option create duplication when region is cursor.

therefore the question mark tool tip comes out for fixing this ambiguity. And the text in parentheses for region is also to clarify the ambiguity. can also place the tooltip to the dropdown if we think it's obvious enough.

veggiesaurus:
@kswang1029 suggests we remove the tooltip entirely

duidae:
> @kswang1029 suggests we remove the tooltip entirely

changed to 
image


@kswang1029  it's the easiest way for implementation, but will this not confuse user (just like @veggiesaurus and me in the beginning)?

kswang1029:
the text label after "Region" is clear enough I think. ok to have a tooltip with the dropdown if you prefer.

veggiesaurus:
> and kind of weird to append 'Image' option because 'Active' option and 'Image' option create duplication when region is cursor.

This is the approach we take with histograms and stats, so doing so would at least be consistent

duidae:
image

@veggiesaurus @kswang1029 added image option, please have a look, thanks.

kswang1029 (approved):

veggiesaurus (approved):

PR #1454, opened 2021-05-25
Jason/line disappear

### LINKED ISSUE #1453: spectral line results disappear if filtered rows are too few
reproduce(according to beta user feedback):
1. open spectral line query widget & query with range 230~231 GHz
2. scroll down to row 50
3. input "Monoxide" in chemical name & press filter
4. the info in the bottom shows "Showing 2 filtered lines..." but see nothing

the other way to reproduce:
3. input "Ketene" in chemical name & press filter

image
image

duidae:
closes #1453

### DIFF
--- a/src/components/SpectralLineQuery/SpectralLineQueryComponent.tsx
+++ b/src/components/SpectralLineQuery/SpectralLineQueryComponent.tsx
@@ -28,6 +28,7 @@ export class SpectralLineQueryComponent extends React.Component<WidgetProps> {
     @observable headerTableColumnWidths: Array<number>;
     private headerTableRef: Table;
     private resultTableRef: Table;
+    private scrollToTopHandle;
 
     public static get WIDGET_CONFIG(): DefaultWidgetConfig {
         return {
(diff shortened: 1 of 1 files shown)


src/components/SpectralLineQuery/SpectralLineQueryComponent.tsx
```diff
@@ -197,12 +197,12 @@ export class SpectralLineQueryComponent extends React.Component<WidgetProps> {
 
     private handleFilter = () => {
         this.widgetStore.filter();
-        this.resultTableRef?.scrollToRegion(Regions.row(0));
+        setTimeout(() => this.resultTableRef?.scrollToRegion(Regions.row(0)), 20);
```
veggiesaurus: @duidae do you think we should use a handle here so that we can `clearTimeout` of the existing scroll instruction?

kswang1029:
The presence of the vertical scrollbar seems delayed by three rows? If there are two rows out of view we cannot scroll down to see those. 

https://user-images.githubusercontent.com/20819712/119917569-662c1a80-bf99-11eb-97ae-739ce1f95fd5.mp4

duidae:
> The presence of the vertical scrollbar seems delayed by three rows? If there are two rows out of view we cannot scroll down to see those.
> 
>  May-28-2021.09-44-17.mp4

there should be some bugs in scroll calculation of BlueprintJS Table, fix with temporary workaround by appending 3 empty rows to ensure that the scroll pops out.

src/components/SpectralLineQuery/SpectralLineQueryComponent.tsx
duidae: @veggiesaurus handle added, but the behavior seems the same before/after the timer handle added, a little bit unsure about why we need it?

veggiesaurus: I suppose its not necessary. I was just worried that we would have multiple `handleFilter` functions resetting the scroll to row 0.

veggiesaurus:
> > The presence of the vertical scrollbar seems delayed by three rows? If there are two rows out of view we cannot scroll down to see those.
> > May-28-2021.09-44-17.mp4
> 
> there should be some bugs in scroll calculation of BlueprintJS Table, fix with temporary workaround by appending 3 empty rows to ensure that the scroll pops out.

Yet another reason to attempt a switch to a more readily maintained and lightweight virtualised table package :see_no_evil:

kswang1029 (approved):

PR #1461, opened 2021-05-27
Jason/multiprofile checkbox

duidae:
comments from beta user:

> Clarifying the "turn on multiple profiles" User Interaction would be helpful. It took me a while to realize I just needed to click on the "Images" button or "Regions" button to select / plot multiple profiles. Maybe just allowing the drop-downs to be multiselect with check marks and have a "Select all" at the bottom. Now that I know about it, it's fine, so maybe this doesn't need to change but it was confusing at first.

make sense & replace button with checkbox


kswang1029 (approved):
this works fine. Could you remove the ":" after Image/Region/Statistic/Stokes?

veggiesaurus (approved):

PR #1468, opened 2021-05-31
Yuhsuan/issue 1037 export selected region

### LINKED ISSUE #1037: support exporting selected regions only
Need to say "export all items" or show a list of regions for the user to select/deselect / right-click on a region and go to export just for that region.

YuHsuan-Hwang:
This PR closes https://github.com/CARTAvis/carta-frontend/issues/1037: exporting selected regions from the file browser dialog.
Supports multiple selection. Changing the active image resets the selection.


YuHsuan-Hwang:
Changing to tab element based on @kswang1029 's suggestion.

YuHsuan-Hwang:
@kswang1029 Added a new tab for selecting regions.

veggiesaurus:
I haven't looked at the code yet, but I do wonder whether we can easily extend this at a later stage to actually allow users to select multiple regions in the image view widget or region list

kswang1029:
> I haven't looked at the code yet, but I do wonder whether we can easily extend this at a later stage to actually allow users to select multiple regions in the image view widget or region list

We need to be careful about this due to Active and linked behavior…🤔

YuHsuan-Hwang:
> I haven't looked at the code yet, but I do wonder whether we can easily extend this at a later stage to actually allow users to select multiple regions in the image view widget or region list

In my current implement, this multiple selection is totally different from multiple active regions. This export selection looks for the index in the active image rather than id (reset easily by switching the active image).

kswang1029 (approved):
Feature-wise this is good. 👍

veggiesaurus (approved):

PR #1471, opened 2021-05-31
Fix info wrapping issue

### LINKED ISSUE #1469: cursor info layout in the spectral profiler
Screen Shot 2021-05-31 at 2 43 58 PM

could you help to see if the cursor info layout can be one per line? thanks.

duidae:
closes #1469


veggiesaurus (approved):
this looks fine. I think we should file a separate issue for automatically adjusting the space below the profile as you add more lines, so that we don't waste space when you only have one profile:

image

kswang1029 (approved):

PR #1491, opened 2021-06-09
Jason/fix double quotes issue

duidae:
eliminate double quotes in commit msg for not breaking mix'n match


veggiesaurus:
@duidae I think it might be time to move this to a separate bash script, then we can just have something like 
```
"git-info": "fill_git_info.sh"
```

duidae:
> @duidae I think it might be time to move this to a separate bash script, then we can just have something like
> 
> ```
> "git-info": "fill_git_info.sh"
> ```

shell script created, please have a look, thanks.

veggiesaurus (approved):

PR #1497, opened 2021-06-09
Angus/prettier

veggiesaurus:
draft PR of the changes when auto-formatting with [Prettier](https://prettier.io/docs/en/install.html).

hint: hide whitespaces when comparing changes in this PR!


duidae:
checked & nice settings that covered out needs :thumbsup:

PR #1503, opened 2021-06-15
Make profiler info height flexible

duidae:
Make the spectral profile info height flexible when in single profile mode(1 line height only).


veggiesaurus (changes requested):
I don't think this approach works, for two reasons:
1. It looks fine for 1 profile, and fine for 3 profiles (at the default size) but looks weird for 2 profiles (still have wasted space).

https://user-images.githubusercontent.com/592504/122545296-fa841d00-d02d-11eb-8995-55160581fe60.mp4

2. Using a percentage-based size means that resizing the widget breaks this approach

https://user-images.githubusercontent.com/592504/122545386-12f43780-d02e-11eb-98ba-59524f520bf2.mp4

Can't we use a fixed size per profile, up to a maximum? For example, 15px per profile, up to a maximum of 5 lines (just a suggestion, it may need some fine tuning)

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -499,10 +499,12 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
             className += " dark-theme";
         }
 
-        let infoHeight = "28px";
-        if (this.plotData?.numProfiles > 1 && this.plotData?.numProfiles <= 5) {
-            infoHeight = `${this.plotData?.numProfiles * 20}px`;
-        } else if (this.plotData?.numProfiles > 5) {
+        const infoMinHeight = "28px";
+        const numProfiles = this.plotData?.numProfiles;
+        let infoHeight = infoMinHeight;
```
veggiesaurus: we don't need to use strings for this. We can just pass a `number` in as `defaultSize` or `minSize`: https://github.com/tomkp/react-split-pane

duidae:
> what about `infoHeight = clamp(numProfiles * 20, INFO_HEIGHT_MIN, INFO_MAX_HEIGHT)`? would that work?

Modified, please have a look.

Issue #1507, opened 2021-06-17
inconsistent theme in mode button of animator widget


duidae:
reproduce:
1. create an animator widget, click mode button
2. switch theme

![image](https://user-images.githubusercontent.com/25971696/122329338-d22fec00-cf63-11eb-8590-1ce0bdb4cd17.png)

PR #1511, opened 2021-06-21
Jason/issue 1046

### LINKED ISSUE #1046: display Stokes value instead of numeric index in animator

duidae:
closes #1046 and #1507

### DIFF
--- a/src/components/Animator/AnimatorComponent.scss
+++ b/src/components/Animator/AnimatorComponent.scss
@@ -14,12 +14,16 @@
             margin-bottom: 5px;
 
             .bp3-radio {
-                flex: 1 1 110px;
+                flex: 1 1 135px;
             }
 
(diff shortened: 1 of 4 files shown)


kswang1029 (approved):
this works fine. 👍 

As a different new issue, to be more self-consistent we should rename "Stokes" to "Polarization" and have the label "I" as "Stokes I". This is because the full polarization lookup table also includes some other polarization modes such as XX, YY, RR, LL etc.

PR #1516, opened 2021-06-22
fade in pixel grid at zoom threshold using smoothstep

### LINKED ISSUE #1515: Display a configurable pixel grid at high zoom levels
Photoshop and a number of other image viewers have a useful feature: displaying a pixel grid when you zoom in past a certain threshold (in photoshop, it appears to be 800%). See example below.

https://user-images.githubusercontent.com/592504/122981864-82558880-d39a-11eb-8752-ede3ac191cab.mp4

I'd like to add a configurable pixel grid, where the user can change the following properties:
- visibility (on/off)
- ~~zoom threshold (default: 800%)~~ edit: I don't think this needs to be configurable
- grid color (default: semi-transparent grey)

Example  of what it would look like with semi-transparent white or black grids, based on a quick and dirty shader hack:

m42 fits-image-2021-06-22-20-55-40

m42 fits-image-2021-06-22-20-57-36

I think we could do one better than photoshop in this regard, and gradually fade the grid in as you zoom in past the threshold, rather than getting that rather jarring appearance of the grid.

veggiesaurus:
Closes #1515

Still to be solved:

- [x] Missing grid lines at tile boundaries
- [x] Proper handling of non-square pixels
- [x] Consistent zoom threshold when using spatial matching
- [x] Add visibility toggle and color to widget settings as an implicit preference

edit: all done, ready for review.

image


kswang1029 (approved):
this looks good. 

As a note: at very high zoom level, the grid rendering suffers from precision issue which also exists with region rendering etc.

duidae (approved):

PR #1521, opened 2021-06-24
Angus/1518 pixel highlight

### LINKED ISSUE #1518: pixel highlight by colorbar cursor value
As an advanced feature (cannot remember if there is any other viewer having the same feature), it would be nice to support pixel highlight by colorbar cursor value. That is, when we use mouse to point a color (hence a pixel value) on the colorbar in the image viewer, CARTA highlights those pixels with values >= the colorbar cursor value. This would serve as a manual version of source identification tool. Example mockups (from @veggiesaurus):
ds91 fits-image-2021-06-23-10-43-21
m51_xray fits-image-2021-06-23-10-36-48

It appears that changing pixels with values < the colorbar cursor value to "grey" provides better visibility and UX. The only drawback is it does not work if the selected colormap is grey. We could consider grey colormap as a special case and implement a different way of pixel highlight (or just forget about it; TBD).

veggiesaurus:
Closes #1518 

Also fixes the blurry lines on the colorbar hover line by clamping the line position to mid-pixel

### DIFF
--- a/src/components/ImageView/Colorbar/ColorbarComponent.tsx
+++ b/src/components/ImageView/Colorbar/ColorbarComponent.tsx
@@ -10,7 +10,7 @@ import {getColorForTheme} from "utilities";
 import "./ColorbarComponent.scss";
 
 @observer
-export class ColorbarComponent extends React.Component {
+export class ColorbarComponent extends React.Component<{onCursorHoverValueChanged: (number) => void}> {
     @observable hoverInfoText: string = "";
     @observable showHoverInfo: boolean = false;
     @observable cursorY: number = -1;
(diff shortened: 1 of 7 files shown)


YuHsuan-Hwang:
looks great :thumbsup:
In the image view setting, "Cursor Info" is for showing/hiding the text and the bar. We can change it to "Interactive" (default true) and include enabling/disabling pixel highlight

veggiesaurus:
> In the image view setting, "Cursor Info" is for showing/hiding the text and the bar. We can change it to "Interactive" (default true) and include enabling/disabling pixel highlight

@YuHsuan-Hwang Ok, I've implemented that. I've also refactored the `showHoverInfo` setting to `interactive` and the `showHoverInfo` observable in the component itself to `isHovering`

YuHsuan-Hwang (approved):

YuHsuan-Hwang (approved):

kswang1029 (approved):
this works nicely 👍 

Later on we can see how we can freeze the colorbar cursor so that users can make a png for presentation for example.
Probably we can consider to use click to enable/disable cursor freeze.

Issue #1149, opened 2020-08-24
frontend support region type: line segment


kswang1029:
the line segment region can be used for
- spatial profiler
- pv generator

related to https://github.com/CARTAvis/carta-backend/issues/610

A line region should be defined in two ways:
1. center + length
2. two points

We will need to be able to define the line region in world coordinates and image coordinates. 

For spatially matched image, we will need to resample the line segment on the reference image with multiple points as a  straight polyline, then apply transformation of this set of points to the spatially matched image for render as a "curve". This is consistent with the shared region design (ie region re-projection) in CARTA. 

science requirements: https://docs.google.com/document/d/1mE-JeZjczT_GEPfllm-aNWWUZ903wTJy4JurxBq537c/edit#heading=h.pfjlp9m9fqp4

kswang1029:
postpone to v3.0 (but needs to happen early when v3.0 dev cycle starts).

YuHsuan-Hwang:
The feature of line regions is merged. Polyline regions remain to be implemented.

PR #1532, opened 2021-06-28
Yuhsuan/1149 line region

YuHsuan-Hwang:
This PR is for https://github.com/CARTAvis/carta-frontend/issues/1149 line regions.
* Requires backend branch `pam/610_line_region`.
* Also refactored `RegionViewComponent` so that the logic and naming are more straightforward. Please also test if all kinds of region creation still works fine.
* Hide length and PA inputs if there are non square pixels.


YuHsuan-Hwang:
Changed to click-and-drag creation mode for better ctrl/cmd key event handling.

kswang1029:
@YuHsuan-Hwang found some issues:
1. if we manually edit the length, sometimes it changes to an unexpected length (and PA becoming 45deg?) after hitting the return key. Sometimes the length is 0 and PA becomes NaN. There seems no way to fix it except deleting it. 

https://user-images.githubusercontent.com/20819712/124538524-621fd200-de4e-11eb-9d84-9d1bfc043c77.mov

2. the line region does not follow the region management rules when un-matching an image. When we unmatch an image, this image will get a copy of the regions from the reference image. If we modify some of the regions and match the image again, the frontend should check redundancy of regions and only copy over those modified regions to the reference image. Please try to create some other regions and see how it works or let me know if you need more details on the rules.

YuHsuan-Hwang:
@kswang1029 Ok. Will work on the issues

YuHsuan-Hwang:
@kswang1029 Fixed both two issues.

kswang1029:
@YuHsuan-Hwang any idea about the size field in the region list widget? The values are weird (should just be the length only).
Screen Shot 2021-07-08 at 9 07 04 AM

YuHsuan-Hwang:
@kswang1029 Forgot to adjust this column ... Will change this part.

kswang1029:
The PA becomes 0deg _after_ unmatching image.

https://user-images.githubusercontent.com/20819712/124847792-2f9ce300-dfce-11eb-9491-b9cc7cc57fd2.mov

YuHsuan-Hwang:
@kswang1029
Fixed the size column (including tooltip). This column will show "NaN" if there are non square pixels.
Also fixed the unmatching image bug.

kswang1029 (approved):
tested. 👍  With this we can start the implementation of 
1. backend pv generator @pford 
2. frontend pv generator UI @TienHao

PR #1527, opened 2021-06-25
show and sort on directory item count and modified date

veggiesaurus:
Closes #1526

PR #1534, opened 2021-06-28
use name field when performing fuzzy search

veggiesaurus:
Fixes issue introduced by #1527 when using fuzzy search


kswang1029 (approved):
confirmed the fix 👍

PR #1541, opened 2021-06-30
allow images without units to have square pixels

### LINKED ISSUE #1540: Improve detection of rectangular pixels
Recent changes to the FITS and HDF5 header delivery from the backend have revealed a bug in the logic for detecting rectangular pixels. A large number of false-positives now seem to exist. This needs to be fixed so that users can still view images correctly.

~~@kswang1029 to assist with pseudocode for better rectangular pixel detection~~

veggiesaurus:
Closes #1540 

Todo:
- [x] Handle situations where images don't have units correctly
- [x] Handle situations where images have CD or PC matrix instead of `CDELT`

### DIFF
--- a/src/stores/FrameStore.ts
+++ b/src/stores/FrameStore.ts
@@ -27,7 +27,7 @@ import {
     Transform2D,
     ZoomPoint
 } from "models";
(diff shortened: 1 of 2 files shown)


veggiesaurus:
The easy part is done... But now we need to also handle situations where the image doesn't have `CDELT` entries (e.g. `hlsp_candels_hst_acs_gn-tot-60mas_f606w_v1.0_drz.fits`)

duidae (approved):

Issue #1372, opened 2021-03-30
adding colorbar setting options to preferences dialog


kswang1029:
Once https://github.com/CARTAvis/carta-frontend/pull/1366 is merged, in the preference dialog, we need a "Colorbar configuration" tab and provide options so that users can configure their own defaults of the colorbar styling.

veggiesaurus:
can we make these settings implicitly set as well?

YuHsuan-Hwang:
I think it may not be a good idea to make the settings implicit because we are only adding some of the settings to the preference dialog. If we make them implicit, some settings in the image view settings will remain when reopening CARTA but some will not. Users may not easily predict this behavior and may take time to be used to it.
I think it’s hard to decide if the user need a setting remain or not because it’s different from person to person and from case to case. It is more simple and clear to me to just use the preference dialog when necessary.

Issue #1169, opened 2020-08-26
set preferences implicitly 


kswang1029:
This is an idea about setting preferences automatically/implicitly when changing some of the options via the GUI. For example, the colormap dropdown would be one case that if we switch to other colormap such as jet in one session, effectively the default colormap in the preferences dialogue is changed so that in a new session, we will see jet as the default colormap. This means for this kind of options, there is no need to have them in the preferences dialogue and we can simplify the GUI a bit (we could consider to keep them too). Open for suggestions. If we think this is a good idea to proceed, please suggest which options can have this property. 

@arusstaylor @Jordatious @jott3077 @low-sky @veggiesaurus

veggiesaurus:
My personal list of what should be implicitly set:
- colormap
- render scale type (linear/log/etc)
- labels on/off
- grid on/off
- colormap legend on/off
- grid colours
- _possibly_ percentile (e.g. if you click 99.9%, the next image will open with 99.9% as well)

In fact, _most_ of the AST settings (title, colours, font types, font sizes, display options) would probably be best served by implicit preferences, rather than an entry in the preferences dialog. We'd still need a "reset to default" button, though

jott3077:
I think this is a great idea!  Maybe the most important one though is the placement and layout of the widgets. Possibly also settings in there, e.g whether a spectrum is shown as connected lines or histogram.

jott3077:
add to the list render histogram per channel or per cube (per cube is really more important but not the default right now due to performance)

veggiesaurus:
@Kechil is there anyone who could possibly get this done for 2.0? it would be really nice to sort this out sooner, rather than later, but it's a nice-to-have

low-sky:
I forgot to mention at original call but something I'm always doing is changing contour styling, so implicit preferences here would be good too.

kswang1029:
I suggest we set the following as implicit:

Global
======
Theme

Render Config
=============
Default scaling
Default colormap
Default percentile ranks

Contour Config
==============
Generator type
Thickness
Default colormap
Default color

Overlay Config
==============
AST color
AST grid visible
Beam color
Beam type
Beam width

Catalog
=======
Displayed columns

Region
======
Color
Line width
Dash length
Region type

Please let me know if there are others needed to be implicit.

low-sky:
[...]

PR #1542, opened 2021-07-02
Yuhsuan/1372 colorbar preference

YuHsuan-Hwang:
This PR is for #1372: adding options to the preference dialog.

I am not sure if we should make the settings implicit:
we are only adding some of the settings to the preference dialog. If we make them implicit, some settings in the image view settings will remain when reopening CARTA but some will not. Users may not easily predict this behavior and may take time to be used to it.
I think it’s hard to decide if the user need a setting remain or not because it’s different from person to person and from case to case. It is more simple and clear to me to just use the preference dialog when necessary.


veggiesaurus:
I think its fine to leave the implicit preference stuff to another feature, as we will need to be consistent across a number of image view settings

YuHsuan-Hwang:
> I think its fine to leave the implicit preference stuff to another feature, as we will need to be consistent across a number of image view settings

@veggiesaurus Ok. I think we can leave it to #1169

YuHsuan-Hwang:
@veggiesaurus @kswang1029 @duidae Changed to formal PR:
* added visible, interactive, position, width, ticks density, and label visible to the preference dialog
* same behavior with AST preference settings: apply the settings when opening / appending images

kswang1029 (approved):
this works fine as far as I can tell. 👍

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -317,6 +317,47 @@ export class PreferenceDialogComponent extends React.Component {
                         onChange={(event: React.FormEvent<HTMLSelectElement>) => preference.setPreference(PreferenceKeys.WCS_OVERLAY_WCS_TYPE, event.currentTarget.value)}
                     />
                 </FormGroup>
+                <FormGroup inline={true} label="Colorbar Visible">
+                    <Switch checked={preference.colorbarVisible} onChange={ev => preference.setPreference(PreferenceKeys.WCS_OVERLAY_COLORBAR_VISIBLE, ev.currentTarget.checked)} />
+                </FormGroup>
+                <FormGroup inline={true} label="Colorbar Interactive">
+                    <Switch checked={preference.colorbarInteractive} onChange={ev => preference.setPreference(PreferenceKeys.WCS_OVERLAY_COLORBAR_INTERACTIVE, ev.currentTarget.checked)} />
+                </FormGroup>
+                <FormGroup inline={true} label="Position">
```
veggiesaurus: To be consistent with the other labels, should this not be "Colorbar Position"?

veggiesaurus (changes requested):

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
YuHsuan-Hwang: @veggiesaurus Yes, thanks. This is changed.

Issue #1432, opened 2021-05-05
Optimize spatial matched catalog performance


panchyo0:
Use control maps to perform approximate transformation on the GPU.

PR #1548, opened 2021-07-07
qi/Issue 1432

panchyo0:
use control maps to perform approximate transformation on the GPU (issue #1432)


kswang1029 (approved):
this works nicely. I see a _**very significant**_ performance boost when matching a catalog (10M) to secondary images! It matches almost instantly. Also do the same with v2.0 and it hangs for a while to see the matched one. 👍 

 Also tested position accuracy and as far as I can tell it works well too. 👍

src/models/ControlMap.ts
```diff
@@ -13,6 +13,13 @@ export class ControlMap {
     private readonly grid: Float32Array;
     private texture: WebGLTexture;
     private gl: WebGLRenderingContext;
+    // catalog spatial matching for webgl2
+    minPoint2: Point2D;
```
veggiesaurus: @panchyo0 doesn't it make more sense to just use a separate control map (and perhaps a separate control map `Map`) for catalogs? The requirements are different: 
- The original control maps are required when viewing contours of one image on another
- The new control maps are required when viewing catalogs of one image on another

We could make a `CatalogControlMap` that extends `ControlMap`, that might make things neater

src/models/CatalogControlMap.ts
```diff
@@ -0,0 +1,101 @@
...
+
+            this.minPoint = {x: minX - paddingX, y: minY - paddingY};
+            this.maxPoint = {x: maxX + paddingX, y: maxY + paddingY};
+            this.grid = AST.getTransformGrid(astTransform, this.minPoint.x, this.maxPoint.x, this.width, this.minPoint.y, this.maxPoint.y, this.height, true);
+            
+            AST.deleteObject(astTransform);
+        }
+    }
+
+    getTextureX2 = (gl2: WebGL2RenderingContext) => {
```
veggiesaurus: renamve to `getTextureX`?

src/services/GLSL/vertex_shader_catalog.glsl
```diff
@@ -129,12 +138,75 @@ vec2 rotate2D(vec2 vector, float theta) {
...
+    float t3 = t * t * t;
+    vec4 a = -A / 2.0 + (3.0 * B) / 2.0 - (3.0 * C) / 2.0 + D / 2.0;
+    vec4 b = A - (5.0 * B) / 2.0 + 2.0 * C - D / 2.0;
+    vec4 c = -A / 2.0 + C / 2.0;
+    vec4 d = B;
+
+    return a * t3 + b * t2 + c * t + d;
+}
+
+vec4 bicubicFilter(sampler2D texture2, vec2 P) {
```
veggiesaurus, line 152: @panchyo0 any idea how we could prevent this sort of repetition between files? I guess glslify might help. Otherwise manually concatenating strings could work :thinking:

panchyo0: glslify looks like the right tech for creating a utilities.glsl

veggiesaurus (approved):

src/services/GLSL/vertex_shader_catalog.glsl
veggiesaurus: @panchyo0 I've made a separate issue for this: https://github.com/CARTAvis/carta-frontend/issues/1574

PR #1553, opened 2021-07-08
Angus/1508 snippets

### LINKED ISSUE #1508: Saving, loading and removing of code snippets
The user should be able to save and load code snippets

### LINKED ISSUE #1551: Better handling of situations where frontend javascript cannot be read
Due to server caching, sometimes `index.html` still points to an old JavaScript file after a new build. We need to display a message indicating that the user should attempt a hard reload if they still encounter this issue. Alternatively, we need to force the refresh ourselves somehow.

veggiesaurus:
Closes #1508. Code snippets feature is hidden behind the "enable code snippets" preference.

Additional tutorials will be handled in #1556.

Additional minor changes: 
- `setActiveFrame(id: number)` renamed `setActiveFrameById`. `setActiveFrame(frame: FrameStore)` introduced. ~~@confluence, can you adjust the python scripting interface accordingly. Should be a one-liner.~~
- interactive alerts now use async/await instead of callback functions.
- Closes #1551 with a bit of static HTML and embedded CSS that shows up while CARTA is loading (see below, emulating slow internet)

https://user-images.githubusercontent.com/592504/126149843-bcc02376-f27b-4359-9d6b-10d475b1d780.mp4


kswang1029:
suggestions for improvements:

1. when we click the "New" button in the snippets dialog, we should ask users to save or discard the snippets in the current editor.
2. the overwrite confirmation popup shows _behind_ the save popup.

https://user-images.githubusercontent.com/20819712/125237324-153d6f00-e318-11eb-9e9b-0a8961001635.mov

3. (no standard use case🙈) if we keep the enter key pressed to create new lines, at some point (~1000 lines with my laptop) we will see it slows down (ie not seeing each new line being created, instead it jumps every few lines)


questions:
1. should we consider to use a widget instead of a dialog? with a widget it might be easier to write a script by referencing the GUI _interactively_. With a dialog, the rest of the GUI is blocked (for example, we cannot see what categories that have been created from the snippets menu. BUT if we do so, we can have multiple snippets widgets activated at the same time.
2. what's the purpose of having "tag" and where to see it other than the save popup?

veggiesaurus:
> 1. when we click the "New" button in the snippets dialog, we should ask users to save or discard the snippets in the current editor.

I'll add a confirmation if the current snippet has not been saved

> 2. the overwrite confirmation popup shows _behind_ the save popup.

will fix that 🙈 

> 3. (no standard use case🙈) if we keep the enter key pressed to create new lines, at some point (~1000 lines with my laptop) we will see it slows down (ie not seeing each new line being created, instead it jumps every few lines)

I think this is an issue with the react-simple-code-editor component. We can file an issue there, I guess.

> 1. should we consider to use a widget instead of a dialog? with a widget it might be easier to write a script by referencing the GUI _interactively_. With a dialog, the rest of the GUI is blocked (for example, we cannot see what categories that have been created from the snippets menu. BUT if we do so, we can have multiple snippets widgets activated at the same time.

It's only blocked when you're in the save dialog. You can still interact with the rest of the GUI while writing code.

> 2. what's the purpose of having "tag" and where to see it other than the save popup?

Future use. But I think we should remove this input until we're actually. making use of this field

@kswang1029 I can't reproduce (2) on mac (chrome, safari) or linux. What were using to test?

kswang1029:
> @kswang1029 I can't reproduce (2) on mac (chrome, safari) or linux. What were using to test?

Yes, it is mac+chrome. I tried to make a custom category, save a custom script, then modify it and try to save again with same name, tag, and category.

veggiesaurus:
> > @kswang1029 I can't reproduce (2) on mac (chrome, safari) or linux. What were using to test?
> 
> Yes, it is mac+chrome. I tried to make a custom category, save a custom script, then modify it and try to save again with same name, tag, and category.

Ok, I can reproduce it now. I think I know what it is.. a z-index issue

veggiesaurus:
@kswang1029 I've resolved (1) and (3) above, and removed the tag input from the save dialog. Please re-review.

kswang1029:
> @kswang1029 I've resolved (1) and (3) above, and removed the tag input from the save dialog. Please re-review.

Tested again with a4679bf 
issue 1: fixed 👍 
issue 2: remains. It appears that initially the z order is correct but later on it is not.

https://user-images.githubusercontent.com/20819712/125717402-5e81a2b9-ec54-41a3-8294-17a330b621ae.mov

issue 3: remains. 

https://user-images.githubusercontent.com/20819712/125717455-7077c956-aac6-4e57-86fa-87934c8ed567.mov

veggiesaurus:
@kswang1029 (2) should be fixed now. (3) will need to be handled by an issue in the `react-simple-code-editor` repo. Doesn't seem urgent to me

YuHsuan-Hwang (commented):
Just found one detail: preference `Enable Code Snippets` need to response to the `Restore defaults` button

veggiesaurus:
> Just found one detail: preference `Enable Code Snippets` need to response to the `Restore defaults` button

@YuHsuan-Hwang fixed, please re-review

YuHsuan-Hwang (approved):

Issue #1484, opened 2021-06-08
Cursor widget to show cursor information for matched images


kswang1029:
When there are multiple spatially and spectrally matched images, it is common to read the cursor information from all the images _in one place at the same time_. The information should include: 
- World coordinates
- Image coordinates of each image
- Spectral coordinate of each image including frequency, velocity, and channel index
- Stokes parameter of each image
- Pixel value with a unit for each image

Please let @kswang1029 know if a UI mockup is needed.

YuHsuan-Hwang:
Separating this issue into two parts
1. implement basic UI and show cursor info of the active frame
2. show cursor info of multiple images
2.1. show explicit WCS for all frames
2.2. update cursor info of matched images when hovering

veggiesaurus:
@YuHsuan-Hwang do you want to create a separate issue for the multi-cursor implementation, or just use this one?

YuHsuan-Hwang:
> @YuHsuan-Hwang do you want to create a separate issue for the multi-cursor implementation, or just use this one?

@veggiesaurus I think we can just use this one

PR #1555, opened 2021-07-09
Yuhsuan/1484 cursor widget

YuHsuan-Hwang:
This PR is for the first part of https://github.com/CARTAvis/carta-frontend/issues/1484: add basic UI and show cursor info of the active frame.
The second part (multi image cursor info) will be implemented in a separate PR.


kswang1029:
We could consider to show the pixel unit in a new line to save space with a narrower default column width
Screen Shot 2021-07-12 at 1 14 17 PM

kswang1029:
We should show Stokes string instead of a numeric index. 

Screen Shot 2021-07-12 at 1 18 33 PM

YuHsuan-Hwang:
> We could consider to show the pixel unit in a new line to save space with a narrower default column width
> Screen Shot 2021-07-12 at 1 14 17 PM

Ok. I will try if this saves more space.

YuHsuan-Hwang:
> We should show Stokes string instead of a numeric index.
> 
> Screen Shot 2021-07-12 at 1 18 33 PM

I followed the stokes column in the image list widget. Maybe we open a new issue for this?

kswang1029:
> > We should show Stokes string instead of a numeric index.
> > Screen Shot 2021-07-12 at 1 18 33 PM
> 
> I followed the stokes column in the image list widget. Maybe we open a new issue for this?

ok https://github.com/CARTAvis/carta-frontend/issues/1562

YuHsuan-Hwang:
image
@kswang1029 Tried Value (140px) -> Value (90px) and Unit (60px). I think it's not saving space because of padding, so I prefer one column.

kswang1029:
> image
> @kswang1029 Tried Value (140px) -> Value (90px) and Unit (60px). I think it's not saving space because of padding, so I prefer one column.

Oh what I mean is that we use one column but have the unit in a new line

YuHsuan-Hwang:
> > image
> > @kswang1029 Tried Value (140px) -> Value (90px) and Unit (60px). I think it's not saving space because of padding, so I prefer one column.
> 
> Oh what I mean is that we use one column but have the unit in a new line

@kswang1029 I see... Changed to two lines

kswang1029 (approved):

src/components/CursorInfo/CursorInfoComponent.tsx
```diff
@@ -0,0 +1,139 @@
...
+        imageCoords[activeFrameIndex] = (
+            <React.Fragment>
+                {toFixed(frame.cursorInfo?.posImageSpace?.x, 3)}
+                <br />
+                {toFixed(frame.cursorInfo?.posImageSpace?.y, 3)}
+            </React.Fragment>
+        );
+
+        let columnsData = new Map<number, any>();
+        columnsData.set(0, {dataType, data: imageNames});
```
veggiesaurus: elsewhere in the codebase we use the `Map` array constructor (e.g. `new Map([[key1, val1], [key2, val2]])`. I'm not sure if that would be neater or not 🤔

src/components/Shared/Tables/SimpleTableComponent.tsx
```diff
@@ -1,20 +1,36 @@
 import * as React from "react";
 import {observer} from "mobx-react";
-import {Cell, Column, Table, SelectionModes, RenderMode} from "@blueprintjs/table";
+import {Cell, Column, Table, SelectionModes, RenderMode, RowHeaderCell} from "@blueprintjs/table";
 import {CARTA} from "carta-protobuf";
-import {ProcessedColumnData} from "models";
+//import {ProcessedColumnData} from "models";
```
veggiesaurus: remove comment

src/components/Shared/Tables/SimpleTableComponent.tsx
```diff
@@ -1,20 +1,36 @@
...
+    defaultRowHeight?: number;
+    enableGhostCells?: boolean;
+    isIndexZero?: boolean;
     updateTableRef?: (ref: Table) => void;
 }
 
 @observer
 export class SimpleTableComponent extends React.Component<SimpleTableComponentProps> {
-    private renderDataColumn(columnName: string, columnData: any) {
+    private widths = this.props.defaultColumnWidths;
```
veggiesaurus: does this work properly? shouldn't this be done in `constructor`?

veggiesaurus (changes requested):
couple of minor queries

src/components/CursorInfo/CursorInfoComponent.tsx
YuHsuan-Hwang: I see. I think it's neater to use `new Map([[key1, val1], [key2, val2]])`. I followed code in https://github.com/CARTAvis/carta-frontend/blob/dev/src/stores/FileBrowserStore.ts#L531. Maybe I also change these lines?

src/components/Shared/Tables/SimpleTableComponent.tsx
YuHsuan-Hwang: I think both works, but more straightforward to initialize it in constructor. I will adjust this part.

src/components/CursorInfo/CursorInfoComponent.tsx
veggiesaurus: Yes please 😊

YuHsuan-Hwang:
@kswang1029 @veggiesaurus @duidae 
Several new changes:
* Show one line in the value column if there's no unit
* Avoid the column widths changing when switching tabs: moved the variables from SimpleTableComponent.tsx to CursorInfoComponent.tsx.
* Avoid catalog header in the import catalog dialog crashing
* Add the widget to layout config

src/components/Shared/Tables/SimpleTableComponent.tsx
YuHsuan-Hwang: moved the variables to CursorInfoComponent.tsx so the issue no longer exists

src/components/CursorInfo/CursorInfoComponent.tsx
YuHsuan-Hwang: Changed both parts

kswang1029:
Two suggestions: 
1) please add a placeholder for the in-app help manual (in this PR please)
2) consider to use bold face text for the active image just like the image list widget (in this or next PR)

YuHsuan-Hwang:
> Two suggestions:
> 
> 1. please add a placeholder for the in-app help manual (in this PR please)
> 2. consider to use bold face text for the active image just like the image list widget (in this or next PR)

1. Ok. I will add a placeholder 2. I think we can leave it to the next PR

veggiesaurus (approved):

YuHsuan-Hwang:
@kswang1029 @veggiesaurus Added a placeholder for help manual

Issue #1426, opened 2021-04-27
Optimize catalog map calculation


panchyo0:
calculate catalog map(like size map) takes about 0.3s for 1M data source.  The bottleneck is `SizeMapData` as it taking 0.28s.

PR #1558, opened 2021-07-10
qi/issue_1426

panchyo0:
related issue #1426
For 1 million data points

> The render time (each 100K points) is improved from 600ms to 250ms.

> The time for calculating catalog map is improved from 280ms to 25ms.


kswang1029 (approved):
Tested with a 5M catalog and I can see significant performance boost when applying size/color/orientation mapping. 👍

veggiesaurus (approved):

PR #1564, opened 2021-07-13
Yuhsuan/1535 distance in matched images

### LINKED ISSUE #1535: distance tool on spatially matched image behaving unexpectedly
It appears that the coordinate transformation is not applied properly for the spatially matched image so that the distance calculation is not correct and varies with different zoom/pan actions.

YuHsuan-Hwang:
1. Closes #1535: show the curves at the right world coordinate whenever we match/unmatch the images.
2. Fixed shortcut bug: switch from distance measuring mode to region creation mode when pressing "c" shortcut

edit: also fixed
switching between region creation mode and pan mode using "c" shortcut;
switching to pan mode using "esc" shortcut


YuHsuan-Hwang:
Fixing major bug in 1.

YuHsuan-Hwang:
@kswang1029 @veggiesaurus @duidae Fixed major bug and changed to formal PR

kswang1029 (approved):

src/stores/OverlayStore.ts
```diff
@@ -96,6 +96,7 @@ export class OverlayGlobalSettings {
         astString.add("Labelling", this.labelType);
         astString.add("Color", AstColorsIndex.GLOBAL);
         astString.add("Tol", toFixed(this.tolerance / 100, 2), this.tolerance >= 0.001); // convert to fraction
+        astString.add("System", this.explicitSystem);
```
veggiesaurus, line 99: does this work ok in situations where `this.explicitSystem` is undefined?

YuHsuan-Hwang: Yes, `astString.add()` checks if the value is undefined

veggiesaurus (approved):

PR #1569, opened 2021-07-14
Angus/1557 package maintenance

### LINKED ISSUE #1557: v3.0-beta.1 package maintenace
Easier to perform package updates for each beta cycle, rather than all at once, as there will be fewer breaking changes. Updates should be done well before code freeze

veggiesaurus:
Closes #1557.

No major changes other than package versions, but @kswang1029 please test things to make sure there's no odd behaviour.


kswang1029:
@veggiesaurus I did a very brief check based on what have been updated as far as I can try. This seems okay.

duidae (approved):

PR #1575, opened 2021-07-16
ajm/adding ci

### LINKED ISSUE #1510: Introduce basic CI checks
We need some CI checks for the frontend:
- build test with node v12, v14 and v16 (can be built on any platform, and should use docker for the EMCC parts, and should be as simple as `npm install; npm run build-libs-docker; npm pack;`)
- formatting check (`npm run checkformat`)

ajm-ska:
@veggiesaurus I almost have a basic version working now.
Some points to note:

1. It reports the green tick or red x back to Github. The developer can click on it to see a summary of each stage, or click "Details" to proceed to Jenkins for full build logs.

2. I needed to modify the `build_wasm_libs_docker.sh` and `build_wasm_wrappers_docker.sh` to remove the `-it` flag because Jenkins does not have a TTY. I found a handy way on StackOverflow to only run it with `-it` if a TTY is present. The scripts will behave just the same as before for users running it with a TTY.

3. I had a lot of trouble trying to get Jenkins to use `nvm`. In the end I have changed to `n` and it seems to work fine.

4. It is currently using our existing ubuntu server to run each stage sequentially. Once our Lustre system upgrade is complete, maybe I will be able to add more servers where we can have a dedicated server for each 'node' version and have them run in parallel.

5. You can see the stages and steps in the Jenkinsfile.
Stage 1: Checkout submodule folder, npm install, runs checkformat.
Stage 2: Runs the WebAssembly compilation in Docker.
Stage 3: Removes node_modules and build folder, switches to node v12, npm install, and then npm run build-docker.
Stage 4: Removes node_modules and build folder, switches to node v14, npm install, and then npm run build-docker.
Stage 5: Removes node_modules and build folder, switches to node v16, npm install, and then npm run build-docker.
Is that OK?

One issue remaining:
You can see there is a format check warning with `src/static/gitInfo.ts`.

> Checking formatting...
> [warn] src/static/gitInfo.ts
> [warn] Code style issues found in the above file(s). Forgot to run Prettier?

That file contains Github commit information:

> const info = {logMessage: "6e82e17 adding node v14 and v16 stages"}; export default info;

Should we do an `rm src/static/gitInfo.ts` before the checkformat command, or could we add that file to an ignore list?

### DIFF
--- a/build_wasm_wrappers_docker.sh
+++ b/build_wasm_wrappers_docker.sh
@@ -1,8 +1,3 @@
 #!/bin/bash
 echo "Building WebAssembly wrappers inside Docker container"
-if [ -t 0 ] && [ -t 1 ]; then
-  docker run -it --rm -u $(id -u):$(id -g) -v `pwd`:/src emscripten/emsdk:2.0.14 /src/wasm_src/build_wrappers.sh
-else
-  # Non interactive version needed by Jenkins as no TTY is available 
-  docker run --rm -u $(id -u):$(id -g) -v `pwd`:/src emscripten/emsdk:2.0.14 /src/wasm_src/build_wrappers.sh
-fi
+docker run --rm -u $(id -u):$(id -g) -v `pwd`:/src emscripten/emsdk:2.0.14 /src/wasm_src/build_wrappers.sh
--- a/build_wasm_wrappers_docker.sh
+++ b/build_wasm_wrappers_docker.sh
@@ -1,3 +1,8 @@
 #!/bin/bash
 echo "Building WebAssembly wrappers inside Docker container"
-docker run -it --rm -u $(id -u):$(id -g) -v `pwd`:/src emscripten/emsdk:2.0.14 /src/wasm_src/build_wrappers.sh
+if [ -t 0 ] && [ -t 1 ]; then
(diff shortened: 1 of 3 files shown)


veggiesaurus:
> Should we do an `rm src/static/gitInfo.ts` before the checkformat command, or could we add that file to an ignore list?

Merge `dev` into this branch. I've added a `.prettierignore` for doing that.

build_wasm_wrappers_docker.sh
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
 echo "Building WebAssembly wrappers inside Docker container"
-docker run -it --rm -u $(id -u):$(id -g) -v `pwd`:/src emscripten/emsdk:2.0.14 /src/wasm_src/build_wrappers.sh
+if [ -t 0 ] && [ -t 1 ]; then
+  docker run -it --rm -u $(id -u):$(id -g) -v `pwd`:/src emscripten/emsdk:2.0.14 /src/wasm_src/build_wrappers.sh
+else
+  # Non interactive version needed by Jenkins as no TTY is available 
+  docker run --rm -u $(id -u):$(id -g) -v `pwd`:/src emscripten/emsdk:2.0.14 /src/wasm_src/build_wrappers.sh
```
veggiesaurus: is there any reason why we should continue to use `-it` under normal circumstances?

veggiesaurus: wait, nevermind. If we don't have `-it` we don't get output, right?

PR #1577, opened 2021-07-16
Jason/issue 1429

### LINKED ISSUE #1429: progress indicator of the spectral profiler

duidae:
closes #1429


kswang1029 (approved):
this works nicely. 👍

src/stores/AppStore.ts
```diff
@@ -2100,4 +2101,17 @@ export class AppStore {
...
     // endregion
+
+    // Reset spectral profile's progress to 0 instead of cleaning the entire out-dated profile to avoid flashy effect in spectral profiler.
+    // Flashy effect: render empty profile and then render the coming profile, repeatedly.
+    public resetCursorRegionSpectralProfileProgress = (fileId: FileId) => {
+        this.spectralProfiles.get(fileId)?.get(CURSOR_REGION_ID)?.resetProfilesProgress();
+    };
+
+    public resetRegionSpectralProfileProgress = (regionId: RegionId) => {
+        const regionProfileStoreMaps = Array.from(this.spectralProfiles?.values());
```
veggiesaurus: you can use a `for` loop here, looks neater:
`for (const [fileId, profile] of this.spectralProfiles)`

Otherwise, can't we also use `this.spectralProfiles?.forEach`?

src/stores/SpectralProfileStore.ts
```diff
@@ -24,8 +24,14 @@ export class SpectralProfileStore {
         return null;
     };
 
-    @action clearProfiles = () => {
-        this.profiles = new Map<string, ObservableMap<CARTA.StatsType, ProcessedSpectralProfile>>();
+    public resetProfilesProgress = () => {
+        this.profiles.forEach(statsProfilesMap => {
```
veggiesaurus, line 28: `this.profiles` is an observable, so shouldn't this still be an `@action` to avoid MobX warnings?

veggiesaurus (changes requested):

veggiesaurus (approved):

PR #1578, opened 2021-07-16
Jason/issue 1130

### LINKED ISSUE #1130: ping pong test for spectral line query
ping pong test to check the status of splatalogue service is available or not initially

check website using curl:
https://www.2daygeek.com/linux-command-check-website-is-up-down-alive/

duidae:
closes #1130, check Splatalogue when widget is loaded.
need companion PR https://github.com/CARTAvis/carta-backend/pull/870

P.S.
Take mature science DB iNaturalist (https://api.inaturalist.org/v1/docs/) for example, restrictions for API connection are specified in their Term of Use:

`Please note that we throttle API usage to a max of 100 requests per minute, though we ask that you try to keep it to 60 requests per minute or lower, and to keep under 10,000 requests per day. If we notice usage that has serious impact on our performance we may institute blocks without notification.`


src/stores/AppStore.ts
```diff
@@ -1304,19 +1304,18 @@ export class AppStore {
             this.apiService.setToken(authTokenParam);
         }
 
-        autorun(() => {
+        autorun(async () => {
             if (this.astReady && this.zfpReady && this.cartaComputeReady && this.apiService.authenticated) {
+                await this.connectToServer();
```
veggiesaurus: Do Async functions work ok with autorun?

veggiesaurus:
I tried to do this with the frontend, and it was elegant and simple. But it doesn't work, due to CORS 🙄 I really wish we could fix that! it makes so much more sense getting results from the frontend.

duidae:
> I tried to do this with the frontend, and it was elegant and simple. But it doesn't work, due to CORS  I really wish we could fix that! it makes so much more sense getting results from the frontend.

yeah the underneath principle of CORS policy is that browser will NEVER allow application to arbitrarily connect to sites over it to prevent unpredictable or even malicious behaviors, pretty strong restriction but also make sense.

src/stores/AppStore.ts
duidae: thanks for spotting this, fixed & please have a look.(ref: https://github.com/mobxjs/mobx/issues/1003)

veggiesaurus:
> yeah the underneath principle of CORS policy is that browser will NEVER allow application to arbitrarily connect to sites over it to prevent unpredictable or even malicious behaviors, pretty strong restriction but also make sense.

I thought there was a workaround by getting "opaque" responses, which would have been enough for an alive-check. Alas, it seems not to be the case.

duidae:
> > yeah the underneath principle of CORS policy is that browser will NEVER allow application to arbitrarily connect to sites over it to prevent unpredictable or even malicious behaviors, pretty strong restriction but also make sense.
> 
> I thought there was a workaround by getting "opaque" responses, which would have been enough for an alive-check. Alas, it seems not to be the case.

the problem is getting `404 not found` is also getting something, still need specific `HTTP 200 OK` to guarantee splatalogue is functional.

kswang1029:
This works fine. Just few minor suggestions:
1. please add a space before (
Screen Shot 2021-07-20 at 1 21 24 PM

2. The messages might be revised as the following:
Screen Shot 2021-07-20 at 1 24 18 PM

> Connecting to Splatalogue failed! The service might be offline or your internet connection is interrupted. Please check and try again.

duidae:
> This works fine. Just few minor suggestions:
> 
> 1. please add a space before (
>    Screen Shot 2021-07-20 at 1 21 24 PM
> 2. The messages might be revised as the following:
>    Screen Shot 2021-07-20 at 1 24 18 PM
> 
> > Connecting to Splatalogue failed! The service might be offline or your internet connection is interrupted. Please check and try again.

msg updated, please have a look, thanks.

kswang1029 (approved):
nice 👍

kswang1029:
As a note here: once there is an offline catalog for query, the current ping-pong mechanism needs to be adjusted otherwise if splatalogue is not available, the UI is blocked for the offline catalog.

src/components/SpectralLineQuery/SpectralLineQuerySplashScreenComponent.scss
```diff
@@ -0,0 +1,29 @@
+@import "../../../node_modules/@blueprintjs/core/lib/scss/variables.scss";
```
veggiesaurus: use `~@blueprintjs`

src/stores/AppStore.ts
```diff
@@ -1328,23 +1347,7 @@ export class AppStore {
...
-                            this.preferenceStore.setPreference(PreferenceKeys.GLOBAL_LAYOUT, PresetLayout.DEFAULT);
-                        }
-                        this.cursorFrozen = this.preferenceStore.isCursorFrozen;
-                        this.connectToServer();
-                    });
-                    this.snippetStore.fetchSnippets();
-                    this.updateASTColors();
-                });
-            }
+            this.initCarta(this.astReady, this.tileService?.zfpReady, this.cartaComputeReady, this.apiService?.authenticated);
```
veggiesaurus, line 1350: nice workaround!

src/stores/widgets/SpectralLineQueryWidgetStore.ts
```diff
@@ -550,9 +557,22 @@ export class SpectralLineQueryWidgetStore extends RegionWidgetStore {
         }
     };
 
+    public pingSplatalogue = async () => {
+        try {
+            this.splataloguePingStatus = SplataloguePingStatus.Checking;
```
veggiesaurus, line 562: please make a `setPingStatus()` `@action` to avoid MobX strict mode warnings

veggiesaurus (changes requested):
couple of minor adjustments

veggiesaurus (approved):

PR #1580, opened 2021-07-20
Add handling for undefined value

### LINKED ISSUE #1563: spectral conversion failed
after the new FITS header parser improvement https://github.com/CARTAvis/carta-backend/pull/827 the supermosaic image has a problem showing velocity info in the cursor info bar. 

Screen Shot 2021-07-12 at 5 29 25 PM

duidae:
closes #1563

### DIFF
--- a/src/stores/FrameStore.ts
+++ b/src/stores/FrameStore.ts
@@ -425,7 +425,7 @@ export class FrameStore {
             const channelInfo = this.channelInfo;
             const spectralType = this.spectralAxis?.type;
             if (spectralType) {
-                spectralInfo.spectralString = `${spectralType.name} (${this.spectralAxis?.specsys}): ${toFixed(channelInfo.values[this.channel], 4)} ${spectralType.unit}`;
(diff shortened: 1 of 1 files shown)


veggiesaurus (approved):

kswang1029:
if we load the supermosaic image, make a region and launch the spectral profiler widget, the frontend will crash.
Screen Shot 2021-07-21 at 11 09 20 AM

kswang1029:
if the spectral unit (CUNIT3) is not explicitly defined in the header, shall we "apply" a default unit for it? in the example of supermosaic image, CTYPE3 is VELO-LSR, so the default unit should be m/s. So at least we will be able to view the spectral profile with proper xlabels and be able to switch to equivalent units as m/s, km/s and so on.  

For frequency we can have the default as Hz. For wavelength, we use m. That is, we use "SI" unit as default. 

As a note here: in the very beginning of the CARTA project, we faced the same issue with the FITS headers which may not follow the standard or some header info is missing. Then we decided to use casacore as a pre-processor to clean up the header before feeding it to AST. Although this seems working fine in most of cases, there are cases that casacore is too aggressive in changing the headers. So that's why we switch back to the way that we feed the native FITS header to AST. It is expected that there will be some other issues (aka exceptions in headers) due to this change.

PR #1581, opened 2021-07-20
Angus/additional preferences

veggiesaurus:
Adds the following preferences:

- Performance -> Limit overlay redraw (defaults to `true`): This controls whether or not we throttle AST overlay redraws to 20 fps.
- Overlay -> Cursor info visible (defaults to `"always"`): This controls visibility of the cursor overlay. There are four options: `always` and `never` are selft-explanatory. `activeImage` only displays the cursor info on the active image, and `hideTiled` hides the cursor overaly whenever you use a tiled view. The last two options will only make a difference once #1259 is implemented.

Also adjusted the preferences text slightly, to remove reference to AST.

### DIFF
--- a/src/stores/PreferenceStore.ts
+++ b/src/stores/PreferenceStore.ts
@@ -2,7 +2,7 @@ import {action, computed, observable, makeObservable} from "mobx";
 import {Colors} from "@blueprintjs/core";
 import {CARTA} from "carta-protobuf";
 import {BeamType, ContourGeneratorType, FileFilteringType, FrameScaling} from "stores";
-import {CompressionQuality, CursorPosition, Event, PresetLayout, RegionCreationMode, SpectralType, Theme, TileCache, WCSMatchingType, WCSType, Zoom, ZoomPoint} from "models";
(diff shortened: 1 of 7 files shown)


kswang1029 (approved):

duidae (approved):

src/stores/PreferenceStore.ts
```diff
@@ -657,7 +671,8 @@ export class PreferenceStore {
                 PreferenceKeys.REGION_CREATION_MODE,
                 PreferenceKeys.WCS_OVERLAY_AST_COLOR,
                 PreferenceKeys.CATALOG_TABLE_SEPARATOR_POSITION,
-                PreferenceKeys.PIXEL_GRID_COLOR
+                PreferenceKeys.PIXEL_GRID_COLOR,
+                PreferenceKeys.WCS_OVERLAY_CURSOR_INFO
```
YuHsuan-Hwang: Do we also add `PERFORMANCE_LIMIT_OVERLAY_REDRAW` in `booleanKeys`?

veggiesaurus: those are only needed for older preferences that we need to upgrade. But that means I can remove `Preferences.WCS_OVERLAY_CURSOR_INFO` from that array as well :) I've added a comment that clarifies this as well

PR #1583, opened 2021-07-23
Yuhsuan/1484 multi cursor info

### LINKED ISSUE #1484: Cursor widget to show cursor information for matched images

YuHsuan-Hwang:
This PR closes #1484: showing cursor info of spatially matched images

Some topics we may want to add new issues for:
1. For spatially matched images, we request for new data only when the cursor moves to a new pixel on the active frame. In this way, the cursor values of non active frames sometimes remain a star symbol
2. I think we can remove the cursor region in the region list widget since it's showing the same information
3. For spatially matched images, the image coord. of the cursor info show `-1.7976931348623157e+308` when hovering out of the world coord. range


veggiesaurus:
> 3\. For spatially matched images, the image coord. of the cursor info show `-1.7976931348623157e+308` when hovering out of the world coord. range

That sounds like the `AST_BAD` value 🤔, which is the same as `-Number.MAX_VALUE`

YuHsuan-Hwang:
> > 3. For spatially matched images, the image coord. of the cursor info show `-1.7976931348623157e+308` when hovering out of the world coord. range
> 
> That sounds like the `AST_BAD` value 🤔, which is the same as `-Number.MAX_VALUE`

I see. I think we can hide this number

veggiesaurus:
> > > 1. For spatially matched images, the image coord. of the cursor info show `-1.7976931348623157e+308` when hovering out of the world coord. range
> > 
> > 
> > That sounds like the `AST_BAD` value thinking, which is the same as `-Number.MAX_VALUE`
> 
> I see. I think we can hide this number

Yes, please hide it

YuHsuan-Hwang:
> > > > 1. For spatially matched images, the image coord. of the cursor info show `-1.7976931348623157e+308` when hovering out of the world coord. range
> > > 
> > > 
> > > That sounds like the `AST_BAD` value thinking, which is the same as `-Number.MAX_VALUE`
> > 
> > 
> > I see. I think we can hide this number
> 
> Yes, please hide it

Sure, but I suggest we file a new issue for it since it also changes the cursor overlay

kswang1029:
~~Regarding issue 1: what are the current update criteria? is it "throttle_time AND new_integer_pixel"? If so, I guess this issue could be fixed by changing AND to OR?~~ @YuHsuan-Hwang @veggiesaurus 

see #1587

YuHsuan-Hwang:
issue 2: found that there is cursor option in the z profiler region dropdown, so I think we can leave it unchanged
issue 3: filed #1588

kswang1029 (approved):

veggiesaurus (approved):

PR #1592, opened 2021-07-30
Yuhsuan/1149 polyline region

### LINKED ISSUE #1149: frontend support region type: line segment

### LINKED ISSUE #1589: Polygon region creation bug
https://user-images.githubusercontent.com/43841102/127272753-d5f65b21-ca28-451f-b589-a48f34b2dc75.mov

Double clicking creates an extra creation point of a polygon region.
~~Occurs on ubuntu but works fine on macOS.~~ Seems to occur after #1569 is merged. May due to the upgrade of the Konva package.

edit: occurs on both ubuntu and macOS

YuHsuan-Hwang:
This PR closes #1149: added polyline region type
Other changes:
* Simplified the code of region buttons by iterating through `AVAILABLE_REGION_TYPES`
* Fixed #1589: set a timeout to clearing creating region so that the mouse up event of a double click will not trigger new region creation

Issues:
* The backend seems to crash (`segmentation fault`) when exporting polyline regions
* Opening the region dialog by double clicking the polyline regions easily creates a new control point. Shall we avoid this when double clicking?


kswang1029:
@YuHsuan-Hwang regarding the issue "Opening the region dialog by double clicking the polyline regions easily creates a new control point. Shall we avoid this when double clicking?" Is it possible to treat the polyline as a polygon with the line segment connecting the 1st and the last points rendered as invisible? This way I guess it is less confusing with adding a control point. 

At the moment, if we double-click inside a polygon (the grey area), the region dialogue is launched. For polyline, the "sensor area" becomes much smaller around the polyline. So the idea is to see if we can treat the polyline as a polygon so that the effective "sensor area" is enlarged.

Screen Shot 2021-08-04 at 3 41 15 PM

veggiesaurus:
> @YuHsuan-Hwang regarding the issue "Opening the region dialog by double clicking the polyline regions easily creates a new control point. Shall we avoid this when double clicking?" Is it possible to treat the polyline as a polygon with the line segment connecting the 1st and the last points rendered as invisible? This way I guess it is less confusing with adding a control point.
> 
> At the moment, if we double-click inside a polygon (the grey area), the region dialogue is launched. For polyline, the "sensor area" becomes much smaller around the polyline. So the idea is to see if we can treat the polyline as a polygon so that the "sensor area" is enlarged.
> 
> Screen Shot 2021-08-04 at 3 41 15 PM

could we not just make the "sensor area" a much wider poly-line?

kswang1029:
> > @YuHsuan-Hwang regarding the issue "Opening the region dialog by double clicking the polyline regions easily creates a new control point. Shall we avoid this when double clicking?" Is it possible to treat the polyline as a polygon with the line segment connecting the 1st and the last points rendered as invisible? This way I guess it is less confusing with adding a control point.
> > At the moment, if we double-click inside a polygon (the grey area), the region dialogue is launched. For polyline, the "sensor area" becomes much smaller around the polyline. So the idea is to see if we can treat the polyline as a polygon so that the "sensor area" is enlarged.
> > Screen Shot 2021-08-04 at 3 41 15 PM
> 
> could we not just make the "sensor area" a much wider poly-line?

That would work too. There will be two sensors, one for creating a new control point, and another _wider_ one for dialog.

YuHsuan-Hwang:
@kswang1029 @veggiesaurus I think avoiding single click event when double clicking may be a simpler solution. One concern is that we have the same issue when a polygon region is small. Users create points when double clicking inside the polygon

kswang1029:
regarding the backend crash issue, please refer to https://github.com/CARTAvis/carta-backend/issues/891

veggiesaurus:
> @kswang1029 @veggiesaurus I think avoiding single click event when double clicking may be a simpler solution.

Yes, that's a good idea, please do that. The rest of the code is fine.

veggiesaurus (approved):
the double-click avoidance works fine :+1: 

Once @kswang1029 has tested this, we can merge it in

YuHsuan-Hwang:
We can decide whether to include this in v3b1, maybe wait until the backend crash issue is closed

kswang1029 (approved):

PR #1595, opened 2021-08-03
Yuhsuan/1562 stokes string

### LINKED ISSUE #1562: show Stokes/polarization string instead of index
to be fixed in 

- [ ] cursor info widget (once it is merged)
- [ ] image list widget

YuHsuan-Hwang:
Fixes #1562: show stokes string in image list and cursor info widgets

### DIFF
--- a/src/components/Animator/AnimatorComponent.tsx
+++ b/src/components/Animator/AnimatorComponent.tsx
@@ -305,7 +305,7 @@ export class AnimatorComponent extends React.Component<WidgetProps> {
                                 showTrackFill={false}
                                 max={activeFrame.frameInfo.fileInfoExtended.stokes - 1}
                                 labelRenderer={(val: number) => {
-                                    return isFinite(val) && val >= 0 && val < activeFrame?.stokesInfo?.length ? `Stokes ${activeFrame.stokesInfo[val]}` : `${val}`;
+                                    return isFinite(val) && val >= 0 && val < activeFrame?.stokesInfo?.length ? activeFrame.stokesInfo[val] : `${val}`;
                                 }}
                                 onChange={this.onStokesChanged}
                                 disabled={appStore.animatorStore.animationActive}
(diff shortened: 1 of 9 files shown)


YuHsuan-Hwang:
@veggiesaurus @duidae 
Additional changes suggested by @kswang1029:
* Besides Stokes IQUV, allow all kinds of stokes value, e.g. XX
* For image list widget, cursor info widget, and cursor overlay, change the format of `Stokes: I` to `Polarization: Stokes I` or `Polarization: XX`

kswang1029:
@YuHsuan-Hwang this looks good mostly. I spotted few other places that need to be fixed:
1. we will need to update the "Stokes" label in the spectral profiler to "Polarization"

Screen Shot 2021-08-05 at 11 01 13 AM

2. file info: "number of Stokes" -> "number of polarization". This requires backend change I think so can be handled separately. 

Screen Shot 2021-08-05 at 11 02 45 AM


3. if we load a 4D cube with the Stokes axis of unity length. Widgets show its polarization label (which is great) but the label is not displayed in the cursor info of the image viewer. Could you also fix this too for consistency. Can be in a separate PR if you think proper. 

Screen Shot 2021-08-05 at 11 05 38 AM

YuHsuan-Hwang:
@kswang1029 I can fix 1 and 3 in this PR, thanks for checking
For 2, seems like we should file a new backend issue
For 3, currently we don't show stokes info when the axis is unity length, regardless of 4D or 3D images. I can change this for both 4D and 3D images for consistency

kswang1029:
> @kswang1029 I can fix 1 and 3 in this PR, thanks for checking
> For 2, seems like we should file a new backend issue
> For 3, currently we don't show stokes info when the axis is unity length, regardless of 4D or 3D images. I can change this for both 4D and 3D images for consistency

for 2: https://github.com/CARTAvis/carta-backend/issues/893

YuHsuan-Hwang:
@kswang1029 @veggiesaurus @duidae

New changes:
* Fixed issue 1 and 3. For issue 1, we also need to change the XY profiler, stats, and histogram widgets after the new features are merged
* Fixed the IQUV stokes dropdown in Save image. Other kinds of polarization is not supported yet. I prefer we file a new issue for this
image

edit: also fixed z profiler stokes dropdown selection. Found that other kinds of polarization is also not supported for z profiler, and may need changes in the backend

veggiesaurus (approved):

Issue #1571, opened 2021-07-15
raster image not updated after matching is just enabled 


kswang1029:
After we have images matched, switching to different images should show similar source features in the following example (matching CO, 13CO and C18O line cubes). 

@veggiesaurus do you think this is due to tile cache PR?

https://user-images.githubusercontent.com/20819712/125720351-3d7d395d-3ebb-4fe3-8db2-8ab8ba5abc8e.mov

veggiesaurus:
Hmm, ok I see the issue. 

What is the expected behaviour though? If you change images, we'll then have to fetch the new channel's images from the server, so there may be a situation where the old images are displayed initially and the new ones are then received

kswang1029:
> Hmm, ok I see the issue.
> 
> What is the expected behaviour though? If you change images, we'll then have to fetch the new channel's images from the server, so there may be a situation where the old images are displayed initially and the new ones are then received

That should be fine for the first time. Once it is cached switching to different images back and forth with fixed FOV should be smooth(?).

veggiesaurus:
> That’s should be fine for the first time. Once it is cached switching to different images back and forth with fixed FOV should be smooth(?).

Yes :+1:

veggiesaurus:
@kswang1029 can you see if this is also fixed by #1596? I think it was my idiotic typo (`=== undefined` instead of `!== undefined` :facepalm: )

kswang1029:
> @kswang1029 can you see if this is also fixed by #1596? I think it was my idiotic typo (`=== undefined` instead of `!== undefined` 🤦 )

Nope, the issue is still there with #1596. When grid mode is on, I see all panels get updated when switching channel but when the grid mode is off (single panel mode), the issue happens.

kswang1029:
@veggiesaurus I suggest we still align the tool bar to the right as if we align to the left, it will block the beam element overlay. 

<img width="1782" alt="Screen Shot 2021-08-09 at 10 05 00 AM" src="https://user-images.githubusercontent.com/20819712/128653768-eb75abf0-f3ca-46b2-8c44-47b5f984bfa1.png">

PR #1596, opened 2021-08-03
Angus/1259 multi panel

### LINKED ISSUE #1259: Multi-panel view with image viewer

veggiesaurus:
Closes #1259. Some bugs will be solved in separate issues. The main aim of this PR is to add an experimental multi-panel view without breaking the existing single-panel view.

- [x] Handle image export in multi-panel view.
- [x] Required tiles are sometimes not updated when zooming in, resulting in pixelated images.
- [x] Beam profile overlays not implemented correctly yet
- [x] The image toolbar should only show up on one panel at a time, and be aligned correctly
- [x] Catalogs should be rendered correctly
- [x] Colorbar's colormap and values should be determined per-image
- [x] Active panel highlight needs to be adjusted to be more obvious
- [x] Disable columns/rows spinner when multi-panel is disabled. Adjust text when in dynamic grid mode
- [x] Align non-ideal state 
- [x] Images sometimes don't show up if you load them simultaneously.

### DIFF
--- a/src/stores/OverlayStore.ts
+++ b/src/stores/OverlayStore.ts
@@ -910,11 +910,11 @@ export class OverlayColorbarSettings {
     }
 
     @computed get numberWidth(): number {
-        const activeFrame = AppStore.Instance.activeFrame;
-        if (!activeFrame) {
-            return 0;
+        let textWidth = 0;
+        for (const frame of AppStore.Instance.visibleFrames) {
+            const frameTextWidth = Math.max(...frame.colorbarStore.texts.map(x => x.length)) * this.textRatio[clamp(Math.floor(this.numberFont / 4), 0, this.textRatio.length)];
+            textWidth = Math.max(textWidth, frameTextWidth);
         }
-        const textWidth = Math.max(...activeFrame.colorbarStore.texts.map(x => x.length)) * this.textRatio[clamp(Math.floor(this.numberFont / 4), 0, this.textRatio.length)];
         return this.numberVisible ? this.numberFontSize * (this.numberRotation || this.position !== "right" ? 1 : textWidth) + this.textGap : 0;
     }
 
--- a/src/stores/OverlayStore.ts
+++ b/src/stores/OverlayStore.ts
@@ -1283,11 +1283,11 @@ export class OverlayStore {
     }
 
     @computed get viewWidth() {
-        return this.fullViewWidth / AppStore.Instance.numColumns;
+        return this.fullViewWidth / AppStore.Instance.numImageColumns;
     }
 
     @computed get viewHeight() {
-        return this.fullViewHeight / AppStore.Instance.numRows;
+        return this.fullViewHeight / AppStore.Instance.numImageRows;
     }
 
     @computed get renderWidth() {
--- a/src/stores/OverlayStore.ts
+++ b/src/stores/OverlayStore.ts
@@ -1283,11 +1283,11 @@ export class OverlayStore {
     }
 
     @computed get viewWidth() {
-        return this.fullViewWidth / (AppStore.Instance.numColumns);
+        return this.fullViewWidth / AppStore.Instance.numColumns;
     }
 
     @computed get viewHeight() {
-        return this.fullViewHeight / (AppStore.Instance.numRows);
+        return this.fullViewHeight / AppStore.Instance.numRows;
     }
 
     @computed get renderWidth() {
--- a/src/stores/OverlayStore.ts
+++ b/src/stores/OverlayStore.ts
@@ -1224,11 +1224,11 @@ export class OverlayStore {
     }
 
     @computed get viewWidth() {
-        return this.fullViewWidth / AppStore.Instance.numImageColumns;
+        return Math.floor(this.fullViewWidth / AppStore.Instance.numImageColumns);
     }
 
     @computed get viewHeight() {
-        return this.fullViewHeight / AppStore.Instance.numImageRows;
+        return Math.floor(this.fullViewHeight / AppStore.Instance.numImageRows);
     }
 
     @computed get renderWidth() {
--- a/src/stores/OverlayStore.ts
+++ b/src/stores/OverlayStore.ts
@@ -1075,8 +1075,11 @@ export class OverlayStore {
     }
 
     // View size options
-    @observable viewWidth: number;
-    @observable viewHeight: number;
+    @observable fullViewWidth: number;
+    @observable fullViewHeight: number;
+
+    @observable numRows: number;
+    @observable numColumns: number;
 
     // Individual settings
     @observable global: OverlayGlobalSettings;
@@ -1102,8 +1105,11 @@ export class OverlayStore {
         this.ticks = new OverlayTickSettings();
         this.colorbar = new OverlayColorbarSettings();
         this.beam = new OverlayBeamSettings();
-        this.viewHeight = 1;
-        this.viewWidth = 1;
+        this.fullViewWidth = 1;
+        this.fullViewHeight = 1;
+
+        this.numRows = 2;
+        this.numColumns = 2;
 
         // if the system is manually selected, set new default formats & update active frame's wcs settings
         autorun(() => {
@@ -1136,8 +1142,16 @@ export class OverlayStore {
     }
 
     @action setViewDimension = (width: number, height: number) => {
-        this.viewWidth = width;
-        this.viewHeight = height;
+        this.fullViewWidth = width;
+        this.fullViewHeight = height;
+    };
+
+    @action setNumColumns = (cols: number) => {
+        this.numColumns = cols;
+    };
+
+    @action setNumRows = (rows: number) => {
+        this.numRows = rows;
     };
 
     @action setFormatsFromSystem() {
@@ -1280,6 +1294,14 @@ export class OverlayStore {
             top: this.paddingTop,
             bottom: this.paddingBottom
         };
+    }
+
+    @computed get viewWidth() {
+        return this.fullViewWidth / this.numColumns;
+    }
+
+    @computed get viewHeight() {
+        return this.fullViewHeight / this.numRows;
     }
 
     @computed get renderWidth() {
--- a/src/stores/OverlayStore.ts
+++ b/src/stores/OverlayStore.ts
@@ -1078,9 +1078,6 @@ export class OverlayStore {
     @observable fullViewWidth: number;
     @observable fullViewHeight: number;
 
-    @observable numRows: number;
-    @observable numColumns: number;
-
     // Individual settings
     @observable global: OverlayGlobalSettings;
     @observable title: OverlayTitleSettings;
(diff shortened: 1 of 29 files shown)


kswang1029:
@veggiesaurus I have done some tests. The following issues/comments are what I found with the current implementation state. **_Not all need to be addressed for v3b1._** 

1. some (or all?) image view settings (eg title, colorbar's colormap/value) need to be generalized to support multiple panels. That is, each image would have its own settings instead of having a global one applied to all images. The UI needs to be revised carefully so that users know which active panel will be configured. KSW's homework...
Title and colorbar are more important for now as they have usability impact. 

2. Depending on the grid configuration, the toolbar at the bottom-right corner of each panel sometimes does not have sufficient space to display in the horizontal direction. KSW's homework on the solution...

3. The active panel highlight is less noticeable. Can we highlight the border of "image-panel-div" instead?

4. If multi-panel mode = "dynamic grid size" and we load two images in one shot via the file browser, the left panel will be empty (raster image not displayed). If we load the two images one by one, then it works fine. Cache issue?

5. In the image view settings, shall we show panel columns/rows spinbox only when multi-panel mode is "fixed grid size"?

6. Depending on which active image it is, the catalog overlays on different images are not rendered correctly (they are scaled). In the single panel mode, it works fine. 

https://user-images.githubusercontent.com/20819712/128120256-f01d500f-df38-4dea-83c3-9b1a41a8749a.mov


7. Tested 4x4 as performance test. We can "feel" delay in both pan and zoom actions but I think this is okay for now.

8. When dynamic grid is default and we load multiple images in one shot, sometimes we see tiles are misplaced in a panel. This displayed content changes with different zoom levels. This likely happens when the loaded images have different sizes in pixels. 
Screen Shot 2021-08-04 at 11 57 52 AM

9. The non-ideal state "No file loaded" of the image viewer is shifted upward. 

Screen Shot 2021-08-04 at 12 02 10 PM

10. The missing tile problem could occur when loading multiple images in one shot with the dynamic grid mode. Example in the screenshot of 8)

11.  Ideally, when we show two matched cubes side by side, change the channel in one should also trigger image update of the other. As for animation, if performance allows and there is no other technical limitation, the behavior should be the same.

12. [Comment] The combination of multi-panel view and the multi-profile view in the spectral profile widget provides very nice UX!
Screen Shot 2021-08-04 at 12 19 50 PM

veggiesaurus:
@kswang1029 here are my responses and suggestions in terms of what gets handled in beta-1:

> 1. some (or all?) image view settings (eg title, colorbar's colormap/value) need to be generalized to support multiple panels. That is, each image would have its own settings instead of having a global one applied to all images. The UI needs to be revised carefully so that users know which active panel will be configured. KSW's homework...

I think the colorbar's colormap/value should be an easy fix if I refactor things a bit to make it associated with the _frame_ rather than the overlay store. I'll try do that for beta-1. The title stuff is a little more complicated, because we currently have a single active WCS system, so all images that have WCS are forced to use the same system as the active image. The only way to prevent this from happening is actually to make the active image one without WCS, and each image reverts to its native system:
image

Custom titles per image will be more difficult and will need a bit of a rethink in terms of UI, so we should add those to beta-2.

> 2. Depending on the grid configuration, the toolbar at the bottom-right corner of each panel sometimes does not have sufficient space to display in the horizontal direction. KSW's homework on the solution...

I'd like the toolbar to only show in a single image at a time. If the image is in the left-most column, then we can align it left, rather than right. e.g:
image
I can add this feature to beta-1. 

> 3. The active panel highlight is less noticeable. Can we highlight the border of "image-panel-div" instead?

Good idea. I'll do this in beta-1.

> 4. If multi-panel mode = "dynamic grid size" and we load two images in one shot via the file browser, the left panel will be 
empty (raster image not displayed). If we load the two images one by one, then it works fine. Cache issue?

Will try to fix this is beta-1. It's got to do with the initial view size, I think. As soon as you move the image it seems to show up again.

> 5. In the image view settings, shall we show panel columns/rows spinbox only when multi-panel mode is "fixed grid size"?

We need to show this when it's in dynamic grid size too, as  this defines the maximum columns. Perhaps I can change the text to be "Max panel columns" when you switch to dynamic grid, and disable when in single-panel. Can fix for beta-1

> 6. Depending on which active image it is, the catalog overlays on different images are not rendered correctly (they are scaled). In the single panel mode, it works fine.

I will try fix this for beta-1. I haven't currently attempted to get catalogs to work in multipanel mode.

> 1. Tested 4x4 as performance test. We can "feel" delay in both pan and zoom actions but I think this is okay for now.

Not sure how to get round this other than severely throttling render redraw when images are not active and the user is panning/zooming.

> 2. When dynamic grid is default and we load multiple images in one shot, sometimes we see tiles are misplaced in a panel. This displayed content changes with different zoom levels. This likely happens when the loaded images have different sizes in pixels.

I noticed that too. Not entirely sure what's going on. I'll try solve this. Does it ever occur if you load images one-by-one?

> 3. The non-ideal state "No file loaded" of the image viewer is shifted upward.

Will fix in beta-1

> 1. The missing tile problem could occur when loading multiple images in one shot with the dynamic grid mode. Example in the screenshot of 8)
> 2. Ideally, when we show two matched cubes side by side, change the channel in one should also trigger image update of the other. As for animation, if performance allows and there is no other technical limitation, the behavior should be the same.

I wonder if this is related to #1571. Do contours update correctly?

veggiesaurus:
Ready for review. Will solve the outstanding issue above ("Required tiles are sometimes not updated when zooming in, resulting in pixelated images.") in a separate issue.

kswang1029:
- [ ] the pixel highlight (mouse on colorbar) feature does not work properly on the non-active panel. The threshold value is adopted from the active panel but then applied to the non-active panel.

https://user-images.githubusercontent.com/20819712/128654330-7b33ca5b-a5ae-4604-996a-6b5cab3475ab.mov

- [ ]  when loading multiple images in one shot, some images do not show full FOV in the panels (all except the bottom-right panel). The zoom levels seem to change panel by panel 🤔. I guess this is because of the dynamic grid generation? The final 2x2 grid is actually generated from 1x1 to 1x2, then 2x2. It would be good if this can be fixed but it is probably okay to keep this "behavior". (PS. missing tiles could happen too 🙈)

https://user-images.githubusercontent.com/20819712/128654595-15ee31ea-dfb1-4025-bb2e-d13a410b6ff6.mov

 - [ ] with a Retina display, the exported png seems to have the deviceRatio issue. What we see on screen is
 
Screen Shot 2021-08-09 at 10 35 01 AM

but the exported png is
dice_one fits-dice_two fits-dice_three fits-dice_four fits-image-2021-08-09-10-34-34

kswang1029:
As a note here: one time I see a panel contains disordered tiles at different zoom levels. It cannot be reproduced so far. I will keep an eye on this.

src/stores/OverlayStore.ts
```diff
@@ -973,7 +910,11 @@ export class OverlayColorbarSettings {
     }
 
     @computed get numberWidth(): number {
-        const textWidth = Math.max(...this.texts.map(x => x.length)) * this.textRatio[clamp(Math.floor(this.numberFont / 4), 0, this.textRatio.length)];
+        const activeFrame = AppStore.Instance.activeFrame;
+        if (!activeFrame) {
+            return 0;
+        }
+        const textWidth = Math.max(...activeFrame.colorbarStore.texts.map(x => x.length)) * this.textRatio[clamp(Math.floor(this.numberFont / 4), 0, this.textRatio.length)];
```
YuHsuan-Hwang: The changes about colorbar looks good to me 👍 
Just found one possible enhancement: we can use the max of `Math.max(...activeFrame.colorbarStore.texts.map(x => x.length))` of all 1~4 frames, instead of just checking the active frame. The `numberWidth` will not be changing when we click different panels in this way

YuHsuan-Hwang:
Minor issues of the toolbar:
The toolbar is placed under the colorbar. may be a z-index issue
We also can not click the buttons if it overlaps with another toolbar
image

veggiesaurus:
> * [ ]   when loading multiple images in one shot, some images do not show full FOV in the panels (all except the bottom-right panel). The zoom levels seem to change panel by panel thinking. I guess this is because of the dynamic grid generation? The final 2x2 grid is actually generated from 1x1 to 1x2, then 2x2. It would be good if this can be fixed but it is probably okay to keep this "behavior". (PS. missing tiles could happen too see_no_evil)

Yes, it's because of the dynamic grid size. Fixed in `05a78c7`.

veggiesaurus:
@kswang1029, I've fixed the issues you raised above. Please re-check. @YuHsuan-Hwang I've implemented the `numberWidth` suggestion you made as well. 

As for the toolbar: it's problematic. I'm not sure how to solve the issue properly. Things behave badly once the toolbar is too long. I've removed the left-alignment for now. I think the only way to really solve this is to have some sort of responsive design that limits the toolbar to a certain width based on the panel size. But I'd rather keep that for another PR. Can we mark it as a know issue for now?

kswang1029:
> @kswang1029, I've fixed the issues you raised above. Please re-check. @YuHsuan-Hwang I've implemented the `numberWidth` suggestion you made as well.
> 
> As for the toolbar: it's problematic. I'm not sure how to solve the issue properly. Things behave badly once the toolbar is too long. I've removed the left-alignment for now. I think the only way to really solve this is to have some sort of responsive design that limits the toolbar to a certain width based on the panel size. But I'd rather keep that for another PR. Can we mark it as a know issue for now?

Sure for another PR. It could be that if the space is not enough, we show a cog or something instead and user can hover over it to expand all the buttons. Surely we need a 2D grid of buttons I guess

veggiesaurus:
> It could be that if the space is not enough, we show a cog or something instead and user can hover over it to expand all the buttons.

That's what I was thinking :+1:

kswang1029:
> > It could be that if the space is not enough, we show a cog or something instead and user can hover over it to expand all the buttons.
> 
> That's what I was thinking 👍

Alternatively when the grid mode is on, we reserve a narrow horizontal space at the bottom of the viewer and put all the buttons there. They apply to the active panel only.

veggiesaurus:
@kswang1029 @YuHsuan-Hwang what about this approach (using a flex wrap):
image

turns into this when resizing:
image

or if necessary: 
image

A more realistic scenario:
image


It's still entirely usable when it wraps across to two lines. I think this seems like a nice solution. We could also shrink the toolbar a bit:

image

but I think that looks a little clumsy.

kswang1029:
> @kswang1029 @YuHsuan-Hwang what about this approach (using a flex wrap):
> image
> 
> turns into this when resizing:
> image
> 
> or if necessary:
> image
> 
> A more realistic scenario:
> image
> 
> It's still entirely usable when it wraps across to two lines. I think this seems like a nice solution. We could also shrink the toolbar a bit:
> 
> image
> 
> but I think that looks a little clumsy.

yes this seems better. However it looks a bit less pleasant due to misalignment. Ok for beta1 for now. We may consider the following approach for beta2. 

https://user-images.githubusercontent.com/20819712/128797306-8ff13feb-9d19-46cf-8bf0-0ca9f414bf5c.mov

The idea is when the space is not enough, we show a button instead and by hovering it all the buttons show up in a popover. This would make the layout cleaner I guess.

kswang1029 (approved):
Another milestone for the ultimate CARTA 👍

duidae (approved):

YuHsuan-Hwang (approved):

TienHao (approved):

PR #1611, opened 2021-08-10
Thchang/issue 1365

### LINKED ISSUE #1365: frontend crash when loading a 1x1 pixel image
This should be considered as an edge case but might accidentally happen when users generate a 1x1 subimage via the CARTA GUI. This only happens with 1x1 image. With 1xn it works fine. 
Screen Shot 2021-03-25 at 11 22 11 AM

TienHao:
Fixed #1365.

Adding `imageData` length check in TileService to avoid range error by small tiles.


kswang1029 (approved):
tested with my test images with special dimensions. 👍

src/services/TileService.ts
```diff
@@ -431,7 +431,7 @@ export class TileService {
         const compressedArray = tile.imageData;
         const workerIndex = this.compressionRequestCounter % this.workers.length;
         const nanEncodings32 = new Int32Array(tile.nanEncodings.slice(0).buffer);
-        let compressedView = new Uint8Array(tile.width * tile.height * 4);
+        let compressedView = new Uint8Array(compressedArray.byteLength > tile.width * tile.height * 4 ? compressedArray.byteLength : tile.width * tile.height * 4);
```
veggiesaurus: wouldn't `Math.max(compressedArray.byteLength, tile.width * tile.height * 4)` be neater?

veggiesaurus: actually, can we not always just use the byte length? For larger tiles, surely the compressed length (`compressedArray.btyteLength`) will be much smaller than the uncompressed length (`tile.width * tile.height * 4`)?

TienHao: yes, using byte length is better.

veggiesaurus (approved):

veggiesaurus:
@kswang1029 please double-check this again, after simplification of the change by @TienHao.

kswang1029:
> @kswang1029 please double-check this again, after simplification of the change by @TienHao.

Nope it is broken. images larger than 3x3 now do not show up as a raster.

veggiesaurus:
wait, sorry. I'm an idiot... `compressedView` needs to be big enough to accomodate the compressed image or the uncompressed. @TienHao please switch to `Math.max(compressedArray.byteLength, tile.width * tile.height * 4)`. Sorry!

Now I actually see why the error is cropping up in the first place!

TienHao:
> wait, sorry. I'm an idiot... `compressedView` needs to be big enough to accomodate the compressed image or the uncompressed. @TienHao please switch to `Math.max(compressedArray.byteLength, tile.width * tile.height * 4)`. Sorry!
> 
> Now I actually see why the error is cropping up in the first place!

That's OK. I didn't check with large image before. It fixed now!

PR #1615, opened 2021-08-10
Angus/1571 raster image not updated

### LINKED ISSUE #1571: raster image not updated after matching is just enabled

veggiesaurus:
Closes #1571 by making sure hidden images' GPU cache is also cleared when changing channels. As @kswang1029 speculated, this was introduced by #1288.

### DIFF
--- a/src/services/TileService.ts
+++ b/src/services/TileService.ts
@@ -243,7 +243,7 @@ export class TileService {
         }
     }
 
-    updateInactiveFileChannel(fileId: number, channel: number, stokes: number) {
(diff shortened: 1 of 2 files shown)


kswang1029 (approved):
confirmed the fix works and does not break something else as far as I can see. 👍

Issue #1572, opened 2021-07-15
region size not displayed 


kswang1029:
This _might_ be due to the recent backend change (https://github.com/CARTAvis/carta-backend/pull/827) on how FITS headers are extract from images. 
![Screen Shot 2021-07-15 at 11 13 13 AM](https://user-images.githubusercontent.com/20819712/125723705-8696971d-af7a-4f23-9cea-cdb98887b2c5.png)

kswang1029:
more investigation is needed to find the root cause as a counter example is found.

kswang1029:
When the size is not displayed, the image list widget is blank. The console says

![Screen Shot 2021-07-15 at 11 35 49 AM](https://user-images.githubusercontent.com/20819712/125725232-fc904b19-43b9-49cc-b747-82dba3ec8a09.png)

kswang1029:
With ALMA image cubes, the region size can be displayed properly. But for the images from elsewhere (eg Hubble, Herschel etc), the region size is empty. I guess this is certainly related to the header as before we used casacore to unify headers but now we use the raw header content. The size calculation (with AST?) needs to be checked.

kswang1029:
My speculation: possibly this is due to missing CDELT in the header. In some images, the pixel size info is stored in the CD matrix.

veggiesaurus:
> My speculation: possibly this is due to missing CDELT in the header. In some images, the pixel size info is stored in the CD matrix.

AST should be able to handle CD and PC matrices :thinking:

kswang1029:
as a note: if we create a line region, the region list widget may crash due to size calculation.

YuHsuan-Hwang:
https://github.com/CARTAvis/carta-frontend/blob/dev/src/stores/FrameStore.ts#L1290
the root cause is that `getWcsSizeInArcsec` returns null if there are no `CDELT1` and `CUNIT1` in the header

veggiesaurus:
> https://github.com/CARTAvis/carta-frontend/blob/dev/src/stores/FrameStore.ts#L1290
> the root cause is that `getWcsSizeInArcsec` returns null if there are no `CDELT1` and `CUNIT1` in the header

Ok, for now we can just guard against `null`. For beta-2 we can make sure we can handle CD and PC matrices

YuHsuan-Hwang:
> > https://github.com/CARTAvis/carta-frontend/blob/dev/src/stores/FrameStore.ts#L1290
[...]

PR #1622, opened 2021-08-11
avoid region list crash

YuHsuan-Hwang:
Fixed crash caused by line region length calculation mentioned in #1572

### DIFF
--- a/src/components/RegionList/RegionListComponent.tsx
+++ b/src/components/RegionList/RegionListComponent.tsx
@@ -149,7 +149,7 @@ export class RegionListComponent extends React.Component<WidgetProps> {
                     if (frame.validWcs) {
                         sizeContent =
                             region.regionType === CARTA.RegionType.LINE ? (
-                                formattedArcsec(length2D(region.wcsSize), WCS_PRECISION)
+                                formattedArcsec(region.wcsSize ? length2D(region.wcsSize) : undefined, WCS_PRECISION)
                             ) : (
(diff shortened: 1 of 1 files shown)


veggiesaurus (approved):

src/components/RegionList/RegionListComponent.tsx
```diff
@@ -158,7 +158,7 @@ export class RegionListComponent extends React.Component<WidgetProps> {
                                 </React.Fragment>
                             );
                     } else {
-                        sizeContent = region.regionType === CARTA.RegionType.LINE ? toFixed(length2D(region.size), 1) : `(${toFixed(region.size.x, 1)}, ${toFixed(region.size.y, 1)})`;
+                        sizeContent = region.regionType === CARTA.RegionType.LINE ? toFixed(region.wcsSize ? length2D(region.size) : undefined, 1) : `(${toFixed(region.size.x, 1)}, ${toFixed(region.size.y, 1)})`;
```
duidae: should be region.`size`?

duidae (changes requested):

src/components/RegionList/RegionListComponent.tsx
YuHsuan-Hwang: right. Changed to `region.size`

kswang1029 (approved):

duidae (approved):

PR #1625, opened 2021-08-13
Thchang/ import and export shortcut in region list 1036

### LINKED ISSUE #1036: Import/export button in the region list widget
adding import and export shortcut buttons to the region list widget

TienHao:
closes #1036.

Added region import and export buttons in the region list widget.

### DIFF
--- a/src/components/RegionList/RegionListComponent.tsx
+++ b/src/components/RegionList/RegionListComponent.tsx
@@ -73,6 +73,10 @@ export class RegionListComponent extends React.Component<WidgetProps> {
         FileBrowserStore.Instance.showFileBrowser(BrowserMode.RegionImport, false);
     };
 
+    private handleRegionExportAllClicked = () => {
+        FileBrowserStore.Instance.showExportRegions();
+    };
+
     private handleRegionListDoubleClick = () => {
         DialogStore.Instance.showRegionDialog();
     };
@@ -222,7 +226,7 @@ export class RegionListComponent extends React.Component<WidgetProps> {
                 exportEntry = (
                     <td style={{width: RegionListComponent.ACTION_COLUMN_DEFAULT_WIDTH}} onClick={ev => this.handleRegionExportClicked(ev, region)}>
                         <Tooltip2 content="Export region" position={Position.BOTTOM}>
-                            <Icon icon="export" />
+                            <Icon icon="cloud-upload" />
                         </Tooltip2>
                     </td>
                 );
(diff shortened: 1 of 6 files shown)


YuHsuan-Hwang:
The feature works nice
About the icons: in the file browser dialog and the z profile fitting widget, we have buttons for "exporting" files:
image
image
Shall we adjust these two buttons (or exchange the icons in the region list) for consistency?

TienHao:
I think the "export" in file browser dialog and z profile fitting is to download files to user's computer, but the "export regions" is to store regions into server. Maybe we don't need to make these three export icons the same, but to change import(cloud: download) and export(cloud: upload) region button into these two icons. 
截圖 2021-08-13 下午2 22 45

YuHsuan-Hwang:
I see the difference. Maybe we can change the locally download icons to those in line plots
image

kswang1029:
@TienHao feature-wise this works well. Based on the current design, I would suggest we also add an export (all) region icon to the right of the import icon at the column head. If there is no region created, it is grey out. 

Just curious: have you considered adding just two buttons (import and export) to the bottom-right corner when mouse hovering over? That might be simpler. 🤔

TienHao:
> @TienHao feature-wise this works well. Based on the current design, I would suggest we also add an export (all) region icon to the right of the import icon at the column head. If there is no region created, it is grey out.

nice suggestion. I have added.

> Just curious: have you considered adding just two buttons (import and export) to the bottom-right corner when mouse hovering over? That might be simpler.

Haven't though that before. The only advantage for the current design is to export inactive regions. I can change to hovered buttons. I think there might need 3 buttons(`import` , `export all`, `export active region`), or we just skip export all?

kswang1029:
> > @TienHao feature-wise this works well. Based on the current design, I would suggest we also add an export (all) region icon to the right of the import icon at the column head. If there is no region created, it is grey out.
> 
> nice suggestion. I have added.
> 
> > Just curious: have you considered adding just two buttons (import and export) to the bottom-right corner when mouse hovering over? That might be simpler.
> 
> Haven't though that before. The only advantage for the current design is to export inactive regions. I can change to hovered buttons. I think there might need 3 buttons(`import` , `export all`, `export active region`), or we just skip export all?

The current design is fine. 👍 I was just thinking that since in the file browser we can select a subset of regions for export so we don’t need to show the export button for each region. But I can see the advantages of the current design so it is okay.

kswang1029 (approved):

YuHsuan-Hwang (approved):

src/components/RegionList/RegionListComponent.tsx
```diff
@@ -89,7 +101,7 @@ export class RegionListComponent extends React.Component<WidgetProps> {
         let nameWidth = RegionListComponent.NAME_COLUMN_DEFAULT_WIDTH;
         const availableWidth = this.width - 2 * padding;
         let fixedWidth =
-            RegionListComponent.ACTION_COLUMN_DEFAULT_WIDTH * 2 +
+            RegionListComponent.ACTION_COLUMN_DEFAULT_WIDTH * 3 +
```
duidae: pleasee help give constants meaningful definition like `OOO_WIDTH`.

src/components/RegionList/RegionListComponent.tsx
```diff
@@ -192,7 +204,8 @@ export class RegionListComponent extends React.Component<WidgetProps> {
                 );
             } else {
                 lockEntry = (
-                    <td colSpan={2} style={{width: RegionListComponent.ACTION_COLUMN_DEFAULT_WIDTH * 2}}>
+                    <td colSpan={3} style={{width: RegionListComponent.ACTION_COLUMN_DEFAULT_WIDTH * 3}}>
```
duidae: Same as above comment.

duidae (changes requested):

src/components/RegionList/RegionListComponent.tsx
TienHao: has changed to `ACTIONS_COLUMN_DEFAULT_WIDTH`.

src/components/RegionList/RegionListComponent.tsx
TienHao: fixed.

duidae (approved):

PR #1627, opened 2021-08-13
Yuhsuan/1605_colorbar_tick_num

### LINKED ISSUE #1605: number of ticks in the colorbar
the number of ticks in the colorbar should be set to 3 so that a meaningful scale can be rendered properly. This is especially important when the multi-panel view is enabled (https://github.com/CARTAvis/carta-frontend/pull/1596) as usually each panel occupies only a small fraction of screen pixels so the computed number of ticks is mostly one or two, resulting a non-ideal colorbar range. 

Screen Shot 2021-08-09 at 11 58 22 AM

YuHsuan-Hwang:
This PR closes #1605:
* Changed the min of tick number to 3. The exact ticks may become 2-4 due to the algorithm of obtaining rounded numbers

Also adjusted the algorithm of obtaining rounded numbers, so please test the correctness of the numbers:
* avoid one tick
* fixed one bug: calculate the smallest number by the exact precision

### DIFF
--- a/src/stores/OverlayStore.ts
+++ b/src/stores/OverlayStore.ts
@@ -891,7 +891,7 @@ export class OverlayColorbarSettings {
 
     @computed get tickNum(): number {
         const tickNum = Math.round((this.height / 100.0) * this.tickDensity);
-        return this.height && tickNum > 1 ? tickNum : 1;
+        return this.height && tickNum > 3 ? tickNum : 3;
     }
 
     @computed get rightBorderPos(): number {
--- a/src/stores/OverlayStore.ts
+++ b/src/stores/OverlayStore.ts
@@ -5,6 +5,7 @@ import {WCSType} from "models";
 import {toFixed, clamp, getColorForTheme} from "utilities";
 
 const AST_DEFAULT_COLOR = "auto-blue";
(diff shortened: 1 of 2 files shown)


kswang1029 (approved):
This works in general. 👍

src/stores/OverlayStore.ts
```diff
@@ -891,7 +891,7 @@ export class OverlayColorbarSettings {
 
     @computed get tickNum(): number {
         const tickNum = Math.round((this.height / 100.0) * this.tickDensity);
-        return this.height && tickNum > 1 ? tickNum : 1;
+        return this.height && tickNum > 3 ? tickNum : 3;
```
duidae: please replace constants with something like `MIN_TICK_NUM` for readibility, thanks.

duidae (changes requested):

src/stores/OverlayStore.ts
YuHsuan-Hwang: Ok. I added `COLORBAR_TICK_NUM_MIN`

duidae (approved):

veggiesaurus (approved):

PR #1633, opened 2021-08-18
Jason/issue 1477

### LINKED ISSUE #1477: Move text of in-app help into static data files?
Would be possible to refactor the existing code to move all the in-help text into dedicated data files, to maintain a better separation between the code and the data? There may be existing libraries which would help with this.

in src/components/HelpDrawer/HelpContent

duidae:
closes #1477


veggiesaurus:
Unless I'm misunderstanding, wasn't #1477 about using static HTML for help content? Not about just moving them into one giant file (I'm not sure what the advantage of that is?)

duidae:
> Unless I'm misunderstanding, wasn't #1477 about using static HTML for help content? Not about just moving them into one giant file (I'm not sure what the advantage of that is?)

I think a more common solution is to:
1. use a file to save html structure of the help contents (i.e. HelpContent.tsx in this PR)
2. save those help text in a individual `en.po` PO file (refers to the static data files) and use Gettext to get corresponding value texts from key
3. have translations like `af.po` `zh_TW.po`... to support multi-languages

but so far we dont have the need of supporting mulit-languages, so step 1 should be sufficient.

ref: https://en.wikipedia.org/wiki/Gettext

veggiesaurus:
> > Unless I'm misunderstanding, wasn't #1477 about using static HTML for help content? Not about just moving them into one giant file (I'm not sure what the advantage of that is?)
> 
> I think a more common solution is to:
> 
> 1. use a file to save html structure of the help contents (i.e. HelpContent.tsx in this PR)
> 2. save those help text in a individual `en.po` PO file and use Gettext to get corresponding key/value texts
> 3. have translations like `af.po` `zh_TW.po`... to support multi-languages
> 
> but so far we dont have the need of supporting mulit-languages, so step 1 should be sufficient.

I don't think (1) is required for (2) and (3), but I don't have a strong opinion on this. I guess it comes down to what @kswang1029 prefers, since he's the one who normally works with the help contents.

kswang1029:
> > > Unless I'm misunderstanding, wasn't #1477 about using static HTML for help content? Not about just moving them into one giant file (I'm not sure what the advantage of that is?)
> > 
> > 
> > I think a more common solution is to:
> > 
> > 1. use a file to save html structure of the help contents (i.e. HelpContent.tsx in this PR)
> > 2. save those help text in a individual `en.po` PO file and use Gettext to get corresponding key/value texts
> > 3. have translations like `af.po` `zh_TW.po`... to support multi-languages
> > 
> > but so far we dont have the need of supporting mulit-languages, so step 1 should be sufficient.
> 
> I don't think (1) is required for (2) and (3), but I don't have a strong opinion on this. I guess it comes down to what @kswang1029 prefers, since he's the one who normally works with the help contents.

I have no preference on having multiple or one file for the help texts. @confluence what's the original motivation?

kswang1029 (approved):

confluence:
The original intention was to move the help text to static files, to introduce a clear separation between code and data, so that it's easier to see whether changes to the frontend are only to data.

I think that multiple data files would make more sense than a single data file -- they would be easier to navigate, and it would be easier to see the scope of changes to the text.

I would suggest implementing up to step 2 in @duidae's proposal -- there is an advantage to setting this up now even though we don't currently have translations.

It would be nice if we could store this content in a more human-writeable format, like markdown, and parse it when we load it, but unless there's a library that would make this a very simple addition I don't think it's critical to do it now.

veggiesaurus:
@kswang1029 since you're the one who maintains the help files, it's up to you whether you want to merge this in or not.

kswang1029:
What are the concerns having or not having multiple  help context files from codebase maintenance perspective? @duidae @confluence We probably need to consider maintainability as the top priority. I can be flexible either way but I may not be always the person who maintains this part in the mid/far future.

duidae:
Restored to separate files. please have a look, thanks.

kswang1029 (approved):

duidae:
@veggiesaurus is this ok to merge?

PR #1636, opened 2021-08-23
Jason/issue 1547

### LINKED ISSUE #1547: Display loading bar when large number of tiles need to be loaded into memory before contours are calculated
HDF5 files can now be loaded progressively. However, when calculating contours we still need the entire full-resolution image loaded into memory. This means that there will be a noticeable delay while generating contours on large HDF5 images

duidae:
closes #1547


kswang1029 (changes requested):
For static image it works fine. When animating, the situation is embarrassing... This happens when the contour dialog is enabled. 

https://user-images.githubusercontent.com/20819712/132150749-886602cc-5513-4baf-9f55-05e447938031.mov

veggiesaurus:
This progress bar only needs to be shown if the backend needs to load a large number of tiles before generating contours. I suggest we only show the progress dialog after a certain amount of time has passed (500 ms?) _and_ only if no contour progress messages have been received.

Pulling @confluence into the conversation here, because she handled the tile reading on the backend. Wonder if we can add some sort of progress message from the backend about tile loading? Specifically when you are forced to load the entire channel for contours

kswang1029:
@duidae I tried 0a5cc92 and now animation playback works as usual without the blinking progress bar. However, by loading a large HDF5 image and generating a set of contours, I cannot trigger the progress bar like before. What is the new condition here in 0a5cc92?

duidae:
> @duidae I tried [0a5cc92](https://github.com/CARTAvis/carta-frontend/commit/0a5cc92cfff9a836708cf87eb145faf35210d814) and now animation playback works as usual without the blinking progress bar. However, by loading a large HDF5 image and generating a set of contours, I cannot trigger the progress bar like before. What is the new condition here in [0a5cc92](https://github.com/CARTAvis/carta-frontend/commit/0a5cc92cfff9a836708cf87eb145faf35210d814)?

it's not done yet, can review after requested.

duidae:
> This progress bar only needs to be shown if the backend needs to load a large number of tiles before generating contours. I suggest we only show the progress dialog after a certain amount of time has passed (500 ms?) _and_ only if no contour progress messages have been received.
> 
> Pulling @confluence into the conversation here, because she handled the tile reading on the backend. Wonder if we can add some sort of progress message from the backend about tile loading? Specifically when you are forced to load the entire channel for contours

For static images we may have ways to identify the trigger point of generating contour images (i.e. pressing Apply in contour dialog, pressing Next/Prev in animator) for customized progress bar of showing up after 500ms, but for animation, frontend just passively waits for the contour stream, so @veggiesaurus I think a simple solution could be that for each contour stream backend just sends a stream of progress 0 with an additional field such as `isLongTask=true` or `estimationTime=1000ms`, so that frontend can show up the progress bar or not according to the field.

veggiesaurus:
> > This progress bar only needs to be shown if the backend needs to load a large number of tiles before generating contours. I suggest we only show the progress dialog after a certain amount of time has passed (500 ms?) _and_ only if no contour progress messages have been received.
> > Pulling @confluence into the conversation here, because she handled the tile reading on the backend. Wonder if we can add some sort of progress message from the backend about tile loading? Specifically when you are forced to load the entire channel for contours
> 
> For static images we may have ways to identify the trigger point of generating contour images (i.e. pressing Apply in contour dialog, pressing Next/Prev in animator) for customized progress bar of showing up after 500ms, but for animation, frontend just passively waits for the contour stream, so @veggiesaurus I think a simple solution could be that for each contour stream backend just sends a stream of progress 0 with an additional field such as `isLongTask=true` or `estimationTime=1000ms`, so that frontend can show up the progress bar or not according to the field.

Ok, I think that's a sensible suggestion. Perhaps @markccchiang could implement this? It's specifically for HDF5 files that need their image cache to be populated.

markccchiang:
> Ok, I think that's a sensible suggestion. Perhaps @markccchiang could implement this? It's specifically for HDF5 files that need their image cache to be populated.

Ok, I will do it.

kswang1029:
@duidae could you check the case below with the attached image? Please try to use min-max generator to get 10 levels and render them. From the console I see all 10 CONTOUR_IMAGE_DATA are all with progress:1 but the progress bar hangs there forever. 
[cluster_01024.fits.zip](https://github.com/CARTAvis/carta-frontend/files/7495093/cluster_01024.fits.zip)


Screen Shot 2021-11-08 at 3 43 28 PM

kswang1029:
@markccchiang @duidae Could you explain when the progress bar will appear? I tried an image with 512x512 pix and there is no progress bar when rendering the contours. With 1024x1024 and above, I see a progress bar then. The UX issue I see here is the progress bar appears and disappears quickly so the browser window flashes. It would be good to show the progress bar _only_ for long contour generation/rendering task. Is it possible to show the progress bar only after 2 or 3 sec if the generation/rendering task does not finish within that time interval? I am not sure which side (frontend/backend) to ask this question so I just post here.

markccchiang:
> @markccchiang @duidae Could you explain when the progress bar will appear? I tried an image with 512x512 pix and there is no progress bar when rendering the contours. With 1024x1024 and above, I see a progress bar then. The UX issue I see here is the progress bar appears and disappears quickly so the browser window flashes. It would be good to show the progress bar _only_ for long contour generation/rendering task. Is it possible to show the progress bar only after 2 or 3 sec if the generation/rendering task does not finish within that time interval? I am not sure which side (frontend/backend) to ask this question so I just post here.

Backend assumes it is a long task if the number of pixels is greater than 1e+6. Since the frontend shows the progress bar at the start of the calculation if `is_long_task` is true. I am not sure is it easy for the frontend to show the progress bar only after 2 or 3 secs if the task does not finish yet. We can estimate the time to finish after the calculation started. But different contour levels run in parallel, their progress reports may confuse the frontend that we tried before. Maybe @duidae has a better idea?

duidae:
> @duidae could you check the case below with the attached image? Please try to use min-max generator to get 10 levels and render them. From the console I see all 10 CONTOUR_IMAGE_DATA are all with progress:1 but the progress bar hangs there forever. [cluster_01024.fits.zip](https://github.com/CARTAvis/carta-frontend/files/7495093/cluster_01024.fits.zip)
> 
> Screen Shot 2021-11-08 at 3 43 28 PM

this one fixed.

duidae:
> > @markccchiang @duidae Could you explain when the progress bar will appear? I tried an image with 512x512 pix and there is no progress bar when rendering the contours. With 1024x1024 and above, I see a progress bar then. The UX issue I see here is the progress bar appears and disappears quickly so the browser window flashes. It would be good to show the progress bar _only_ for long contour generation/rendering task. Is it possible to show the progress bar only after 2 or 3 sec if the generation/rendering task does not finish within that time interval? I am not sure which side (frontend/backend) to ask this question so I just post here.
> 
> Backend assumes it is a long task if the number of pixels is greater than 1e+6. Since the frontend shows the progress bar at the start of the calculation if `is_long_task` is true. I am not sure is it easy for the frontend to show the progress bar only after 2 or 3 secs if the task does not finish yet. We can estimate the time to finish after the calculation started. But different contour levels run in parallel, their progress reports may confuse the frontend that we tried before. Maybe @duidae has a better idea?

current logic of showing progress bar is depending on the backend msg, if the backend tells this contour takes time (isLongTask=true), then frontend shows the progress bar, so it's inevitable that there are always some contour streams falling into the flickering situation. But I think the real problem is why we choose that masking progress bar in the beginning as a solution just according to some sloppy thoughts, if we choose another approach to show progress like YT, it doent not mask the whole UI and the msg of the backend can be more flexible. 


https://user-images.githubusercontent.com/25971696/140876423-2657ec19-e5b6-4f87-b58a-95f8d4fabef0.mp4

veggiesaurus:
> But I think the real problem is why we choose that masking progress bar in the beginning as a solution just according to some sloppy thoughts, if we choose another approach to show progress like YT, it doent not mask the whole UI and the msg of the backend can be more flexible.

We have less invasive progress indicators for the tiles and contours already. This blocking progress bar should only be shown when we're sure that there's a long-running task that should not be interrupted (e.g. closing a file, etc).

You can use `setTimeout` to show the progress bar after 500 ms or something, and then hide it as soon as the contour messages start to arrive. Remember, this is purely necessary for the HDF5 case where you have to load a large amount of data from disk _before_ calculating contours. Once the contours are being calculated and delivered, we show them progressively, don't we?

duidae:
> > But I think the real problem is why we choose that masking progress bar in the beginning as a solution just according to some sloppy thoughts, if we choose another approach to show progress like YT, it doent not mask the whole UI and the msg of the backend can be more flexible.
> 
> We have less invasive progress indicators for the tiles and contours already. This blocking progress bar should only be shown when we're sure that there's a long-running task that should not be interrupted (e.g. closing a file, etc).
> 
> You can use `setTimeout` to show the progress bar after 500 ms or something, and then hide it as soon as the contour messages start to arrive. Remember, this is purely necessary for the HDF5 case where you have to load a large amount of data from disk _before_ calculating contours. Once the contours are being calculated and delivered, we show them progressively, don't we?

does that HDF5 stuff need backend modification?

veggiesaurus:
> does that HDF5 stuff need backend modification?

Isn't that what @markccchiang is implementing?

duidae:
> > does that HDF5 stuff need backend modification?
> 
> Isn't that what @markccchiang is implementing?

kswang1029:
@duidae @veggiesaurus @markccchiang I suggest we do the following and please comment if you see problems:
- for FITS/CASA/MIRIAD, `is_long_task` is always `false` regardless the image size as the image data is already cached. Image smoothing and contour vertices calculations should be fast enough in ram. 
- for HDF5, if the image size is greater than 8000x8000 (~250 MB; this is assuming the data I/O is with a HDD), then we set `is_long_task` as `true` so that the frontend can display a popup message like "loading" and once the backend starts to deliver contour data, the frontend hides the popup and users will see the contours being progressively rendered. 
- when animating an HDF5 image with contours, we don't show the popup.

duidae:
> @duidae @veggiesaurus @markccchiang I suggest we do the following and please comment if you see problems:
> 
> * for FITS/CASA/MIRIAD, `is_long_task` is always `false` regardless the image size as the image data is already cached. Image smoothing and contour vertices calculations should be fast enough in ram.
> * for HDF5, if the image size is greater than 8000x8000 (~250 MB; this is assuming the data I/O is with a HDD), then we set `is_long_task` as `true` so that the frontend can display a popup message like "loading" and once the backend starts to deliver contour data, the frontend hides the popup and users will see the contours being progressively rendered.

@markccchiang for this scenario frontend's logic without extra msg modification would be: 

1. receiving a contour of progress 0 && isLongTask=True, then shows a progress like: 
image
2. then close the progress when receiving contour stream of progress > 0

> * when animating an HDF5 image with contours, we don't show the popup.

kswang1029:
Correct me if I think in the wrong direction: what I proposed could be achieved by the frontend implementation only it seems. If we are requesting contour rendering from an hdf5 with the spatial dimension greater then 8000x8000 pixels, then immediately after the generate button is pressed we can show the loading bar. Once the frontend receives the first contour data message, the loading bar can be removed. Then we will start to see contours being progressively rendered. 🤔

veggiesaurus:
> Correct me if I think in the wrong direction: what I proposed could be achieved by the frontend implementation only it seems. If we are requesting contour rendering from an hdf5 with the spatial dimension greater then 8000x8000 pixels, then immediately after the generate button is pressed we can show the loading bar. Once the frontend receives the first contour data message, the loading bar can be removed. Then we will start to see contours being progressively rendered. thinking

@kswang1029 I think you're right.

duidae:
> > Correct me if I think in the wrong direction: what I proposed could be achieved by the frontend implementation only it seems. If we are requesting contour rendering from an hdf5 with the spatial dimension greater then 8000x8000 pixels, then immediately after the generate button is pressed we can show the loading bar. Once the frontend receives the first contour data message, the loading bar can be removed. Then we will start to see contours being progressively rendered. thinking
> 
> @kswang1029 I think you're right.

then what was those detours above? can we save those time & energy if the requirement is precise enough in the very beginning?

PR #1638, opened 2021-08-30
Yuhsuan/1587 update cursor value criteria

### LINKED ISSUE #1587: revisiting SET_CURSOR criteria
The SET_CURSOR message is only triggered when the cursor position as "whole pixel" is updated. So if there are matched  images with very different pixel sizes, we will see relevant info such as cursor value (cf @YuHsuan-Hwang 's work on the cursor info widget), cursor spatial profile or cursor spectral profile not being updated.  

Taken the following figure as an example, if we move the cursor inside the pixel highlighted with thick red line, on the matched image (with black color) there are multiple smaller pixels. So in principle if we move the cursor around inside the red pixel, we should still be able to see data update of the cursor hits a new black pixel.

Screen Shot 2021-07-27 at 11 54 17 AM

A possible solution would be we refine the criteria of SET_CURSOR to support fractional pixel down to 1/1000 (not perfect as the same issue still happens at a smaller scale). This would extend the use cases significantly I suppose.

YuHsuan-Hwang:
This PR closes #1587. Changes:
* Use unrounded positions to obtain matched image positions
* Update cursor value of the hovered image instead of the active image
* For the cursor info widget, show info of the hovered image instead of the active image

### DIFF
--- a/src/components/CursorInfo/CursorInfoComponent.tsx
+++ b/src/components/CursorInfo/CursorInfoComponent.tsx
@@ -120,7 +120,7 @@ export class CursorInfoComponent extends React.Component<WidgetProps> {
     render() {
         const appStore = AppStore.Instance;
         const frameNum = appStore.frames.length;
-        const frame = appStore.activeFrame;
+        const frame = appStore.hoveredFrame ?? appStore.activeFrame;
 
         if (frameNum <= 0) {
             return (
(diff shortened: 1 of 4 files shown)


kswang1029 (commented):
@YuHsuan-Hwang do you think it is possible to show more digits in the world coordinates in this PR too? There is a situation that world coordinates are displayed inconsistently. 

Screen Shot 2021-09-06 at 10 24 42 AM

YuHsuan-Hwang:
> @YuHsuan-Hwang do you think it is possible to show more digits in the world coordinates in this PR too? There is a situation that world coordinates are displayed inconsistently.
> 

~~Sure, I will adjust the digits~~ seems better to file a new issue for this

kswang1029 (approved):

veggiesaurus (approved):

PR #1646, opened 2021-09-06
Jason/performance import regions and fps

duidae:
PR includes:
1. add progress bar during importing large numbers of regions via code splitting, related to https://github.com/CARTAvis/carta/issues/90
a

2. introduce FPS meter for development
b

3. minor refactorings


src/index.tsx
```diff
@@ -54,3 +54,25 @@ async function fetchConfig() {
 fetchConfig().then(() => {
     console.log("Configuration complete. Rendering frontend");
 });
+
+// Add FPS meter for development
+if (process.env.NODE_ENV === "development") {
```
veggiesaurus: I think we should make this optional, and independent of dev mode. We may want to benchmark things in prod builds as well. Can you make this a preference? You don't have to add it to the preference dialog, because we can use a code snippet for those sort of diagnostic things:

```javascript
app.preferences.setPreference("statsPanelEnabled", true);
app.preferences.setPreference("statsPanelMode", 0);

**Review by veggiesaurus — CHANGES_REQUESTED** [review]
one minor change: change to a preference for stats panel

File: src/index.tsx (hunk 1 above)
Comment (duidae) [reply src/index.tsx]: ok

Comment (veggiesaurus) [reply src/index.tsx]: @kswang1029 are you ok with this solution? I think we can hide debug/diagnostic settings from the preferences dialog

Comment (kswang1029) [reply src/index.tsx]: 👌

**Review by kswang1029 — COMMENTED** [review]
I tested this improvement a bit. Comments/issues below (issues may be addressed in the same PR or later).

1. A casa region file containing 4000 point regions can be loaded in this branch successfully. 🚀
2. the resource monitoring tool is handy 👍
3. It appears that the loading performance scales like t ~ n^2. Note this is tested with point regions only. Not sure if other regions (eg rectangle) scale differently. 
n_points(time)
250(3.8s)
500(13s)
1000(52s)
2000(220s; 3m40s)
4000(927s; 15m27s) 
4. while loading a large N region file, initially the loading performance is better. Once there are more regions loaded, the loading performance decreases (significantly if N is large)
5. regions are loaded in a batch of 10. Any idea about the choice here? Would this introduce more overhead?

**Comment by duidae** [comment]
the batch number is just an interval for the main thread to respond to UI, numbers resulting in enough FPS are sufficient.

**Review by veggiesaurus — APPROVED** [review]

File: src/stores/AppStore.ts [hunk 2]
```diff
@@ -929,30 +931,39 @@ export class AppStore {
...
+        const batchEnd = Math.min(count + IMPORT_REGION_BATCH_SIZE, regions.length);
+        for (; count < batchEnd; count++) {
+            const [regionIdString, regionInfo] = regions[count];
+            const styleInfo = regionStyleMap.get(regionIdString);
+            frame.regionSet.addExistingRegion(regionInfo.controlPoints as Point2D[], regionInfo.rotation, regionInfo.regionType, parseInt(regionIdString), styleInfo?.name, styleInfo?.color, styleInfo?.lineWidth, styleInfo?.dashList);
+        }
+        this.fileBrowserStore.updateLoadingState(count / regions.length, count, regions.length);
+
+        if (count < regions.length) {
+            setTimeout(this.addRegionsInBatch, 0, regions, regionStyleMap, count);
```
veggiesaurus: I had no idea you could pass additional arguments to `setTimeout` like this :+1: 

But I do wonder if moving the settimeout out of this function and into the calling function (as something like `await delay(0)` would be clearer

src/stores/AppStore.ts
```diff
@@ -929,30 +931,39 @@ export class AppStore {
...
+                this.addRegionsInBatch(regions, regionStyleMap, 0);
             }
-            this.fileBrowserStore.hideFileBrowser();
         } catch (err) {
             console.error(err);
             AppToaster.show(ErrorToast(err));
         }
     };
 
+    addRegionsInBatch = (regions: [string, CARTA.IRegionInfo][], regionStyleMap: Map<string, CARTA.IRegionStyle>, count: number) => {
```
veggiesaurus: shouldn't this be an `@action` in order to batch side-effects of adding each region?

veggiesaurus:
> I tested this improvement a bit. Comments/issues below (issues may be addressed in the same PR or later).
> 
> 1. A casa region file containing 4000 point regions can be loaded in this branch successfully. rocket
> 2. the resource monitoring tool is handy +1
> 3. It appears that the loading performance scales like t ~ n^2. Note this is tested with point regions only. Not sure if other regions (eg rectangle) scale differently.
>    n_points(time)
>    250(3.8s)
>    500(13s)
>    1000(52s)
>    2000(220s; 3m40s)
>    4000(927s; 15m27s)
> 4. while loading a large N region file, initially the loading performance is better. Once there are more regions loaded, the loading performance decreases (significantly if N is large)
> 5. regions are loaded in a batch of 10. Any idea about the choice here? Would this introduce more overhead?

@kswang1029 are you saying that performance is worse than in `dev`?
I think the reason it scales with n^2 is due to allocating memory and using `push` for a large number of items. If we pre-allocate the array and fill it, that might work better.

@duidae we might be able to solve this by pre-allocating the region array in `RegionSetStore` to the correct number of regions, and then filling the region as we go, and doing things in batches. If using a larger batch size improves performance, we should do that, even if it means we don't maintain interactive frame times while loading.

veggiesaurus (changes requested):
see comments above

src/stores/AppStore.ts
duidae: it does not modify any observable directly, so that mobx did not give any warning.

duidae:
> @duidae we might be able to solve this by pre-allocating the region array in `RegionSetStore` to the correct number of regions, and then filling the region as we go, and doing things in batches. If using a larger batch size improves performance, we should do that, even if it means we don't maintain interactive frame times while loading.

code splitting (batch stuff) is a common technique for handling long tasks, and it has nothing to do with performance in our case, can take a look if interested: https://javascript.plainenglish.io/how-to-optimize-long-tasks-blocking-javascript-in-browsers-d49508f72c9

veggiesaurus:
> > @duidae we might be able to solve this by pre-allocating the region array in `RegionSetStore` to the correct number of regions, and then filling the region as we go, and doing things in batches. If using a larger batch size improves performance, we should do that, even if it means we don't maintain interactive frame times while loading.
> 
> code splitting (batch stuff) is a common technique for handling long tasks, and it has nothing to do with performance in our case, can take a look if interested: https://javascript.plainenglish.io/how-to-optimize-long-tasks-blocking-javascript-in-browsers-d49508f72c9

I think you're misunderstanding what I'm saying. The _O(N^2)_ performance is because each time you `push` an element to the array, the side-effects are being re-run, and the runtime of the side effects are _O(N)_. For example (run times are arbitrary):

- Start with 0 regions. Add one region. Side effect (component render/`computed`/`autorun`) completes in e.g. 1 ms
- Add another region. Now there are two regions. Side effect now completes in 2 ms.
- Add another region. Side effect now completes in 3 ms
- ...
- Add Nth region. Side effect now completes in N ms

So you have 0 + 1 + 2 + ... + _N_ = _N_(_N_+1) / 2. i.e runtime _O(N^2)_. 

If you prevent the side-effects from running, then the only cost would be the `push` operations. _N_ `push` calls would be amortized to _O(N)_. Increasing the batch size (and wrapping in an `@action` would reduce the number of times the side-effects are calculated, taking us closer to _O(N)_, rather than _O(N^2)_.

duidae:
> If you prevent the side-effects from running, then the only cost would be the `push` operations. _N_ `push` calls would be amortized to _O(N)_. Increasing the batch size (and wrapping in an `@action` would reduce the number of times the side-effects are calculated, taking us closer to _O(N)_, rather than _O(N^2)_.

this batch `rendering` to reduce total loading time is a separate issue to this PR, and now I am working on improving the usability of loading thousands of regions, I think this may have higher priority than reducing the loading time to user.

em66z-0wc45

veggiesaurus:
> this batch `rendering` to reduce total loading time is a separate issue to this PR, and now I am working on improving the usability of loading thousands of regions, I think this may have higher priority than reducing the loading time to user.

I agree, as long as the performance hasn't been degraded by this PR. Waiting for @kswang1029's feedback on that, but I can't see why it would be.

While I agree that making the frontend usable with thousands of regions is high priority, taking more than a minute to load 2000 regions is probably in "unusable" territory, especially when it takes a few seconds in DS9. Obviously we'll have some overhead compared to DS9 due to our client/server arch and our react-based rendering, but our users don't really care about that.

veggiesaurus:
> em66z-0wc45

This looks _very_ promising :+1:

kswang1029:
> @kswang1029 are you saying that performance is worse than in `dev`?
> I think the reason it scales with n^2 is due to allocating memory and using `push` for a large number of items. If we pre-allocate the array and fill it, that might work better.

comparing to dev, I don't see significant performance degradation (checked the 500 and 1000 regions cases).

veggiesaurus:
> comparing to dev, I don't see significant performance degradation (checked the 500 and 1000 regions cases).

Just a note: the loading time is _far_ higher if you have the region list open. 

Wrapping the region list loading loop (in the `dev` branch) in a `runInAction` call takes the loading time for 4000 regions down to ~5 seconds.

PR #1647, opened 2021-09-07
Yuhsuan/1306 velocity reference

### LINKED ISSUE #1306: new feature: allow user to change the velocity reference for WCS matching of cubes
We can match cubes in the spectral axis either by frequency, by channel, or by radio or optical velocity.  The latter are matched to the cube by using either a redshift for the object, or a rest frequency for a line (I suspect its by RESTFREQ). 

Problems arise when trying to register cubes of the same object that have incompatible values for the redshift, or a missing redshift. In this case, the spectral features in one cube will not properly overlay with the other. I will give an example later.

This ticket is a request to display the reference in the image list widget where the WCS registration/matching is enabled, and allow the user to change it. 

An example dataset where problems occur is from the ALMA archive, project code 2017.1.00079.S, where the 12m and 7m cubes of the 12CO(1-0) emission do not overlap if matching by Vrad. 
e.g. 12m SBname=M83_e_03_TM1
product = member.uid___A001_X1295_X35.M83_sci.spw21.cube.I.pbcor.fits
(header stays VELREF = 257, RESTFREQ=1.1500000E+11 Hz ALTRVAL=2.902229E+06 ALTRPIX=1.0)
register with 7m SBname=M83_e_03_7M
product = member.uid___A001_X1295_X37.M83_sci.spw16.cube.I.pbcor.fits
(header stays VELREF = 257, RESTFREQ=1.15271202E+11 Hz ALTRVAL=8.44435E+05 ALTRPIX=1.0)

If you contour the 7m, and append the 12m with z registration using Vrad, the contours do not overlay, presumably b/c they have different values for RESTFREQ. They do overlay if you change your preferences to make spectral matching=Frequency.

YuHsuan-Hwang:
This PR closes #1306:
* added "Set Rest Frequency" to the right-click menu in the layer widget for opening the rest frequency dialog
* this feature is only enabled for 3D images with frequency/wavelength native spectral type
* after saving custom rest frequency, the converted velocity in the z profile and the spectrally matched channels will be updated


kswang1029 (commented):
This works nicely. 👍 Image matching is also supported after a new rest frequency is defined.

Do you think for the PV image defined as OFFSET VS "FREQ" we can have the same capability? It would be great to support it too.


One minor improvement is we add the rest frequency defined in header in the dialog like " Header RESTFRQ = 1.152712018000E+11 / Rest Frequency (Hz) " somewhere so that it is easier to know if RESTFRQ has been overwritten or not. Or another way to let user know the current RESTFRQ has been overwritten.

YuHsuan-Hwang:
@kswang1029 
> Do you think for the PV image defined as OFFSET VS "FREQ" we can have the same capability? It would be great to support it too.

Yes, added this feature to PV images. Please test this part, thanks

> One minor improvement is we add the rest frequency defined in header in the dialog like " Header RESTFRQ = 1.152712018000E+11 / Rest Frequency (Hz) " somewhere so that it is easier to know if RESTFRQ has been overwritten or not. Or another way to let user know the current RESTFRQ has been overwritten.

Added the information in the tooltip. The reset button is disabled if RESTFRQ is not provided in the header

kswang1029 (commented):
Found two usability issues:
1. If we load an image cube then immediately set a new rest frequency, the displayed velocity in the cursor info is not updated accordingly. We will need to change to a new channel to make the new rest frequency effective in the velocity conversion. 
2. If we load an image cube then immediately set a new rest frequency, then we load the same image cube again in append mode and match the two image cubes, we will see initially the matched channel indices are identical. We should see they are different as now their rest frequencies are different. This is probably related to the first issue.

kswang1029:
Another usability improvement: we should disable the "set rest frequency" menu when animation playback is on, just to simplify things to prevent unforeseen side effects.

kswang1029:
> Found two usability issues:
> 
> 1. If we load an image cube then immediately set a new rest frequency, the displayed velocity in the cursor info is not updated accordingly. We will need to change to a new channel to make the new rest frequency effective in the velocity conversion.
> 2. If we load an image cube then immediately set a new rest frequency, then we load the same image cube again in append mode and match the two image cubes, we will see initially the matched channel indices are identical. We should see they are different as now their rest frequencies are different. This is probably related to the first issue.

my bad. the second issue is non issue. The behavior is correct.

YuHsuan-Hwang:
@kswang1029 
> 1. If we load an image cube then immediately set a new rest frequency, the displayed velocity in the cursor info is not updated accordingly. We will need to change to a new channel to make the new rest frequency effective in the velocity conversion.

Fixed this issue. The velocity string is automatically updated

> Another usability improvement: we should disable the "set rest frequency" menu when animation playback is on, just to simplify things to prevent unforeseen side effects.

I disabled the save button if the animation is active (in case the dialog is already open)

kswang1029 (approved):
nice. 

Later on we can also add a new tab in the image list setting dialog and let user configure rest frequency of loaded images when possible. The current context menu can be severed as a shortcut then.

src/components/Dialogs/RestFreqDialog/RestFreqDialogComponent.tsx
```diff
@@ -0,0 +1,154 @@
...
+        } else if (restFreq >= 1e6) {
+            return {value: restFreq / 1e6, unit: FrequencyUnit.MHZ};
+        } else if (restFreq >= 1e3) {
+            return {value: restFreq / 1e3, unit: FrequencyUnit.KHZ};
+        } else {
+            return {value: restFreq, unit: FrequencyUnit.HZ};
+        }
+    };
+
+    private saveRestFreq = () => {
```
veggiesaurus: perhaps make this an `@action` to bundle the called actions together

veggiesaurus (changes requested):

src/components/Dialogs/RestFreqDialog/RestFreqDialogComponent.tsx
YuHsuan-Hwang: I'm not sure if it's necessary because only `hideRestFreqDialog()` is an `@action` function. or do we consider also add `@action` to `updateCustomRestFreq()`?

veggiesaurus: yes, that should also be an `@action`, since it calls multiple actions. Functions that call multiple actions should also be actions, so that we reduce the number of reactions

YuHsuan-Hwang: @veggiesaurus added `@action` to `saveRestFreq()` and `updateCustomRestFreq()`. The features work fine after the change

PR #1650, opened 2021-09-09
Thchang/multiple panel view buttons

### LINKED ISSUE #1614: add control buttons for single/multiple panel view
Possible place for the single/multiple panel buttons is the top-right corner of the widget. 

Screen Shot 2021-08-10 at 2 44 07 PM

TienHao:
closes #1614. Adding control buttons for single/multiple panel in Image Viewer.

An `autorun` is added in AppStore to update panel page buttons through jQurry.

### DIFF
--- a/src/stores/WidgetsStore.ts
+++ b/src/stores/WidgetsStore.ts
@@ -672,15 +672,16 @@ export class WidgetsStore {
                         stackHeaderControlButtons.children[0].remove();
                     }
 
+                    // add image panel control buttons when tab is image-view
                     if (component === "image-view") {
-                        const nextPageButton = $(`<li class="lm-multi-panel-forward" title="next page"><span class="bp3-icon-standard bp3-icon-step-forward" style/></li>`);
-                        nextPageButton.on("click", this.onNextPagePinedClick);
+                        const nextPageButton = $(`<li class="lm-image-panel-next" title="next page"><span class="bp3-icon-standard bp3-icon-step-forward" style/></li>`);
+                        nextPageButton.on("click", this.onNextPageClick);
                         stack.header.controlsContainer.prepend(nextPageButton);
-                        const multiPanelControlButton = $(`<li class="lm-multi-panel" title="switch multi-panel Mode"><span class="bp3-icon-standard bp3-icon-grid-view"/></li>`);
-                        multiPanelControlButton.on("click", this.onMultiPanelPinedClick);
-                        stack.header.controlsContainer.prepend(multiPanelControlButton);
-                        const previousPageButton = $(`<li class="lm-multi-panel-backward" title="previous page"><span class="bp3-icon-standard bp3-icon-step-backward" style/></li>`);
-                        previousPageButton.on("click", this.onPreviousPagePinedClick);
+                        const imagePanelButton = $(`<li class="lm-image-panel" title="${this.getImagePanelButtonTooltip(PreferenceStore.Instance.imagePanelMode)}"><span class="bp3-icon-standard bp3-icon-grid-view"/></li>`);
+                        imagePanelButton.on("click", this.onImagePanelButtonClick);
+                        stack.header.controlsContainer.prepend(imagePanelButton);
+                        const previousPageButton = $(`<li class="lm-image-panel-previous" title="previous page"><span class="bp3-icon-standard bp3-icon-step-backward" style/></li>`);
+                        previousPageButton.on("click", this.onPreviousPageClick);
                         stack.header.controlsContainer.prepend(previousPageButton);
                     }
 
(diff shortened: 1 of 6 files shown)


src/stores/WidgetsStore.ts
```diff
@@ -806,6 +821,61 @@ export class WidgetsStore {
...
+    @action onPreviousPageClick = () => {
+        const appStore = AppStore.Instance;
+        if (appStore.currentImagePage > 0) {
+            appStore.setActiveFrameByIndex((appStore.currentImagePage - 1) * appStore.numImageColumns * appStore.numImageRows);
+        }
+    };
+
+    updateImagePanelPageButtons = () => {
+        const appStore = AppStore.Instance;
+        $(".lm_goldenlayout")
```
veggiesaurus: I think we could split this up into multiple lines to make it more readable.

veggiesaurus (changes requested):
This works even better than I'd imagined. One minor code ccomment, and then:
- Can you make the tooltip say "next image" when in single panel mode, instead of "next page"?
- Can we use a different icon for the different modes?
- (question for @kswang1029) Do we want to switch between dynamic and fixed with this button, or just between single and multi-panel?

kswang1029:
> * (question for @kswang1029) Do we want to switch between dynamic and fixed with this button, or just between single and multi-panel?

@veggiesaurus I think single<->multiple is sufficient. Fixed<->Dynamic sounds like a property of the "multiple" mode so it makes sense to me to use the setting dialog for configuration.

TienHao:
@kswang1029: which multiple mode(Fixed/Dynamic) do we want to switch between single and multi-panel by the switch button?

kswang1029:
@TienHao what’s your opinion of having (or not) the option of fixed/dynamic at the top right corner?

kswang1029 (commented):
This works quite nicely 🚀

I would suggest we 
1. align the icon with the current view configuration. If it is single panel, we show 口 and if it is multiple panel, we show 田. (@veggiesaurus I am using Chinese characters 😉). The tooltip needs to be adjusted too. 
2. set the icon to support only two states (single panel vs multiple panel). As it is dynamic multiple panel or fixed multiple panel, we let users set in the setting dialog. The main reasoning is as it is dynamic or fixed mode, users will only tell the difference from the image viewer _after_ loading a new image. So having these two options in the click cycle feels a bit odd in my point of view.

TienHao:
> 2. set the icon to support only two states (single panel vs multiple panel). As it is dynamic multiple panel or fixed multiple panel, we let users set in the setting dialog. The main reasoning is as it is dynamic or fixed mode, users will only tell the difference from the image viewer _after_ loading a new image. So having these two options in the click cycle feels a bit odd in my point of view.

I suggest that we add a toggle named **enable multi-panel** in the setting, and remove the option `Single panel only` in the multi panel mode dropdown. 
When the **enable multi-panel** toggle is off, the multi panel mode dropdown is disabled and the image viewer apply single panel mode.

veggiesaurus:
That sounds like a good plan

src/stores/WidgetsStore.ts
```diff
@@ -806,6 +822,63 @@ export class WidgetsStore {
         }
     };
 
+    @action onImagePanelButtonClick = () => {
```
YuHsuan-Hwang: perhaps we could remove `@action` here because we don't directly modify the observables

src/stores/WidgetsStore.ts
```diff
@@ -806,6 +822,63 @@ export class WidgetsStore {
...
+        this.setImageMultiPanelEnabled(!PreferenceStore.Instance.imageMultiPanelEnabled);
+    };
+
+    @action setImageMultiPanelEnabled = (multiPanelEnabled: boolean) => {
```
YuHsuan-Hwang: same comment

src/stores/WidgetsStore.ts
```diff
@@ -806,6 +822,63 @@ export class WidgetsStore {
...
+
+    private getImagePanelButtonTooltip = (imagePanelMode: ImagePanelMode) => {
+        return imagePanelMode === ImagePanelMode.None ? "switch to multi-panel" : "switch to single panel";
+    };
+
+    private getImagePanelButtonIcon = (imagePanelMode: ImagePanelMode) => {
+        return imagePanelMode === ImagePanelMode.None ? "bp3-icon-square" : "bp3-icon-grid-view";
+    };
+
+    @action onNextPageClick = () => {
```
YuHsuan-Hwang: same comment

src/stores/WidgetsStore.ts
```diff
@@ -806,6 +822,63 @@ export class WidgetsStore {
...
+        const appStore = AppStore.Instance;
+        const firstIndexInNextPage = (appStore.currentImagePage + 1) * appStore.imagesPerPage;
+        if (appStore.frames?.length > firstIndexInNextPage) {
+            appStore.setActiveFrameByIndex(firstIndexInNextPage);
+        }
+    };
+
+    @action onPreviousPageClick = () => {
```
YuHsuan-Hwang: same comment

YuHsuan-Hwang (commented):
The feature works good 👍  I have minor questions about `@action`

src/stores/WidgetsStore.ts
veggiesaurus: Yes, I agree. We only need `@action` decorators when we do either the following:
- Modify `@observable` properties directly
- Call multiple `@action` functions from within the same function. Decorating the calling function with `@action` means that any `@computed` or `autorun` dependencies will only be run once, _after_ the calling function has exited.

TienHao: thanks for your good reminder

kswang1029 (approved):
works nicely 👍

Issue #1243, opened 2020-10-26
Catalog tool: Online catalog query (SIMBAD)


low-sky:
Use online query based on TAP to retrieve remote catalog data to display on region of the sky being examined.

Edit: issue now specifically focused on SIMBAD queries.

kswang1029:
tagged as v3.0 for now

kswang1029:
@low-sky could you come up quickly a doc about the details?

Issue #1593, opened 2021-07-30
Support boolean columns in catalogue table


veggiesaurus:
The catalogue table needs to support displaying and filtering boolean columns
- [x] Handle boolean types from the backend
- [x] Display boolean types correctly
- [x] Filter boolean types correctly

Related to https://github.com/CARTAvis/carta-backend/issues/882

veggiesaurus:
@panchyo0 looks like one of your recent PRs broke this, can you try fix it? it might be due to a bad merge?

PR #1651, opened 2021-09-09
Qi/online catalog (SIMBAD)

panchyo0:
related issue #1243
1. `search radius` range is 0 - 90 deg according simbad.
2. [In Simbad, the coordinate system parameter is never interpreted. All coordinates MUST be expressed in the ICRS. coordinate system ](http://simbad.u-strasbg.fr/simbad/tap/help/adqlHelp.html#geometricalFunctions), so only ICRS available for SIMBAD as `center coordinate` system.
3. `center coordinate` will auto update once user frozen curser at image viewer.


kswang1029:
I tested this PR and found the following issues/suggestions/comments so far:

- when rendering the retrieved catalog as an image overlay, it seems always to have position offset and the sources are not matched to the features in the image. The following example is the famous M51 galaxy, but the returned source (setting the max number of objects = 1) appears not correct. 

Screen Shot 2021-09-13 at 2 06 19 PM

- Can we use a button named as "Resolve" to the right of the text field? It would be more intuitive than the current implementation I think. We might consider to use an icon next to the button to show if the source is found (with v) or not (with x).

- for the "Search Radius" entry, can we support unit conversion too? That is, if I enter 1 with unit deg and switch to arcmin via the dropdown, the value changes to 60.

- for the "Search Radius" entry, it is better to keep the unit persist as it was after pressing the "set to image" button. Now it is always be degree.

- for the "center coordinates", we should also support the sexagesimal format. 

- for the "center coordinates", the coordinate reference should be better aligned with the image. That is, if the image is in GAL, then the default should be GAL for better UX. 

- could you explain a bit on how the "search radius" is derived when "set to image" button is clicked? (maybe it should be "set to viewer" to be more precise?)

- the long title is not displayed well with the catalog setting dialog. it is a global issue when the title is long...

Screen Shot 2021-09-13 at 2 16 25 PM

- I suggest we set the following as the default displayed columns from SIMBAD.

Screen Shot 2021-09-13 at 2 32 14 PM

src/components/Dialogs/CatalogQueryDialog/CatalogOnlineQueryDialogComponent.tsx
```diff
@@ -0,0 +1,305 @@
...
+import {ClearableNumericInputComponent, SafeNumericInput} from "components/Shared";
+import {CatalogSystemType} from "models";
+import "./CatalogOnlineQueryDialogComponent.scss";
+import {ApiService} from "services";
+
+@observer
+export class CatalogQueryDialogComponent extends React.Component {
+    private static readonly DefaultWidth = 550;
+    private static readonly DefaultHeight = 500;
+    private static readonly DBMap = new Map<CatalogDatabase, {type: CatalogSystemType[]; prefix: string}>([
```
veggiesaurus: please move anything interacting with an external API via HTTP to a separate `Service` class

src/components/Dialogs/CatalogQueryDialog/CatalogOnlineQueryDialogComponent.tsx
```diff
@@ -0,0 +1,305 @@
...
+        );
+    }
+
+    private query = () => {
+        const configStore = CatalogOnlineQueryConfigStore.Instance;
+        // In Simbad, the coordinate system parameter is never interpreted. All coordinates MUST be expressed in the ICRS coordinate system
+        const baseUrl = CatalogQueryDialogComponent.DBMap.get(configStore.catalogDB).prefix;
+        const query = `SELECT Top ${configStore.maxObject} * FROM basic WHERE CONTAINS(POINT('ICRS',ra,dec),CIRCLE('ICRS',${configStore.centerCoord.x},${configStore.centerCoord.y},${configStore.radiusInDeg}))=1 AND ra IS NOT NULL AND dec IS NOT NULL`;
+        configStore.setQueryStatus(true);
+        AppStore.Instance.appendOnlineCatalog(baseUrl, query, this.cancelTokenSource)
```
veggiesaurus: `appendOnlineCatalog` should be moved to a service class

src/models/Processed.ts
```diff
@@ -169,3 +169,60 @@ export class ProtobufProcessing {
         return dataMap;
     }
 }
+
+enum SIMBADTYPE {
```
veggiesaurus: `SimbadType`

src/models/Processed.ts
```diff
@@ -169,3 +169,60 @@ export class ProtobufProcessing {
...
+    CHAR = "CHAR",
+    SHORT = "SHORT",
+    FLOAT = "FLOAT",
+    LONG = "LONG",
+    INT = "INT",
+    DOUBLE = "DOUBLE"
+}
+
+export class APIProcessing {
```
veggiesaurus: separate file

src/services/ApiService.ts
```diff
@@ -505,4 +505,13 @@ export class ApiService {
             }
         }
     };
+
+    public getSimbad = (apiAddress: string, query: string, cancelTokenSource?: CancelTokenSource): Promise<AxiosResponse<any>> => {
```
veggiesaurus: move into separate class. `ApiService` is forinteracting with the CARTA backend/server API. We can rename this class if you want

src/services/ApiService.ts
```diff
@@ -505,4 +505,13 @@ export class ApiService {
...
+        if (cancelTokenSource) {
+            return axios.get(`${apiAddress}${query}`, {
```
veggiesaurus: you can set the base URL in a new axios instance

src/stores/AppStore.ts
```diff
@@ -873,6 +848,97 @@ export class AppStore {
         });
     };
 
+    // Online Catalog Query
```
veggiesaurus: you can move some of this to the newly created service class. 

Can you not just use `await` to simplify this?

src/stores/CatalogOnlineQueryProfileStore.ts
```diff
@@ -0,0 +1,222 @@
...
+    // do not need infinite scroll for API data
+    @computed get shouldUpdateData(): boolean {
+        return false;
+    }
+
+    @computed get loadOntoImage() {
+        return this.loadingData;
+    }
+
+    @computed get maxRows(): number {
```
veggiesaurus: what's the point of having a `@computed` wrapper around an `@observable`? is this just so that it adheres to the parent class? If so, I'd make the `@observable` values private

src/stores/CatalogOnlineQueryProfileStore.ts
```diff
@@ -0,0 +1,222 @@
...
+                        const bNumber = catalogColumn.data[b] as number;
+                        return direction * (aNumber < bNumber ? -1 : 1);
+                    });
+                    break;
+            }
+        }
+    }
+
+    @action initSortedIndexMap() {
+        this.sortedIndexMap = [];
```
veggiesaurus, line 125: shouldn't `sortedIndexMap` be a `Map<>`?

src/utilities/parsing.ts
```diff
@@ -61,7 +61,11 @@ export function findDeep(obj: any, pred: (obj: any) => boolean) {
 
 // parsing filter string for TableComponent filter function
 function getNumberFromFilterString(filterString: string): number {
-    return Number(filterString.replace(/[^0-9.+-.]+/g, ""));
+    const n = filterString.replace(/[^0-9.+-.]+/g, "");
```
veggiesaurus, line 64: can we not turn this into a single regex with the comparison operator? @confluence is our regex expert here :laughing:

veggiesaurus (changes requested):
making great progress :sunglasses:

src/stores/CatalogOnlineQueryProfileStore.ts
panchyo0: the order of this array is represented the sort order. And the value is the index in the original catalog data 😄. Should we rename it to `sortedIndexArray`

panchyo0:
Hi @kswang1029, I fixed those issues, Please check again, thanks
> when rendering the retrieved catalog as an image overlay, it seems always to have position offset and the sources are not matched to the features in the image. The following example is the famous M51 galaxy, but the returned source (setting the max number of objects = 1) appears not correct.

I think the offset comes from the `search radius`, SIMBAD does not return the data according the distance to the source. To solve this, I added another column `dist` as an extra column to the returned data and set the default order according `dist`

> ould you explain a bit on how the "search radius" is derived when "set to image" button is clicked? (maybe it should be "set to viewer" to be more precise?)

when user click the `set to viewer` button, the `center coordinate` will auto update to the current view center. Then I get the pixel coordinates for X1(xmax, ymax) and X2 (xmin, ymin), and convert X1 and X2 to ICRS coordinates (degree). After that I will calculate the distance between X1 and X2. Then the `search radius = the distance between X1 and X2 / 2` Not sure if this is the right way to do it 🙈

kswang1029:
> when user click the `set to viewer` button, the `center coordinate` will auto update to the current view center. Then I get the pixel coordinates for X1(xmax, ymax) and X2 (xmin, ymin), and convert X1 and X2 to ICRS coordinates (degree). After that I will calculate the distance between X1 and X2. Then the `search radius = the distance between X1 and X2 / 2` Not sure if this is the right way to do it 🙈

this sounds sensible I think. 🙂 As the image field might have a rotation or have significant projection distortion, I think we should also check the other diagonal distance (xmin, ymax) (xmax, ymin) then take their max. Ideally (probably overkill...), we should check
Screen Shot 2021-09-16 at 1 03 09 PM

src/utilities/parsing.ts
panchyo0: This regex will return the numerical part from the filter string. Not sure what we should do at here 🤔

kswang1029:
Few more feedback:

- not all the entries in the "xxx_bibcode" columns are hyperlink, e.g.,
Screen Shot 2021-09-24 at 10 25 51 AM

column index of xxx_bibcode: [1, 11, 18, 24, 30, 42, 52] 

- usability improvement: the "dist" column is now displayed in degree which may not be that friendly to some users. It would be better if we use arcsec as the unit, same as SIMBAD, so that users feel familiar with.

- usability improvement: at the moment ra and dec are displayed in degree. It would be good to construct a _new_ column as string to show the the coordinate (ra, dec) in sexagesimal format (note sorting in the column will become meaning less but it is fine).

- I tried to use the distance measuring tool to test the search radius estimate. In the following example, the search radius is estimated as 8.689197478226912 deg but from the distance measuring tool, the distance from bottom-left to top-right corner (roughly as I cannot click the true corners due to the blockage of the tool bar and cursor info elements) is ~25 deg. So the radius should be ~12 deg-ish. Any idea?

Screen Shot 2021-09-24 at 11 01 53 AM


ps sorting and filtering work fine 👍

src/services/CatalogApiService.ts
```diff
@@ -0,0 +1,109 @@
...
+import {APIProcessing, CatalogInfo, CatalogType} from "models";
+import {AppStore, CatalogOnlineQueryConfigStore, CatalogOnlineQueryProfileStore} from "stores";
+// import {makeObservable} from "mobx";
+
+export enum CatalogDatabase {
+    SIMBAD = "SIMBAD"
+}
+
+export class CatalogApiService {
+    public static readonly SimBadHyperLink: {bibcode: string; mainId: string} = {bibcode: "https://ui.adsabs.harvard.edu/abs/", mainId: "https://simbad.u-strasbg.fr/simbad/sim-id?Ident="};
```
veggiesaurus: lower-case `B`

src/services/CatalogApiService.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    static get Instance() {
+        if (!CatalogApiService.staticInstance) {
+            CatalogApiService.staticInstance = new CatalogApiService();
+        }
+        return CatalogApiService.staticInstance;
+    }
+
+    constructor() {
+        // makeObservable(this);
```
veggiesaurus: remove comment

src/services/CatalogApiService.ts
```diff
@@ -0,0 +1,109 @@
...
+    // Online Catalog Query
+    public appendOnlineCatalog = async (query: string): Promise<number> => {
+        const frame = AppStore.Instance.activeFrame;
+        if (!frame) {
+            AppToaster.show(ErrorToast("Please load the image file"));
+            throw new Error("No image file");
+        }
+
+        const fileId = AppStore.Instance.catalogNextFileId;
+        const dataSize = await this.getSimbad(query)
```
veggiesaurus: confused as to why you are awaiting _and_ using `.then`. 

why not do 

```typescript
let dataSize: number;
try {
const res = await this.getSimbad(query);
if (...) {
   dataSize = response.data?.data?.length;
}
}
catch(err) {
...
}
```

src/services/CatalogApiService.ts
```diff
@@ -0,0 +1,109 @@
...
+    constructor() {
+        // makeObservable(this);
+        this.cancelTokenSource = axios.CancelToken.source();
+        this.axiosInstanceSimbad = axios.create({
+            baseURL: CatalogApiService.DBMap.get(CatalogDatabase.SIMBAD).baseURL,
+            cancelToken: this.cancelTokenSource.token
+        });
+    }
+
+    public getSimbad = (query: string): Promise<AxiosResponse<any>> => {
```
veggiesaurus: rename `getSimbadCatalog` or `getSimbadResponse`

veggiesaurus (changes requested):

panchyo0:
@kswang1029 could you check again about the `search radius `  bug?

kswang1029:
> @kswang1029 could you check again about the `search radius ` bug?

Yes, now the search radius should be correct. 👍 

Minor issues:
Screen Shot 2021-09-28 at 10 33 37 AM

panchyo0:
@kswang1029  I have fixed those minor issues, please check.

kswang1029 (approved):
Nice work 👍

src/components/Dialogs/CatalogQueryDialog/CatalogOnlineQueryDialogComponent.tsx
```diff
@@ -0,0 +1,376 @@
...
+        const val = parseFloat(ev.currentTarget.value);
+        const configStore = CatalogOnlineQueryConfigStore.Instance;
+        if (isFinite(val) && val <= configStore.maxRadius && val >= 0) {
+            configStore.setSearchRadius(val);
+        } else {
+            ev.currentTarget.value = clamp(val, 0, configStore.maxRadius).toString();
+        }
+    };
+
+    private handleCenterWCSXChange = ev => {
```
veggiesaurus: rename this and other functions `Wcs` instead of `WCS`

src/components/Shared/ClearableNumericInput/ClearableNumericInputComponent.tsx
```diff
@@ -77,7 +78,7 @@ export class ClearableNumericInputComponent extends React.Component<ClearableNum
     render() {
         let value = this.props.displayExponential && !this.isFocused ? toExponential(Number(this.props.value), 3) : this.props.value;
         return (
-            <FormGroup className={this.props.className} label={this.props.label} inline={true} disabled={this.props.disabled}>
+            <FormGroup className={this.props.className} label={this.props.label} inline={this.props.inline === undefined ? true : false} disabled={this.props.disabled}>
```
veggiesaurus: ` ? true : false` is unnecessary, isn't it?

src/models/CatalogApiProcessed.ts
```diff
@@ -0,0 +1,106 @@
...
+enum SimbadType {
+    CHAR = "CHAR",
+    SHORT = "SHORT",
+    FLOAT = "FLOAT",
+    LONG = "LONG",
+    INT = "INT",
+    DOUBLE = "DOUBLE"
+}
+
+export class APIProcessing {
```
veggiesaurus: please match the file name

src/services/CatalogApiService.ts
```diff
@@ -0,0 +1,106 @@
...
+                        description: "Online Catalog",
+                        coosys: [coosy]
+                    };
+                    let catalogInfo: CatalogInfo = {
+                        fileId,
+                        fileInfo: catalogFileInfo,
+                        dataSize: response.data?.data?.length,
+                        directory: ""
+                    };
+                    let catalogWidgetId = AppStore.Instance.updateCatalogProfile(fileId, frame);
```
veggiesaurus: use `const appStore = AppStore.Instance`

src/services/CatalogApiService.ts
```diff
@@ -0,0 +1,106 @@
...
+    public resetCancelTokenSource(error: any) {
+        if (axios.isCancel(error)) {
+            this.cancelTokenSource = axios.CancelToken.source();
+            this.axiosInstanceSimbad.defaults.cancelToken = this.cancelTokenSource.token;
+        }
+    }
+
+    // Online Catalog Query
+    public appendOnlineCatalog = async (query: string): Promise<number> => {
+        const frame = AppStore.Instance.activeFrame;
```
veggiesaurus: `const appStore = ...`

veggiesaurus (changes requested):

src/components/Dialogs/CatalogQueryDialog/CatalogOnlineQueryDialogComponent.tsx
panchyo0: there are some other functions like `getFormattedWCSPoint, getPixelValueFromWCS, isWCSStringFormatValid` in utilities, should I rename all of them? 🤔

src/components/Shared/ClearableNumericInput/ClearableNumericInputComponent.tsx
panchyo0: the default value is `false or undefined` from blueprintjs, in our case we need the default to be `true` 🙈 this is a one line solution, we also can add `inline = true` to all `ClearableNumericInputProps`, not sure which one is better?

kswang1029 (approved):
I have also tested the sorting and filtering functions and compared with the backend sorting and filtering functions. All looking good. 

one thing I noticed is the "dist" calculation has ~0.01-0.1 arcsec difference comparing to the downloaded votable. It would be good to try to make the dist calculation as precise as possible. This minor issue is not critical though.

src/components/Shared/ClearableNumericInput/ClearableNumericInputComponent.tsx
veggiesaurus: `this.props.inline === undefined` returns true or false already

src/components/Dialogs/CatalogQueryDialog/CatalogOnlineQueryDialogComponent.tsx
```diff
@@ -0,0 +1,376 @@
...
+
+    @action setObjectSize(objectSize: number) {
+        this.objectSize = objectSize;
+    }
+
+    @computed get resultInfo(): string {
+        let resultStr = undefined;
+        const configStore = CatalogOnlineQueryConfigStore.Instance;
+        if (configStore.isQuerying || configStore.isObjectQuerying) {
+            resultStr = `Querying ${configStore.catalogDB}`;
```
veggiesaurus: replace with just `return` instead of assigning `resultStr`

src/components/Dialogs/CatalogQueryDialog/CatalogOnlineQueryDialogComponent.tsx
```diff
@@ -0,0 +1,376 @@
...
+            this.setResultSize(dataSize);
+        });
+    };
+
+    private handleObjectUpdate = () => {
+        const configStore = CatalogOnlineQueryConfigStore.Instance;
+        const query = `SELECT basic.* FROM ident JOIN basic ON ident.oidref = basic.oid WHERE id = '${configStore.objectName}'`;
+        configStore.setObjectQueryStatus(true);
+        CatalogApiService.Instance.getSimbadCatalog(query)
+            .then(response => {
```
veggiesaurus, line 258: why don't you use await here instead?

src/models/CatalogApiProcessed.ts
```diff
@@ -0,0 +1,106 @@
...
+enum SimbadType {
+    CHAR = "CHAR",
+    SHORT = "SHORT",
+    FLOAT = "FLOAT",
+    LONG = "LONG",
+    INT = "INT",
+    DOUBLE = "DOUBLE"
+}
+
+export class CatalogApiProcessing {
```
veggiesaurus: These don't belong belong under `models`. It's just a bunch of static functions. perhaps something in `utilities`?

I guess the same is true for the `ProtobufProcessing` class, so perhaps move that one to utilities too?

src/services/CatalogApiService.ts
```diff
@@ -0,0 +1,107 @@
...
+        const fileId = appStore.catalogNextFileId;
+        let dataSize: number = 0;
+        try {
+            const response = await this.getSimbadCatalog(query);
+            if (frame && response?.status === 200 && response?.data?.data?.length) {
+                runInAction(() => {
+                    const configStore = CatalogOnlineQueryConfigStore.Instance;
+                    const headers = CatalogApiProcessing.ProcessSimbadMetaData(response.data?.metadata);
+                    const columnData = CatalogApiProcessing.ProcessSimbadData(response.data?.data, headers);
+                    const coosy: CARTA.ICoosys = {system: configStore.coordsType};
```
veggiesaurus: typo?

veggiesaurus (changes requested):

veggiesaurus (approved):

src/components/Shared/Tables/FilterableTableComponent.tsx
```diff
@@ -162,37 +168,42 @@ export class FilterableTableComponent extends React.Component<FilterableTableCom
...
-
-        let contents: any;
-        if (rowIndex < columnData.length) {
-            contents = columnData[rowIndex];
-            if (typeof contents === "boolean") {
-                contents = contents.toString();
-            }
-        } else {
-            contents = "";
+        let rowIndex = index;
```
veggiesaurus, line 178: @panchyo0 here's the issue with #1593 ...see the handling of boolean columns in the previos commit has been removed by this PR

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -394,113 +388,10 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
...
-        const filters = this.profileStore.catalogControlHeader;
-        filters.forEach((value, key) => {
-            if (value.filter !== undefined && value.display) {
-                let filter = new CARTA.FilterConfig();
-                const dataType = this.profileStore.catalogHeader[value.dataIndex].dataType;
-                filter.columnName = key;
-                if (dataType === CARTA.ColumnType.String) {
-                    filter.subString = value.filter;
-                    userFilters.push(filter);
-                } else if (dataType === CARTA.ColumnType.Bool) {
```
veggiesaurus, line 408: this has also now been removed @panchyo0, and breaks #1593

PR #1654, opened 2021-09-14
add flux density unit when image file is moment 0

### LINKED ISSUE #1632: missing unit for statistic = flux density
Reported by a user via the Helpdesk. 

If the input image is a moment 0 image with the pixel unit as Jy/beam km/s, after defining a region, flux density can be derived. The unit should be "Jy km/s" as the per beam part is integrated out. 

Screen Shot 2021-08-18 at 11 12 46 AM

acdo2002:
Fixed #1632 
I add if statement when the image file is moment 0 and then update the unitString for flux density.

### DIFF
--- a/src/components/Stats/StatsComponent.tsx
+++ b/src/components/Stats/StatsComponent.tsx
@@ -163,9 +163,9 @@ export class StatsComponent extends React.Component<WidgetProps> {
                 } else if (type === CARTA.StatsType.SumSq) {
                     unitString = `(${unit})^2`;
                 } else if (type === CARTA.StatsType.FluxDensity) {
-                    if (unit.toLowerCase().includes("beam") && !unit.toLowerCase().includes("km/s")) {
+                    if (!!unit.match(/beam/gi) && !unit.match(/km\/s/gi)) {
                         unitString = unit.split("/")[0];
-                    } else if (unit.toLowerCase().includes("km/s")) {
+                    } else if (!!unit.match(/km\/s/gi)) {
                         unitString = unit.split("/")[0] + ".km/s";
                     }
                 } else {
--- a/src/components/Stats/StatsComponent.tsx
+++ b/src/components/Stats/StatsComponent.tsx
@@ -163,11 +163,7 @@ export class StatsComponent extends React.Component<WidgetProps> {
                 } else if (type === CARTA.StatsType.SumSq) {
                     unitString = `(${unit})^2`;
                 } else if (type === CARTA.StatsType.FluxDensity) {
-                    if (/beam/gi.test(unit) && !/km\/s/gi.test(unit)) {
-                        unitString = unit.split("/")[0];
-                    } else if (/beam.km\/s/gi.test(unit)) {
-                        unitString = unit.split("/")[0] + ".km/s";
-                    }
+                    unitString = unit.replace(/\/beam/i, "");
                 } else {
                     unitString = unit;
                 }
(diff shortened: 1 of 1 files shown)


acdo2002:
I unify the unit for all different styles from different data cubes.
"Janky" => "Jy"; The first character uses upper case, others use lower case.

kswang1029:
> I unify the unit for all different styles from different data cubes.
> "Janky" => "Jy"; The first character uses upper case, others use lower case.

we better not change the format as defined in the header. 
Screen Shot 2021-09-16 at 12 18 24 PM

mJy and Mjy are different things.

acdo2002:
Keep the format as defined in the header.
I add three conditions based on the request:
"JY/BEAM" => "JY"
"jy/beam" => "jy"
"Jy/Beam" => "Jy"

acdo2002:
I modify the method to extract the unit.
please check it, now it should even fine with mJy unit

src/components/Stats/StatsComponent.tsx
```diff
@@ -163,8 +163,10 @@ export class StatsComponent extends React.Component<WidgetProps> {
                 } else if (type === CARTA.StatsType.SumSq) {
                     unitString = `(${unit})^2`;
                 } else if (type === CARTA.StatsType.FluxDensity) {
-                    if (unit === "Jy/beam") {
-                        unitString = "Jy";
+                    if (unit.toLowerCase().includes("beam") && !unit.toLowerCase().includes("km/s")) {
```
veggiesaurus: might be more elegant to use regex here?

acdo2002: I change to the regex way to find the matching string.

veggiesaurus: I was thinking of doing it in one line with regex and groups. If you use separate `match` calls and `!!`, then it's simpler to just use `includes`.

veggiesaurus:
is this logic correct? for example, if the unit is `km/s`, then the resultant string is `km.km/s` :thinking:

src/components/Stats/StatsComponent.tsx
acdo2002: Should I move back to the previous method: using unit.toLowerCase().include("beam")?

acdo2002:
the original unit = Jy/beam.km/s
I use unit.split("/")[0] + ".km/s"
then the displayed unitString would be "Jy.km/s" (because I only extract the part before "/")

veggiesaurus:
> the original unit = Jy/beam.km/s
> I use unit.split("/")[0] + ".km/s"
> then the displayed unitString would be "Jy.km/s" (because I only extract the part before "/")

I mean I tested the above code, with `unit = "km/s"` and it results in `km.km/s`.

src/components/Stats/StatsComponent.tsx
confluence: _Edit: I moved this to the review comment._

src/components/Stats/StatsComponent.tsx
```diff
@@ -163,8 +163,10 @@ export class StatsComponent extends React.Component<WidgetProps> {
                 } else if (type === CARTA.StatsType.SumSq) {
                     unitString = `(${unit})^2`;
                 } else if (type === CARTA.StatsType.FluxDensity) {
-                    if (unit === "Jy/beam") {
-                        unitString = "Jy";
+                    if (/beam/gi.test(unit) && !/km\/s/gi.test(unit)) {
+                        unitString = unit.split("/")[0];
+                    } else if (/beam.km\/s/gi.test(unit)) {
+                        unitString = unit.split("/")[0] + ".km/s";
```
confluence: You can replace this with `unitString = unit.replace(/\/beam/i, "");`

confluence (changes requested):

confluence (approved):

veggiesaurus (approved):

kswang1029 (approved):

PR #1655, opened 2021-09-15
Thchang/spectral matching type ux 1302

### LINKED ISSUE #1302: enhancement of spectral matching reference selections
ref: https://github.com/CARTAvis/carta/issues/56

in addition to improving the user manual about the default spectral reference (radio velocity), we could consider to add possible options to the context menu (via right-click) of the "matching" column of the image list widget OR we add options in the "settings" dialog (not exist yet) of the image list widget. The could help a bit the UX I guess. @jehibbard @veggiesaurus  thoughts?

TienHao:
closes #1302. Improving spectral matching type selection.

- added a sub-menu spectral matching type in the right-click menu of image list widget.
- updating matched images channels instantly after the spectral matching type was changed.

### DIFF
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -12,7 +12,7 @@ import {CARTA} from "carta-protobuf";
 import {DraggableDialogComponent} from "components/Dialogs";
 import {ScalingSelectComponent} from "components/Shared/ScalingSelectComponent/ScalingSelectComponent";
 import {ColormapComponent, ColorPickerComponent, AutoColorPickerComponent, SafeNumericInput} from "components/Shared";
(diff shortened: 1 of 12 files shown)


kswang1029 (approved):
this works nicely 👍 
open question: do you think if we show the options explicitly in the widget would improve the UX? Usually we match cubes in radio velocity so our default should work for most of  use cases. Sometimes, however, matching in frequency or channel is intended (eg band scan data cubes). At the moment the options are hidden in the context menu (ie right click). I wonder if we should show them in a more easy-to-spot place such as a setting dialog.

TienHao:
@kswang1029: After second thought, I think that showing options in the right-click menu is not as easy-to-spot as showing in a setting dialog. I will change it. Thanks for your suggestion.

veggiesaurus (approved):

YuHsuan-Hwang (approved):

kswang1029 (approved):

PR #1658, opened 2021-09-20
Angus/88 loading files spinner

### LINKED ISSUE #88: Directory labelled as empty for too long
This is related to https://github.com/CARTAvis/carta/issues/87 but it technically a different issue.

Like many CARTA users, the directory that opens by default when CARTA is launched contains many images, so the file browser takes a long time to generate the list, and in the mean-time, labels the directory as empty. After perhaps 5 seconds, a loading bar pops up and one can cancel the loading and browse to a different directory of choice.

Discussing this with @veggiesaurus, we need a better non-ideal state. We don't want the loading bar to come up for a short read, but we also don't want it to hang for so long. We discussed a 1-2 second delay for the loading bar would be much better. I'd vote for 1 second!

veggiesaurus:
closes https://github.com/CARTAvis/carta/issues/88

### DIFF
--- a/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
+++ b/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
@@ -8,7 +8,7 @@ import {Popover2, Tooltip2} from "@blueprintjs/popover2";
 import {CARTA} from "carta-protobuf";
 import {FileInfoComponent, FileInfoType} from "components/FileInfo/FileInfoComponent";
 import {FileListTableComponent} from "./FileListTable/FileListTableComponent";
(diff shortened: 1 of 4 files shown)


YuHsuan-Hwang (commented):
the non-ideal states work nice
I would prefer a background without the blur filter because there is no similar design elsewhere in Carta

kswang1029 (approved):
This works well. I am not sure what to do with the blurry effect. Maybe it looks better with no blurry effect or even-more blurry effect. 

UPDATE:
no blur
Screen Shot 2021-09-23 at 12 02 15 PM

blur 10px
Screen Shot 2021-09-23 at 11 59 12 AM

no blur seems better. 🤔

veggiesaurus:
ok, I've removed the blur :+1:

YuHsuan-Hwang (approved):

PR #1659, opened 2021-09-22
hide regions from raster view / list while loading

veggiesaurus:
Adjusts batched imports as follows:
- increased batch size to 100
- wrap batch import in `@action` decorator to limit O(N^2) behaviour.
- hide regions from region list and image view while loading

### DIFF
--- a/src/components/ImageView/RegionView/RegionViewComponent.tsx
+++ b/src/components/ImageView/RegionView/RegionViewComponent.tsx
@@ -535,12 +535,15 @@ export class RegionViewComponent extends React.Component<RegionViewComponentProp
     };
 
     render() {
+        const appStore = AppStore.Instance;
         const frame = this.props.frame;
         const regionSet = frame.regionSet;
         const className = classNames("region-stage", {docked: this.props.docked});
         let regionComponents = null;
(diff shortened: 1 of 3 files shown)


duidae:
nice :+1:

duidae (approved):

kswang1029 (approved):
Loading 4000 regions is now faster. 👍 
Just a video note for the remaining known issues:
- after the progress bar is gone, there is a few-second gap to see regions get rendered. 
- enable the region list takes time.
- zoom/pan is struggling (not recorded in the video)

https://user-images.githubusercontent.com/20819712/134453868-6206d279-e1c5-445b-a999-a94b9a654325.mov

veggiesaurus:
> Loading 4000 regions is now faster. 
> Just a video note for the remaining known issues:
> 
> * after the progress bar is gone, there is a few-second gap to see regions get rendered.
> * enable the region list takes time.
> * zoom/pan is struggling (not recorded in the video)
> 
>  Screen.Recording.2021-09-23.at.12.12.32.PM2.mov

that should be fixed once the region view and rendering is sorted out

PR #1660, opened 2021-09-24
Yuhsuan/1656 line region pa

### LINKED ISSUE #1656: Position angle of the line region
to be consistent with the definition in CASA, a vertical line should have the position angle as 0 degree. At the moment in CARTA, a horizontal line has the position angle as 0 degree. We will need to shift the reference by 90 degrees.

YuHsuan-Hwang:
This PR closes #1656:
* adjusted the definition of position angle of line regions by 90 degrees
* removed the duplicate code for setting position angle of line regions


kswang1029 (approved):
yes now the pa definition is "more" consistent with casaviewer. Note that casaviewer has negative pa if the line goes from bottom-left to top-right.

veggiesaurus (approved):

PR #1664, opened 2021-10-07
Jason/performance improve region view pan zoom line region

### LINKED ISSUE #90: Poor performance when many regions loaded
One awkward thing about CARTA (a fast visualisation tool for big data) is how it doesn't handle many regions. A standard workflow would include overlaying a ds9 region file output from a source finder like PyBDSF or Aegean, which could include hundreds or thousands of regions, which CARTA cannot handle. I believe I've seen CARTA either crash or freeze for such cases. Although the catalogue tool could handle this, it does not replicate the region files that show the exact shape of the fitted sources. 

It may be worth considering limiting the region functionality above some threshold of a large number of regions (10? - assuming that's the limiting factor). As a user, I don't need all the specific region functionality when I have 100 of them loaded. And to remove all the regions loaded, one does not want to go through and manually remove every single one. 

ds9 can handle this and also enables selecting all regions to remove or change colour, etc. Perhaps CARTA could go beyond this by putting regions within groups that can be labelled? A simple shift- / control-click to select multiple and group them would work.

My apologies if this is an issue already submitted, or if it should really be two issues!

duidae:
closes https://github.com/CARTAvis/carta/issues/90


src/components/ImageView/ImagePanel/ImagePanelComponent.tsx
```diff
@@ -29,6 +29,8 @@ export class ImagePanelComponent extends React.Component<ImagePanelComponentProp
     @observable imageToolbarVisible: boolean = false;
     readonly activeLayer: ImageViewLayer;
 
+    private regionViewRef;
```
veggiesaurus: shouldn't this be typed?

src/stores/FrameStore.ts
```diff
@@ -65,6 +65,8 @@ export class FrameStore {
     private cachedTransformedWcsInfo: AST.FrameSet = -1;
     private zoomTimeoutHandler;
 
+    public requiredFrameViewForRegionRender;
```
veggiesaurus: can you add type info here?

veggiesaurus (approved):
Code looks good to me. Two minor typing additions requested, but other than that looks fine. A lot of code changes though, so careful testing by @kswang1029 needed

kswang1029 (changes requested):
I see an out-of-sync issue when we have matched images.


https://user-images.githubusercontent.com/20819712/136878103-0669f2ca-d9ad-4dec-ad85-b4e4e35dfadf.mov

kswang1029:
found an out-of-sync issue when panning the image. Please focus on the center part of the movie and focus on the relative positions between the regions and image features. If we pan slightly, sometimes we see the image and regions are synchronized but sometimes they are not. The example is with 2000 point regions. 

https://user-images.githubusercontent.com/20819712/139632147-5d21b6a5-7902-4401-824f-d1a7e62c9ca1.mov

duidae:
> found an out-of-sync issue when panning the image. Please focus on the center part of the movie and focus on the relative positions between the regions and image features. If we pan slightly, sometimes we see the image and regions are synchronized but sometimes they are not. The example is with 2000 point regions.
> 
Fixed, please have a look.

kswang1029 (approved):
this looks good as far as I can tell. 👍

PR #1666, opened 2021-10-08
Mylin/set the cursor type to be default when only cursor region

### LINKED ISSUE #1599: cursor type remains as + after deleting a polygon or polyline region if the cursor is on a side
https://user-images.githubusercontent.com/20819712/128282426-32ffa39e-9d98-483c-8b30-be2a187562f7.mov

acdo2002:
Fixed #1599 
I set the cursor type to be "default" when the selectedRegion only remain the cursor region itself or the selectedRegion to be null.


YuHsuan-Hwang (approved):
works fine 👍

veggiesaurus (approved):

kswang1029 (approved):

PR #1670, opened 2021-10-13
Angus/637 telemetry

### LINKED ISSUE #637: Optional user telemetry (usage statistics unique user count)
Note: this issue will now focus on the telemetry side of things, rather than the news feed. The initial suggestion is shown below for context.

I suggest we do the following:
- When the frontend starts, it checks if a `uuid` value exists in  preferences / `localStorage`. If not, it generates a new value.
- When the frontend splash screen is opened. it makes an HTTP `GET` request to a URL defined at build-time by `REACT_APP_NEWS_FEED_URL`. It passes the `uuid` as a parameter, as well as the version. 
- The news feed checks a mongoDB for news items for the given version
- The news feed HTTP server checks if the `uuid` is listed in the current month's list of active users. If not, it adds them to the list, along with the version they're running. If the version passed to the server is newer than the version stored for the `uuid`, it updates it. 
- At the end of the month, it consolidates the number of entries in the list, and the distribution of versions. The aggregated statistics are stored in a separate mongoDB collection, and the current month's stats are cleared.

This approach has minimal impact on user privacy, as individual users' app behaviour is not tracked in any way, only the launching of the app itself. In addition, the `uuid` is entirely anonymous, and is not tracked past the current month.

If we want to extend the analytics slightly, we could also record an approximate (country-level) geo-location, but that has some user privacy questions.

veggiesaurus:
Closes #637 

Note: Telemetry server (repo: [CARTAvis/carta-telemetry](https://github.com/CARTAvis/carta-telemetry)) is a work-in-progress, and currently just logs things to `stdout`. This will be implemented next.

Telemetry can be disabled altogether at build time using env files (`REACT_APP_SKIP_TELEMETRY=true`) or by passing `skipTelemetry=1` as a URL parameter.

@ajm-asiaa you might want to add the build-time env variable to mix 'n match once this is merged into `dev`.


kswang1029:
@veggiesaurus where can I track for this new feature other than from the console?
Screen Shot 2021-10-29 at 1 08 46 PM

kswang1029:
Do we expect to have a new tab "Telemetry" here (based on commit a736f3d)? 
Screen Shot 2021-10-29 at 1 15 31 PM

veggiesaurus:
> Do we expect to have a new tab "Telemetry" here (based on commit a736f3d)? 
> Screen Shot 2021-10-29 at 1 15 31 PM
> 

Yes. Did you opt in to the telemetry initially? You may have to refresh your browser afterwards 🤔

kswang1029:
> > Do we expect to have a new tab "Telemetry" here (based on commit [a736f3d](https://github.com/CARTAvis/carta-frontend/commit/a736f3d1f0c76e94e5f9f16fa62d6b08a2d43fc6))?
> > Screen Shot 2021-10-29 at 1 15 31 PM
> 
> Yes. Did you opt in to the telemetry initially? You may have to refresh your browser afterwards 🤔

do I need to start the backend with a special flag beforehand? I started as usual and can see the platform info from the console message. 

I also built CARTA with a ubuntu VM and the telemetry tab is missing too.

veggiesaurus:
> > > Do we expect to have a new tab "Telemetry" here (based on commit [a736f3d](https://github.com/CARTAvis/carta-frontend/commit/a736f3d1f0c76e94e5f9f16fa62d6b08a2d43fc6))?
> > > Screen Shot 2021-10-29 at 1 15 31 PM
> > 
> > 
> > Yes. Did you opt in to the telemetry initially? You may have to refresh your browser afterwards thinking
> 
> do I need to start the backend with a special flag beforehand? I started as usual and can see the platform info from the console message.
> 
> I also built CARTA with a ubuntu VM and the telemetry tab is missing too.

no, it should depend only on the frontend settings. That's odd. So you didn't encounter any popup when you first started CARTA? can you open the dev console and run `app.preferenceStore.resetTelemetrySettings()`

kswang1029:
> no, it should depend only on the frontend settings. That's odd. So you didn't encounter any popup when you first started CARTA? can you open the dev console and run `app.preferenceStore.resetTelemetrySettings()`

Nope, there is no such popup. If I run `app.preferenceStore.resetTelemetrySettings()` and yes I immediately see the popup. Even if I clicked yes to enable telemetry, there is still no telemetry tab in the preference dialog. ~~Trying to build from scratch now.~~ no luck when building from scratch. 

Is it because of this?
Screen Shot 2021-11-02 at 12 35 46 PM

panchyo0 (approved):

src/stores/PreferenceStore.ts
```diff
@@ -692,6 +719,10 @@ export class PreferenceStore {
         this.clearPreferences([PreferenceKeys.CATALOG_DISPLAYED_COLUMN_SIZE, PreferenceKeys.CATALOG_TABLE_SEPARATOR_POSITION]);
     };
 
+    @action resetTelemetrySettings = () => {
+        this.clearPreferences([PreferenceKeys.TELEMETRY_CONSENT_SHOWN, PreferenceKeys.TELEMETRY_MODE, PreferenceKeys.TELEMETRY_UUID, PreferenceKeys.TELEMETRY_LOGGING]);
```
YuHsuan-Hwang: I went through the code although I'm not familiar with `axios` and `idb`
I have one question about the user id design: do we expect that the user generates a new user ID by clicking the `restore defaults` button?

veggiesaurus: @YuHsuan-Hwang thanks, I have fixed this. It should now generate and display a new user ID

veggiesaurus:
@kswang1029 I've adjusted the logic when skipping telemetry. Can you re-test after running `app.preferenceStore.resetTelemetrySettings()` and then refreshing your browser?

src/stores/PreferenceStore.ts
YuHsuan-Hwang: @veggiesaurus sorry, I mean if we allow user to easily generate a new ID, will this causes the number of IDs to be much greater than the actual number of users?

veggiesaurus: yes, but I'm not really sure what the alternative is. Not allowing the user to generate a new uuid might be problematic :thinking: I'm trying to see what Atom does.

If they just delete the `uuid` entry in their `preferences.json` it will also generate a new uuid.

veggiesaurus: perhaps the middle ground is to _not_ generate a new token when you click the "reset defaults" button, and users could always delete the entry in their preferences manually to get a new token if they really wanted to? (or run `app.preferenceStore.clearPreferences(["telemetryUuid"])` as a snippet)

YuHsuan-Hwang: we can emphasize how to reset the ID in in-app help

veggiesaurus: :+1: ok. Or at least in the privacy policy. I think this is definitely more of an "expert user" thing, so I agree with you

veggiesaurus: I've made the required changes, please have a look again

YuHsuan-Hwang (approved):

kswang1029:
when we click restore default, do we expect to see a new user id? Or is there a way to regenerate a user id? When we enable log telemetry output, where can we see it?

Screen Shot 2021-11-22 at 5 57 52 PM

Further question, if we uninstall carta or even wipe out os then reinstall carta, will we have the same user id?

veggiesaurus:
> when we click restore default, do we expect to see a new user id? Or is there a way to regenerate a user id? 

The user ID will stay the same, as per @YuHsuan-Hwang's suggestion. You can clear the user ID by deleting the `telemetryUuid` field in the `preferences.json` file, or by running the following code snippet and refreshing the browser (but this is not something a user would ever normally need to do):

```javascript
app.preferenceStore.clearPreferences(["telemetryUuid"])
```

> When we enable log telemetry output, where can we see it?

It should show in the javascript console if you change the log level to "verbose".
image
I kept it there because telemetry might be very verbose, but can move it to the default log level if you'd prefer.

> Further question, if we uninstall carta or even wipe out os then reinstall carta, will we have the same user id?

The user ID is stored in `preferences.json`. I don't think we delete those if you uninstall CARTA, but if you delete the `carta` folder manually or reinstall the OS, a new user ID will be generated.

kswang1029 (approved):

PR #1672, opened 2021-10-14
Qi/switch to webgl2 for image view and contours

### LINKED ISSUE #1350: Switch to WebGL2 rendering
We should attempt to switch to WebGL2 for raster and contour rendering

### LINKED ISSUE #1574: Reduce repetition in GLSL shader code
We should use something like glslify to share shader code between the different shader programs (catalog, raster & contour)

panchyo0:
This PR should be merged after PR #1668 

related issue #1350 #1574

- [x] switch to webgl2 for image view and contours
- [x] reduce repetition code in shaders with template literals


src/services/GLSL/CatalogShaders.ts
```diff
@@ -0,0 +1,54 @@
...
+#define TRIANGLE_LINED_DOWN 13
+#define HEXAGON_FILLED_2 14
+#define HEXAGON_LINED_2 15
+#define CROSS_FILLED 16
+#define CROSS_LINED 17
+#define X_FILLED 18
+#define X_LINED 19
+#define LineSegment_FILLED 20
+`;
+const macorsVertex = `
```
veggiesaurus: do you mean `macros`?

src/services/GLSL/CatalogShaders.ts
```diff
@@ -0,0 +1,54 @@
...
+#define COS_0 1.0
+#define COS_45 0.70710678118
+#define SIN_60 0.86602540378
+#define COS_60 0.5
+#define SIN_90 1.0
+#define COS_90 0.0
+`;
+
+export const catalogShaders = {
+    vertexShader: `${utilities.version300}\n${macors}\n${macorsVertex}\n${utilities.imageToGL}\n${utilities.bicubicFilter}\n${utilities.rotate2D}\n${catalogVertexShader}`,
```
veggiesaurus: surely there's a neater way to do this? for example can we just export an array of strings instead of a single string? I think `gl.shaderSource` accepts a string array, that might be neater?

```javacript
{
vertexShader: [utilities.versionString, sharedMacros, vertexMacros, utilities.imageToGL, utilities.bicubicFilter, utilities.rotate2D, catalogVertexShader]`,
}
```

We could also just have a utilities.glsl file instead of splitting each function into a separate file I guess, but I'm not sure how much neater that would be. There's no major problem in including bits of the shader that don't get used, since compilation will be very quick anyway.

veggiesaurus (changes requested):

src/services/GLSL/CatalogShaders.ts
veggiesaurus: I still think we should consider using glsify for this

panchyo0: the definition from typescript for `gl.shaderSource` is [shaderSource(shader: WebGLShader, source: string): void;](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.webglrenderingcontext.html#shadersource) I tried to disable the type with `// @ts-ignore` then I got `ERROR: ',' : syntax error` when compile the code. 🤔

src/services/GLSL/utilities/utilities.ts
```diff
@@ -0,0 +1,74 @@
+export const rotate2D = /* glsl */ `
```
veggiesaurus: can't we just have a single `utilities.glsl` and import that? It doesn't matter if we import more than we need.

Having multiple GLSL snippets in a typescript file looks pretty messy

kswang1029:
@panchyo0 @veggiesaurus just to report that so far I don't see odd things in this PR.

veggiesaurus (approved):

PR #1673, opened 2021-10-15
Qi/online catalog vizier

panchyo0:
This PR should go after PR #1651 
Catalog query feature for vizier


kswang1029:
@panchyo0 could you update protobuf to 6571a44afc486902b322fbd68f81a12e96eb075e?

kswang1029:
@panchyo0 a quick question: when we make a query to VizieR, does the frontend retrieve _all_ matched table to the frontend first? Then we use the multi-select menu to select which catalogs to be displayed?

kswang1029:
Question: when the "Key words (Catalog title)" is filled, what are the query parameters here? Are the "Key words", plus the coordinate info and search radius all together as the query parameters?

Screen Shot 2021-11-01 at 3 21 07 PM

panchyo0:
> we make a query to VizieR, does the frontend retrieve _all_ matched table to the frontend first? Then we use the multi-select menu to select which catalogs to be displayed?

That's right

Yes, with `key words` vizier will return all tables that match the `search radius` and `center coordinates` with the `key words` inside the table title.
without `key words` we use `votable?-c=${point.x} ${point.y}&-c.eq=J2000&-c.${radiusUnits}=${radius}&-sort=_r&-out.max=${max}&-corr=pos&-out.add=_r,_RA,_DE&-oc.form=d&-out.meta=hud` as parameters
with `key words` we use `votable?-c=${point.x} ${point.y}&-c.eq=J2000&-c.${radiusUnits}=${radius}&-sort=_r&-out.max=${max}&-corr=pos&-out.add=_r,_RA,_DE&-oc.form=d&-out.meta=hud&-words=${keyWords}` as parameters

kswang1029:
> > we make a query to VizieR, does the frontend retrieve _all_ matched table to the frontend first? Then we use the multi-select menu to select which catalogs to be displayed?
> 
> That's right

Ok. My real question is actually performance. It is quite slow to retrieve all catalogs before selecting which to display. Is it possible to just get the titles of catalogs first? Then based on user selection the frontend retrieves data afterwards. 🤔

panchyo0:
> > > we make a query to VizieR, does the frontend retrieve _all_ matched table to the frontend first? Then we use the multi-select menu to select which catalogs to be displayed?
> > 
> > 
> > That's right
> 
> Ok. My real question is actually performance. It is quite slow to retrieve all catalogs before selecting which to display. Is it possible to just get the titles of catalogs first? Then based on user selection the frontend retrieves data afterwards. 🤔

Thanks for the suggestion, now we only load tables names at the first step. it's about 2 times faster than before. But sometime loading data for user selected tables is time consuming. That is out of our control 😞

kswang1029:
> > > > we make a query to VizieR, does the frontend retrieve _all_ matched table to the frontend first? Then we use the multi-select menu to select which catalogs to be displayed?
> > > 
> > > 
> > > That's right
> > 
> > 
> > Ok. My real question is actually performance. It is quite slow to retrieve all catalogs before selecting which to display. Is it possible to just get the titles of catalogs first? Then based on user selection the frontend retrieves data afterwards. 🤔
> 
> Thanks for the suggestion, now we only load tables names at the first step. it's about 2 times faster than before. But sometime loading data for user selected tables is time consuming. That is out of our control 😞

Thanks @panchyo0. Yes now it performs better. 👍 

Overall this looks good and will be handy for users. Just one minor concern about "Max Number of Objects" field. It makes sense when making a query to SIMBAD but it is less sensible for VizieR. At the moment it still sets as the upper limit when loading a specific catalog from VizieR. However we cannot see from the catalog title how many sources there are per catalog. So it might be the case that I only partially load one catalog. I would suggest for VizieR catalog, we don't apply such "Max Number of Objects" limitation but instead we load the full catalog for completeness. This should be more realistic use case when a user would like to study the content for a given catalog. So my suggestion would be when VizieR is selected, we don't show that field and we don't apply any cutoff when loading a VizieR catalog.

panchyo0:
>  So my suggestion would be when VizieR is selected, we don't show that field and we don't apply any cutoff when loading a VizieR catalog.

@kswang1029 There is a performance issue when we load large catalog from vizier via internet. I added a `Cutoff` switch button, so user can have option to choose use the cutoff or not. Please check. 

load 15 tables without limit, it takes about 9mins to download those catalogs
Screen Shot 2021-11-03 at 4 38 44 PM

load same table with `Max number of Objects =10` only takes 153 ms for downloading part
Screen Shot 2021-11-03 at 4 31 31 PM

kswang1029:
@panchyo0 I see. Good point. Then I would suggest to remove the toggle but just change the title from "Max number of objects" to "Max number of objects per catalog" when VizieR is selected. So if a user wants full catalog, a very large number can be used (eg 10000000).

 When it is SIMBAD, we use "Max number of objects"

panchyo0:
@kswang1029  please, check again

src/stores/CatalogOnlieQueryConfigStore.ts
```diff
@@ -78,6 +87,33 @@ export class CatalogOnlineQueryConfigStore {
...
+
+    @action removeVizierSelectedTable(table: string) {
+        this.vizierSelectedTableName = this.vizierSelectedTableName.filter(element => table !== element.name);
+    }
+
+    @action resetVizierSelectedTable() {
+        this.vizierSelectedTableName = [];
+    }
+
+    @action resetVizirR() {
```
veggiesaurus: typo?

src/utilities/CatalogApiProcessed.ts
```diff
@@ -2,16 +2,37 @@ import * as AST from "ast_wrapper";
...
     SHORT = "SHORT",
     FLOAT = "FLOAT",
     LONG = "LONG",
     INT = "INT",
-    DOUBLE = "DOUBLE"
+    DOUBLE = "DOUBLE",
+    UNSIGNEDBYTE = "UNSIGNEDBYTE"
 }
 
+export type VizieResource = {
```
veggiesaurus: typo? `VizieRResource`?

I think perhaps we should be consistent and use a lowercase `r` in `vizier` everywhere except user-facing strings.

veggiesaurus (changes requested):

kswang1029 (approved):

veggiesaurus (approved):

PR #1674, opened 2021-10-18
Mylin/1667 fix pa change when set length

### LINKED ISSUE #1667: line PA changes when manually setting a new length
https://user-images.githubusercontent.com/20819712/136578919-43775186-21e9-4d9b-a234-afeca5f9d11e.mov

acdo2002:
fixed #1667 

I found when given a length, the sin and cos in x and y is different(i.e. opposite) to setRotation action in carta-frontend/src/stores/RegionStore.ts
I make them consistent to setRotation action then the issue fixed.

I also found the existingValue in handleLengthChange should be the length of 2D point rather than just region.size.x, otherwise the existingValue could be negative.


veggiesaurus (approved):

kswang1029 (commented):
It would be good to add a comment why sin and cos need to be swapped. The usual rotation matrix requires an additional offset to meet the PA=0 definition in astronomy.

acdo2002:
The setLength will calculate the length in x and y direction based on the current PA.
I add description: rotation is in degrees counter-clockwise of north(N).
PA = 0 in North direction and increased counter-clockwise.

Thus the length in x-direction should be sin(rotation) and length in y-direction should be cos(rotation).
The calculation in PA is correct, however I have no idea why in setLength, the calculation changes to cos(rotation) in x-direction and sin(rotation) in y-direction?
I just fixed it then the setLength works normal as we expect.

src/components/Dialogs/RegionDialog/LineRegionForm/LineRegionForm.tsx
```diff
@@ -167,8 +167,8 @@ export class LineRegionForm extends React.Component<{region: RegionStore; frame:
         const existingValue = this.props.region.size.x;
```
YuHsuan-Hwang: also change this line to `length2D(this.props.region.size)`

kswang1029:
> The setLength will calculate the length in x and y direction based on the current PA. I add description: rotation is in degrees counter-clockwise of north(N). PA = 0 in North direction and increased counter-clockwise.
> 
> Thus the length in x-direction should be sin(rotation) and length in y-direction should be cos(rotation). The calculation in PA is correct, however I have no idea why in setLength, the calculation changes to cos(rotation) in x-direction and sin(rotation) in y-direction? I just fixed it then the setLength works normal as we expect.

This is because the rotation matrix is different from the usual one we learned from math course. The usual rotation matrix is
cos -sin
sin cos
if 0 deg is aligned with +x axis and the angle increases counter-clockwise. In astronomy, however, the 0 deg is aligned with +y axis and the angle increases counter-clockwise. So the rotation matrix needs to be modified with the additional 90deg offset hence we see cos<->sin with a proper +/-1 applied.

acdo2002:
@kswang1029 & @YuHsuan-Hwang 
I add the short comment to explain the PA and why cos<->sin with a proper +/-1 applied.
please check it.

YuHsuan-Hwang (approved):

kswang1029 (changes requested):
// rotation is in degrees counter-clockwise of north(N)
// additional 90 deg offset hence we see cos<->sin with a proper +/-1 applied

the comments may be revised as the following:

the rotation angle is defined to be 0 at North (mostly in +y axis) and increases counter-clockwisely. This is different from the usual definition in math where 0 degree is in the +x axis. The extra 90-degree offset swaps cos and sin with a proper +/-1 constant applied.

acdo2002:
@kswang1029 
Thanks for the suggestion, the comments have been revised now!

kswang1029:
the pa and length part are fixed. Just noticed that if we change the center manually, PA will be changed too...

acdo2002:
Same problem but in the src/stores/RegionStore.ts.
Now it should be fixed when changing length & centre position (both image and world coordinate), the PA will not be changed.

kswang1029 (approved):

YuHsuan-Hwang (commented):
shall we also add the same comments for line 433 in `RegionStore.ts` `setRotation()`?

acdo2002:
@YuHsuan-Hwang Yes, `setRotation()` in `RegionStore.ts` is the original function I followed.
I put the description in the line 433.

YuHsuan-Hwang (approved):

veggiesaurus (approved):

Issue #1662, opened 2021-10-01
The size of catalog overlay on Apple M1 and with Retina display


YuHsuan-Hwang:
The size of catalog overlay does not increase as expected when size is larger than around 30:
![Screen Recording 2021-09-30 at 3 42 16 PM](https://user-images.githubusercontent.com/43841102/135563215-32212f71-9cac-4461-b6ec-545a10b76808.gif)

Can be reproduced only on Apple M1 and with Retina display (with `devicePixelRatio = 2`). This may be due to the difference of webgl 2.0, particularly the `Aliased Point Size Range`.

Webgl 2.0 report on Apple M1 which I am using:
<img src="https://user-images.githubusercontent.com/43841102/135562151-ea863752-675d-4b4f-ad0a-ca0fed3451d3.png" width=50% height=50%>

kswang1029:
FYI: iPad Pro 13 M1 is free of this issue (ios 14.8)
Aliased point size range is [1, 511]

veggiesaurus:
This should eventually be fixed by #1802

PR #1678, opened 2021-10-21
Yuhsuan/1267 high resolution png

### LINKED ISSUE #1267: Support image export - high resolution png
as part of the epic #1009 

This is to provide the ability to export high resolution png (600dpi?) in addition to the screen resolution.

YuHsuan-Hwang:
The PR closes #1267.

Procedure of exporting a png image:

1. Click save image and select a resolution
2. Disable highlighting pixels and saving another image
3. Wait for 0.5 sec
3-1. set the canvas size to x1/x2/x4. Request new data if necessary.
3-2. check Webgl parameters `drawingBufferWidth` and `drawingBufferHeight` of the raster canvas. If they are smaller than expected, we may have requested a size larger than the `Max Render Buffer Size` or the limit of the memory. In this case, try to set the canvas size to x(n-1) instead (back to 3-1).
4. Wait until the data tiles and the contours are ready
5. Wait until the raster canvas and the contour canvas is not updated within 0.5 sec
6. Download the image
7. Enable highlighting pixels and saving images
8. Easily set the canvas size back to x1 when the image view is re-rendered (this includes zooming and panning, hovering the colorbar, change the active layer, …)

Known issues:

- Size of the catalog overlay may change due to the limit of `Aliased Point Size Range`. See #1662 for example.


kswang1029 (commented):
This works well. 👍 Just two suggestions for improvements:
1) I suggest we remove the shortcut alt+E as now it is ambiguous in resolution. Unless we add the default in preferences?
2) when requesting 400% and if the resulting canvas exceeds the limit, we see a warning toast at the bottom. I suggest we explicitly mention this is due to canvas size limitation instead of just saying failed to export as this is frustrating and does not provide additional info why.

YuHsuan-Hwang:
@kswang1029 
> 1. I suggest we remove the shortcut alt+E as now it is ambiguous in resolution. Unless we add the default in preferences?

How about we move "alt+E" to the sub-menu: "Normal (100%) alt + E"
> 2. when requesting 400% and if the resulting canvas exceeds the limit, we see a warning toast at the bottom. I suggest we explicitly mention this is due to canvas size limitation instead of just saying failed to export as this is frustrating and does not provide additional info why.

Ok, I will adjust the the warning

kswang1029:
> @kswang1029
> 
> > 1. I suggest we remove the shortcut alt+E as now it is ambiguous in resolution. Unless we add the default in preferences?
> 
> How about we move "alt+E" to the sub-menu: "Normal (100%) alt + E"

that works too. 

> 
> > 1. when requesting 400% and if the resulting canvas exceeds the limit, we see a warning toast at the bottom. I suggest we explicitly mention this is due to canvas size limitation instead of just saying failed to export as this is frustrating and does not provide additional info why.
> 
> Ok, I will adjust the the warning

YuHsuan-Hwang:
@kswang1029 adjusted the sub-menu and the warning. also removed "(alt + E)" from the toolbar tooltip

kswang1029 (approved):

veggiesaurus:
@YuHsuan-Hwang can you resolve the conflicts after @duidae's region view changes?

veggiesaurus:
@YuHsuan-Hwang I'm just wondering whether we need a progress bar of sorts while saving? If I save a 400% image from mix 'n match it takes a good 30 seconds before the image is saved properly

YuHsuan-Hwang:
@veggiesaurus 
> @YuHsuan-Hwang can you resolve the conflicts after @duidae's region view changes?

Ok. the calculation of zooming is changed, and I may take some time to resolve it

> @YuHsuan-Hwang I'm just wondering whether we need a progress bar of sorts while saving? If I save a 400% image from mix 'n match it takes a good 30 seconds before the image is saved properly

I think yes, since it could also take a long time if the network condition is bad. I prefer we file a issue because this PR is already complicated

veggiesaurus:
> @veggiesaurus
> 
> > @YuHsuan-Hwang can you resolve the conflicts after @duidae's region view changes?
> 
> Ok. the calculation of zooming is changed, and I may take some time to resolve it
> 
> > @YuHsuan-Hwang I'm just wondering whether we need a progress bar of sorts while saving? If I save a 400% image from mix 'n match it takes a good 30 seconds before the image is saved properly
> 
> I think yes, since it could also take a long time if the network condition is bad. I prefer we file a issue because this PR is already complicated

ok, will you file an issue for it then? Once you've solved the merge conflicts I'll merge this into `dev`

YuHsuan-Hwang:
> ok, will you file an issue for it then? Once you've solved the merge conflicts I'll merge this into `dev`

Ok, I will file one. we may have to re-test the regions after my update

YuHsuan-Hwang:
@veggiesaurus
I have resolved the conflicts; made some additional changes so that the region canvas updates as expected when exporting images.
Also added new issue for the progess bar: #1695

veggiesaurus (approved):

PR #1680, opened 2021-10-26
Yuhsuan/1621 complex polygon fix

### LINKED ISSUE #1621: imported polyline is considered as "complex polygon" so the color is not correct
When a polyline region is imported from a region file, the color may be in magenta if the polyline is considered as complex when we treat it as a polygon. The coloring logic for complex polygon should not be applied to the polyline in any case. 

Screen Shot 2021-08-11 at 3 35 52 PM

YuHsuan-Hwang:
Closes #1621: set `isSimplePolygon = true` by default and call `simplePolygonTest()` for only polygon regions

### DIFF
--- a/src/stores/RegionStore.ts
+++ b/src/stores/RegionStore.ts
@@ -322,6 +322,7 @@ export class RegionStore {
         } else {
             this.coordinate = RegionCoordinate.Image;
         }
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
issue fixed. 👍

PR #1687, opened 2021-10-29
ajm/jenkinsfile update

ajm-ska:
@veggiesaurus I have moved the carta-frontend CI to a new server, at least temporarily (Running Ubuntu 20.04, but not on SSDs yet). Now we can have multiple executors (I set it to 8) and that means we can build the different node versions stages in parallel. 

To do that, I noticed we need results of the WebAssembly stage for each parallel executor. Rather than building the WebAssembly stage on each executor, I have tried the stash/unstash feature.

I notice that we don't really need the entire contents of wasm_libs, so I am just stashing `wasm_libs/built`, and found we also needed `wasm_libs/zstd/build/standalone_zstd.bc`. I think we need to stash the `protobuf` folder too.
It appears to work. Is that acceptable? Do you think we would ever need other files from `wasm_libs`?

Regarding the WebAssembly stage; It is still built using node v14 for each commit. I was considering a persistent stash with `preserveStashes()` so that we could skip the WebAssembly stage. It would save ~2.5 minutes. But I guess it may be a good idea to keep the WebAssembly stage in case a commit has modified it?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@ pipeline {
     stages {
         stage('Format check') {
             agent {
-                label "ubuntu-1"
+                label "ubuntu-2004"
             }
             steps {
                 catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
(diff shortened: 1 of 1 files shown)


veggiesaurus:
> I notice that we don't really need the entire contents of wasm_libs, so I am just stashing `wasm_libs/built`, and found we also needed `wasm_libs/zstd/build/standalone_zstd.bc`. I think we need to stash the `protobuf` folder too. It appears to work. Is that acceptable? Do you think we would ever need other files from `wasm_libs`?

I think this is fine. However, if we _do_ need additional files later on, we will need to adjust the CI. But in general, I guess we'll move things into `built`.

> 
> Regarding the WebAssembly stage; It is still built using node v14 for each commit. I was considering a persistent stash with `preserveStashes()` so that we could skip the WebAssembly stage. It would save ~2.5 minutes. But I guess it may be a good idea to keep the WebAssembly stage in case a commit has modified it?

We make regular changes to the WebAssembly wrappers, but seldom adjust the libs. However, I'm sure we'll add libs in future. In which case you'll have to remove the stash?

veggiesaurus:
if a new commit comes in, won't all three compilation tests (node 12, 14 and 16) start at the same time, and all be missing the stashed WebAssembly stage? After which they'll _all_ attempt to build it?

ajm-ska:
> if a new commit comes in, won't all three compilation tests (node 12, 14 and 16) start at the same time, and all be missing the stashed WebAssembly stage? After which they'll _all_ attempt to build it?

No. The check format stage and the WebAssembly stage are done first with one executor. Once that is complete, the parallel part kicks in and two other executors join the first one. That's why I needed to stash the files from wasm_libs folder, so that the 2nd and 3rd executors could copy in the contents.

ajm-ska:
> We make regular changes to the WebAssembly wrappers, but seldom adjust the libs. However, I'm sure we'll add libs in future. In which case you'll have to remove the stash?

OK. Then I guess the safest and easiest option is to leave in the WebAssembly libs build section.

veggiesaurus (approved):

PR #1692, opened 2021-11-04
Qi/ catalog scatter linear regression

### LINKED ISSUE #1240: Catalog tool: Simple regressions and summary statistics
- Be able to report the slope and intercept of a linear fit to a selected region or subpopulation in a bivariate plot.
- Be able to report summary statistics for data in bivariate or histogram plots (e.g., mean flux density of selected sources, where flux density is a column in the catalog).

panchyo0:
Add linear regression fitting function for catalog scatter plot
1. fitting will apply to all scatter data if no data points selected, double click on plots will remove the linear regression
2. display the linear function, variance-covariance matrix(cov00/cov01/cov11) and the sum of squares of the residuals as rss at the bottom information area.
3. performance: About 70ms to do the calculation and plot for 1M data source (i5 3GHz, 16 GB memory)
4. add statistics (mean/min/max/rms/std/count) for user selected column in scatter or histogram plot
https://user-images.githubusercontent.com/22226157/140429146-a2bdc83b-4e2d-4213-b335-b3c3d73cb97e.mp4


kswang1029:
The linear fit function works fine. Few improvement to consider:
1) the "statistic" dropdown is not consistent with the one in other widgets such as spectral profile widget. Maybe we can call it "show statistics".
2) the presentation of the fitting results (showing as %.4e) could be improved by displaying them in a new line
3) instead of just showing mean statistic, I suggest we also compute and show other common ones such as min/max/rms/std etc. The presentation would be in a new line too.

2 and 3 might be improved by using a dedicated box but I am not sure where to put it properly. For 2, maybe we can show those info as an annotation on the plot directly at the top-left corner. 


Screen Shot 2021-11-08 at 8 06 08 PM

4) for the "statistic" dropdown in the histogram plot widget, I am not sure if we need that. It would be good to also call it as "show statistics" as a _toggle_ and once it is enabled, we show the statistics of the column being used to generate the histogram.

panchyo0:
@kswang1029  I updated the linear info as an annotation on the top left corner, Also added min/max/rms/std/count for statistics. Please check. 
> for the "statistic" dropdown in the histogram plot widget, I am not sure if we need that. It would be good to also call it as "show statistics" as a toggle and once it is enabled, we show the statistics of the column being used to generate the histogram.

I renamed this dropdown to `Statistic Source`. `can plot one column (DEC) and then show the statistic for another column (flux…).` is a feature request. So we might need to keep the dropdown 🤔

veggiesaurus:
@kswang1029 please re-review

kswang1029:
@panchyo0 it appears that if there are NaN data in a column, linear fit does not work, neither the statistics.

Screen Shot 2021-11-16 at 10 55 36 AM

kswang1029 (approved):
This now works nicely. One last change is to display numbers as %.4e instead of the current %.2e for better UX. This applies to the displayed fitting results and statistics.

kswang1029 (approved):

src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
```diff
@@ -611,6 +703,25 @@ export class CatalogPlotComponent extends React.Component<WidgetProps> {
             </FormGroup>
         );
 
+        const renderStatisticSelect = (
+            <FormGroup inline={true} label="Statistic Source">
```
YuHsuan-Hwang, line 707: Statistic "Source" is hidden in the histogram panel. We could either set a larger default width or try to adjust the column widths
image

src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
```diff
@@ -702,13 +813,53 @@ export class CatalogPlotComponent extends React.Component<WidgetProps> {
...
+                {
+                    type: "line",
+                    layer: "above",
+                    x0: minMaxX.minVal,
+                    y0: fitting.intercept + fitting.slope * minMaxX.minVal,
+                    x1: minMaxX.maxVal,
+                    y1: fitting.intercept + fitting.slope * minMaxX.maxVal,
+                    line: {
+                        color: Colors.GREEN2,
+                        width: 5
```
YuHsuan-Hwang: change to `2.5 * ratio` so that the line width looks the same on a devicePixelRatio = 1 screen
image

src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
```diff
@@ -702,13 +813,53 @@ export class CatalogPlotComponent extends React.Component<WidgetProps> {
...
+                    yref: "paper",
+                    x: 0,
+                    xanchor: "left",
+                    y: 1,
+                    yanchor: "top",
+                    align: "left",
+                    text: widgetStore.fittingResultString,
+                    showarrow: false,
+                    font: {
+                        size: 18,
```
YuHsuan-Hwang: same comment: `9 * ratio`

panchyo0: Thanks 👍

YuHsuan-Hwang (approved):

veggiesaurus (approved):

PR #1700, opened 2021-11-16
Yuhsuan/rest freq UI

YuHsuan-Hwang:
Adjustment of #1306 UI: changing the rest freq dialog to a tab in the image list settings panel
Also updated issue 1306 to the change log

With the new UI, users are able to:
* Open the rest freq tab and set focus on the input via the shortcut in the right-click menu
* View and edit the rest freq of all the images in one interface when image num <= 10
* Use a dropdown to select a image when image num > 10


kswang1029 (changes requested):
I suggest we set default unit to MHz and have more digits such as %.4f. The reason is the big two line catalogs (JPL and CDMS) use MHz as the default. So this would improve UX a bit. 


UPDATE: never mind, after discussions, I see the current way has its pros too.

YuHsuan-Hwang:
@kswang1029 currently the default unit is decided by the header, and is MHz if no rest freq info in the header. also there's no limit to the number of digits

YuHsuan-Hwang:
added minor improvement:
1. also set focus on the input when we use the same shortcut twice
2. select the whole number instead of moving the cursor at the end so that long numbers will not be hidden

kswang1029 (commented):
Ready for code review while user further testing is ongoing

kswang1029 (approved):
tested the transition between the two layouts and it works nicely. The rest frequency is displayed properly with the new fix. Images with non-editable frequency are grey out properly.

kswang1029:
No code review?

veggiesaurus:
> No code review?

:facepalm: oops Sorry, I clicked _merge_ before actually approving the PR :see_no_evil:

PR #1707, opened 2021-11-22
Jason/issue 648 unit conv

### LINKED ISSUE #648: unit conversion of intensity axis (y) in the spectral profile widget (c.f. casaviewer)
sub task of epic #643 

The spectral profile widget should provide the ability of unit conversion in the y axis.
The most common conversion is Jy/beam <-> K.

The complexity here is some units under certain conditions cannot be converted to other units. For example, if an image with K as unit but does not have information of beam size and frequency, we cannot convert K to Jy/beam.

A full set of units and their conversions will be provided later (@kswang1029).

duidae:
closes #648


kswang1029 (changes requested):
Did a quick check of converting Jy/beam to Kelvin. The result is not consistent with the casaviewer. In the example below, the peak should be ~10 K but the converted value in CARTA is ~2.2x10^-3 K. 


Screen Shot 2021-11-22 at 1 11 38 PM

kswang1029 (changes requested):
cdelt1 from the header is usually a negative number so we need to apply abs() to it, otherwise the profile will get flipped up side down.

kswang1029 (changes requested):
When the input image is in Jy/beam, the conversion to Jy/pixel should be forbidden because this conversion involves image deconvolution.

duidae:
> Did a quick check of converting Jy/beam to Kelvin. The result is not consistent with the casaviewer. In the example below, the peak should be ~10 K but the converted value in CARTA is ~2.2x10^-3 K.
> 
> Screen Shot 2021-11-22 at 1 11 38 PM

tried correcting the value of bmaj/bmin in this file,
bmaj = 1.884508960777E-04
bmin = 1.530597441726E-04
and applies the formula 
image

but there is still a 1e-11 scale difference, and also see this formula in NRAO's site (https://science.nrao.edu/facilities/vla/proposing/TBconv)
image

guessing the difference 1e-11 is from 1e3(typo of formula) * 1e4(scale of bmaj) * 1e4(scale bmin), @kswang1029 not sure the logic is correct or not.

kswang1029:
> > Did a quick check of converting Jy/beam to Kelvin. The result is not consistent with the casaviewer. In the example below, the peak should be ~10 K but the converted value in CARTA is ~2.2x10^-3 K.
> > Screen Shot 2021-11-22 at 1 11 38 PM
> 
> tried correcting the value of bmaj/bmin in this file, bmaj = 1.884508960777E-04 bmin = 1.530597441726E-04 and applies the formula image
> 
> but there is still a 1e-11 scale difference, and also see this formula in NRAO's site (https://science.nrao.edu/facilities/vla/proposing/TBconv) image
> 
> guessing the difference 1e-11 is from 1e3(typo of formula) * 1e4(scale of bmaj) * 1e4(scale bmin), @kswang1029 not sure the logic is correct or not.

the formula with 10^3 instead of 10^6 from NRAO is because the input I is mJy instead of Jy. So we see a factor 1000 difference. Could you check if theta_a and theta_b are in "arcsec" unit? BMAJ and BMIN from header is in degree mostly. We will need to convert from degree to arcsec (1 deg = 60 arcmin, 1 arcmin = 60 arcsrc, so 1 deg = 3600 arcsec). Also the  "nu" should be in GHz.

duidae:
> the formula with 10^3 instead of 10^6 from NRAO is because the input I is mJy instead of Jy. So we see a factor 1000 difference. Could you check if theta_a and theta_b are in "arcsec" unit? BMAJ and BMIN from header is in degree mostly. We will need to convert from degree to arcsec (1 deg = 60 arcmin, 1 arcmin = 60 arcsrc, so 1 deg = 3600 arcsec). Also the "nu" should be in GHz.

how can we know the unit of bmaj/bmin? through which header?

kswang1029:
> > the formula with 10^3 instead of 10^6 from NRAO is because the input I is mJy instead of Jy. So we see a factor 1000 difference. Could you check if theta_a and theta_b are in "arcsec" unit? BMAJ and BMIN from header is in degree mostly. We will need to convert from degree to arcsec (1 deg = 60 arcmin, 1 arcmin = 60 arcsrc, so 1 deg = 3600 arcsec). Also the "nu" should be in GHz.
> 
> how can we know the unit of bmaj/bmin? through which header?

BMAJ and BMIN (and BPA) are not included in the FITS standard. Mostly people use degree as the implicit unit when creating the FITS file. We will have to assume it is in degree. Sometimes, the unit is included as a comment in the same line as BMAJ starting with a slash but we cannot trust comments in general.

duidae:
> BMAJ and BMIN (and BPA) are not included in the FITS standard. Mostly people use degree as the implicit unit when creating the FITS file. We will have to assume it is in degree.

ok thanks.

kswang1029:
@duidae if we load a cube with the unit as Jy/beam, when we launch the spectral profiler, the ylabel is displayed as "Value" without the unit in the (). Could you fix this regression please?

kswang1029:
@duidae the conversion from Jy/beam (the input cube) to Jy/arcsec^2 seems not correct. 
Screen Shot 2021-11-25 at 1 46 49 PM

Could you double check if the unit is correct? The calculation of Omega (beam solid angle) requires the theta_a and theta_b in "rad".

kswang1029:
@duidae do you think with the same PR we can support unit conversion when the multiple profile plot mode is enabled? For example, there are two images loaded, one is in Jy/beam and the other is in K. Then when the two images are matched and show region spectral profiles in the multiple profile plot mode, if the user selects K, then the profile with Jy/beam needs to be converted to K before plotting, and vice versa. This feature is a common use case in radio astronomy but casaviewer does not handle this well. If you think this is too much change, then it is fine to leave it for now.

duidae:
Jy/beam <=> Jy/arcsec^2, ylabel fixed, please test. @kswang1029 can you remember to use "request Change" for notification?

src/models/SpectralDefinition.ts
```diff
@@ -124,3 +139,229 @@ export const SPECTRAL_COORDS_SUPPORTED = new Map<string, {type: SpectralType; un
...
+    return forward ? 1 / omega : omega;
+};
+
+const JySrToJyArcsec2 = (forward: boolean = true): number => {
+    const constant = 2.350443 * 1e-11;
+    return forward ? constant : 1 / constant;
+};
+
+const JyPixelToJyArcsec2 = (cdelta1: number, cdelta2: number, forward: boolean = true): number => {
+    const coefficient = (cdelta1 * cdelta2) / (2.350443 * 1e-11);
```
kswang1029, line 281: please check the units of `cdelta1` and `cdelta2`. They should be in rad, not degree.

src/stores/widgets/SpectralProfileWidget/SpectralProfileWidgetStore.ts
```diff
@@ -320,6 +336,37 @@ export class SpectralProfileWidgetStore extends RegionWidgetStore {
         });
     }
 
+    @computed private get intensityConfig(): IntensityConfig {
+        const frame = this.effectiveFrame;
+        if (frame) {
+            let config: IntensityConfig = {nativeIntensityUnit: frame.unit};
+            if (frame.frameInfo?.beamTable?.length > 0) {
+                const beam = frame.frameInfo.beamTable[0];
```
kswang1029: If the input image has per-plane-beam table (ie there are bmaj, bmin, and bpa _per_ channel), do you consider the beam size _as a function of channel_ in the Jy/beam <-> K conversion? For line 344, do you quote the 0th elements in the beamTable as a representative bmaj and bmin for the conversion? Comparing with casaviewer, I see a significant offset per channel.

kswang1029 (approved):
checked all the conversion pairs (with and without required headers) and the per-plane-beam case. All now work fine.

veggiesaurus:
code looks fine to me. Just one question: I see it's relying on `CDELT` header entries. What about situations where the WCS is specified using `PC` or `CD` matrices?

kswang1029:
> code looks fine to me. Just one question: I see it's relying on `CDELT` header entries. What about situations where the WCS is specified using `PC` or `CD` matrices?

we might use the same approach as https://github.com/CARTAvis/carta-frontend/pull/1719 ? 🤔

@YuHsuan-Hwang do we have pixel size info stored elsewhere already?

YuHsuan-Hwang:
> we might use the same approach as #1719 ? 🤔
> 
> @YuHsuan-Hwang do we have pixel size info stored elsewhere already?

Yes, stored in private `regionUnitSize` under `frameStore`. Can be used by public functions `getWcsSizeInArcsec`, `getImageXValueFromArcsec`, and `getImageYValueFromArcsec`. We may change `regionUnitSize` to a more general name if we are using it for other purposes.

duidae:
> > we might use the same approach as #1719 ? thinking
> > @YuHsuan-Hwang do we have pixel size info stored elsewhere already?
> 
> Yes, stored in private `regionUnitSize` under `frameStore`. Can be used by public functions `getWcsSizeInArcsec`, `getImageXValueFromArcsec`, and `getImageYValueFromArcsec`. We may change `regionUnitSize` to a more general name if we are using it for other purposes.

change the source of cdelta1 cdelta2 to regionUnitSize, please check.

veggiesaurus (approved):

veggiesaurus:
looks good to me. @kswang1029 can you just check that it still behaves correctly with this change?

kswang1029 (approved):

kswang1029:
> looks good to me. @kswang1029 can you just check that it still behaves correctly with this change?

@veggiesaurus yes, conversions involving pixel size are still correct.

PR #1711, opened 2021-11-23
Jason/issue 1697

### LINKED ISSUE #1697: Region anchors on Retina display
Using a `devicePixelRatio = 2` screen:

1. the anchors jump to twice farther when changing the region size

https://user-images.githubusercontent.com/43841102/140694913-641e63f6-cbe9-46e3-9950-c4708843a0f6.mov

2. the offset of the rotation anchor is twice farther
`devicePixelRatio = 1`:
image
`devicePixelRatio = 2`:
image

duidae:
closes #1697


veggiesaurus (approved):

YuHsuan-Hwang (approved):

kswang1029 (changes requested):
could you check the case when holding "shift" key and drag the corner control point? On a Retina display, we can see resizing that way has the same issue. 

https://user-images.githubusercontent.com/20819712/143198425-04ba062d-a6f6-415c-ab19-a6959fe0e56f.mov

kswang1029 (approved):

PR #1714, opened 2021-11-23
determine apiAddress correctly when running electron app

### LINKED ISSUE #106: MacOS electron build needs to use preferences.json file
The MacOS electron build currently reverts to `localStorage`, rather than storing the preferences and layouts in `$HOME/.carta/config/preferences.json`.

We need to ensure the frontend is using the appropriate REST endpoint (I think it would be `http://localhost:<port>/api`). This path is currently determined by the contents of the `GET` request to `/config` from the frontend, so perhaps we can just manually edit the `config` file in the built version to include the following:

```javascript
{"apiAddress":"/api"}
```

veggiesaurus:
Closes https://github.com/CARTAvis/carta/issues/106 by adjusting how the api address is determined.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -28,7 +28,7 @@ window["React"] = React; // tslint:disable-line
 window["ReactDOM"] = ReactDOM; // tslint:disable-line
 
 async function fetchConfig() {
-    const baseUrl = window.location.href.replace(window.location.search, "");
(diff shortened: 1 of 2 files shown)


ajm-ska (approved):
Confirmed to work on the MacOS Electron version. Still appears to work fine on the Mix 'N Match page so appears to be OK.

duidae (approved):

YuHsuan-Hwang (approved):

PR #1719, opened 2021-11-26
Yuhsuan/1572 region unit size

### LINKED ISSUE #1572: region size not displayed

YuHsuan-Hwang:
This PR closes #1572. Applied a new approach to calculate the unit size of a region:

* Calculate the geodesic distance of `(CRPIX1, CRPIX2), (CRPIX1 + 1, CRPIX2)` and `(CRPIX1, CRPIX2), (CRPIX1, CRPIX2 + 1)` by AST `astDistance` to obtain the x/y unit size
* The x/y unit size (in arcsec) is rounded to 6 digits
* The x/y unit size are used for wcs width/height (or major/minor-axis for ellipse, or delta x/y for line) respectively. No need to re-calculate the size after the regions are rotated because regions can not be rotated in a non square pixel image.
* The calculated sizes are used for numbers in the region dialog and region info in exported tsv/txt files


kswang1029 (approved):
tested with images that had issues in the size calculation. also tested with images with non-square pixels. All work fine.

veggiesaurus (approved):

Issue #1718, opened 2021-11-26
colorbar tick mark offset


kswang1029:
The tick mark of the colorbar seems to have a small offset in the expected location. In this example, the pixel value varies from 0, 1, 2, ..., 8, 9 and NaN from left to right. If we configure the colorbar range from -0.5 to 9.5, we should expect to see the tick marks at 0, 1, 2, and so on are centered at corresponding color blocks. Apparently there is a small offset in each.  

![Screen Shot 2021-11-26 at 1 07 00 PM](https://user-images.githubusercontent.com/20819712/143530430-0b75e42c-9a1f-4c27-b15c-c8bc8e279f68.png)

kswang1029:
based on further test, this is due to the inconsistency of the color blocks. So I guess it is related to #1717 

<img width="363" alt="Screen Shot 2021-11-26 at 3 25 47 PM" src="https://user-images.githubusercontent.com/20819712/143542796-8fca7df4-7f90-4fbe-ab39-837e84080aa9.png">

PR #1720, opened 2021-11-26
Yuhsuan/1717 1718 colorbar gradient fix

### LINKED ISSUE #1717: horizontal colorbar and vertical colorbar are rendered inconsistently
With a test image (attached) with 11 pixels varying from 0, 1, 2, ... 8, 9, NaN in the x direction, the colorbar at different orientation is rendered differently. 

Screen Shot 2021-11-26 at 12 47 38 PM
Screen Shot 2021-11-26 at 12 47 14 PM
[pixel_shader_test.fits.zip](https://github.com/CARTAvis/carta-frontend/files/7606341/pixel_shader_test.fits.zip)

YuHsuan-Hwang:
Closes #1717 and #1718: fixed the positions of the colorbar gradient

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 * Fixed crash when opening the image view configuration dialog before opening an image [(#1705)](https://github.com/CARTAvis/carta-frontend/issues/1705).
 * Fixed panning and zooming when opening a new image in distance measuring mode [(#1665)](https://github.com/CARTAvis/carta-frontend/issues/1665).
+* Fixed incorrent color gradient of the colorbar ([#1717](https://github.com/CARTAvis/carta-frontend/issues/1717) and [#1718](https://github.com/CARTAvis/carta-frontend/issues/1718)).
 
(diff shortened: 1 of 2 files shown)


veggiesaurus (approved):

kswang1029 (changes requested):
Thanks for the quick fix. Now the color blocks and tick marks are consistent. Additional issue I spotted (not sure if they are related):
- 1px mis-alignment between coordinate box and colorbar box (this might be the same root cause as the next one). This is also observed when the colorbar is vertical.
- the boundary between adjacent color blocks is not sharp (cf the pixel rendering which is sharp between the pixel boundary

Screen Shot 2021-11-26 at 7 30 27 PM

YuHsuan-Hwang:
> * 1px mis-alignment between coordinate box and colorbar box (this might be the same root cause as the next one). This is also observed when the colorbar is vertical.

Fixed: avoided the 1 px border and ticks rendered with 2 px on Retina displays

> * the boundary between adjacent color blocks is not sharp (cf the pixel rendering which is sharp between the pixel boundary

This is expected in our current approach. We generate a list of colors and positions, and render a gradient in a box. Will need special handling for the discrete color scale if we want to remove the blurry boundary.

kswang1029 (approved):

PR #1735, opened 2021-12-02
Jason/issue 1731 2

### LINKED ISSUE #1731: Incorrect positions of region anchors
Region anchors are at incorrect positions after we view other images.

One way to reproduce:
1. disable multi-panel mode
2. open two images
3. create a region in the first image (used file: HD163296_CO_2_1.fits)
image
4. set the active frame to the second image
5. set the active frame to the first image
image
6. the anchors move back to the correct position if we pan the region

duidae:
closes #1731

### DIFF
--- a/src/components/ImageView/RegionView/shared.ts
+++ b/src/components/ImageView/RegionView/shared.ts
@@ -33,14 +33,24 @@ export function canvasToTransformedImagePos(canvasX: number, canvasY: number, fr
     return imagePos;
 }
 
-export function transformedImageToCanvasPos(imageX: number, imageY: number, frame: FrameStore, layerWidth: number, layerHeight: number) {
-    let imagePos = {x: imageX, y: imageY};
-    if (frame.spatialReference) {
-        imagePos = frame.spatialTransform.transformCoordinate(imagePos, true);
-        return imageToCanvasPos(imagePos.x, imagePos.y, frame.spatialReference.requiredFrameViewForRegionRender, layerWidth, layerHeight, frame.spatialTransform);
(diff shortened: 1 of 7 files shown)


src/components/ImageView/RegionView/shared.ts
```diff
@@ -33,14 +33,24 @@ export function canvasToTransformedImagePos(canvasX: number, canvasY: number, fr
...
+            const frameView = origin.x === 0 && origin.y === 0 && zoom === 1 ? frame.unitFrameView : frame.requiredFrameViewForRegionRender;
+            canvasPos = imageToCanvasPos(imagePos.x, imagePos.y, frameView, layerWidth, layerHeight);
+        }
+        if (canvasPos) {
+            return {x: (canvasPos.x - origin.x) / zoom, y: (canvasPos.y - origin.y) / zoom};
+        }
     }
-
-    return imageToCanvasPos(imagePos.x, imagePos.y, frame.requiredFrameViewForRegionRender, layerWidth, layerHeight, frame.spatialTransform);
+    return undefined;
```
veggiesaurus: do we always check that the result of this isn't `undefined` before use elsewhere in the codebase?

veggiesaurus: previously `imageToCanvasPos` always returned an `{x: number, y: number}` object. I just want to make sure we're not introducing any hard-crashes when the result of that function is `undefined`

duidae: fixed.

acdo2002 (approved):
I tested single/multi plots and set single/many region(s), then switch different images and check whether the regions rendered in the proper position (and check the selected region area).
So far I did not find any weird behavior, beside the confirmed issue #1737 and #1740

veggiesaurus (approved):

PR #1745, opened 2021-12-08
Regression fix: Dynamic spatial matching

### LINKED ISSUE #1737: regression: image spatial matching is not dynamically picking up image viewer center to perform matching
It appears that matching images spatially is not dynamically adopting the image view center to perform matching calculations. The matching process now is static and adopts image center(?) only once. This can be seen clearly only with full sky images. For smaller field images the difference is too small to be observed.

veggiesaurus:
Closes #1737 by reverting a change made in #1664.

@duidae, I'm not sure whether the changes were required somewhere else? [This commit](https://github.com/CARTAvis/carta-frontend/commit/2807546c636bebd59b5015796da8b69bcfa060bf) is where the changes occurred. 

The code was changed to always use the image center, rather than the `FrameView` center, as the reference point for spatial matching. I've reverted this change, so it now updates dynamically. But we will need to be sure that there are no other regressions with this change. @duidae can you remember why this change was originally neccessary?

### DIFF
--- a/src/components/ImageView/Overlay/OverlayComponent.tsx
+++ b/src/components/ImageView/Overlay/OverlayComponent.tsx
@@ -39,8 +39,10 @@ export class OverlayComponent extends React.Component<OverlayComponentProps> {
     }
 
     updateImageDimensions() {
-        this.canvas.width = this.props.overlaySettings.viewWidth * devicePixelRatio * AppStore.Instance.imageRatio;
-        this.canvas.height = this.props.overlaySettings.viewHeight * devicePixelRatio * AppStore.Instance.imageRatio;
+        if (this.canvas) {
+            this.canvas.width = this.props.overlaySettings.viewWidth * devicePixelRatio * AppStore.Instance.imageRatio;
+            this.canvas.height = this.props.overlaySettings.viewHeight * devicePixelRatio * AppStore.Instance.imageRatio;
+        }
     }
 
     renderCanvas = () => {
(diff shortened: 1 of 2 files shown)


veggiesaurus:
I've removed the other change that I accidentally added to the PR :see_no_evil: (bug fix for crash when switching images too quickly). Will make a separate PR for it

kswang1029:
@acdo2002 please help to test too. Thanks.

duidae:
@veggiesaurus originally it is to support massive regions of matched images by removing referencing the observable(spatialReference.center), since it will cause re-render inside the tracking function. Reverting it back is ok and the side effect will be that the region views of `matched images` becomes freezed like before when importing massive regions.

kswang1029 (approved):
the revert works and now spatial matching works as before. The PR passes all rendering tests and matching tests we have so far. The performance impact on matched image is observed. At this stage for v3b2 release, I would suggest we move forward and see if that part can be improved in the next release.

acdo2002 (approved):
I have tested "`M17_SWex.fits", "HH211_IQU.hdf5", "HD163296_13CO_2-1.fits", and "HD163296_C18O_2-1.fits"` (as the first reference image) matching with all sky images (as second or third image) `"ovro_mollweide_lwa_sky_map_36.528MHz.fits", "Halpha_map.fits", "mx_mjd55060_55800_ssc_a.fits"`.

The spatially matching region now is in the viewable location in the frame (the current dev branch, there is a gap, so I have to move the viewable frame to find the matched region) and when I zoom in/out in the second/third image, the first image dynamically zoom in/out.

duidae (approved):

PR #1747, opened 2021-12-08
ensure canvas ref is defined before resizing or assigning to AST

### LINKED ISSUE #1746: Crash when switching images just after spatial matching is enabled
Crash can be reproduced by using the following code snippet (should occur for any two images with a valid spatial transform):

```javascript
const frame1 = await app.openFile("fits/wcs_dev_images/Gaussian_SE2.fits");
const frame2 = await app.appendFile("fits/wcs_dev_images/Gaussian2.fits");
app.setSpatialMatchingEnabled(frame2, true);
frame2.setZoom(16);
frame2.setCenter(598, 382);
// uncommenting this delay seems to avoid the crash
// await app.delay(100);
app.setActiveFrame(frame1);
```

Results in a hard crash because the `canvas` ref ins undefined in `OverlayComponent`.

veggiesaurus:
Closes #1746 by checking that the `canvas` ref is defined.

### DIFF
--- a/src/components/ImageView/Overlay/OverlayComponent.tsx
+++ b/src/components/ImageView/Overlay/OverlayComponent.tsx
@@ -39,8 +39,10 @@ export class OverlayComponent extends React.Component<OverlayComponentProps> {
     }
 
     updateImageDimensions() {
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):

YuHsuan-Hwang (approved):
Although I could not reproduce the issue, the changes look fine to me

Issue #1623, opened 2021-08-12
correct "hard reload" shortcut suggestions


veggiesaurus:
Chrome + MacOS should be Cmd + Shift + R.

Need to double-check other combinations too.

PR #1757, opened 2022-01-10
fix hard reload shortcuts

kswang1029:
This addresses #1623.

Screen Shot 2022-01-10 at 12 24 43

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -134,9 +134,10 @@
                     this site in your browser using the shortcut below:
                 </p>
                 <pre class="static-pre">
-                MacOS: Cmd + R
-                Linux: Ctrl + Shift + R
-                Windows: Ctrl + F5
+                macOS : option + command + R (Safari)
+                macOS : shift + command + R (Chrome, Firefox)
+                Linux : shift + control + R
+                Windows : Ctrl + Fn + F5
             </pre>
             </div>
         </div>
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -134,9 +134,10 @@
                     this site in your browser using the shortcut below:
                 </p>
                 <pre class="static-pre">
-                MacOS: Cmd + R
-                Linux: Ctrl + Shift + R
-                Windows: Ctrl + F5
+                macOS : option + command + R (Safari)
+                macOS : shift + command + R (Chrome, Firefox)
+                Linux : shift + control + R
+                Windows : Ctrl + Fn + F5
```
veggiesaurus: Shouldn't this be control as well, instead of Ctrl?

Also, the function modifier is just for laptops in order to actually use function keys, so it's implied, and also implementation-dependent (sometimes Fn modifier is used to disable the function keys)

veggiesaurus (changes requested):

public/index.html
kswang1029: @veggiesaurus for windows, it is changed to control + F5. I actually don't have a windows system here... could you confirm "control+F5" is _hard_ reload instead of the regular reload?

veggiesaurus (approved):

PR #1760, opened 2022-01-20
Remove moving stage from onClick event

### LINKED ISSUE #1751: region re-rendering with click to pan method
Reported by a user using v3b2 release.

To reproduce:
1) load an image and create few region
2) hold command (Mac) or control (linux) and click to pan the image (or use preferences dialog to enable click-to-pan as default pan method)  
3) we should see regions are not rendered at their correct locations.

https://user-images.githubusercontent.com/20819712/146339612-8c9c217f-9fb3-4b2c-bdde-b253ad0e64a6.mov

duidae:
closes #1751


veggiesaurus (approved):

acdo2002 (approved):
I have tested with Mac and Ubuntu. 
After clicking to pan the image, the regions now show in the proper location.
In addition, I tested with multi-images on multi-panels, even switch the images, the behavior works fine so far.

Issue #1749, opened 2021-12-09
initial spectral range of the moment generator


YuHsuan-Hwang:
The initial spectral range of the moment generator should be the actual range of the spectral axis instead of (0, 0).

tested images: `HD163296_CO_2_1.fits`, `GV_Tau_sci.spw0.cube.I.manual.image.pbcor.10ch.fits`

![image](https://user-images.githubusercontent.com/43841102/145347962-a7d62483-4b6c-4787-bd9a-e5ae79d78087.png)

PR #1763, opened 2022-01-24
restore moment generator initial range update

kswang1029:
This is to address https://github.com/CARTAvis/carta-frontend/issues/1749 by restoring the approach that v2.0 adopts. ~~This fix also fixes the issue that if there are two images loaded and if we change the intensity unit of the current one then switch to the other and switch back, we see the intensity unit resets to the default (undesired).~~

### DIFF
--- a/src/stores/widgets/SpectralProfileWidget/SpectralProfileWidgetStore.ts
+++ b/src/stores/widgets/SpectralProfileWidget/SpectralProfileWidgetStore.ts
@@ -1,4 +1,4 @@
-import {action, autorun, computed, observable, makeObservable, override} from "mobx";
+import {action, autorun, computed, observable, makeObservable, override, reaction} from "mobx";
 import {IOptionProps, NumberRange} from "@blueprintjs/core";
 import {CARTA} from "carta-protobuf";
 import {PlotType, LineSettings, VERTICAL_RANGE_PADDING, SmoothingType} from "components/Shared";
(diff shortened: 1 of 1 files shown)


src/stores/widgets/SpectralProfileWidget/SpectralProfileWidgetStore.ts
```diff
@@ -316,16 +316,12 @@ export class SpectralProfileWidgetStore extends RegionWidgetStore {
         this.intensityConversion = undefined;
         this.setIntensityUnit(this.effectiveFrame?.unit);
 
-        reaction(
-            () => this.effectiveFrame,
-            frame => {
-                if (frame) {
-                    this.momentRegionId = RegionId.ACTIVE;
-                    this.setIntensityUnit(frame.unit);
```
YuHsuan-Hwang, line 324: is there any effect after removing this line?

src/stores/widgets/SpectralProfileWidget/SpectralProfileWidgetStore.ts
```diff
@@ -316,16 +316,12 @@ export class SpectralProfileWidgetStore extends RegionWidgetStore {
...
-                    this.updateRanges();
-                }
+        autorun(() => {
+            if (this.effectiveFrame) {
+                this.updateRanges();
+                this.momentRegionId = RegionId.ACTIVE;
```
YuHsuan-Hwang: may change to `this.selectMomentRegion(RegionId.ACTIVE)` to remove mobx warning
image

YuHsuan-Hwang (commented):
tested that issue is fixed 👍

src/stores/widgets/SpectralProfileWidget/SpectralProfileWidgetStore.ts
TienHao: I found that the intensity unit options in conversion and the unit of y axis in spectral profiler won't be updated when switching images. Should add the removing line back.

src/stores/widgets/SpectralProfileWidget/SpectralProfileWidgetStore.ts
kswang1029: revised

src/stores/widgets/SpectralProfileWidget/SpectralProfileWidgetStore.ts
kswang1029: revised

YuHsuan-Hwang (approved):

TienHao (approved):

PR #1773, opened 2022-02-07
prevent crash when plot data is empty

### LINKED ISSUE #1772: render config widget crashes when the image pixels are all NaN
**Describe the bug**
render config widget crashes when the image pixels are all NaN

**To Reproduce**
1. load an image with all pixels as NaN
2. we should see the render config widget is empty and the console shows
Screen Shot 2022-02-07 at 09 22 22

**Expected behavior**
The render config widget should not crash.

**Screenshots or videos**
N/A

 - OS: macOS Monterey 12.2
 - Browser: macOS v3b2 electron app
 - Version: macOS v3b2 electron app

**Additional context**
N/A

veggiesaurus:
Closes #1772 by adding a simple check to see if `data` is empty/undefined.


kswang1029 (approved):
confirmed the fix. 👍

PR #1775, opened 2022-02-07
Angus/1774 million channels

### LINKED ISSUE #1774: frontend hangs when loading a cube with 1000000 (one million) channels
**Describe the bug**
(credit to Felix Stoehr who found this issue)The GUI and the entire browser become unresponsive when loading a cube with 1 million channels. The test image is attached (10x10x1e6x1).

**To Reproduce**
1. load the attached test image
2. then the browser (tested with Firefox and google chrome) will hang with CPU usage 100%
3. a few minutes later the image loads....

**Expected behavior**
The expected behavior is when loading such cubes with that many channels, the image should load quickly just like the cube has 1000 channels only.

**Screenshots or videos**

 - OS: tested with macOS Monterey
 - Browser: chrome, firefox
 - Version: ..

**Additional context**
A test image (10x10x1000000x1; 400MB) can be downloaded at https://drive.google.com/file/d/1bhva81mCJocf2Y13lNxdUvRUmfsGl_o0/view?usp=sharing

veggiesaurus:
Closes #1774 by adding a new function to the AST wrapper `transformSpectralPointArray`, which converts an entire array between spectral configs, rather than calling `transformSpectralPoint` a million times.

@kswang1029 can you please test to ensure the following:
1. The loading time is now almost instant.
2. The spectral values in normal files are still the same.

### DIFF
--- a/src/stores/FrameStore.ts
+++ b/src/stores/FrameStore.ts
@@ -1114,7 +1114,7 @@ export class FrameStore {
 
     private convertSpectral = (values: Array<number>): Array<number> => {
         const N = values?.length;
-        if (!N) {
+        if (!N || !this.spectralFrame) {
             return null;
         }
 
--- a/src/stores/FrameStore.ts
+++ b/src/stores/FrameStore.ts
@@ -1113,7 +1113,17 @@ export class FrameStore {
     };
 
     private convertSpectral = (values: Array<number>): Array<number> => {
-        return values && values.length > 0 ? values.map(value => this.astSpectralTransform(this.spectralType, this.spectralUnit, this.spectralSystem, value)) : null;
+        const N = values?.length;
+        if (!N) {
+            return null;
+        }
+
+        console.log(`Converting ${N} spectral values`);
+        const spectralArray = new Array<number>(N);
+        for (let i = 0; i < N; i++) {
+            spectralArray[i] = this.astSpectralTransform(this.spectralType, this.spectralUnit, this.spectralSystem, values[i]);
+        }
+        return spectralArray;
     };
 
     private astSpectralTransform = (type: SpectralType, unit: SpectralUnit, system: SpectralSystem, value: number): number => {
--- a/src/stores/FrameStore.ts
+++ b/src/stores/FrameStore.ts
@@ -1118,12 +1118,8 @@ export class FrameStore {
             return null;
         }
 
-        console.log(`Converting ${N} spectral values`);
-        const spectralArray = new Array<number>(N);
-        for (let i = 0; i < N; i++) {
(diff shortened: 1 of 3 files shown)


kswang1029 (commented):
@veggiesaurus thanks for the quick fix! Now a 1M-ch cube can be loaded quickly as usual cubes. Spectral conversion is identical to the previous implementation. 👍 Few performance issues spotted:
1. the spectral profile plot is very not responsive to zoom action. This is expected with chartjs I think. Or perhaps this is due to the frontend profile decimation calculations? 🤔
2. the spectral profile settings dialog would take ~7s to initialize on my laptop
3. spectral conversion is quite slow. It took ~10s for a conversion such as from frequency to velocity. Do you think we can apply the same fix to this part too?
4. cube histogram for a cube as 10x10x1e6x1 (400MB) would take 30 mins on my laptop

I guess 1, 2, and 4 could be addressed later.

arusstaylor:
Wow! A million channel cube!!!

veggiesaurus:
> 3\. spectral conversion is quite slow. It took ~10s for a conversion such as from frequency to velocity. Do you think we can apply the same fix to this part too?

It only takes about 3 seconds for me :thinking: But either way, I don't think this is caused by the AST conversion, but by other issues in the code (`getDataPointsAndProperties`). So I suggest we create separate issues for these. I suspect (3) is actually related to (1).

(4) should be filed as a backend issue I guess.

src/stores/FrameStore.ts
```diff
@@ -1113,7 +1113,13 @@ export class FrameStore {
     };
 
     private convertSpectral = (values: Array<number>): Array<number> => {
-        return values && values.length > 0 ? values.map(value => this.astSpectralTransform(this.spectralType, this.spectralUnit, this.spectralSystem, value)) : null;
```
YuHsuan-Hwang, line 1116: `this.astSpectralTransform` checks if `this.spectralFrame` and `value` are valid. Maybe we could add the checks back in `convertSpectral` (at least for `this.spectralFrame`)

veggiesaurus: I've added a check to `this.spectralFrame`. If `values` is undefined, then `values?.length` will also be undefined, so this already checks that `values` is valid.

YuHsuan-Hwang (approved):

kswang1029 (approved):
Retested performance of image loading and spectral value accuracy and they are good. 👍 

Performance about spectral profile rendering of million channels at some point can be revisited. With a test cube of 10x10x1M the backend can deliver point region spectrum quickly. The frontend is struggling on processing and rendering I think.

PR #1776, opened 2022-02-07
Mylin/1762 add cursor info to histogram widget

### LINKED ISSUE #1762: missing cursor info in the histogram widget
Like the spatial or spectral profiler, the histogram widget should also show cursor info (pointed x and y values) at the bottom-left corner. The histogram in the render config widget does show the cursor info though.

acdo2002:
Fixed #1762 
I follow the spatial and spectral widget method to add the cursor information to the left bottom of the histogram widget.

### DIFF
--- a/src/components/Histogram/HistogramComponent.tsx
+++ b/src/components/Histogram/HistogramComponent.tsx
@@ -194,7 +194,7 @@ export class HistogramComponent extends React.Component<WidgetProps> {
             if (this.widgetStore.isMouseMoveIntoLinePlots) {
                 let numberString;
                 // Switch between standard and scientific notation
-                if (this.widgetStore.cursorX < 1e-2) {
+                if (this.widgetStore.cursorX < 1e-5) {
                     numberString = toExponential(this.widgetStore.cursorX, 5);
                 } else {
                     numberString = toFixed(this.widgetStore.cursorX, 5);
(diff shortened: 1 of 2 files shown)


kswang1029 (changes requested):
@acdo2002 thanks for implementing the cursor info for the histogram widget. 👍 
We will need to show "Cursor: ..." only. We don't need to support "Data: ..." as it does not make any sense to show the cursor pointed pixel info there. The "Cursor: ..." part needs to show both x and y values of the histogram, such as (1.23 Jy/beam, 25 counts).

acdo2002:
@kswang1029 I follow your suggestion, adding x(value) and y(count) values into the bottom left of the histogram, please check it.

kswang1029 (changes requested):
now it works nicely and thanks for taking care of singular/plural terms. 
Just one issue I spotted: when the cursor is outside the x-range of the data, the count stays as the first or the last count. I guess this is due to binary search? In such case I think it is better to show "NaN" as counts. 


https://user-images.githubusercontent.com/20819712/152951878-4a7a67d6-21f5-4bed-96c8-46ee54a23e18.mov

kswang1029 (approved):
feature-wise this works nicely 👍

YuHsuan-Hwang (approved):

PR #1783, opened 2022-02-15
Upgrade chart-js to 3.7.1

### LINKED ISSUE #1752: stokes widget does not handle Nan channels correctly in line or step style plotting
In latest beta release 3.00 b2  the Stokes widget is not doing the right thing for Nan channels It is showing a value which looks it is the close value of the actually data. This is not the desired behaviour.   This only occurs for line or step style plotting.  It works correctly for scatter plots (i.e.points).  

Screenshot 2021-12-16 at 10 28 22

acdo2002:
Fixed #1752 
I found the problem after migrate the chart-js from 2.9 to 3.5.1: the NaN in Stoke will be plotted in the widget.

Then I used the newer version (3.7.1) and found the behavior back to 2.9 version, the NaN will not be plotted in the Stoke widget.

So far, I have tested that the chart-js related plots (Histogram, Spectral, Spatial in X, Y) of 3.7.1 working fine and NaN will not be plotted in Stokes, Spectral and Spatial (X,Y).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,7 @@
                 "@types/tinycolor2": "^1.4.2",
                 "ajv": "^8.6.1",
                 "axios": "^0.21.1",
-                "chart.js": "^3.5.1",
+                "chart.js": "^3.7.1",
                 "classnames": "^2.3.1",
                 "fuzzy-search": "^3.2.1",
                 "glob-to-regexp": "^0.4.1",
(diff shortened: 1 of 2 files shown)


veggiesaurus (approved):
@acdo2002 in future please make sure to run `npm install` after making changes to `package.json`, so that you also commit the updated `package-lock.json` file

kswang1029 (changes requested):
I spotted a minor coloring issue in the stokes widget. Please refer to the attached screenshot. The first two (second half of channel 0 and first half of channel 1) channels are rendering in black color which is not consistent with the other channels. 
Screen Shot 2022-02-17 at 11 32 28

kswang1029:
> I spotted a minor coloring issue in the stokes widget. Please refer to the attached screenshot. The first two (second half of channel 0 and first half of channel 1) channels are rendering in black color which is not consistent with the other channels.
> Screen Shot 2022-02-17 at 11 32 28

this issue will be handled as a separate issue.

Issue #1501, opened 2021-06-11
Default file names too long for saving


marcinglowacki:
When trying to save images from the Z Profile window with multiple spectra displayed, the default file name can become too long (>255 characters) to be accepted for download by the browser, due to combination of the opened image file names. Issue already reproduced by Angus. 
![Screenshot from 2021-06-11 09-28-15](https://user-images.githubusercontent.com/29977213/121653436-fb9cd380-ca9c-11eb-96f2-f97e64cec512.png)

PR #1789, opened 2022-02-21
ksw/limit filename length before timestamp up to 200

kswang1029:
This is to address https://github.com/CARTAvis/carta-frontend/issues/1501 by trimming the filename string before the timestamp up to 200 char. This applies to the spectral profiler when saving the profile plot as a png and image viewer when saving the image as a png.

### DIFF
--- a/src/components/Shared/LinePlot/LinePlotComponent.tsx
+++ b/src/components/Shared/LinePlot/LinePlotComponent.tsx
@@ -629,7 +629,8 @@ export class LinePlotComponent extends React.Component<LinePlotComponentProps> {
 
         composedCanvas.toBlob(blob => {
             const link = document.createElement("a") as HTMLAnchorElement;
-            link.download = `${imageName}-${plotName.replace(" ", "-")}-${getTimestamp()}.png`;
+            // Trim filename before timestamp to 200 characters to prevent browser errors
+            link.download = `${imageName}-${plotName.replace(" ", "-")}`.substring(0, 200) + `-${getTimestamp()}.png`;
             link.href = URL.createObjectURL(blob);
             link.dispatchEvent(new MouseEvent("click"));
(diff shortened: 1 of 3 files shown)


kswang1029:
@acdo2002 please help on verification. thanks.

veggiesaurus:
Shouldn't we trim to 196 characters, because of the trailing ".png"?

kswang1029:
> Shouldn't we trim to 196 characters, because of the trailing ".png"?

Trim is applied to the pre-string including filenames and plot_modes. Then we append a timestamp and .png string.

YuHsuan-Hwang:
Do we have the same issue when downloading tsv and txt files?

kswang1029:
> Do we have the same issue when downloading tsv and txt files?

For tsv export it is one tsv file per profile. So we don’t have the problem having a too long concatenated image name.

acdo2002 (approved):
I have tested it with long name file 
`(e.g. M17_SWexAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABBBB.fits)`
and exported the .png. 
The exported name only catch the first 200 characters (then + timeStamp.png).

This behavior has confirmed with 
(1) export image in spectral profiler  
(2) export image in spatial profiler
(3) export image in Stoke widget, scatter plot
(4) main canvas: image viewer

YuHsuan-Hwang (approved):

PR #1799, opened 2022-03-03
Thchang/fix image viewer tab

### LINKED ISSUE #1686: reset tab title of image viewer when we close all images
Screen Shot 2021-10-29 at 12 03 44 PM

TienHao:
closes #1686. Clear image viewer title after all images closed.


kswang1029 (changes requested):
Please try the following case to see the issue:
1) load two images, say A.fits and B.fits in order.
2) we should see the title is B.fits
3) close B.fits either via the file menu or using the context menu (right click) of the image list widget
4) we should see the title gets updated to A.fits as it is the active image now but we still see B.fits as the title.

TienHao:
> Please try the following case to see the issue:
> 
> 1. load two images, say A.fits and B.fits in order.
> 2. we should see the title is B.fits
> 3. close B.fits either via the file menu or using the context menu (right click) of the image list widget
> 4. we should see the title gets updated to A.fits as it is the active image now but we still see B.fits as the title.

fixed

YuHsuan-Hwang (commented):
https://github.com/CARTAvis/carta-frontend/blob/dev/src/stores/AppStore.ts#L1439
Could we remove these lines in `autorun` if title update is handled in `changeActiveFrame` and `closeFile`?

TienHao:
> https://github.com/CARTAvis/carta-frontend/blob/dev/src/stores/AppStore.ts#L1439 Could we remove these lines in `autorun` if title update is handled in `changeActiveFrame` and `closeFile`?

I think it's OK to remove these lines.

kswang1029 (approved):
Tested. Now the tab title of the image viewer shows correct image name. 👍

YuHsuan-Hwang (approved):

PR #1800, opened 2022-03-03
Thchang/1676 fix extra multi panel control buttons

### LINKED ISSUE #1676: show multiple-panel switch of the image viewer only when needed
If we have other tab sharing the same box with the image viewer, that tab will also have the multiple-panel switch. If we switch tabs back and forth, we see more and more switches. 

https://user-images.githubusercontent.com/20819712/138095636-17546baf-d8f1-44ea-b275-5784a7f549b9.mov

also if it is the animator docked as a tab, we still see a cog. Animator widget does not have a setting dialog (yet)

TienHao:
closes #1676. Removed extra multi-panel control buttons and refactored code of docked widget panel control buttons.


kswang1029 (approved):
this works properly now. 👍

veggiesaurus (approved):

YuHsuan-Hwang (approved):

PR #1803, opened 2022-03-07
generated field on fileOpen

veggiesaurus:
adds a `generated` field to file open telemetry entries when reporting on moment map or PV images


veggiesaurus:
(will need to update the CARTA telemetry metrics at some stage to make use of the new field)

Issue #1397, opened 2021-04-13
GUI for 2D image fitting 


kswang1029:
the relevant backend issue is https://github.com/CARTAvis/carta-backend/issues/150

Requirement document:
https://docs.google.com/document/d/1E9kL7FOhVtkDDit8VIYle61tU5PzSYXjnLpxZBCHliY/edit?usp=sharing

Implementation plan:
- [x] Add a new dialog and UI for basic single/multiple gaussian fitting and text outputs
- [x] Add image/world coordinate conversion for the outputs
- [ ] Convert output FWHM size to arcmin or degree 
- [ ] Add image/world coordinate conversion for the inputs
- [x] Add buttons for fixing initial values
- [x] use current FOV as a subimage for fitting
---- target v3b3 ----
- [x] Add progress bar dialog
- [ ] UI for set a region as the data for fitting
- [x] UI for appending model/residual images
- [x] Export results as txt files
- [ ] support a background level in the fitting
- [ ] create ellipse regions from the results
- [ ] support different solvers in the fitting
- [X] ~guess initial value~ moved to new issue: #2095

PR #1805, opened 2022-03-15
Yuhsuan/1397 2d fitting

YuHsuan-Hwang:
This PR is for the first item of #1397: add a new dialog and UI for basic single/multiple gaussian fitting and text outputs

Features:
* Input initial values of multiple Gaussian components and request fitting the current channel of the image
* Disable fit button when parameters are not all set or when a fitting task is in progress
* show a simple dialog and block the UI when a fitting task is in progress
* Resize the result text column and the component slider with the dialog size

Other changes:
* remove unused variable `restFreqDialogVisible`

Corresponding backend PR: https://github.com/CARTAvis/carta-backend/pull/1055


kswang1029 (commented):
Few suggestions to consider for this PR. They can be part of the next step of implementation if you think proper.

1. I suggest we use an icon like the following, which _may_ be better as a representation of "2D image fitting". Open to other alternatives. :)
Screen Shot 2022-03-16 at 13 28 17

2. we will need to also show and define values in world coordinates. If world coordinate and angular size are adopted, we will then convert them to image coordinate first before setting the protobuf message (assuming the backend image fitter can only accept image coordinates)

3. need a help button too. 
Screen Shot 2022-03-16 at 14 34 59

4. disable label cursor selection. This happens when we try to increase the number of components _quickly_.
Screen Shot 2022-03-16 at 14 43 35

src/stores/AppStore.ts
```diff
@@ -1102,6 +1104,27 @@ export class AppStore {
...
 
+    requestFitting = async (message: CARTA.IFittingRequest) => {
+        if (!message) {
+            return;
+        }
+
+        try {
+            const ack = await this.backendService.requestFitting(message);
+            if (ack.success) {
+                this.getFrame(message.fileId).setFittingResult(ack.results);
```
veggiesaurus: rather use `const frame = this.getFrame(message.fileId)` and then reuse

src/stores/ImageFittingStore.ts
```diff
@@ -0,0 +1,157 @@
...
+    constructor() {
+        makeObservable(this);
+        this.center = {x: null, y: null};
+        this.amplitude = null;
+        this.fwhm = {x: null, y: null};
+        this.pa = null;
+    }
+
+    @computed get validParams() {
+        return Number.isFinite(this.center.x) && Number.isFinite(this.center.y) && Number.isFinite(this.amplitude) && Number.isFinite(this.fwhm.x) && Number.isFinite(this.fwhm.y) && Number.isFinite(this.pa);
```
veggiesaurus: can't you just use `isFinite` without `Number.`? Perhaps just in case include `this.center && this.fwhm &&` to ensure the points exist

src/stores/ImageFittingStore.ts
```diff
@@ -0,0 +1,157 @@
...
+        this.fwhm.y = val;
+    };
+
+    @action setPa = (val: number) => {
+        this.pa = val;
+    };
+
+    constructor() {
+        makeObservable(this);
+        this.center = {x: null, y: null};
```
veggiesaurus: shouldn't these be `NaN` at first?

veggiesaurus (changes requested):
couple of very minor changes. This looks very good otherwise

veggiesaurus:
> Few suggestions to consider for this PR. They can be part of the next step of implementation if you think proper.
> 
> 1. I suggest we use an icon like the following, which _may_ be better as a representation of "2D image fitting". Open to other alternatives. :)
>    Screen Shot 2022-03-16 at 13 28 17
> 2. we will need to also show and define values in world coordinates. If world coordinate and angular size are adopted, we will then convert them to image coordinate first before setting the protobuf message (assuming the backend image fitter can only accept image coordinates)
> 3. need a help button too.
>    Screen Shot 2022-03-16 at 14 34 59
> 4. disable label cursor selection. This happens when we try to increase the number of components _quickly_.
>    Screen Shot 2022-03-16 at 14 43 35

I suggest we do (3) and (4) in this PR, as they are minor changes. And then perhaps @TienHao or @Michaelazzz can make a custom icon for this. (2) seems like something that should be done in a separate PR.

src/stores/ImageFittingStore.ts
YuHsuan-Hwang: I used `Number.isFinite` because `isFinite(null)` returns true. I think we can change the initial value from `null` to `NaN` as you suggested, and change here to `isFinite`

src/stores/AppStore.ts
YuHsuan-Hwang: fixed

src/stores/ImageFittingStore.ts
YuHsuan-Hwang: changed to `isFinite` and added optional chaining to check if the points exist

src/stores/ImageFittingStore.ts
YuHsuan-Hwang: fixed

YuHsuan-Hwang:
> Few suggestions to consider for this PR. They can be part of the next step of implementation if you think proper.
> 
> 1. I suggest we use an icon like the following, which _may_ be better as a representation of "2D image fitting". Open to other alternatives. :)
> 2. we will need to also show and define values in world coordinates. If world coordinate and angular size are adopted, we will then convert them to image coordinate first before setting the protobuf message (assuming the backend image fitter can only accept image coordinates)
> 3. need a help button too.
> 4. disable label cursor selection. This happens when we try to increase the number of components _quickly_.

I will implement 2. in the next PR.
3. and 4. are fixed.
We may need more ideas for 1. ...

changing channels during fitting causes problem with the new approach in backend so I also added a blocking dialog.

kswang1029 (approved):
This is good 👍 Just one minor issue when setting an initial amplitude if it is negative. For example, try to type in -0.01. This can be addressed any time. Good to get this code-reviewed and merged first.

kswang1029:
Just wonder if this https://www.svgrepo.com/svg/24161/target-concentric-circles-symbol is a sensible icon for 2d image fitting? 🤔️

YuHsuan-Hwang:
changed icons of image fitting and z profile fitting based on @kswang1029's suggestion.

veggiesaurus (approved):

PR #1809, opened 2022-03-22
Ajm/remove circleci format test

ajm-ska:


ajm-ska:
@veggiesaurus Actually, I forgot that the _frontend_ format check was not actually using CircleCI! It was actually running through Jenkins. So the title here is misleading!

Anyway, the Jenkins format check is now removed.  It seems to work OK without the "n 12 / n 14 / n16" lines so maybe they are no longer necessary. 

For the Github actions, I think 

> on:
>   push:
>     branches:
>       - '**'

will be enough to run on every commit and likely pull requests too (Let's see).

.github/workflows/continuous_integration.yml
```diff
@@ -1,10 +1,9 @@
 name: Continuous Integration
 
 on:
-  pull_request:
   push:
     branches:
-      - dev
+      - '**'
```
veggiesaurus: same comment as the other PR. You can just use `on: push` here, to trigger on all branches. See [here](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-events-to-trigger-workflows) for details.

veggiesaurus (changes requested):

ajm-ska:
I thought I would also shorten the name so it fits on the list better if that is OK with you?
I changed `Continuous Integration` to `CI`.
Screen Shot 2022-03-23 at 8 43 09 PM

ajm-ska:
Yes. Looks better now
Screen Shot 2022-03-23 at 8 47 39 PM

PR #1813, opened 2022-03-29
Update RegionListComponent.tsx

chenblackeagler:
(Draft version for recruit assignment) Adopt a new-imported package "react-scrollable-feed" and try to make the region list auto-scrollable.


kswang1029:
need `npm install react-scrollable-feed` first before building

kswang1029:
@chenblackeagler with a local build here, I see the region list widget is corrupted (table is not rendered properly). Do we need a special build instruction to get it work or this is still in progress?

PR #1814, opened 2022-03-30
Thchang/spectral line query reset button 1741

### LINKED ISSUE #1741: reset button does not work in the spectral line query widget
to reproduce:
1) enable the spectral line query widget
2) set frequency range 349370 MHz to 349500 MHz and get the line table back (there should be 419 lines)
3) in the Species column, set a filter as CH3CNv=0 (there should be 4 lines)
4) press the reset button and we should see the table is reset to the no filter state with 419 lines but this does not happen

This might be a regression after updating the blueprint table lib.

TienHao:
closes #1741. Fixed error by `cloneDeep` function of lodash when resetting filtered data of spectral line query.


veggiesaurus (approved):

kswang1029 (approved):
The issue was found when I was preparing relevant e2e test. I confirm the fix works by manual test and by the e2e test. 👍

YuHsuan-Hwang (approved):

PR #1824, opened 2022-04-11
Load complex images

### LINKED ISSUE #1492: support complex image visualization as raster
related to the backend issue: https://github.com/CARTAvis/carta-backend/issues/520

frontend should provide options to visualize a complex-valued image as
- real
- imaginary
- amplitude
- phase

Potentially we can use the animator for the UI. To be discussed further.

jolopezl:
Open complex images using LELified image expressions:

- [x] ~Load/append image on double click, default on double click is amplitude.~ (to be supported from the backend)
- [x] Load/append image selecting complex component from menu.
- [x] Check that scripting works.

### DIFF
--- a/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
+++ b/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
@@ -23,8 +23,6 @@ export class FileBrowserDialogComponent extends React.Component {
     @observable defaultHeight: number;
     @observable enableImageArithmetic: boolean = false;
     @observable imageArithmeticString: string = "";
-    @observable imageArithmeticStringMap: Map<string, string> = new Map<string, string>();
-    @observable isComplexImage: boolean = false;
     private readonly imageArithmeticInputRef: React.RefObject<HTMLInputElement>;
 
     constructor(props: any) {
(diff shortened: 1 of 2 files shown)


kswang1029 (changes requested):
@jolopezl few issues to be addressed:
1. the "Load as" and "Append as" button need to be colored in blue to be consistent with the "Load" and "Append" buttons. 
2. If we multi-select images in the file list, we should show "Load selected" or "Append selected" buttons instead of "Load as" or "Append as" buttons and if a complex image is part of the multi-selected images, we load it as amplitude (same assumption with double-click to load). The current behavior is if there is a complex image in the multi-selected image list, "Load as" (or "Append as") button shows up and will load the complex image only. Other selected regular images are not loaded.

kswang1029 (approved):

veggiesaurus (approved):

src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
```diff
@@ -97,6 +97,24 @@ export class FileBrowserDialogComponent extends React.Component {
         return frame;
     };
 
+    private loadComplexImage = async (filename: string, expression: string) => {
+        this.imageArithmeticString = `${expression}("${filename}")`;
```
YuHsuan-Hwang: We need `@action` when changing the observable.
I suggest we instead define a new variable here ``const imageArithmeticString = `${expression}("${filename}")`;`` and remove `this.clearArithmeticString(); this.setEnableImageArithmetic(false);` because we don't need to change the status of the UI.

YuHsuan-Hwang (approved):

Issue #1155, opened 2020-08-24
support vector field rendering


kswang1029:
In order to display polarized intensity and angle properly, the frontend needs to support vector field rendering. 
For each line segment, length and orientation should be supported. For example, orientation is derived from one image (pixel value as angle) and length is derived from another image (normalized pixel value as the rendered length). In addition, block averaging (done by the backend) should be also supported to enhance signal-to-noise ratio.

counterpart backend issue: https://github.com/CARTAvis/carta-backend/issues/1002

Google doc [here](https://docs.google.com/document/d/1S8nNvonlowNwX_uPoPspxugUEnw5wRxbAS6y2HScTzs/edit#)

veggiesaurus:
if we're smart, I think we can just adjust the mip level of the raster data requested from the backend. Then as users zoom in, they'll get more arrows per pixel, but the same _total_ number of arrows. 

For example, if the frontend needs one arrow every 32px, then the backend will downsample the image 32x32 when the image is at 1.0x zoom, and 16x16 when at 2.0x zoom, and so on

veggiesaurus:
> if we're smart, I think we can just adjust the mip level of the raster data requested from the backend. Then as users zoom in, they'll get more arrows per pixel, but the same _total_ number of arrows.

Update: we've opted to generate the overlay at full resolution and stream to the frontend tile-by-tile, in a similar approach to the contour overlay.

Issue #1016, opened 2020-07-20
Make popup image tool buttons user configurable


Kechil:
Display on main image or on a toolbar? They can get in the way,

PR #1829, opened 2022-04-13
Angus/1155 vector fields

veggiesaurus:
~~Draft PR for #1155. Still to do:~~ Ready for review

- [x] Correct handling of intensity/angle Stokes coordinates
- [x] Adding overlay toggle icon to the image view widget
- [x] Support spatial matched images (if identical FoV/projection) when not in tiled view
- [x] Include vector overlay output when saving images

Known issues:
- ~~Line thickness and length is incorrect when spatially matched images have different pixel sizes~~
- ZFP compression is not supported. This will be added in a separate PR

Depends on [backend PR #1016](https://github.com/CARTAvis/carta-backend/pull/1016)


kswang1029:
@veggiesaurus @markccchiang here is a set of issues I spotted so far. Most of them are minor.  

[frontend] When debiasing toggle is on, the Apply button is not greyed out after a click.

[frontend] Pixel Averaging text input should correct user input if it is not an allowed value (eg 1, 3 or -1 etc).

[frontend] default Line Length Max to 20 (better UX in my opinion). 

[backend / frontend] possibly there are vector components rendered on top of empty pixels (ie outside the image) or on top of NaN pixels. 

[frontend] UI element alignment in the Styling tab. The ones in the Configuration tab are well aligned. 

[frontend] when the mode is "Angle Only (lines)", the radio buttons of "Polarization intensity" need to be greyed out.

[frontend / backend] I see the default threshold is 0 and is set in the SET_VECTOR_OVERLAY_PARAMETERS message and the backend applies it to the vector field data calculation. The default threshold should be NaN so that the backend does not apply a threshold in the calculations. 

[frontend] default contour color becomes white? (Should be green IIRC)

[frontend] could you add a new UI field as "extra rotation" in degrees please? Then we will need to apply this angle offset to the computed PA. The use case is to turn polarization orientation to magnetic field orientation (here it is 90 degrees offset).

[frontend] refine the checker for threshold text input field. Usually the value is quite small such as 0.0001. At the moment it is tricky to refine the value from 0.0001 to 0.0002 for example. 

[frontend] for line length (px) min and max, it will have better UX if it is a spin box so users can continuously change length by multiple clicking the spin up/down button and see the results changing.

veggiesaurus:
@kswang1029 

- [x] [frontend] When debiasing toggle is on, the Apply button is not greyed out after a click.
- [x] [frontend] default Line Length Max to 20 (better UX in my opinion).
- [x] [frontend] UI element alignment in the Styling tab. The ones in the Configuration tab are well aligned.
- [x] [frontend] when the mode is "Angle Only (lines)", the radio buttons of "Polarization intensity" need to be greyed out.
- [x] [frontend / backend] I see the default threshold is 0 and is set in the SET_VECTOR_OVERLAY_PARAMETERS message and the backend applies it to the vector field data calculation. The default threshold should be NaN so that the backend does not apply a threshold in the calculations.
- [x] [frontend] default contour color becomes white? (Should be green IIRC)
- [x] [frontend] could you add a new UI field as "extra rotation" in degrees please? Then we will need to apply this angle offset to the computed PA. The use case is to turn polarization orientation to magnetic field orientation (here it is 90 degrees offset).
- [x] [frontend] refine the checker for threshold text input field. Usually the value is quite small such as 0.0001. At the moment it is tricky to refine the value from 0.0001 to 0.0002 for example.
- [x] [frontend] for line length (px) min and max, it will have better UX if it is a spin box so users can continuously change length by multiple clicking the spin up/down button and see the results changing.
- [ ] fix width/length of markers when using spatial matching

> [frontend] Pixel Averaging text input should correct user input if it is not an allowed value (eg 1, 3 or -1 etc).

Can we fix this in a separate PR? It's more related to the actual input element's display. The value itself gets corrected.

> [backend / frontend] possibly there are vector components rendered on top of empty pixels (ie outside the image) or on top of NaN pixels.

Confirmed non-issue

veggiesaurus:
~~The one remaining issue before marking as ready for review is the incorrect length and thickness of markers when images have different pixel sizes~~ edit: fixed

kswang1029:
@veggiesaurus found a critical issue, likely is from our old friend "deviceRatio" 🙈
On a Retina display the spatial matching does not work properly. 

HD142527_Stokes_IQU fits-HD142527_Stokes_IQU_lpoli image fits-image-2022-04-22-11-19-32

kswang1029:
@veggiesaurus found another odd behavior when spatial matching is on (this is again with retina display):
The example here is I have two sources on the sky (one at North, the other at South), and I have two images with each centered at different sources. 
Screen Shot 2022-04-22 at 13 35 48

Now I load them, match them and use the first image as the data source to generate the vector field layers. As we can see the vector field is not matched properly on the second image. 

Screen Shot 2022-04-22 at 13 30 21

if we enable threshold, the vector field on the second image becomes even odd.
Screen Shot 2022-04-22 at 13 34 10


UPDATE: the above issue can only be reproduced with a Retina display.

kswang1029:
@veggiesaurus other minor comments to consider:

1. [optional UX enhancement] auto-select "Computed PA" and "Computed PI" if the Data Source has Stokes QU

2. Allowing pixel averaging factor options to include 1

3. Set the threshold unit to the image pixel unit instead of % when fractional is selected. I had some experiments with the help from Mark to see if we can use fractional % for masking but the result is bad. So we will need to use Stokes I (or PI, TBD) value for masking in order to get good results.

4. if it is Fractional selected, the "Intensity (Jy/beam)" in Styling tab should be revised to "Fractional (%)" I think. 

5. [question] I haven't tried this extreme case but can we have layers of one raster + multiple contours + multiple catalogs + multiple vector field layers at the same time? I remembered Qi mentioned a limitation of WebGL instances a while ago when he was trying to use WebGL acceleration from plotly for all plots?

6. Just for consistency: "Angular source" -> "Angular Source", "Intensity source" -> "Intensity Source". Question: is it angular source or angle source more sensible? 

7. When streaming vector data, do we show the green cloud indicator at the top right corner too?

8. when the color mode is color-mapped, do we map Intensity Min and Intensity Max (as defined in the same Styling tab) to the colormap?

src/components/Dialogs/VectorOverlayDialog/VectorOverlayDialogComponent.tsx
```diff
@@ -0,0 +1,410 @@
...
+                </FormGroup>
+                <FormGroup inline={true} label="Intensity source">
+                    <HTMLSelect value={this.intensitySource} onChange={this.handleIntensitySourceChanged}>
+                        <option value={VectorOverlaySource.None}>None</option>
+                        <option value={VectorOverlaySource.Current}>Current image</option>
+                        <option value={VectorOverlaySource.Computed}>Computed PI (if available)</option>
+                    </HTMLSelect>
+                </FormGroup>
+                <FormGroup inline={true} label="Pixel Averaging">
+                    <SafeNumericInput placeholder="Pixel Averaging" min={2} max={64} value={this.pixelAveraging} majorStepSize={2} stepSize={2} onValueChange={this.handlePixelAveragingChanged} />
```
YuHsuan-Hwang: Shall we also allow `pixelAveraging = 1`?

src/components/Dialogs/VectorOverlayDialog/VectorOverlayDialogComponent.tsx
```diff
@@ -0,0 +1,410 @@
...
+        const dataSource = appStore.activeFrame;
+        const intensityOnly = dataSource.vectorOverlayConfig.angularSource === VectorOverlaySource.None;
+
+        const configPanel = (
+            <div className="vector-overlay-config-panel">
+                <FormGroup inline={true} label="Angular source">
+                    <HTMLSelect value={this.angularSource} onChange={this.handleAngularSourceChanged}>
+                        <option value={VectorOverlaySource.None}>None</option>
+                        <option value={VectorOverlaySource.Current}>Current image</option>
+                        <option value={VectorOverlaySource.Computed}>Computed PA (if available)</option>
```
YuHsuan-Hwang: Would it be more clear to the users if we hide this option when PA is not available?

src/components/Dialogs/VectorOverlayDialog/VectorOverlayDialogComponent.tsx
```diff
@@ -0,0 +1,410 @@
...
+                    </HTMLSelect>
+                </FormGroup>
+                <FormGroup inline={true} label="Intensity source">
+                    <HTMLSelect value={this.intensitySource} onChange={this.handleIntensitySourceChanged}>
+                        <option value={VectorOverlaySource.None}>None</option>
+                        <option value={VectorOverlaySource.Current}>Current image</option>
+                        <option value={VectorOverlaySource.Computed}>Computed PI (if available)</option>
```
YuHsuan-Hwang: Same comment when PI is not available

src/components/Dialogs/VectorOverlayDialog/VectorOverlayDialogComponent.tsx
```diff
@@ -0,0 +1,410 @@
...
+                    </HTMLSelect>
+                </FormGroup>
+                <FormGroup inline={true} label="Pixel Averaging">
+                    <SafeNumericInput placeholder="Pixel Averaging" min={2} max={64} value={this.pixelAveraging} majorStepSize={2} stepSize={2} onValueChange={this.handlePixelAveragingChanged} />
+                </FormGroup>
+                <FormGroup inline={true} label="Polarization Intensity" disabled={this.intensitySource === VectorOverlaySource.None}>
```
YuHsuan-Hwang, line 294: The unit can be displayed with `labelInfo`. Same comments for all the other `FormGroup` elements.

src/components/Dialogs/VectorOverlayDialog/VectorOverlayDialogComponent.tsx
```diff
@@ -0,0 +1,410 @@
...
+        if (!frame) {
+            return null;
+        }
+
+        const config = frame.vectorOverlayConfig;
+        const intensityMin = isFinite(config.intensityMin) ? config.intensityMin : frame.vectorOverlayStore.intensityMin;
+        const intensityMax = isFinite(config.intensityMax) ? config.intensityMax : frame.vectorOverlayStore.intensityMax;
+
+        return (
+            <FormGroup label={`Intensity (${frame.unit})`} inline={true}>
```
YuHsuan-Hwang: Shall we disable these inputs when intensity source is None?
Also the column widths seems to change after applying the vectors.
image
image

src/components/Dialogs/VectorOverlayDialog/VectorOverlayDialogComponent.tsx
```diff
@@ -0,0 +1,410 @@
...
+        }
+    }
+
+    constructor(props: {appStore: AppStore}) {
+        super(props);
+        makeObservable(this);
+        this.setDefaultVectorOverlayParameters();
+    }
+
+    @action setDefaultVectorOverlayParameters() {
```
YuHsuan-Hwang: may change to an arrow function

src/components/Dialogs/VectorOverlayDialog/VectorOverlayDialogComponent.tsx
```diff
@@ -0,0 +1,410 @@
...
+
+        return (
+            <FormGroup label={intensityOnly ? "Block Width (px)" : "Line Length (px)"} inline={true}>
+                <div className="parameter-container">
+                    <div className="parameter-line parameter-length">
+                        <FormGroup inline={true} label="Min">
+                            <SafeNumericInput min={0} max={config.lengthMax} disabled={angleOnly} value={config.lengthMin} onValueChange={val => config.setLengthRange(val, config.lengthMax)} />
+                        </FormGroup>
+                        <FormGroup inline={true} label="Max">
+                            <SafeNumericInput min={config.lengthMin} disabled={angleOnly} value={config.lengthMax} onValueChange={val => config.setLengthRange(config.lengthMin, val)} />
```
YuHsuan-Hwang: Seems that max length is applied when `angleOnly`. Shall we allow users change this value when `angleOnly`?

src/components/Dialogs/VectorOverlayDialog/VectorOverlayDialogComponent.tsx
```diff
@@ -0,0 +1,410 @@
...
+                    </FormGroup>
+                    <Tabs defaultSelectedTabId={VectorOverlayDialogTabs.Configuration} renderActiveTabPanelOnly={false}>
+                        <Tab id={VectorOverlayDialogTabs.Configuration} title="Configuration" panel={configPanel} panelClassName="vector-overlay-config-panel" />
+                        <Tab id={VectorOverlayDialogTabs.Styling} title="Styling" panel={stylingPanel} panelClassName="vector-overlay-styling-panel" />
+                    </Tabs>
+                </div>
+                <div className={Classes.DIALOG_FOOTER}>
+                    <div className={Classes.DIALOG_FOOTER_ACTIONS}>
+                        <AnchorButton intent={Intent.WARNING} onClick={this.handleClearOverlay} disabled={!dataSource.vectorOverlayConfig.enabled} text="Clear" />
+                        <AnchorButton intent={Intent.SUCCESS} onClick={this.handleApplyOverlay} disabled={!this.configChanged && dataSource.vectorOverlayConfig.enabled} text="Apply" />
```
YuHsuan-Hwang: maybe we could disable Apply button when both angle and intensity source is None

YuHsuan-Hwang (commented):
I have some minor comments about the dialog UI

veggiesaurus:
> @veggiesaurus found another odd behavior when spatial matching is on (this is again with retina display): The example here is I have two sources on the sky (one at North, the other at South), and I have two images with each centered at different sources.
> UPDATE: the above issue can only be reproduced with a Retina display.

Weird, it seems to work fine for me, with `devicePixelRatio = 2` or `1` :thinking: 

comparison_dev_pixel_ratio_2

src/components/Dialogs/VectorOverlayDialog/VectorOverlayDialogComponent.tsx
veggiesaurus: fixed :+1:

veggiesaurus:
> @veggiesaurus other minor comments to consider:
> 


- [x] 1. [optional UX enhancement] auto-select "Computed PA" and "Computed PI" if the Data Source has Stokes QU
- [x] 2. Allowing pixel averaging factor options to include 1
- [x]  3. Set the threshold unit to the image pixel unit instead of % when fractional is selected. I had some experiments with the help from Mark to see if we can use fractional % for masking but the result is bad. So we will need to use Stokes I (or PI, TBD) value for masking in order to get good results.
- [x]  4. if it is Fractional selected, the "Intensity (Jy/beam)" in Styling tab should be revised to "Fractional (%)" I think.
- [x]  6. Just for consistency: "Angular source" -> "Angular Source", "Intensity source" -> "Intensity Source". Question: is it angular source or angle source more sensible?
- [x] 7. When streaming vector data, do we show the green cloud indicator at the top right corner too?

> 5. [question] I haven't tried this extreme case but can we have layers of one raster + multiple contours + multiple catalogs + multiple vector field layers at the same time? I remembered Qi mentioned a limitation of WebGL instances a while ago when he was trying to use WebGL acceleration from plotly for all plots?

Shouldn't be a problem. We only have a single webgl instance for each feature (raster, contour, catalog, overlay), versus per-image.

> 8. when the color mode is color-mapped, do we map Intensity Min and Intensity Max (as defined in the same Styling tab) to the colormap?

We scale from [intensity min, intensity max] to [0, 1] linearly and then map that to the colormap. Same process as contours (where we scale from [lowest contour level, highest contour level] to [0, 1]

kswang1029:
@veggiesaurus 
> * [x]  2. Allowing pixel averaging factor options to include 1

If we click the spin down button to set the pixel averaging factor to 1 and click spin up button, we will see odd factors. Maybe we need a toggle to disable pixel averaging instead (factor =1)? 🤔

kswang1029:
@veggiesaurus @YuHsuan-Hwang 
Safari (Version 15.4 (17613.1.17.1.13)) is naughty!  It appears that the rendered line segments have an upper limit in total number. Following is with AMD GPU (so no my weird matching issue from my intel  GPU) at different averaging factors. With Chrome it is free of this issue.

Screen Shot 2022-04-29 at 11 04 04

Screen Shot 2022-04-29 at 11 04 16

Screen Shot 2022-04-29 at 11 03 23

Screen Shot 2022-04-29 at 11 03 48

kswang1029:
> We scale from [intensity min, intensity max] to [0, 1] linearly and then map that to the colormap. Same process as contours (where we scale from [lowest contour level, highest contour level] to [0, 1]

@veggiesaurus ok. If there is one image only, then color-mapped vector field rendering works as expected. However, if there is a matched image, it does not. The scale changes in an unexpected way and it appears changing to a different colormap does not work properly and it seems broken irreversibly.

veggiesaurus:
> If we click the spin down button to set the pixel averaging factor to 1 and click spin up button, we will see odd factors. Maybe we need a toggle to disable pixel averaging instead (factor =1)? thinking

ok, we can do that instead.

kswang1029:
@veggiesaurus I had some tests with the latest commit [d647780](https://github.com/CARTAvis/carta-frontend/pull/1829/commits/d647780c57d298bf4ceaac4632dcad50c94ce7ca) and here is probably the last set of comments:

1. The unit of Threshold should be the same as the unit of the image when Polarization intensity is Absolute or Fractional. Typo in the unit: {Jy/beam)

2. When there is a matched image, the color mode "color-mapped" is not working (eg changing a different colormap does not work correctly, you can try grey but we will see yellow vector segments) 

3. When Polarization Intensity is Fractional, in the Styling tab the unit of the Intensity Min/Max should be %.

veggiesaurus:
> When there is a matched image, the color mode "color-mapped" is not working (eg changing a different colormap does not work correctly, you can try grey but we will see yellow vector segments)

@kswang1029 I will just try to fix this and then re-request reviews.

kswang1029:
@veggiesaurus @YuHsuan-Hwang happy to report that with [1785db7](https://github.com/CARTAvis/carta-frontend/pull/1829/commits/1785db751117fe83bc2eba9a782482c715b7d146) the spatial matching issue with my local intel GPU is gone~

kswang1029:
> @veggiesaurus @YuHsuan-Hwang Safari (Version 15.4 (17613.1.17.1.13)) is naughty! It appears that the rendered line segments have an upper limit in total number. Following is with AMD GPU (so no my weird matching issue from my intel GPU) at different averaging factors. With Chrome it is free of this issue.
> 
> Screen Shot 2022-04-29 at 11 04 04
> 
> Screen Shot 2022-04-29 at 11 04 16
> 
> Screen Shot 2022-04-29 at 11 03 23
> 
> Screen Shot 2022-04-29 at 11 03 48

To add more clues on this, if we load an image as 512 x 512 and render the vector field overlay with an averaging factor of 2 (so there are 256 x 256 elements), it works fine. But if there is no averaging, meaning to rendering 512 x 512 elements, then it does not work properly. 

webglreport
Screen Shot 2022-05-05 at 17 58 34

kswang1029:
@veggiesaurus shall we add a "lock" icon next to the data source dropdown too?

Screen Shot 2022-05-06 at 10 28 15

veggiesaurus:
> @veggiesaurus shall we add a "lock" icon next to the data source dropdown too?
> 
> Screen Shot 2022-05-06 at 10 28 15

can we add this as a separate issue?

kswang1029:
> > @veggiesaurus shall we add a "lock" icon next to the data source dropdown too?
> > Screen Shot 2022-05-06 at 10 28 15
> 
> can we add this as a separate issue?

ok

src/components/Dialogs/VectorOverlayDialog/VectorOverlayDialogComponent.tsx
```diff
@@ -0,0 +1,421 @@
...
+    @action private handleIntensitySourceChanged = (ev: React.ChangeEvent<HTMLSelectElement>) => {
+        this.intensitySource = parseInt(ev.currentTarget.value) as VectorOverlaySource;
+    };
+
+    @action private handlePixelAveragingEnabledChanged = (ev: React.ChangeEvent<HTMLInputElement>) => {
+        this.pixelAveragingEnabled = ev.currentTarget.checked;
+    };
+
+    @action private handlePixelAveragingChanged = (value: number) => {
+        this.pixelAveraging = value;
```
YuHsuan-Hwang: We may still need `Math.floor(value * 0.5) * 2.0` when users input non integer values.

src/stores/Frame/FrameStore.ts
```diff
@@ -821,6 +826,17 @@ export class FrameStore {
...
     }
 
+    @computed get hasLinearStokes(): boolean {
+        const options = this.stokesOptions;
+        const labelQ = POLARIZATION_LABELS.get("Q");
+        const labelU = POLARIZATION_LABELS.get("U");
+        return !!options.find(o => o.label === labelQ) && !!options.find(o => o.label === labelU);
+    }
+
+    @computed get requiredStokesName(): string {
```
YuHsuan-Hwang: looks like an unused variable

src/components/ImageView/VectorOverlayView/VectorOverlayView.tsx
```diff
@@ -0,0 +1,269 @@
...
+            } else {
+                console.error("Could not generate control map for vector overlay");
+            }
+            this.gl.activeTexture(GL2.TEXTURE1);
+            this.gl.bindTexture(GL2.TEXTURE_2D, controlMap.getTextureX(this.gl));
+            this.gl.uniform1i(shaderUniforms.ControlMapTexture, 1);
+        }
+
+        this.gl.uniform1i(shaderUniforms.DataTexture, 0);
+        this.gl.uniform1f(shaderUniforms.CanvasSpaceLineWidth, lineThickness);
```
YuHsuan-Hwang: Could we simplify to `pixelRatio * frame.vectorOverlayConfig.thickness` and remove `lineThickness`?

kswang1029 (approved):

YuHsuan-Hwang (changes requested):
The shaders look good to me. I have only minor comments.
About file naming, would it be clearer if we rename the files as "..._shader_vectors.glsl"?

veggiesaurus:
> The shaders look good to me. I have only minor comments. About file naming, would it be clearer if we rename the files as "..._shader_vectors.glsl"?

Can we do this in a separate branch? Since this means changes across multiple features

veggiesaurus:
@YuHsuan-Hwang I've implemented the changes you suggested

YuHsuan-Hwang:
> > The shaders look good to me. I have only minor comments. About file naming, would it be clearer if we rename the files as "..._shader_vectors.glsl"?
> 
> Can we do this in a separate branch? Since this means changes across multiple features

Ok

YuHsuan-Hwang (approved):

PR #1830, opened 2022-04-15
Yuhsuan/1653 save restfreq

### LINKED ISSUE #1653: support reset rest frequency when saving as a subimage
The relevant backend issue is https://github.com/CARTAvis/carta-backend/issues/918

The current subimage export adopts the same RESTFRQ defined the in header of the input cube. This feature is to support the ability to set a new rest frequency of the subimage. This frontend part needs to provide a new text input field for this purpose and send this information to the backend.

YuHsuan-Hwang:
Closes #1653 with some code maintenance:

* Add rest frequency input in the image save panel
    * update the value after adjusting custom rest freq in the image layer settings
    * set to header rest freq by the reset button
* Add `Freq` model for `FileBrowserStore` and `RestFreqStore`
* Refactor `ImageSaveComponent`
    * Remove `renderSaveImageControl` and `const stokesOptions`
    * change to `Formgroup` to avoid manually setting margins in CSS
* Add `headerRestFreq` in `FrameStore`

Companion PRs:
https://github.com/CARTAvis/carta-backend/pull/1088
https://github.com/CARTAvis/carta-protobuf/pull/69

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+* The ability to set a custom rest frequency for saving subimages. ([#1653](https://github.com/CARTAvis/carta-frontend/issues/1653)).
+
 ## [3.0.0-beta.2]
 
 ### Added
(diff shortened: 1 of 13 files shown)


kswang1029:
@YuHsuan-Hwang a quick question, if the loaded image has a shape as [100, 100, 1, 1] (ie one channel and one stokes) and the header contains restfrq, can we also allow user to set a custom rest frequency when saving a subimage? This would support additional use cases I suppose. If the required change is large, we can leave it for now. 

UPDATE: never mind for now. I will do some more test to see if the above is really necessary.

kswang1029 (approved):
tested 👍 @veggiesaurus @TienHao please proceed code review.

veggiesaurus (approved):

PR #1835, opened 2022-04-22
Jrhosk/1703 spectral profile secondary axis

jrhosk:
### Implemented:
Secondary axis in spectral profile plots. Axis can be added by navigating to the spectral profile _Settings_, _Styling_ and click "Show Secondary Axis". Under the _Conversion_ tab there will now be an additional axis option that includes everything but `channel`.

### Pending Issues
**`Channel` is not implemented on secondary axis**. The problem seems to be that there is no good way to make the conversion. To explain further, the values on the secondary axis are set by getting the ticks from the primary axis and converting them to the desired units for the secondary axis. The conversion goes as follows: `(top (axis unit) => AST.transformSpectralPoint(..., false)` to convert to native coordinates, `AST.transformSpectralPoint(...) => toFormattedNotation(...) -> (secondary unit)`. This convert the native axis to the users desired secondary unit and then does some simple formatting. The issue here is that the `AST.transformSpectralPoint(...)` function does not support the conversion from `frequency` to `channel`. For the primary axis this is handled by getting the channel from the `channelInfo` function, which can be done for the secondary axis, but in this case the ticks will not align because the primary axis formatting is handled by `Chartjs`. 

**Conversion from `Freq`(lower axis) to `wavelength` (upper axis)** seems to have slightly different rounding for each tick versus `Freq`(upper axis) to `wavelength` (lower axis).  As an example, the following shows a set of conversions in both directions.
**Freq(lower) to Vacuum(upper):**
372.672 (GHz)(lower) ===⇒ 0.804441 (mm)(upper)
372.672 (GHz)(lower) ===⇒ 804440.521 (nm)(upper)

**Vacuum(upper) to Freq(lower):**
804440 (nm)(lower) ===⇒ 372.672 (GHz)(upper) 

**Vacuum(mm) to Vacuum (nm):**
0.80444 (mm)(lower) ===⇒ 804440 (mm)(upper)

So the values are consistent within rounding for each corresponding tick but there are cases where lower axis truncates the values instead of rounding it.


veggiesaurus:
@jrhosk not sure why `package-lock.json` is removed in this PR :innocent:

jrhosk:
Not intentionally. I have not idea, should I add it?

veggiesaurus:
> Not intentionally. I have not idea, should I add it?

You should roll back the changes to this file, unless you've added or removed packages or upgraded them

jrhosk:
I have not personally changed it to my knowledge. I assumed it was due to changes from when I merged dev into my local branch.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -1,19 +1,22 @@
...
+import {Tick} from "chart.js";
 import {Colors, NonIdealState} from "@blueprintjs/core";
 import ReactResizeDetector from "react-resize-detector";
 import SplitPane, {Pane} from "react-split-pane";
 import {LineMarker, LinePlotComponent, LinePlotComponentProps, LinePlotSelectingMode, SmoothingType, PlotType} from "components/Shared";
 import {MultiPlotProps, TickType} from "../Shared/LinePlot/PlotContainer/PlotContainerComponent";
 import {SpectralProfilerToolbarComponent} from "./SpectralProfilerToolbarComponent/SpectralProfilerToolbarComponent";
 import {ProfileInfo, SpectralProfilerInfoComponent} from "./SpectralProfilerInfoComponent/SpectralProfilerInfoComponent";
-import {WidgetProps, HelpType, AnimatorStore, WidgetsStore, AppStore, DefaultWidgetConfig} from "stores";
+import {WidgetProps, HelpType, AnimatorStore, WidgetsStore, SpectralProfileStore, AppStore, DefaultWidgetConfig /*, ASTSettingsString, OverlayStore*/} from "stores";
```
veggiesaurus: please remove commented code

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -49,6 +54,15 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
         console.log("can't find store for widget");
         return new SpectralProfileWidgetStore();
     }
+    // not sure if I need this
+    @computed get profileStore(): SpectralProfileStore {
+        const widgetStore = this.widgetStore;
+        if (widgetStore.effectiveFrame) {
+            //const profileKey = `${widgetStore.effectiveFrame.frameInfo.fileId}-${widgetStore.effectiveRegionId}`;
```
veggiesaurus, line 62: please remove commented code

kswang1029 (changes requested):
@jrhosk thanks for implementing this feature 👍 
Here I report a layout issue
Screen Shot 2022-04-25 at 10 16 50
Screen Shot 2022-04-25 at 10 15 30

1. extra space is observed in spatial profiler widget, stokes analysis widget, histogram widget, render config widget, and contour dialog etc. Basically, _all_ widgets/dialogs containing a chartjs chart are affected.

2. except the spectral profiler widget which has a switch to enable/disable the secondary axis, all widgets/dialogs containing a chartjs chart have the secondary axis enabled and cannot be switched off. Those should not show the secondary axis and there is no need to support the secondary axis.

kswang1029:
@jrhosk Other findings from my test:
1. the spectral profiler (when docked) or the entire GUI (when undocked) may crash when enabling the secondary axis.
Screen Shot 2022-04-25 at 10 29 44

2. to have a better UX, we would need to automatically adjust the displayed precision of the 2nd axis so that each tic value is distinct.
Screen Shot 2022-04-25 at 10 36 20

3. when the 1st axis is channel, the x2 label becomes None.
Screen Shot 2022-04-25 at 10 40 51

4. if we have a fresh session and enable the spectral profiler with the 2nd axis enabled, then if we load an image the GUI will crash.

kswang1029:
@jrhosk @veggiesaurus @YuHsuan-Hwang I suggest we find a time to talk about this a bit. I have an alternative idea but not sure if it is worth to do and feasible.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
jrhosk: This is done.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
jrhosk: I need to look at this more closely. This might be why @kswang1029 sees the crash when undocking. For now I have left it commented because the code crashes without the function, does nothing but works when the code is commented and fails when it is uncommnented. I didn't write this and don't understand how to fix the issue yet.

```
/export/home/fornax/jhoskins/Development/CARTA/carta-frontend/src/components/SpectralProfiler/SpectralProfilerComponent.tsx
TypeScript error in /export/home/fornax/jhoskins/Development/CARTA/carta-frontend/src/components/SpectralProfiler/SpectralProfilerComponent.tsx(63,13):
Type 'ObservableMap<number, SpectralProfileStore>' is missing the following properties from type 'SpectralProfileStore': fileId, regionId, profiles, getProfile, and 2 more.  TS2740

    61 |         if (widgetStore.effectiveFrame) {
    62 |         //    const profileKey = `${widgetStore.effectiveFrame.frameInfo.fileId}-${widgetStore.effectiveRegionId}`;
  > 63 |             return AppStore.Instance.spectralProfiles.get(widgetStore.effectiveFrame.frameInfo.fileId);
       |             ^
    64 |         }
    65 |         return undefined;
    66 |     }
```

jrhosk:
> @jrhosk Other findings from my test:
> 
> 1. the spectral profiler (when docked) or the entire GUI (when undocked) may crash when enabling the secondary axis.
>    Screen Shot 2022-04-25 at 10 29 44
> 2. to have a better UX, we would need to automatically adjust the displayed precision of the 2nd axis so that each tic value is distinct.
>    Screen Shot 2022-04-25 at 10 36 20
> 3. when the 1st axis is channel, the x2 label becomes None.
>    Screen Shot 2022-04-25 at 10 40 51
> 4. if we have a fresh session and enable the spectral profiler with the 2nd axis enabled, then if we load an image the GUI will crash.

1.  & 4. I think these have to do with the `@get profileStore()` issue above. 
2. I might have fixed this but I need to find an .image with similar values (1.3000x).
3 . Is easy enough to fix, I'll work on it today.

YuHsuan-Hwang:
Will revisit this issue after #1883 is merged.

PR #1836, opened 2022-04-25
Yuhsuan/1827 image save panel

### LINKED ISSUE #1827: Add NonIdealState for image save panel
**Describe the bug**
Need `NonIdealState` for the image save panel when there is no active images.

**To Reproduce**
1. Open a file
2. Open Save image dialog from the menu
3. Close the file by the menu or the image list widget
4. See crash

**Expected behavior**
See NonIdealState

**Screenshots or videos**
image

 - OS: MacOS
 - Browser: chrome
 - Version: latest dev branch

YuHsuan-Hwang:
* Closes #1827:
   * added error handling in `ImageSaveComponent` to avoid crashing
   * added `NonIdealState` when no active images

image

* Added directories for `ImageSaveComponent` and `RegionSelectComponent`


kswang1029 (approved):

src/components/Dialogs/FileBrowser/ImageSave/ImageSaveComponent.tsx
```diff
@@ -152,9 +152,9 @@ export class ImageSaveComponent extends React.Component {
                   })
                 : [];
         // Calculate a small step size
-        const numChannels = activeFrame.numChannels;
-        const min = activeFrame.channelValueBounds?.min;
-        const max = activeFrame.channelValueBounds?.max;
+        const numChannels = activeFrame?.numChannels;
+        const min = activeFrame?.channelValueBounds?.min;
```
veggiesaurus: if `activeChannel` is undefined, then `min` and `max` are both undefined. So `delta` will be `NaN`. Perhaps we should make min and max default to `0` and `1` using the `??` operator?

Unless passing `NaN` as step size works fine?

src/components/FileInfo/FileInfoComponent.tsx
```diff
@@ -237,6 +238,9 @@ export class FileInfoComponent extends React.Component<{
             // Here is only controls for save, no need to wait file info
             case FileInfoType.SAVE_IMAGE:
             case FileInfoType.SELECT_REGION:
+                if (!AppStore.Instance.activeFrame) {
+                    return <NonIdealState icon={"folder-open"} title={"No file loaded"} description={"Load a file using the menu"} />;
+                }
                 break;
```
veggiesaurus, line 244: don't need `break` after return

veggiesaurus (changes requested):
minor changes requested

src/components/Dialogs/FileBrowser/ImageSave/ImageSaveComponent.tsx
YuHsuan-Hwang: `NaN` works fine because all elements is hidden when no activeFrame, but I think it's safer to set to 0 and 1. will change this part.

src/components/FileInfo/FileInfoComponent.tsx
YuHsuan-Hwang: we still need `break` when `(!AppStore.Instance.activeFrame)` is false

Issue #1716, opened 2021-11-24
add warning to the region dialog if the displayed size differs significantly to the actual geodesic distance


kswang1029:
The region rendering approach now in CARTA assuming the image is flat which is not always true for wide field images. The displayed angular size at the moment is estimated by the product of n_pixel and angular_size_per_pixel. Ideally angular_size_per_pixel should not be a constant due to projection effect. 

The request here is we add a new check by taking the corners in pixel coordinates and convert them to world coordinates then compute the geodesic distance using AST. If we see >10% difference between the geodesic distance and the product of n_pixel and angular_size_per_pixel, we display a warning elsewhere in the dialog.

PR #1839, opened 2022-04-29
Yuhsuan/1397 world coord result

YuHsuan-Hwang:
This PR is for the second item of #1397: adding fitting results in world coordinates.
* [Fitting Result] shows the result in world coordinates, while [Full Log] shows the result in both px and world coordinates and shows more digits.
* If world coordinate info in the header is not sufficient, [Fitting Result] and [Full Log] only show the result in px.
* The position format is either `XX:XX:XX.XX...` or `XXX... (deg)`; the size format is in arcsec.
* The size calculation follows the same approach for regions (unit size estimated by the geodesic distance near `[CRPIX1, CRPIX2]`, see PR #1719). We may set warnings when the projection effect is large in the future (see #1716). If such cases are common, we may also consider calculating the geodesic distance near the Gaussian center.

### DIFF
--- a/src/components/Dialogs/FittingDialog/FittingDialogComponent.tsx
+++ b/src/components/Dialogs/FittingDialog/FittingDialogComponent.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import {observer} from "mobx-react";
-import {action, computed, makeObservable, observable} from "mobx";
+import {action, makeObservable, observable} from "mobx";
 import {AnchorButton, Classes, Dialog, FormGroup, HTMLSelect, Icon, IDialogProps, Intent, NonIdealState, Position, Pre, Slider, Tab, Tabs, Text} from "@blueprintjs/core";
 import {Tooltip2} from "@blueprintjs/popover2";
 import classNames from "classnames";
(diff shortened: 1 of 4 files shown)


kswang1029 (approved):
Tested with images with equatorial and galactic coordinates, and linear coordinate as well. All work well. 👍 

One minor UX improvement is for FWHM size, we can report the size in a proper unit as arcsec, arcmin, or degree depending on the value ranges. Please refer to the section "Unit of the offset axis" of https://docs.google.com/document/d/1Z1fCDj9IujGvV6ER_bzjjL2j_bF5ykc34GRhPdOiXI4/edit or refer to the list below
We have such feature in the generated PV image (ie the Offset axis). 

- **milliarcsecond** if FWHM < 2 milliarcsecond
- **arcsecond** if 2 milliarcsecond <= FWHM < 2 arcmin
- **arcminute** if 2 arcmin <= FWHM < 2 degree
- **degree** if 2 degree <= FWHM


Note for the world coordinate error, having it in arcsec is just fine.

veggiesaurus (approved):

kswang1029:
> Tested with images with equatorial and galactic coordinates, and linear coordinate as well. All work well. 👍
> 
> One minor UX improvement is for FWHM size, we can report the size in a proper unit as arcsec, arcmin, or degree depending on the value ranges. Please refer to the section "Unit of the offset axis" of https://docs.google.com/document/d/1Z1fCDj9IujGvV6ER_bzjjL2j_bF5ykc34GRhPdOiXI4/edit or refer to the list below We have such feature in the generated PV image (ie the Offset axis).
> 
> * **milliarcsecond** if FWHM < 2 milliarcsecond
> * **arcsecond** if 2 milliarcsecond <= FWHM < 2 arcmin
> * **arcminute** if 2 arcmin <= FWHM < 2 degree
> * **degree** if 2 degree <= FWHM
> 
> Note for the world coordinate error, having it in arcsec is just fine.

@YuHsuan-Hwang please consider the "smart unit" as future enhancement. Now it is good for v3-beta3.

YuHsuan-Hwang:
> @YuHsuan-Hwang please consider the "smart unit" as future enhancement. Now it is good for v3-beta3.

Added to the image fitting UI issue.

PR #1840, opened 2022-05-03
Yuhsuan/controlmap lookup texelfetch

YuHsuan-Hwang:
Replace `texture()` by `texelFetch()` to obtain texture data in a safer way guaranteeing no interpolation is applied. This change will affect the catalog and contour overlay on a spatially matched image.
Manually tested the performance of rendering 6000 catalog sources and see no significant difference.


veggiesaurus (approved):

kswang1029 (approved):
@YuHsuan-Hwang I have done some tests with the current dev branch and this branch by comparing png images rendered from the two branches. 

old
new

old2
new2

old3
new3

The comparison is done by checking the pixel RGBA values and looking for inconsistent pixels. 

new.png old.png 
match rate 99.9993806620527%  19/3067792 not matched.

new2.png old2.png
match rate 99.99687071352947% 96/3067792 not matched.

new3.png old3.png
match rate 99.99654474618879% 106/3067792 not matched.

These minor inconsistent pixels are due to antialiasing I suppose. Based on this test and some other manual tests, I think we can assume that the refactoring does not break contour and catalog matching.

kswang1029:
@veggiesaurus this new improvement should also affect the vector field matching feature I suppose. Not sure if when we should merge in this branch to dev. 🤔

YuHsuan-Hwang:
I prefer we first merge the vector overlay and recheck the rendering here. Seems easier to track the discussion

kswang1029 (approved):
Tested again to see if matched vector field rendering works fine and it seems so. 👍

PR #1849, opened 2022-05-12
add missing vector overlay and image fitting options to the View menu

kswang1029:
The missing vector overlay and image fitting options are added to the View menu.


acdo2002 (approved):
It works fine on my MacOS laptop with Chrome and Firefox browsers.

YuHsuan-Hwang (approved):

PR #1850, opened 2022-05-13
Yuhsuan/1397 size unit

YuHsuan-Hwang:
This PR is for the third item in #1397: convert output FWHM size to arcmin or degree.

changes:
* added angular size conversion to marcsec/arcmin/deg in the fitting results
* changed `FWHM X/Y` to `FWHM Major/Minor Axis` to match the terms used in initial value setting
* refactored `formattedArcsec` to reuse the conversion
* removed unused conversion in `getAngleInRad`. It was implemented for reading header entries in #1707, but then became  unused after we changed the approach in the same branch.

to be tested:
* size conversions in the fitting results
* region inputs work as usual; tsv file headers unchanged
* the spectral profile intensity `/beam` and `/arcsec^2` conversion works as usual

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [3.0.0-beta.4]
 
+### Added
+* Size conversion in the image fitting results ([#1397](https://github.com/CARTAvis/carta-frontend/issues/1397)).
 ### Fixed
 * Added missing vector overlay and image fitting options in the View menu ([#1848](https://github.com/CARTAvis/carta-frontend/issues/1848))
 * Hide code snippet option in the View menu when code snippet is disabled in the preferences ([#1856](https://github.com/CARTAvis/carta-frontend/issues/1856))
--- a/src/models/AngularSize.ts
+++ b/src/models/AngularSize.ts
@@ -1,0 +1,47 @@
+export enum AngularSizeUnit {
+    MILLIARCSEC = "marcsec",
+    ARCSEC = "arcsec",
+    ARCMIN = "arcmin",
+    DEG = "deg"
+}
+
+export class AngularSize {
+    value: number;
+    unit: AngularSizeUnit;
+
+    public static convertValueFromArcsec = (arcsec: number, dstUnit: AngularSizeUnit): number => {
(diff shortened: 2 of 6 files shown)


kswang1029 (approved):
@YuHsuan-Hwang I have checked 
- size conversions in the fitting results
- region inputs work as usual; tsv file headers unchanged
- the spectral profile intensity /beam and /arcsec^2 conversion works as usual

and they seem to work fine. 

Just spotted a UX issue regarding the smart unit, which also exists in the region config dialog so this can be addressed later: at the moment the smart unit is determined per FWHM so it is possible to see 

FWHM Major Axis = 1.014267 +/- 0.002858 (marcsec)
FWHM Minor Axis = 0.003010 +/- 0.000009 (arcsec)

for example. It happens when one FWHM value is above the threshold and the other is not. It would be better if the adopted unit can be considered jointly.

YuHsuan-Hwang:
> Just spotted a UX issue regarding the smart unit, which also exists in the region config dialog so this can be addressed later: at the moment the smart unit is determined per FWHM so it is possible to see
> 
> FWHM Major Axis = 1.014267 +/- 0.002858 (marcsec)
> FWHM Minor Axis = 0.003010 +/- 0.000009 (arcsec)
> 
> for example. It happens when one FWHM value is above the threshold and the other is not. It would be better if the adopted unit can be considered jointly.

@kswang1029 do we show marcsec for both in this case?
For the region dialog, seems like we only need to adjust rectangle and ellipse region width/height units. I can try to implement this in the next PR.

kswang1029:
> > Just spotted a UX issue regarding the smart unit, which also exists in the region config dialog so this can be addressed later: at the moment the smart unit is determined per FWHM so it is possible to see
> > FWHM Major Axis = 1.014267 +/- 0.002858 (marcsec)
> > FWHM Minor Axis = 0.003010 +/- 0.000009 (arcsec)
> > for example. It happens when one FWHM value is above the threshold and the other is not. It would be better if the adopted unit can be considered jointly.
> 
> @kswang1029 do we show marcsec for both in this case? For the region dialog, seems like we only need to adjust rectangle and ellipse region width/height units. I can try to implement this in the next PR.

yes I suggest we use the smaller FWHM to determine the displayed unit .

veggiesaurus (approved):

PR #1851, opened 2022-05-13
1742 persistant export image tooltip

### LINKED ISSUE #1742: persistent export image tooltip
This only happens after 100% resolution is clicked.

Screen Shot 2021-12-07 at 1 49 04 PM

Michaelazzz:
closes #1742

### DIFF
--- a/src/stores/AppStore.ts
+++ b/src/stores/AppStore.ts
@@ -2240,6 +2240,7 @@ export class AppStore {
                         link.dispatchEvent(new MouseEvent("click"));
                     }, "image/png");
                 }
+            }).then(() => {
                 this.setIsExportingImage(false);
             });
         }
--- a/src/stores/AppStore.ts
+++ b/src/stores/AppStore.ts
@@ -2227,22 +2227,24 @@ export class AppStore {
 
             this.setIsExportingImage(true);
             this.setImageRatio(imageRatio);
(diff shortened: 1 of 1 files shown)


src/stores/AppStore.ts
```diff
@@ -2227,21 +2227,24 @@ export class AppStore {
...
-                        // Trim filename before timestamp to 200 characters to prevent browser errors
-                        link.download = `${joinedNames}-image`.substring(0, 200) + `-${getTimestamp()}.png`;
-                        link.href = URL.createObjectURL(blob);
-                        link.dispatchEvent(new MouseEvent("click"));
-                    }, "image/png");
-                }
-                this.setIsExportingImage(false);
-            });
+            this.waitForImageData()
+                .then(() => {
```
veggiesaurus, line 2231: might be simpler if we just switch this to an `async` function and use `await`

kswang1029:
@Michaelazzz it seems the tooltip is still persistent. This is with Chrome v101

https://user-images.githubusercontent.com/20819712/168520273-d056d282-8b8d-4e6b-ace8-faf68511673e.mov

kswang1029:
#1742 will be fixed via #1898

kswang1029:
Fixed via #1898

PR #1861, opened 2022-05-19
OIDC login

daikema:
This is intended to be paired with / in support of https://github.com/CARTAvis/carta-controller/pull/119 where I've tried to elaborate a little bit on the changes made.

My guess is that for the most part it's probably useful to try to consolidate most of the discussion regarding the changes there as that's where most of them were made.

### DIFF
--- a/src/services/ApiService.ts
+++ b/src/services/ApiService.ts
@@ -181,7 +181,7 @@ export class ApiService {
                     this.setToken(response.data.access_token, response.data.expires_in || Number.MAX_VALUE);
 
                     // Update refresh token
-                    if (response.data.refresh_token?) {
+                    if (response.data.refresh_token) {
                         localStorage.setItem("oidc_refresh_token", response.data.refresh_token)
                     }
                     return true;
@@ -230,6 +230,7 @@ export class ApiService {
 
             localStorage.removeItem("oidc_refresh_token")
             localStorage.removeItem("oidc_id_token")
+            localStorage.removeItem("authenticationType")
 
             window.open(ApiService.RuntimeConfig.logoutAddress + "?" + usp.toString(), "_self")
         } else if (ApiService.RuntimeConfig.logoutAddress) {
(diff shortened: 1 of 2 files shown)


src/services/ApiService.ts
```diff
@@ -183,16 +219,30 @@ export class ApiService {
...
+        } else if (ApiService.RuntimeConfig.oidcClientId) {
+            let usp = new URLSearchParams();
+            usp.set("id_token_hint", localStorage.getItem("oidc_id_token"));
+            usp.set("post_logout_redirect_uri", window.location.href);
+
+            localStorage.removeItem("oidc_refresh_token");
+            localStorage.removeItem("oidc_id_token");
+            localStorage.removeItem("authenticationType");
+
+            window.open(ApiService.RuntimeConfig.logoutAddress + "?" + usp.toString(), "_self");
```
veggiesaurus: I'd prefer `${ApiService.RuntimeConfig.logoutAddress}?${usp.toString()}` but I don't have a strong opinion about it :slightly_smiling_face:

src/services/ApiService.ts
```diff
@@ -158,6 +160,40 @@ export class ApiService {
             } catch (e) {
                 return false;
             }
+        } else if (ApiService.RuntimeConfig.oidcClientId) {
+            try {
+                if (!this.oidcAxiosInstance) {
+                    //console.log("Creating OIDC Axios instance")
```
YuHsuan-Hwang, line 166: Please remove comment.

I wonder what the changes in `package-lock.json` are for?

veggiesaurus: > I wonder what the changes in package-lock.json are for?

I was wondering that too :thinking:

daikema:
As per suggestion from @veggiesaurus updated this to a draft as I'm working on redoing the login doing more on the controller side, which hopefully won't require any changes to the frontend code.  That's still work in progress though.

YuHsuan-Hwang:
Implemented in #2016.

PR #1865, opened 2022-05-23
Thchang/1546 exporting full resolution data in the spatial profiler

### LINKED ISSUE #1546: Fix exported spatial profiles when using mipmapped data
Fix bug introduced by #1310:

> @veggiesaurus could you try the s255 cube? Following is what I see from the tsv file when the frontend decimation is applied for visualization.
> 
> Screen Shot 2021-07-06 at 10 41 59 AM
> 
> If we maximize the spatial profiler and make a big enough browser window so that there is no frontend decimation, we will see the full resolution data in the tsv.
> 
> Screen Shot 2021-07-06 at 10 42 51 AM

TienHao:
closes #1546: exporting full resolution data when showing decimated data in the plotting of the spatial profiler

changes:
- added `exportData` in LinePlotComponentProps to export data which are different from the plotting `data`.
- applied `exportData` to contain full resolution(un-decimated) data in the spatial profiler.

### DIFF
--- a/src/components/Shared/LinePlot/LinePlotComponent.tsx
+++ b/src/components/Shared/LinePlot/LinePlotComponent.tsx
@@ -122,7 +122,7 @@ export class LinePlotComponentProps {
     insideTexts?: LinePlotInsideTextMarker[];
     order?: number;
     multiPlotPropsMap?: Map<string, MultiPlotProps>;
-    exportData?: {x: number; y: number; z?: number}[];
+    fullResolutionData?: {x: number; y: number; z?: number}[];
 }
 
 // Maximum time between double clicks
@@ -661,7 +661,7 @@ export class LinePlotComponent extends React.Component<LinePlotComponentProps> {
             // data part
             rows.push("# x\ty");
             const useScientificForm = plotName === "histogram" || this.props.tickTypeX === TickType.Scientific;
-            const data = this.props.exportData?.length > 0 ? this.props.exportData : this.props.data;
+            const data = this.props.fullResolutionData?.length > 0 ? this.props.fullResolutionData : this.props.data;
             rows = rows.concat(data.map(o => (useScientificForm ? `${toExponential(o.x, 10)}\t${toExponential(o.y, 10)}` : `${o.x}\t${toExponential(o.y, 10)}`)));
 
             exportTsvFile(imageName, plotName, `${comment}\n${rows.join("\n")}\n`);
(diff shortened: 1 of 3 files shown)


src/components/Shared/LinePlot/LinePlotComponent.tsx
```diff
@@ -660,7 +661,8 @@ export class LinePlotComponent extends React.Component<LinePlotComponentProps> {
             // data part
             rows.push("# x\ty");
             const useScientificForm = plotName === "histogram" || this.props.tickTypeX === TickType.Scientific;
-            rows = rows.concat(this.props.data.map(o => (useScientificForm ? `${toExponential(o.x, 10)}\t${toExponential(o.y, 10)}` : `${o.x}\t${toExponential(o.y, 10)}`)));
+            const data = this.props.exportData?.length > 0 ? this.props.exportData : this.props.data;
```
veggiesaurus: I'd prefer it if we used something like `fullResolutinoData` instead of `exportData`, as we may reuse this elsewhere.

veggiesaurus (changes requested):

src/components/Shared/LinePlot/LinePlotComponent.tsx
```diff
@@ -122,6 +122,7 @@ export class LinePlotComponentProps {
     insideTexts?: LinePlotInsideTextMarker[];
     order?: number;
     multiPlotPropsMap?: Map<string, MultiPlotProps>;
+    exportData?: {x: number; y: number; z?: number}[];
```
YuHsuan-Hwang: Maybe not related to the fix, but I wonder if we can remove `z` and use `Point2D[]` for `exportData` and also `data`. Seems that `z` is not used, and it was for inheritance in `ScatterPlot`.

kswang1029 (approved):
👍

src/components/Shared/LinePlot/LinePlotComponent.tsx
TienHao: modified!

src/components/Shared/LinePlot/LinePlotComponent.tsx
TienHao: modified!

YuHsuan-Hwang (approved):

src/components/SpatialProfiler/SpatialProfilerComponent.tsx
```diff
@@ -116,12 +116,14 @@ export class SpatialProfilerComponent extends React.Component<WidgetProps> {
             let yCount = 0;
 
             let values: Array<{x: number; y: number}>;
+            let fullResolutionValues: Array<{x: number; y: number}>;
```
veggiesaurus: Can't we use Point2D here?

veggiesaurus (changes requested):

src/components/SpatialProfiler/SpatialProfilerComponent.tsx
TienHao: fixed

veggiesaurus (approved):

PR #1868, opened 2022-05-25
Set default histogram in contour widget is per-channel if stokes changed

markccchiang:
This is related to the [backend issue 1013](https://github.com/CARTAvis/carta-backend/issues/1013). The default histogram option in the contour widget should be `per-channel` when stokes is changed. Just like the histogram option in the render configuration widget.


kswang1029 (approved):
yes the issue is gone 👍

acdo2002 (approved):
Follow the reproduce steps in https://github.com/CARTAvis/carta-backend/issues/1013

The current backend and frontend dev branch, the ICD sequences would be:
image


However in the branch and the corresponding backend branch `mark/1013_avoid_auto_per_cube_histogram`.
The ICD sequences become:
image

The additional
image
has been removed.

Issue #1855, opened 2022-05-17
incorrect image statistics after a region is deleted where there are multiple statistics widget configured explicitly to display image region statistics


kswang1029:
**Describe the bug**
When there are two statistics widgets configured explicitly to display image region statistics, for example, one for image01 and region1 and the other for image02 and region1. Region1 is the only region created in CARTA. If we delete region1, the statistics widget for image01 shows the correct full image statistics as expected. But the other widget for image02 displays incorrect statistics. Based on the ICD message flow, it seems the frontend does not set the requirement properly in this case. Please refer to the video as a demo of the issue.

**UPDATE**: after few more tests, histogram widget has the same issue. Spatial profiler has the issue. Spectral profiler does not have the issue.

**To Reproduce**
1. download the test images at https://drive.google.com/file/d/1uv2f-quEVsf1r9Imt_SwNS6EFv1NvHmO/view?usp=sharing
2. load image01.fits and image02.fits and match them spatially
3. load the region file polygon.crtf
4. load two statistics widget
5. configure one to show image01.fits and the region explicitly and configure the other one to show image02.fits and the region explicitly too
6. delete the region
7. the NumPixel count should be ~10^8 but one of the widget did not update correctly and still shows the region stat.

**Expected behavior**
statistics requirement is set correctly after a region is deleted so that the two statistics can display correct full image statistics for image01 and image02 respectively.

**Screenshots or videos**

https://user-images.githubusercontent.com/20819712/168733122-8d4f684c-bdc7-46b7-9e18-26bb3683f6ff.mov

**Platform info (please complete the following information):**
 - OS [e.g. macOS Monterey]: macOS Monterey
 - Browser [e.g. chrome, safari, electron app]: electron app
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: v3b3
 - Frontend branch [e.g. dev, v3b2 release]: v3b3

YuHsuan-Hwang:
fixed via #1869

PR #1869, opened 2022-05-26
Thchang/1831 fixed no updating of spatial profilers after region deleting

### LINKED ISSUE #1831: spatial profile not updating after a region is deleted
**Describe the bug**
When a point region (should not be just due to point region I think) is selected and a spatial profiler is configured to display its profile, if we delete the point region, the spatial profiler does not update profile to the "active" region.

**To Reproduce**
1. load an image
2. configure the two spatial profiler widgets to display same profile (eg X direction)
3. create a point region
4. configure one of the spatial profiler widget so the the active image is explicitly selected in the image dropdown and the point region is explicitly selected in the region dropdown.
5. update position of the point region and we should see the two spatial profilers display the same profile as expected6. 
6. delete the point region and move cursor around7. 
7. now only one spatial profiler is updating spatial X profile. The other one (which we had explicit configuration before) does not update profiles anymore. 

**Expected behavior**
Profile should update based on the selections in the image and region dropdown, especially when the selection becomes "active".

**Screenshots or videos**

https://user-images.githubusercontent.com/20819712/163699796-4291a93a-a005-4eeb-84dc-ec8bda8e43c4.mov

 - OS: macOS Monterey
 - Browser: macOS electron app
 - Version carta v3b2

TienHao:
this PR closes #1831, #1855. 

The spatial profiler did not update profile from a non-cursor region(selected in the region dropdown) to the cursor region after the non-cursor region was deleted. This was caused by not removing deleted region id from the `regionIdMap` of `SpatialProfilerWidgetStore`, a subclass of `RegionWidgetStore`. I also found this problem in the PV generator for same reason.

changes:
- removing the deleted region id from the `regionIdMap` of all widget stores which extend `RegionWidgetStore`
- refactoring code of closing image and removing frame from `RegionWidgetStore`


kswang1029 (changes requested):
@TienHao Could you investigate the following case on why the image 1 with active region (ie cursor) does not update a profile after a region is deleted? Image 0 with active region (cursor) does update its profile.


https://user-images.githubusercontent.com/20819712/170918213-6068fd05-1a09-415f-80cf-eeacd2292ba8.mov

TienHao:
> @TienHao Could you investigate the following case on why the image 1 with active region (ie cursor) does not update a profile after a region is deleted? Image 0 with active region (cursor) does update its profile.

This is related to issue #1855. The regionMaps of `RegionWidgetStore` did not clear previous region id after image matching/unmatching or regions of spatial matching were deleted.

kswang1029 (approved):
tested manually with the spatial profiler, histogram, and statistics widgets. The behavior is now correct. 👍 e2e tests are all passing.

YuHsuan-Hwang (approved):

PR #1872, opened 2022-06-02
Thchang/1313 fixed stokes scatter plot indicator

### LINKED ISSUE #1313: stokes scatter plot point indicator not working
Feb-04-2021 10-38-53
only works if coordinate type is `Channel`

TienHao:
**Description**
This PR closes #1313: unable to switch channel by clicking scatter plot in stokes widgets when coordinate type is not `channel`.

change: fixed bugs caused by PR #744 and #827 in `onScatterChannelChanged`.

**Checklist**


kswang1029:
@TienHao confirmed this PR does not result in any regressions that the current e2e tests are capable to capture. Manual tests are on the way before approving the PR.

kswang1029 (changes requested):
Hmm the issue remains. Only it works only when the spectral convention is channel. The expected behavior is when we click a given point in the scatter plot, effectively we switch to that channel so we see the correct channel image and the red vertical lines in the profile plots are rendered at the correct channel. 


https://user-images.githubusercontent.com/20819712/172110658-017df89f-47e4-47e5-a300-4204ef986abd.mov

kswang1029 (approved):
confirmed the fix works. no regression is observed from the e2e tests.

YuHsuan-Hwang (approved):

veggiesaurus (approved):

PR #1876, opened 2022-06-06
Angus/splatalog service

veggiesaurus:
**Description**
Moves the splatalogue interaction from backend to frontend. Also fixes a bug with column alignment that was affecting a few lines of the response from the `"SLAIM"` line list.

**Checklist**
- [x] changelog updated
- [x] e2e test passing
- [x] no protobuf update needed


kswang1029 (approved):
no regression is found via the current set of e2e tests. Also done some manual tests to compare the queries from CARTA and from the Splatalogue website. No odd thing is observed so far.

src/stores/widgets/SpectralLineQueryWidgetStore.ts
```diff
@@ -558,15 +557,8 @@ export class SpectralLineQueryWidgetStore {
     };
 
     @action pingSplatalogue = async () => {
-        try {
-            this.splataloguePingStatus = SplataloguePingStatus.Checking;
```
YuHsuan-Hwang, line 562: only one minor question: do we keep this line for reseting status?

veggiesaurus: good point. I've added it back, but also switch from an `@action` to a `flow` (see here for [reasoning](https://mobx.js.org/actions.html#using-flow-instead-of-async--await-). We should be doing this throughout the codebase to avoid modifying observables outside of actions when using async functions).

veggiesaurus:
@kswang1029 please double-check this after the async -> flow change :innocent:

kswang1029 (approved):
e2e tests run fine. Manual tests appears fine too.

YuHsuan-Hwang (approved):

PR #1882, opened 2022-06-13
use sass instead of node-sass

veggiesaurus:
**Description**
Minor CI changes


Jenkinsfile
```diff
@@ -59,8 +59,6 @@ pipeline {
                             sh 'rm -rf node_modules build'
                             sh 'n 16'
                             sh 'n exec 16 node -v'
-                            sh 'n exec 16 npm uninstall node-sass --legacy-peer-deps' 
-                            sh 'n exec 16 npm install --save-dev --legacy-peer-deps sass'
```
ajm-ska, line 63: It needs a `n exec 16 npm install` here. 
But I wonder if the blueprintjs problem would return?

ajm-ska (changes requested):

Jenkinsfile
ajm-ska: Or rather `sh 'n exec 16 npm install'`

veggiesaurus: `n exec 16 npm install --legacy-peer-deps` should do the trick

PR #1883, opened 2022-06-14
Jrhosk/1837 optional spectral profiler info

### LINKED ISSUE #1837: provide an optional second spectral conversion to the cursor info in the spectral profiler
given the time pressure of the v3 release, an alternative solution for the feature https://github.com/CARTAvis/carta-frontend/issues/1703 (see implementation challenges in https://github.com/CARTAvis/carta-frontend/pull/1835) is to just add an additional spectral value based on the user selection to the cursor info in the spectral profiler.

Screen Shot 2022-04-26 at 15 55 32

In the above example, at the moment we show (frequency, intensity). If it is enabled from the settings dialog (default off) and velocity is selected, then we show (frequency, velocity, intensity) instead. 

The reference image for freq-velo conversion should be the “active” or user-selected image. This is relevant when multi-profile plot mode is enabled. 

quick mockup
Screen Shot 2022-04-26 at 16 44 23

Attached is a doc I compiled from AST user manual regarding spectral and spatial conversion functions. Hope it helps.
[carta_ast_specConvert_wcConvert.pdf](https://github.com/CARTAvis/carta-frontend/files/8561046/carta_ast_specConvert_wcConvert.pdf)

jrhosk:
**Description**
Implemented optional cursor information for spectral profile. Optional display can be toggled via spectral profile settings.

(updates by Yu-Hsuan)
Closes #1837. Please note that currently we do not support secondary unit for the smoothed data. This needs extra calculation, and I prefer we file a new issue if needed.
Changes:
* Added `channelSecondaryValues`, `spectralTypeSecondary`, and `spectralUnitSecondary` for the converted x data of the spectral profile.
* Added toggle and dropdown in the spectral profiler settings to enable the feature.
* Implemented `toFormattedNotationByDiff` to calculate the required digits of the cursor info (see https://github.com/CARTAvis/carta-frontend/pull/1883#issuecomment-1171323434 for the details).

Minor change:
* Renamed `startEndIndexes` to `dataIndexes` in `SpectralProfileWidgetStore`

**Checklist**

- [X] ZenHub issue connection, board status, and estimate updated

For the pull request:
- [X] ZenHub estimate, milestone, and release (if needed) added
- [X] e2e test passing / ~corresponding fix added~


jrhosk:
@YuHsuan-Hwang I made some change according to our slack discussion. I think this is ready for testing.

kswang1029:
@jrhosk I had a quick look and found an issue about channel index. If we set the main axis as channel and the 2nd axis as velocity, when we zoom in a profile we see the displayed channel index for a specific channel is changing with different zoom levels. It appears that the channel indices count the channels in the field of view only. The channel index per channel should be invariant at different profile zoom levels. 


https://user-images.githubusercontent.com/20819712/176071279-e7233be7-f6b5-4b0f-a084-978a0b8f4a75.mov

jrhosk:
> @jrhosk I had a quick look and found an issue about channel index. If we set the main axis as channel and the 2nd axis as velocity, when we zoom in a profile we see the displayed channel index for a specific channel is changing with different zoom levels. It appears that the channel indices count the channels in the field of view only. The channel index per channel should be invariant at different profile zoom levels.
> 
>  Screen.Recording.2022-06-28.at.09.47.51.mov

Hi @kswang1029 I fixed this. I am still testing a bit on my side but I think it is about ready to switch to full testing.

jrhosk:
Yu asked me to include an explanation of the difference between `formattedNoation` and `toFormattedNotation`. The main difference is that the former should provide dynamic precision versus the prior which has a few levels of exponential precision.
I added the new function because I found that in some cases the cursor rounding was such that there was no change in the displayed values between ticks, i.e. the cursor shows 1.3 when the ticks are varying as 1.3x. The new function bases the precision on the delta between the nearest tick values by iterating through the decimal value and comparing each decimal place to determine the minimum precision needed. It also trims additional trailing zeros. This makes the displayed values a bit longer but beyond that seems to work in initial testing. I look forward to seeing what others think.

kswang1029:
@jrhosk I have some manual tests and have run e2e test. The e2e tests show some test failures due to the change of the displayed number of digits which are understandable based on what you have described. For this, can we force to display two extra digits based on what you have derived? This is to avoid confusions of rounding error of adjacent channels. In some cases, for example, we see the step is 0.1 but some times 0.2 due to rounding differences.

Other issues spotted:
1. the displayed value of the first channel is not processed to display proper number of digits. Only the 2nd channel and so on, including the last channel are processed. 

https://user-images.githubusercontent.com/20819712/178406813-c5f1c2f7-3d56-4a0d-a80b-c19dce44a446.mov

2. when the secondary cursor info is enabled, we see an extra "channel" string to the 2nd cursor value.
Screen Shot 2022-07-12 at 12 13 58

src/stores/Frame/FrameStore.ts
```diff
@@ -914,6 +936,11 @@ export class FrameStore {
         }
     }
 
+    // return spectralFrame
```
acdo2002: Could you explain this public returnSpectralFrame()? Because I don't understand why do we need this?

acdo2002 (commented):
I have the same comment to Kuo-Song's second comment.
It quite confuse me the secondary Info shows "channel", I suggest change the label to "Secondary Info" or show the current "SpectralType (e.g. VRAD, VOPT...etc)". Either way would be more clear to me.

src/stores/Frame/FrameStore.ts
jrhosk: Good catch, it is not needed anymore. I just missed it after refactoring recently. It is now removed.

jrhosk:
> I have the same comment to Kuo-Song's second comment. It quite confuse me the secondary Info shows "channel", I suggest change the label to "Secondary Info" or show the current "SpectralType (e.g. VRAD, VOPT...etc)". Either way would be more clear to me.

This wasn't a feature, it was a bug. After consolidating the output string this got lost in the logic. It should be fixed now.

jrhosk:
> @jrhosk I have some manual tests and have run e2e test. The e2e tests show some test failures due to the change of the displayed number of digits which are understandable based on what you have described. For this, can we force to display two extra digits based on what you have derived? This is to avoid confusions of rounding error of adjacent channels. In some cases, for example, we see the step is 0.1 but some times 0.2 due to rounding differences.
> 
> Other issues spotted:
> 
> 1. the displayed value of the first channel is not processed to display proper number of digits. Only the 2nd channel and so on, including the last channel are processed.
> 
>  Screen.Recording.2022-07-12.at.12.06.06.mov 
> 2. when the secondary cursor info is enabled, we see an extra "channel" string to the 2nd cursor value.
>    Screen Shot 2022-07-12 at 12 13 58

@kswang1029 I am pushing an updated version now that I think fixes these issues. While I don't understand your e2e issues, I added two decimal places to the precision and hopefully that will fix your issue.

kswang1029:
@jrhosk thanks for addressing the issues found in my previous tests. Just spotted one more:
In some cases, the wavelength in mm as the 2nd info does not have sufficient digits to distinguish adjacent channels. 

https://user-images.githubusercontent.com/20819712/178889441-7ebe3050-306e-4706-a087-9687000c81ea.mov

src/components/Shared/SpectralSettings/SpectralSettingsComponent.tsx
```diff
@@ -39,6 +42,17 @@ export class SpectralSettingsComponent extends React.Component<{
                         onChange={(event: React.FormEvent<HTMLSelectElement>) => this.props.onSpectralCoordinateChange(event.currentTarget.value as string)}
                     />
                 </FormGroup>
+                {this.props.secondaryAxisCursorInfoVisible && (
+                    <FormGroup label={"Opt. Coord."} inline={true} disabled={disableCoordinateSetting}>
```
YuHsuan-Hwang: I personally prefer we use "Secondary Coordinate" for label. We can increase the label width if necessary.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -35,6 +36,8 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
         };
     }
 
+    private profileIndex: number = 0;
```
YuHsuan-Hwang: Please remove unused variable.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -152,21 +155,59 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
         this.widgetStore.setCursor(x);
     }, 33);
 
-    private genCursoInfoString = (data: Point2D[], cursorXValue: number, cursorXUnit: string, label: string): string => {
-        let cursorInfoString = undefined;
+    private precisionFormatting = (nearest: {point: Point2D; index: number}, data: number, diff: number, spectralType: SpectralType): string => {
```
YuHsuan-Hwang: Looks like we can remove `nearest` in this function.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -152,21 +155,59 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
...
+        if (spectralType === SpectralType.CHANNEL) {
```
YuHsuan-Hwang: can be simplified to `return spectralType === SpectralType.CHANNEL ?? toFixed(data) : toFormattedNotation(data, diff);`

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -152,21 +155,59 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
...
+
+            // We calculate the difference between neighboring values to get and estimate of the precision.
+            diffLeft = nearest.index - 1 >= 0 ? Math.abs(data[nearest.index].x - data[nearest.index - 1].x) : Math.abs(data[nearest.index + 1].x - data[nearest.index].x);
+
+            // Use precision to determine the proper rounding and zero suppression for displayed value. Data and secondary
+            // are handled idfferently because they have different structures.
+            primaryXStr = this.precisionFormatting(nearest, data[nearest.index].x, diffLeft, frame.spectralType);
+
+            if (this.widgetStore.secondaryAxisCursorInfoVisible) {
+                let secondary = this.widgetStore.effectiveFrame.channelSecondaryValues.slice(start, end);
```
YuHsuan-Hwang: can be simplified to `frame`

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -152,21 +155,59 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
...
+
+                // Use precision to determine the proper rounding and zero suppression for displayed value. Data and secondary
+                // are handled idfferently because they have different structures.
+                diffLeft = nearest.index - 1 >= 0 ? Math.abs(data[nearest.index].x - data[nearest.index - 1].x) : Math.abs(data[nearest.index + 1].x - data[nearest.index].x);
```
YuHsuan-Hwang: shall `data` here be `secondary` instead?

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -152,21 +155,59 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
...
+            return toFormattedNotation(data, diff);
+        }
+    };
+
+    private genCursoInfoString = (data: Point2D[], cursorXValue: number, cursorXUnit: string, label: string, start: number, end: number): string => {
+        const frame = this.widgetStore.effectiveFrame;
+
+        let diffLeft: number = undefined;
+        let secondaryXUnit = "";
+        let channelString = "";
```
YuHsuan-Hwang: maybe rename to `secondaryChannelString`

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -152,21 +155,59 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
...
+                // are handled idfferently because they have different structures.
+                diffLeft = nearest.index - 1 >= 0 ? Math.abs(data[nearest.index].x - data[nearest.index - 1].x) : Math.abs(data[nearest.index + 1].x - data[nearest.index].x);
+
+                // Use precision to determine the proper rounding and zero suppression for displayed value.
+                const secondaryXStr = this.precisionFormatting(nearest, secondary[nearest.index], diffLeft, frame.spectralTypeSecondary);
+
+                if (frame.spectralTypeSecondary !== SpectralType.CHANNEL) secondaryXUnit = frame.spectralUnitSecondary;
+                else channelString = "Channel ";
+
+                const xLabel =
```
YuHsuan-Hwang: Maybe we can apply `let xLabel = primary string` before if block and then `xLabel += secondary string` in the if block to avoid duplicate code.

src/components/SpectralProfiler/SpectralProfilerSettingsPanelComponent/SpectralProfilerSettingsPanelComponent.tsx
```diff
@@ -72,6 +77,14 @@ export class SpectralProfilerSettingsPanelComponent extends React.Component<Widg
         });
     }
 
+    @action setCursorInfo(state: boolean) {
```
YuHsuan-Hwang: remove unused functions `setCursorInfo` and `handleSecondaryAxisCursorInfoChanged`

src/components/SpectralProfiler/SpectralProfilerSettingsPanelComponent/SpectralProfilerSettingsPanelComponent.tsx
```diff
@@ -22,6 +22,8 @@ export enum SpectralProfilerSettingsTabs {
 
 @observer
 export class SpectralProfilerSettingsPanelComponent extends React.Component<WidgetProps> {
+    secondaryAxisCursorInfoVisible?: boolean;
```
YuHsuan-Hwang: remove unused variable

src/stores/Frame/FrameStore.ts
```diff
@@ -698,6 +711,15 @@ export class FrameStore {
         return label;
     }
 
+    @computed get secondarySpectralLabel(): string {
```
YuHsuan-Hwang: remove unused variable

src/stores/Frame/FrameStore.ts
```diff
@@ -110,6 +110,7 @@ export class FrameStore {
     public readonly colorbarStore: ColorbarStore;
 
     public spectralCoordsSupported: Map<string, {type: SpectralType; unit: SpectralUnit}>;
+    public secondarySpectralCoordsSupported: Map<string, {type: SpectralType; unit: SpectralUnit}>;
```
YuHsuan-Hwang: perhaps we can remove `secondarySpectralCoordsSupported` and use `spectralCoordsSupported` for both

src/utilities/units.ts
```diff
@@ -43,6 +43,88 @@ export function toFixed(val: number, decimals: number = 0): string {
     return String(val);
 }
 
+export function trimTrailingDecimals(value: string): string {
+    var splitValue: string[] = value.split(".");
```
YuHsuan-Hwang: Please use `let` instead of `var`.

YuHsuan-Hwang (changes requested):
I listed some suggestions. Most of them should be minor.

src/components/SpectralProfiler/SpectralProfilerSettingsPanelComponent/SpectralProfilerSettingsPanelComponent.tsx
jrhosk: This isn't unused. It is used in 6 places in that file.

YuHsuan-Hwang: Those are either for initialization, or in unused functions, or `widgetStore.secondaryAxisCursorInfoVisible`.

jrhosk:
I think I have fixed all the issues listed and added Yu's format changes. A quick tests shows things working.

src/utilities/units.ts
```diff
@@ -43,6 +43,88 @@ export function toFixed(val: number, decimals: number = 0): string {
...
+    return decimals[0];
+}
+
+export function getVariablePrecision(value: number): number {
+    // Estimates the precision of input tick value. Input provides
+    // delta between neighboring tick values and iterates through
+    // up to 14 decimal places to determine the approxmiate
+    // precision.
+
+    var decimalPlacement = 0.1;
```
YuHsuan-Hwang: Shall `decimalPlacement` be 0.001 because `precision` is 3?
Please use `let/const` instead of `var` for all the variables in this file.

src/utilities/units.ts
```diff
@@ -43,6 +43,88 @@ export function toFixed(val: number, decimals: number = 0): string {
...
+    if (value === null || isNaN(value)) {
+        return null;
+    }
+    // Determine approximate precision
+    var precision = getVariablePrecision(Math.abs(delta));
+    var frontDecimalLength = value.toString().split(".")[0].length;
+    var totalPrecision = frontDecimalLength + precision;
+
+    // Trim trailing zeros
+    var trimmedValue = trimTrailingZeros(value.toPrecision(totalPrecision));
```
YuHsuan-Hwang: Maybe we can use `toFix` instead of `toPrecision`, so we don't need to calculate `frontDecimalLength`.

src/utilities/units.ts
```diff
@@ -43,6 +43,88 @@ export function toFixed(val: number, decimals: number = 0): string {
...
+
+    return value < 1 ? trimTrailingDecimals(trimmedValue) : trimTrailingDecimals(trimmedValue);
```
YuHsuan-Hwang: Can be simplify to `return trimTrailingDecimals(trimmedValue);`.
Perhaps we don't need to apply `trimTrailingDecimals` here because we have removed the zeros in `trimTrailingZeros`?

src/components/SpectralProfiler/SpectralProfilerSettingsPanelComponent/SpectralProfilerSettingsPanelComponent.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from "react";
-import {computed, autorun} from "mobx";
+import {computed, autorun, /*observable,*/ makeObservable} from "mobx";
```
YuHsuan-Hwang: remove comment code.

src/components/SpectralProfiler/SpectralProfilerSettingsPanelComponent/SpectralProfilerSettingsPanelComponent.tsx
```diff
@@ -58,6 +58,8 @@ export class SpectralProfilerSettingsPanelComponent extends React.Component<Widg
 
     constructor(props: WidgetProps) {
         super(props);
+        makeObservable(this);
+
```
YuHsuan-Hwang: Do we also remove `makeObservable`?

src/utilities/units.ts
```diff
@@ -43,6 +43,88 @@ export function toFixed(val: number, decimals: number = 0): string {
...
+            precision++;
+        } else {
+            return precision;
+        }
+    }
+
+    return precision;
+}
+
+export function toFormattedNotation(value: number, delta: number): string {
```
YuHsuan-Hwang: We need a better naming to distinguish `toFormattedNotation` and `formattedNotation`. Perhaps `formattedNotationByDiff`?

YuHsuan-Hwang (changes requested):
The main changes look good to me 👍
I just have some questions related to formatting. Seems that after changing from  `toFormattedNotation` to `formattedNotation`, we don't use scientific notation. Not sure if this affects user experience.

src/utilities/units.ts
jrhosk: > Maybe we can use `toFix` instead of `toPrecision`, so we don't need to calculate `frontDecimalLength`.

I will need to look at it again but there was a reason I did this. If I recall, using toFixed() screws up the channel values formatting.

src/utilities/units.ts
jrhosk: > Shall `decimalPlacement` be 0.001 because `precision` is 3? Please use `let/const` instead of `var` for all the variables in this file.

Yes, I think I changed this after trying to fix another error that was quoted that came about after some of the cosmetic changes you requested. I'll fix it.

src/utilities/units.ts
jrhosk: I would like to suggest a best practices document for these things. If there was a documented naming convention it would have saved half the changes requests. The purpose of naming it `toFormattedNotation` was to match the standard `toFixed` format, which I think is actually better than `formattedNotation`. I had intended to remove the later if my function was ok'd.

I'll make all these changes on Monday when I am back in the office.

kswang1029:
We have a set of e2e tests for the cursor info check. There are some test failures due to less significant digits.
The test would expect
`Cursor: (86.748221 GHz, 1.69e-2)
`but it sees
`Cursor: (86.7482 GHz, 1.69e-2)
`
The channel width from the image header is
`CDELT3 = -2.442377011414E+05 Hz
`
Or the test would expect
`Cursor: (-2.972 km/s, -5.99e-4)
`but it sees
`Cursor: (-3 km/s, -5.99e-4)
`
The channel width from the image header is
`CDELT3 = -2.441261094971E+05 Hz
`
The rest frequency is
`RESTFRQ = 2.203986800000E+11 Hz
`

I wonder if we can add additional digits (3) on top of the current approach? The reason is there are two groups of radio astronomers who are used to MHz or GHz, respectively. When it is MHz, usually we display the value with 3 or 4 decimal digits. So when we display in GHz, we need more digits (the above example). We cannot use fixed number of decimal digits as in some cases, the channel width can be very small so it is possible that adjacent channels would result in identical displayed value. So the solution I am proposing is to add additional 3 digits on top of the current approach.

Note there is also one e2e test failure about region matching. Merging dev to this branch should make the test passing.

jrhosk:
> We have a set of e2e tests for the cursor info check. There are some test failures due to less significant digits. The test would expect `Cursor: (86.748221 GHz, 1.69e-2) `but it sees `Cursor: (86.7482 GHz, 1.69e-2) ` The channel width from the image header is `CDELT3 = -2.442377011414E+05 Hz ` Or the test would expect `Cursor: (-2.972 km/s, -5.99e-4) `but it sees `Cursor: (-3 km/s, -5.99e-4) ` The channel width from the image header is `CDELT3 = -2.441261094971E+05 Hz ` The rest frequency is `RESTFRQ = 2.203986800000E+11 Hz `
> 
> I wonder if we can add additional digits (3) on top of the current approach? The reason is there are two groups of radio astronomers who are used to MHz or GHz, respectively. When it is MHz, usually we display the value with 3 or 4 decimal digits. So when we display in GHz, we need more digits (the above example). We cannot use fixed number of decimal digits as in some cases, the channel width can be very small so it is possible that adjacent channels would result in identical displayed value. So the solution I am proposing is to add additional 3 digits on top of the current approach.
> 
> Note there is also one e2e test failure about region matching. Merging dev to this branch should make the test passing.

Hi Kusong, that is unfortunate, this was working fine before. It is probably due to the recently requested changes I made. I will take a look at them tomorrow, revert some of the changes and hopefully get it working for you again. Is there a way I can see that the e2e tests are failing? That way I can only ping you after they are all passing.

kswang1029:
e2e tests are currently run manually for a fair of frontend+backend commits. They are run automatically on dev+dev. 

Please refer to the two images adopted in the e2e tests for reproducing the issues. (M17 for frequency, HD163296 for velocity)
https://drive.google.com/file/d/1PQRptRzYEWvgN081mIMX-ZytXJ68gV6X/view?usp=sharing
https://drive.google.com/file/d/1GBkof53vxeJb4GfELAj1_uIwJzOyxJqf/view?usp=sharing

jrhosk:
> e2e tests are currently run manually for a fair of frontend+backend commits. They are run automatically on dev+dev.
> 
> Please refer to the two images adopted in the e2e tests for reproducing the issues. (M17 for frequency, HD163296 for velocity) https://drive.google.com/file/d/1PQRptRzYEWvgN081mIMX-ZytXJ68gV6X/view?usp=sharing https://drive.google.com/file/d/1GBkof53vxeJb4GfELAj1_uIwJzOyxJqf/view?usp=sharing

Thanks Kusong, I'll work this out this week.

jrhosk:
> > e2e tests are currently run manually for a fair of frontend+backend commits. They are run automatically on dev+dev.
> > Please refer to the two images adopted in the e2e tests for reproducing the issues. (M17 for frequency, HD163296 for velocity) https://drive.google.com/file/d/1PQRptRzYEWvgN081mIMX-ZytXJ68gV6X/view?usp=sharing https://drive.google.com/file/d/1GBkof53vxeJb4GfELAj1_uIwJzOyxJqf/view?usp=sharing
> 
> Thanks Kusong, I'll work this out this week.

Alright, hopefully it will pass now. I didn't see any issues using the files you posted.

src/utilities/units.ts
```diff
@@ -96,7 +96,7 @@ export function getVariablePrecision(value: number): number {
     // precision.
 
     let decimalPlacement = 0.1;
-    let precision = 1;
+    let precision = 3;
```
YuHsuan-Hwang: @jrhosk I would prefer we add a comment here, explaining that the precision here is 3 instead of 1 because we add extra 2 digits for better UX.
Could you also update this branch with the latest dev, and add an item to `Added` in CHANGELOG.md?

kswang1029:
I spotted a frontend crash.
Screen Shot 2022-08-05 at 08 30 30

To reproduce:
1) load an image
2) make a region and enable a spectral profiler to see the spectral profile
3) without changing anything (not even enable the 2nd spec info), try to zoom the profile all the way in
4) frontend will crash

jrhosk:
> I spotted a frontend crash. Screen Shot 2022-08-05 at 08 30 30
> 
> To reproduce:
> 
> 1. load an image
> 2. make a region and enable a spectral profiler to see the spectral profile
> 3. without changing anything (not even enable the 2nd spec info), try to zoom the profile all the way in
> 4. frontend will crash

@kswang1029 I think I fixed this problem. The issue is that the precision is based on the difference between two consecutive tick points, when zooming in to the point where only one point is still shown, there is no "difference" to be calculated. The fix I implemented just checks the data length being displayed and when there is only one point,  it defaults to a precision of 1e-9. A better fix might be to limit the zoom to two points but that might be a bigger fix. Not sure what the use case is for zooming in to only 1 or even 0 points, which you can do. That would be my preferred fix.

kswang1029:
> > I spotted a frontend crash. Screen Shot 2022-08-05 at 08 30 30
> > To reproduce:
> > 
> > 1. load an image
> > 2. make a region and enable a spectral profiler to see the spectral profile
> > 3. without changing anything (not even enable the 2nd spec info), try to zoom the profile all the way in
> > 4. frontend will crash
> 
> @kswang1029 I think I fixed this problem. The issue is that the precision is based on the difference between two consecutive tick points, when zooming in to the point where only one point is still shown, there is no "difference" to be calculated. The fix I implemented just checks the data length being displayed and when there is only one point, it defaults to a precision of 1e-9. A better fix might be to limit the zoom to two points but that might be a bigger fix. Not sure what the use case is for zooming in to only 1 or even 0 points, which you can do. That would be my preferred fix.

The spectral profiler allows multi-profile plotting. Sometimes we compare profiles with quite different channel widths so it is possible and necessary to zoom profiles at such high zoom levels.

jrhosk:
> > > I spotted a frontend crash. Screen Shot 2022-08-05 at 08 30 30
> > > To reproduce:
> > > 
> > > 1. load an image
> > > 2. make a region and enable a spectral profiler to see the spectral profile
> > > 3. without changing anything (not even enable the 2nd spec info), try to zoom the profile all the way in
> > > 4. frontend will crash
> > 
> > 
> > @kswang1029 I think I fixed this problem. The issue is that the precision is based on the difference between two consecutive tick points, when zooming in to the point where only one point is still shown, there is no "difference" to be calculated. The fix I implemented just checks the data length being displayed and when there is only one point, it defaults to a precision of 1e-9. A better fix might be to limit the zoom to two points but that might be a bigger fix. Not sure what the use case is for zooming in to only 1 or even 0 points, which you can do. That would be my preferred fix.
> 
> The spectral profiler allows multi-profile plotting. Sometimes we compare profiles with quite different channel widths so it is possible and necessary to zoom profiles at such high zoom levels.

That explains it, good to know.

kswang1029:
@jrhosk when there are multiple profiles from matched cubes, we see the 2nd spectral info is Null. In the video I am setting the three matched images as active one by one. 

https://user-images.githubusercontent.com/20819712/183230441-5ab73435-7c23-4f8e-afbc-48bad025c7ab.mov

test images are attached 
https://drive.google.com/file/d/1xrvn0Xss2h_ns-zQh2PPBWD-Bo1Ztdk4/view?usp=sharing

jrhosk:
> @jrhosk when there are multiple profiles from matched cubes, we see the 2nd spectral info is Null. In the video I am setting the three matched images as active one by one.
> 
>  Screen.Recording.2022-08-06.at.10.40.53.mov 
> test images are attached https://drive.google.com/file/d/1xrvn0Xss2h_ns-zQh2PPBWD-Bo1Ztdk4/view?usp=sharing

Hi @kswang1029 I can't reproduce this error. I opened all three files you sent (one one and append the other two) then I made a region with the rect selection box and used the wcs matching. I opened the spectral profile window and selected _image_ to get all three profiles showing and added the second coordinate. When zooming in and scrolling through the profile as you did, I don't see and discontinuity.  I don't see a null value anywhere on the profile. Could you post a detailed, step-by-step on how to reproduce this issue please?

kswang1029:
@jrhosk please try the following steps:
1. load HD163296_CO_2_1.fits, HD163296_13CO_2-1.fits and HD163296_C18O_2-1.fits in the sequence
2. match them
3. make a region to see a region spectral profile 
4. enable "Image" multi-profile plot and we will see three profiles
5. zoom to about -16km/s to +2km/s
6. enable the 2nd spectral info and set it to frequency
7. set HD163296_CO_2_1.fits as the active image and check cursor info (this appears fine)
8. set HD163296_13CO_2-1.fits as the active image and we see null (esp. ~0:13 in the video where at the cursor location there are three profiles)
9. set HD163296_C18O_2-1.fits as the active image and we see null


https://user-images.githubusercontent.com/20819712/188762185-e67a2d38-c4a8-4bd7-8462-d731ab71a1c2.mov

src/components/Shared/SpectralSettings/SpectralSettingsComponent.tsx
```diff
@@ -39,6 +42,17 @@ export class SpectralSettingsComponent extends React.Component<{
                         onChange={(event: React.FormEvent<HTMLSelectElement>) => this.props.onSpectralCoordinateChange(event.currentTarget.value as string)}
                     />
                 </FormGroup>
+                {this.props.secondaryAxisCursorInfoVisible && (
+                    <FormGroup label={"Secondary Coordinate"} inline={true} disabled={disableCoordinateSetting}>
+                        <HTMLSelect
+                            disabled={disableCoordinateSetting}
+                            value={frame && frame.spectralCoordinateSecondary ? frame.spectralCoordinateSecondary : ""}
+                            options={spectralCoordinateOptions}
+                            onChange={(event: React.FormEvent<HTMLSelectElement>) => this.props.onSpectralCoordinateChangeSecondary(event.currentTarget.value as string)}
```
veggiesaurus: can use implicit typing here (`{event=>this.props.onSpectral...}`)

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -152,22 +153,50 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
...
-            if (diffLeft > 0 && diffLeft < 1e-6) {
-                floatXStr = formattedNotation(nearest.point.x);
-            } else if (diffLeft >= 1e-6 && diffLeft < 1e-3) {
-                floatXStr = toFixed(nearest.point.x, 6);
-            } else {
-                floatXStr = toFixed(nearest.point.x, 3);
+            let primaryXStr: string = "";
+
+            // We calculate the difference between neighboring values to get and estimate of the precision.
+            diffLeft = data.length === 1 ? 1e-9 : nearest.index > 0 ? Math.abs(data[nearest.index].x - data[nearest.index - 1].x) : Math.abs(data[nearest.index + 1].x - data[nearest.index].x);
```
veggiesaurus: I think this would be more readable if you assigned variables to the data points themselves (e.g `const currentData = data[nearest.index]`), but just a suggestion :innocent:

src/stores/widgets/SpectralProfileWidget/SpectralProfileWidgetStore.ts
```diff
@@ -34,6 +34,7 @@ export type MultiPlotData = {
     yMax: number;
     yMean: number;
     yRms: number;
+    startEndIndexes: {startIndex: number; endIndex: number}[];
```
veggiesaurus: this is a very confusing variable name :thinking:

src/utilities/units.ts
```diff
@@ -43,6 +43,84 @@ export function toFixed(val: number, decimals: number = 0): string {
...
+            if (splitValue[1][i] !== "0") {
+                return value;
+            }
+        }
+        return splitValue[0];
+    }
+}
+
+export function trimTrailingZeros(value: string): string {
+    // Trims the trailing zeros from the input decimal value. If
```
veggiesaurus: this seem like something that could be done quite easily with a regex replace using something like `/\d?\.\d+?(0+)/gm` (@confluence will have a better solution :innocent: )

veggiesaurus (changes requested):

src/components/Shared/SpectralSettings/SpectralSettingsComponent.tsx
jrhosk: Would you want all the similar lines to be this way as well? All I did was copy the style of whoever wrote the rest of the code.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
jrhosk: Well yes but what about `data[nearest.index + 1]` and `data[nearest.index - 1]`?  Would you want a new name for every index used?

src/stores/widgets/SpectralProfileWidget/SpectralProfileWidgetStore.ts
jrhosk: I agree and I am not sure how ti came to be honestly. Maybe a find-replace error on my part. I changed it to `dataIndexes`

src/components/Shared/SpectralSettings/SpectralSettingsComponent.tsx
YuHsuan-Hwang: applied implicit typing

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
YuHsuan-Hwang: applied `currentIndex` and `neighborIndex` for better readability

src/utilities/units.ts
YuHsuan-Hwang: applied regex replace in `trimTrailingDecimals` (`trimTrailingZeros` is unused and is removed)

YuHsuan-Hwang:
The information of this PR is updated in the first comment.

New applied changes:
* moved data processing from component to store
* renamed `secondary`/`secondaryData` to `secondaryXData`
* resolved conflict with the smoothed data cursor info feature. Please note that currently we do not support secondary unit for the smoothed data. This needs extra calculation, and I prefer we file a new issue if needed.
* adjustments according to the comments: applied implicit typing in `SpectralSettingsComponent`; `currentIndex` and `neighborIndex` in `SpectralProfilerComponent` for better readability; regex replace in `trimTrailingDecimals`

YuHsuan-Hwang (approved):

YuHsuan-Hwang:
@kswang1029 This is ready for testing, thanks.

YuHsuan-Hwang:
fixing formatting issue observed by @kswang1029

acdo2002 (approved):
I have tested that opening the four images and set four different combinations of first/second coordinates. So far the new function (the first and the second coordinates) works fine and smooth.

YuHsuan-Hwang:
new changes:
* fixed incorrect regex replace for removing trailing zeros
* fixed incorrect slicing; now null will not be shown at the end of data
* fixed conversion when primary unit is native unit
* adjusted the width of column in the setting panel

kswang1029 (approved):
@YuHsuan-Hwang this now works nicely! 👍  No regression is observed via current e2e tests.

acdo2002 (approved):

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -152,25 +152,40 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
         this.widgetStore.setCursor(x);
     }, 33);
 
-    private genCursoInfoString = (data: Point2D[], smoothedData: Point2D[], cursorXValue: number, cursorXUnit: string, label: string): string => {
-        let cursorInfoString = undefined;
+    private precisionFormatting = (data: number, diff: number, spectralType: SpectralType): string => {
+        return spectralType === SpectralType.CHANNEL ? toFixed(data) : toFormattedNotationByDiff(data, diff);
+    };
+
+    private genCursoInfoString = (data: Point2D[], smoothedData: Point2D[], secondaryXData: number[], cursorXValue: number, cursorXUnit: string, label: string): string => {
```
crocka: Is this a typo? Is it genCursorInfoString?

kswang1029: Eagle eye 🦅

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
```diff
@@ -152,25 +152,40 @@ export class SpectralProfilerComponent extends React.Component<WidgetProps> {
...
+            smoothedFloatXStr = this.precisionFormatting(nearestSmooth?.point.x, diffLeft, frame.spectralType);
+
+            let xLabel = cursorXUnit === "Channel" ? `Channel ${primaryXStr}` : `${primaryXStr}${cursorXUnit ? ` ${cursorXUnit}` : ""}`;
+
+            if (this.widgetStore.secondaryAxisCursorInfoVisible) {
+                diffLeft = data.length === 1 ? 1e-9 : Math.abs(secondaryXData[currentIndex] - secondaryXData[neighborIndex]);
+                const secondaryXStr = this.precisionFormatting(secondaryXData[currentIndex], diffLeft, frame.spectralTypeSecondary);
+
+                if (frame.spectralTypeSecondary !== SpectralType.CHANNEL) secondaryXUnit = frame.spectralUnitSecondary;
+                else secondaryChannelString = "Channel";
```
crocka: Angus has told me in my previous PR that he wants to always have brackets for if-statements.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
YuHsuan-Hwang: good catch! Typo fixed.

src/components/SpectralProfiler/SpectralProfilerComponent.tsx
YuHsuan-Hwang: fixed

crocka (approved):

veggiesaurus (approved):

PR #1890, opened 2022-06-22
disable splash screen when showing an alert

### LINKED ISSUE #1808: Bad behavior when coming back from sleep
I'm using the standalone app v3.0.0-beta.1b on my mac laptop.  When I put the computer to sleep (close it), it appears to clear the GPU cache.  CARTA then goes into a waiting mode where it asks whether to reload the session.  As long as it's in this waiting mode, it uses ~50% CPU according to activity monitor, using up battery & causing the fans to go strong - i.e., it's doing a lot of processing for nothing.  As soon as I hit 'Retry', the problem clears up.  It would be helpful to avoid this behavior.

veggiesaurus:
Closes #1808 (Hopefully). Can we see if this solves the issue? I simply hide the splash screen when an alert is visible. This hides the spinner, which seems to improve CPU usage

### DIFF
--- a/src/components/SplashScreen/SplashScreenComponent.tsx
+++ b/src/components/SplashScreen/SplashScreenComponent.tsx
@@ -13,7 +13,7 @@ export class SplashScreenComponent extends React.Component {
         const className = classNames("splash-screen", {"bp3-dark": appStore.darkTheme});
 
         return (
-            <Overlay className={Classes.OVERLAY_SCROLL_CONTAINER} autoFocus={false} canEscapeKeyClose={false} canOutsideClickClose={false} isOpen={appStore.splashScreenVisible} usePortal={true}>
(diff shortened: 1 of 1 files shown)


kswang1029:
given the issue described in https://github.com/CARTAvis/carta/issues/153 it is hard to tell if this improves the situation unfortunately, unless @ajm-asiaa can help on building it with an older electron wrapper (chrome 96?) to see the effect.

veggiesaurus:
> given the issue described in [CARTAvis/carta#153](https://github.com/CARTAvis/carta/issues/153) it is hard to tell if this improves the situation unfortunately, unless @ajm-asiaa can help on building it with an older electron wrapper (chrome 96?) to see the effect.

one way to test this is to simply kill the backend process and see if the CPU usage of the frontend is reduced in comparison to `dev` after a while

kswang1029:
The test is done with Chrome Version 103.0.5060.53 (Official Build) (x86_64)
It appears that there is a little improvement in the cpu usage.

this branch
Screen Shot 2022-06-23 at 11 52 31

dev
Screen Shot 2022-06-23 at 11 56 20

Issue #940, opened 2020-06-08
crowded indices of the Frame axis in the animator


kswang1029:
<img width="557" alt="Screen Shot 2020-06-08 at 4 37 52 PM" src="https://user-images.githubusercontent.com/20819712/84014639-3ca36900-a9ad-11ea-9abc-68749a1f403d.png">

YuHsuan-Hwang:
fixed via #1891.

PR #1891, opened 2022-06-22
make animator frame indices less crowded

### LINKED ISSUE #1892: misalignment of channel slider indices and the slider knot index
**Describe the bug**
When loading a cube with small and even channel count, we see the channel slider indices and the slider knot index are misaligned. 

Screen Shot 2022-06-23 at 11 25 45

test images: https://drive.google.com/file/d/1WIEAc2B3VmhT0AklkMN0zrdezPLo8azb/view?usp=sharing

**To Reproduce**
1. load M17_SWex_0_13.fits
2. click the next button in the animator
3. observe the slider indices and the knot index

**Expected behavior**
No offset is observed.

 - Browser version [e.g. 22]: 103
 - Backend branch [e.g. dev, v3b2 release]: dev, v3b3
 - Frontend branch [e.g. dev, v3b2 release]: dev, v3b3

I-Chenn:
**Description**
This PR is for issue #940 "**crowded indices of the Frame axis in the animator**". The original indices of Frame slider are like:
animator_image_is_crowded

The issue is fixed with the similar concept of Channel slider. The result is shown below:
animator_image_not_crowded

**Checklist**
Since I'm not quite aware of our standard procedure of making PR, thus I'm not quite sure that should I do the e2e test by myself before making pull request? As for changelog, should I do that after the PR review?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed issue with exporting decimated data instead of full resolution data in spatial profiler ([#1546](https://github.com/CARTAvis/carta-frontend/issues/1546))
 * Fixed larger position errors of projected contours, catalog overlays, and vector overlays near the border ([#1843](https://github.com/CARTAvis/carta-frontend/issues/1843)).
 * Fixed no updating of spatial profile after region deleting ([#1831](https://github.com/CARTAvis/carta-frontend/issues/1831), [#1855](https://github.com/CARTAvis/carta-frontend/issues/1855)).
+* Fixed issues of crowded Frame idices in the animator and misalignment of channel slider indices ([#940](https://github.com/CARTAvis/carta-frontend/issues/940)), ([#1892]https://github.com/CARTAvis/carta-frontend/issues/1892)
 
 ## [3.0.0-beta.3]
 
--- a/src/components/Animator/AnimatorComponent.tsx
+++ b/src/components/Animator/AnimatorComponent.tsx
@@ -216,7 +216,7 @@ export class AnimatorComponent extends React.Component<WidgetProps> {
         if (appStore.frames.length > 1) {
             const frameIndex = appStore.frames.findIndex(f => f.frameInfo.fileId === activeFrame.frameInfo.fileId);
             const numIndices = 5;
-            const frameStep = numFrames > 10 ? Math.floor( (numFrames - 1) / (numIndices - 1) ) : 1;
+            const frameStep = numFrames > 10 ? Math.floor((numFrames - 1) / (numIndices - 1)) : 1;
             frameSlider = (
                 <div className="animator-slider">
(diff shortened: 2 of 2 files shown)


kswang1029:
It is always @kswang1029 handling e2e test check item. You can leave it. 

Changelog should be part of PR and review process.

YuHsuan-Hwang (commented):
The changes look good 👍 Please also update the `CHANGELOG.md` file

kswang1029 (changes requested):
Sometimes the ending index is not displayed. Try to load 20+ images and close them one by one and observe the slider indices. 

Screen Shot 2022-06-23 at 11 13 45

kswang1029:
there is a bug in the current implementation of the channel slider (so adopting the same may have the same bug).

This is a cube with 14 channels and we see the displayed indices and the slider index do not match. I think this is a tricky problem of even/odd and equal partition... 
Screen Shot 2022-06-23 at 11 25 45

kswang1029 (approved):
given the issue described in #1892, the current implementation is fine.  No regression is observed via e2e tests.

I-Chenn:
The channel bug is due to this line:
channel
You can see that the channelStep defined in this way may be with decimal (not divisible during the calculation), and this may cause the slider not to locate on the exact position. However, since there's usually great numbers of channels, this bug doesn't "appear" too often. It "appears" only when there's less number of channels.

I somehow fixed this bug for "frame" slider with the adoption of **floor**:
frame
Nevertheless, this is probably the reason that causes the "ending index not displaying" issue. I will try come up with an idea to solve both issues in my next commit.

> there is a bug in the current implementation of the channel slider (so adopting the same may have the same bug).
> 
> This is a cube with 14 channels and we see the displayed indices and the slider index do not match. I think this is a tricky problem of even/odd and equal partition... Screen Shot 2022-06-23 at 11 25 45

kswang1029:
@I-Chenn I suggest we keep the PR in the current state and address the alignment issue in #1892 plus the missing last index  (if there is a good solution). @YuHsuan-Hwang would you agree?

YuHsuan-Hwang:
> @I-Chenn I suggest we keep the PR in the current state and address the alignment issue in #1892 plus the missing last index (if there is a good solution). @YuHsuan-Hwang would you agree?

Both ok with me.

I-Chenn:
@kswang1029 @YuHsuan-Hwang
Excuse me, I would like to double check again if I misunderstood what you mean during the weekly meeting.
I should fix the new opened issue #1892 (the channel alignment issue) by adopting **floor** still in **this branch**, and then the "ending index" issue in another branch in the future, right?

YuHsuan-Hwang:
> @kswang1029 @YuHsuan-Hwang Excuse me, I would like to double check again if I misunderstood what you mean during the weekly meeting. I should fix the new opened issue #1892 (the channel alignment issue) by adopting **floor** still in **this branch**, and then the "ending index" issue in another branch in the future, right?

Yes. We can file another new issue for the ending index after this PR is merged.

YuHsuan-Hwang (approved):

PR #1895, opened 2022-06-29
Angus/remove css animation

### LINKED ISSUE #1808: Bad behavior when coming back from sleep

veggiesaurus:
**Description**
Closes #1808 and https://github.com/CARTAvis/carta/issues/153 by:
- removing the CSS animation on the loading indicator
- hiding the splash screen (with animated spinner) when an alert is displayed

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,13 +19,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed larger position errors of projected contours, catalog overlays, and vector overlays near the border ([#1843](https://github.com/CARTAvis/carta-frontend/issues/1843)).
 * Fixed no updating of spatial profile after region deleting ([#1831](https://github.com/CARTAvis/carta-frontend/issues/1831), [#1855](https://github.com/CARTAvis/carta-frontend/issues/1855)).
 * Fixed unable to switch channel by clicking scatter plot in stokes analysis widgets ([#1313](https://github.com/CARTAvis/carta-frontend/issues/1313)).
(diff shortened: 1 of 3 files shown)


kswang1029:
@ajm-asiaa tested with Chrome 103 and the high CPU/GPU issue is gone! I guess we can switch back to the latest electron lib.

kswang1029 (approved):
no regression spotted via e2e tests. It appears fine when performing manual tests. When CARTA is idle, CPU and GPU usages as seen in Chrome related processes are basically 0. 👍

YuHsuan-Hwang (approved):
If there's a performance issue with the spinner, shall we plan to also replace the others? Carta may be busy with calculations when we show a spinner ...

kswang1029:
This is the CPU/GPU usage of Chrome 103 when we load the spinner page from blueprintjs user manual. 

Screen Shot 2022-06-30 at 10 20 07

If the duration of rendering the spinner is not too long, it might be still okay to use it as a "in progress" indicator. However, if there are better and power-efficient solutions, we should try those.

veggiesaurus:
> If the duration of rendering the spinner is not too long, it might be still okay to use it as a "in progress" indicator. However, if there are better and power-efficient solutions, we should try those.

I think using it for in progress is fine. The main problem is idle usage when reconneting

Issue #1897, opened 2022-06-30
toolbar in the image viewer: hide tooltip when option menu is enabled


kswang1029:
**Describe the bug**
The buttons of the toolbar in the image viewer have a tooltip each to indicate the functionality. This is good for a single button without extra option menu (such as toggle grid). However, the region button, the wcs matching button, the WCS button, and the export image button have extra option menu when we click again. When the option menu shows up, the tooltip may block the user from selecting an option (observed with the region button), or the tooltip is blocked by the option menu so that we cannot read the context of the tooltip properly.

![Screen Shot 2022-06-30 at 10 02 23](https://user-images.githubusercontent.com/20819712/176577422-8026963d-0f3d-483a-84d3-699072361379.png)

![Screen Shot 2022-06-30 at 10 02 40](https://user-images.githubusercontent.com/20819712/176577532-ec470cca-716d-4348-b424-28dd9e0c992d.png)

![Screen Shot 2022-06-30 at 10 02 55](https://user-images.githubusercontent.com/20819712/176577548-6c49a7ec-7fcd-40b2-b9b5-1c7ae894fd19.png)

I suggest we hide the tooltip of the three buttons _when_ their option menus are enabled. This approach would also solve the issue #1742 I guess.

**To Reproduce**
1. load an image
2. mouse over the image viewer to see the toolbar
3. mouse over the region button to see the tooltip
4. click the region button to see the option menu and we may see the tooltip blocks the option menu.
5. mouse over the WCS button to see the tooltip
6. click the WCS button to see the option menu and we may see the option menu blocks the tooltip

**Expected behavior**
Hide the tooltip of the region, the wcs matching button, WCS, and export image buttons _when_ their option menus are enabled.

**Platform info (please complete the following information):**
 - OS [e.g. macOS Monterey]: macOS Monterey
 - Browser [e.g. chrome, safari, electron app]: chrome
 - Browser version [e.g. 22]: 103
 - Backend branch [e.g. dev, v3b2 release]: dev
 - Frontend branch [e.g. dev, v3b2 release]: dev

kswang1029:
Fixed via #1898

PR #1898, opened 2022-06-30
ksw/show tooltip when needed

### LINKED ISSUE #1742: persistent export image tooltip

kswang1029:
**Description**
This is to address https://github.com/CARTAvis/carta-frontend/issues/1897 and also fixes https://github.com/CARTAvis/carta-frontend/issues/1742.

Thanks for the suggestion from @YuHsuan-Hwang based on the example code from Blueprintjs. The root cause (sort of) is the nested structure of popover2 and tooltip2. If we place them in the correct order, the tooltip issue is gone. 

The following buttons+tooltips in the image viewer are revised
- region
- match wcs
- grid wcs type
- export image

**Checklist**
- [X] e2e test passing / ~added corresponding fix~


acdo2002 (approved):
The tooltip now is not blocked by the extended menu!

YuHsuan-Hwang (approved):

veggiesaurus (approved):

PR #1912, opened 2022-07-18
Yuhsuan/1780 migrate region fix

### LINKED ISSUE #1780: missing regions when the image is unmatched to the reference and gets a copy of the reference region list
To reproduce:

1) load an image and a region file with 2000 regions
2) append the same image and match it to the reference image
3) keep the region list widget active in the view
4) unmatch the 2nd image and it will get a copy (slow~) of the region list from the 1st image. We will see lots of SET_REGION_ACK but the total count in the region list is incorrect. In this case we expect to see 2000 region registered but there are 1813 only.

YuHsuan-Hwang:
**Description**
Closes #1780 missing regions when the image is matched or unmatched to the reference.

When migrating regions, we check existing time stamps and skip duplicate ones. The root cause is that regions imported from a file may have same time stamps. Some regions are skipped when compared with regions that are just migrated. Fixed by copying existing time stamps before migrating regions to avoid comparison between the group of regions.

**Checklist**


veggiesaurus (approved):

kswang1029 (approved):
confirmed that no region is missing. all e2e tests are passing. Manual tests are fine.

PR #1919, opened 2022-07-20
ajm/jenkinsfile update

ajm-ska:
**Description**
This makes two changes to the Jenkinsfile:
1. It uses the slackSend Jenkins plugin to send a final success or failure status to the #jenkins-build-status Slack channel.
2. Previously we only had one server doing the web assembly compilation and then the three build stages for node v12, v14, and v16 occurred parallel. An ongoing issue has been that occasionally two of the three parallel stages fail to contact Github for some unknown reason. This PR attempts to circumvent that issue by performing the three parallel build stages on three different servers instead of only one. Let's try it for a while. If the problem remains, then we can remove the parallel stage.

**Checklist**
- [v] no changelog update needed
- [v] e2e test passing - This should have no effect on the e2e tests
- [v] no protobuf update needed

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,6 +27,8 @@ pipeline {
                     steps {
                         unstash "protobuf"
                         unstash "wasm_libs"
+                        sh "uname -a"
+                        sh "lsb_release -a"
                         sh 'rm -rf node_modules build'
                         sh 'n 12'
                         sh 'n exec 12 node -v'
@@ -41,6 +43,8 @@ pipeline {
                     steps {
                         unstash "protobuf"
                         unstash "wasm_libs"
+                        sh "uname -a"
+                        sh "lsb_release -a"
                         sh 'rm -rf node_modules build'
                         sh 'n 14' 
                         sh 'n exec 14 node -v'
@@ -55,6 +59,8 @@ pipeline {
                     steps {
                         unstash "protobuf"
                         unstash "wasm_libs"
+                        sh "uname -a"
+                        sh "sw_vers"
                         sh 'rm -rf node_modules build'
                         sh 'n 16'
                         sh 'n exec 16 node -v'
@@ -67,10 +73,10 @@ pipeline {
     }
     post {
         success {
-            slackSend color: 'good', message: "carta-frontend - Success - ${env.BRANCH_NAME} ${COMMIT_ID} (<${env.RUN_DISPLAY_URL}|open>)";
+            slackSend color: 'good', message: "carta-frontend - Success - ${env.BRANCH_NAME} <${env.RUN_DISPLAY_URL}|${COMMIT_ID}>";
         }
         failure {
-            slackSend color: 'danger', message: "carta-frontend - Fail - ${env.BRANCH_NAME} ${COMMIT_ID} (<${env.RUN_DISPLAY_URL}|open>)";
+            slackSend color: 'danger', message: "carta-frontend - Fail - ${env.BRANCH_NAME} <${env.RUN_DISPLAY_URL}|${COMMIT_ID}>";
         }
     }
 }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,14 +9,20 @@ pipeline {
                 label "ubuntu2004-agent"
             }
             steps {
-                catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
-                    script {
-                        COMMIT_ID = sh(returnStdout: true, script: 'git rev-parse --short HEAD').trim()
-                    }
-                    sh "git submodule update --init --recursive"
-                    sh "n exec 14 npm run build-libs-docker"
-                    stash includes: "protobuf/**/*", name: "protobuf" 
-                    stash includes: "wasm_libs/built/**/*,wasm_libs/zstd/build/standalone_zstd.bc", name: "wasm_libs"
+                script {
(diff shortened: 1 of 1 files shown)


YuHsuan-Hwang (approved):

kswang1029 (approved):

veggiesaurus:
I don't have a problem in general with this PR, but I don't think introducing another moving part (the jenkins-slack connection) is the right thing to do. We should rather fix the underlying issue or request help fixing it from a devops specialist.

ajm-ska:
@veggiesaurus The main purpose of this PR is an attempt to fix the failure where sometimes the 3 parallel stages are unable to contact Github simultaneously. Previously, the entire build took place on one server, but this update spreads the parallel stage over three different servers. The idea is that perhaps three different IP addresses will have less trouble contacting Github.
 
I have just made another change today. At first, I was using an old MacPro as one of the servers, but I see the `npm run build-docker` step takes ~6 minutes whereas the same step on the Linux servers takes ~3.5 minutes. I was surprised because the Mac is powerful with dual 6-core 3.46GHZ Intel Xeon CPUs and 64GB RAM. Yes, I did adjust the Docker Desktop settings to add more RAM and cores than the default, but it is still slow. It appears such Docker builds with a lot of file I/O are slow on Mac due to the way Docker mounts and accesses the local filesystem (Although it may no longer be an issue on newer Macs where all components are integrated on the SoC). A fix could be [docker-sync](https://github.com/EugenMayer/docker-sync). But for now, I have added a 3rd Linux node to Jenkins instead.

YuHsuan-Hwang (approved):

veggiesaurus:
@ajm-asiaa what version of macOS are you using? Newer versions support virtio which dramatically speeds up io in docker

ajm-ska:
@veggiesaurus It is macOS Big Sur 11.6.1. Although there isn't really any need to use macOS here. I only chose it at first because we already had it set up as a Jenkins node to be used in the ICD tests. I didn't expect it would be the slowest. 
But we can easily use a Linux server instead. Now the Jenkinsfile is using 3 different Linux servers for the parallel build. Looking at the last run, the total time of the parallel stage is 4 minutes 20 seconds:
```
- WebAssembly compilation (2 min 34 sec)
- Build (4 min 20 sec)
   - Build with node v12 (1 ms)
     - Build with node v12 (4 min 14 sec)
   - Build with node v14 (1 ms)
     - Build with node v14 (4 min 10 sec)
   - Build with node v16 (3 min 47 sec)
     - Build with node v16 (3 min 43 sec)
```

veggiesaurus:
> @veggiesaurus It is macOS Big Sur 11.6.1. Although there isn't really any need to use macOS here. I only chose it at first because we already had it set up as a Jenkins node to be used in the ICD tests. I didn't expect it would be the slowest. 
> But we can easily use a Linux server instead. Now the Jenkinsfile is using 3 different Linux servers for the parallel build. Looking at the last run, the total time of the parallel stage is 4 minutes 20 seconds:
> ```
> - WebAssembly compilation (2 min 34 sec)
> - Build (4 min 20 sec)
>    - Build with node v12 (1 ms)
>      - Build with node v12 (4 min 14 sec)
>    - Build with node v14 (1 ms)
>      - Build with node v14 (4 min 10 sec)
>    - Build with node v16 (3 min 47 sec)
>      - Build with node v16 (3 min 43 sec)
> ```

Just out of interest, could you enable the experimental features in docker desktop on macOS and see how it does?

ajm-ska:
> Just out of interest, could you enable the experimental features in docker desktop on macOS and see how it does?

I can only test the "Use the new Virtualiztion framework" feature. I can not turn on "Enable VirtioFS accelerated directory sharing" on that machine because it says it requires macOS 12.2 or later. I don't have any Intel Mac running Monterey.

But either way, I did the test with "Use the new Virtualiztion framework".
(I ran `rm -rf node_modules` and `n exec 16 npm install --legacy-peer-deps` first for both the before and after tests)
Before: `time n exec 16 npm run build-docker`  took `cpu 6:02.47 total`
After: `time n exec 16 npm run build-docker ` took `cpu 6:22.44 total`
So it seems "Use the new Virtualiztion framework" by itself actually slows things down a bit.

I do have Monterey on the M1 MacMini, but unfortunately, I can not try the VirtioFS feature on that either as it is still on macOS 12.0.1! I'll try updating it later.

ajm-ska:
@veggiesaurus I was able to test the Docker VirtioFS feature in Docker v4.11.0 on an updated Monterey M1 Mac now.

Without VirtioFS the `npm run build-docker` step took:
system 23% cpu 9:15.63 total
system 22% cpu 9:14.40 total
system 22% cpu 9:06.92 total

With VirtioFS the `npm run build-docker` step took:
system 24% cpu 8:32.84 total
system 24% cpu 8:33.09 total
system 24% cpu 8:29.08 total

So there is a performance boost using VirtioFS.

Although I know this isn't really an ideal test. M1 is a bit of a mystery with its 4 efficiency and 4 performance cores. It is not clear which cores it is using (In this Docker Desktop I had 6 CPUs and 12GB RAM assigned). 

I am unable to do the same test on the older Intel MacPro because we are using it for macOS 11 ICD testing and so can not update the OS to 12.3 or later in order to try the Docker VirtioFS feature. 

But just to note again, this is not an issue for this PR. The parallel stage runs on 3 different Linux servers. We are not using MacOS now.

veggiesaurus (approved):

PR #1923, opened 2022-07-22
carli/1884_better_precision_handling_for_catalog_plots_axis

crocka:
**Description** 
linked issues and companion PRs (if there are), what is implemented or fixed, how to test it, … 
Created a formatTickValues method that checks the range of the yaxis values shown on the plot and adjusts the significant digit as needed. 

**Checklist** 
- [x] ~~e2e test passing~~ / added corresponding fix  

### DIFF
--- a/src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
+++ b/src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
@@ -875,7 +875,6 @@ export class CatalogPlotComponent extends React.Component<WidgetProps> {
             layout.xaxis.tickformat = this.formatTickValues(layout.xaxis.range);
         } else {
             data = this.histogramData.data;
-            //layout.xaxis.tickformat = this.formatTickValues(layout.xaxis.range);
             let border;
             if (widgetStore.isHistogramAutoScaledX) {
                 border = this.histogramData.border;
@@ -883,6 +882,7 @@ export class CatalogPlotComponent extends React.Component<WidgetProps> {
                 border = widgetStore.histogramBorder;
             }
             layout.xaxis.range = [border.xMin, border.xMax];
+            layout.xaxis.tickformat = this.formatTickValues(layout.xaxis.range);
             layout.yaxis.range = [this.histogramY?.yMin, this.histogramY?.yMax];
             layout.yaxis.fixedrange = true;
             // autorange will trigger y axis range change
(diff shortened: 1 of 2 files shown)


kswang1029:
@crocka to have better UX and readability, could you try to display the tick values like %g? So if abs(value) <= 1E-5, we use scientific notation. Otherwise, we use usual decimal notation.

crocka:
The tickvalues are formatted based on its value and the difference between the max and min value of the tickvalues shown on the plot axis at that moment. If the value is > 1e-5 and < 1e+5, then it is displayed as decimal number, and the exponent of the difference is used to set the digit after the decimal point. If the value is > 1e+5, the format is set to 'e'. If the value is < 1e-5, we use the exponent of the difference to determine how many significant digit is needed.

kswang1029:
@crocka spotted these cases near the threshold we set for changing formatter
Screen Shot 2022-07-27 at 15 48 27
Screen Shot 2022-07-27 at 15 49 36

kswang1029:
@crocka good to turn this to a PR 👍

kswang1029 (approved):
this looks good. e2e tests are all passing except the tests for scatter plot, histogram plot and linked visualization. This is due to the padding width change. I have created a branch to fix these e2e tests. Once this PR is merged, I will also merge test fix as well.

src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
```diff
@@ -600,6 +600,22 @@ export class CatalogPlotComponent extends React.Component<WidgetProps> {
...
+    private formatTickValues = (range: number[]): string => {
+        const difference = range[1] - range[0];
+        const exponential = difference.toExponential(2);
+        const power = parseFloat(exponential.split("e")[1]);
+        const maxPower = parseFloat(range[1].toExponential(1).split("e")[1]);
+        const minPower = parseFloat(range[0].toExponential(1).split("e")[1]);
+        if (maxPower >= 5) {
+            return `e`;
+        } else if (minPower <= -5) {
+            console.log(power, minPower, maxPower);
```
YuHsuan-Hwang: Please remove `console.log`.

YuHsuan-Hwang (commented):
Sometimes we still have extra zeros or insufficient digits when scrolling around. But I guess this is hard to fix because we do not know the exact tick step in our approach, so I'm ok with the current changes.
image
image

kswang1029:
> Sometimes we still have extra zeros or insufficient digits when scrolling around. But I guess this is hard to fix because we do not know the exact tick step in our approach, so I'm ok with the current changes.
> image
> image

uh, extra zero is okay. missing digit is more critical though. 🤔

YuHsuan-Hwang (approved):
Tested that the insufficient digit is solved.
@crocka Please update the branch with the latest dev branch.

PR #1927, opened 2022-07-26
Yuhsuan/1852 check new release

### LINKED ISSUE #1852: check for new CARTA release
This is considered as a good way to notify existing users that there is a new CARTA release. Mostly useful for UDM I guess.
- the frontend checks regularly (daily?) if there is a new CARTA release. Maybe there is a way by checking Github using curl? If there is a new version available, show a popup and guide users where to get it.
- in the Help manual, add an option as "Check for update" so users can trigger the check manually anytime too.

YuHsuan-Hwang:
**Description**

Closes #1852: show a notification when there is a new CARTA release.

Implemented:
* When the frontend is initialized (with preferences updated), check the release tags on GitHub.
(ref: https://docs.github.com/en/rest/releases/releases#list-releases)
* If the latest release tag is not the same as the latest release version stored in the preference (default: `"v" + version info` in package.json), "send a message" to the users:
  
  image  image

* Click the message button to show the notification:
  beta release
  image
  stable release
  image
* Click the message button again or else where to close the notification. The message will remain at the corner.
* Click `OK` button to dismiss the message. The message will not be shown again until another new release is available.
* If `don't show` toggle is on when clicking the `OK` button, the message will never be shown again.
* Manually reset this feature by running `app.preferenceStore.clearPreferences(["checkNewRelease", "latestRelease"]);` in the code snippet or the console.
* Adjusted the alignment of top right corner indicators
  image  image

How to test it:
* Manually set the compared version by `app.preferenceStore.setPreference("latestRelease", "xxx");`
* Simulate a Github release update by `app.updateNewRelease("xxx");`

@veggiesaurus @ajm-asiaa Please check if the version comparison works, thanks.

~ToDo: we may want to skip this feature in some situations, e.g. read-only mode?~
The preference still works with local storage in a read-only mode, so we can keep this feature.

**Checklist**
- [ ] e2e test passing / added corresponding fix


kswang1029:
> ToDo: we may want to skip this feature in some situations, e.g. read-only mode?

For the ALMA archive deployment, it is the UDM version deployed (ie no controller). It will depend on the ALMA observatory to deploy a new version during their regular release cycles. Not sure what to do with such situation. I guess it is still okay to show the popup if there is a new version available. Users can click "don't show again" button when they want to. 

For the controller version, I guess we may disable the check? @veggiesaurus ?

kswang1029:
@YuHsuan-Hwang is it possible to add an option in the "Help" menu so users can use it to check for update manually?

Maybe above "About"?

Screen Shot 2022-07-26 at 16 19 35


Here is a minor enhancement to consider:
Screen Shot 2022-07-26 at 16 37 51
It might be handy to show additional info in the popup too such as 

> Current release: v3.0.0-beta.3
> New release: v3.0.0-stable

YuHsuan-Hwang:
> For the ALMA archive deployment, it is the UDM version deployed (ie no controller). It will depend on the ALMA observatory to deploy a new version during their regular release cycles. 

Maybe we don't need to show the notification when the users can not upgrade the program? Also, "don't show again" button may not work in a read-only mode.

> @YuHsuan-Hwang is it possible to add an option in the "Help" menu so users can use it to check for update manually?
> 
> Maybe above "About"?

Maybe we can add a button in the about dialog. We will have to add a "no new release" alert though.

kswang1029:
> Maybe we don't need to show the notification when the users can not upgrade the program? Also, "don't show again" button may not work in a read-only mode.

that's true. 🤔 is there an easy to bypass the update check? do we need a new backend flag --no_update_check to do so? Or we keep the check flag in localstorage so that users just need to click once.

veggiesaurus:
If we make the notification far less invasive (e.g. just a tiny icon next to the connection info that shows the alert when clicked), then this becomes less of an issue, I guess?

src/stores/AppStore.ts
```diff
@@ -1303,12 +1303,40 @@ export class AppStore {
...
+                    this.checkNewRelease();
+                }
             } catch (err) {
                 console.error(err);
             }
         }
     };
 
+    private checkNewRelease = () => {
+        fetch("https://api.github.com/repos/CARTAvis/carta/releases", {headers: {Accept: "application/vnd.github+json"}})
```
veggiesaurus: I suggest you use `axios` instead, and also make this an `async` function so we can use `await`

src/stores/AppStore.ts
```diff
@@ -1303,12 +1303,40 @@ export class AppStore {
...
+                    throw new Error(response.statusText);
+                }
+                return response.json();
+            })
+            .then(async releases => {
+                const currentRelease = "v" + CARTA_INFO.version;
+                const latestRelease = releases[0].tag_name;
+
+                if (currentRelease !== latestRelease) {
+                    console.log("new release available: ", latestRelease);
```
veggiesaurus: we need quite a bit more logic to determine this. For example:

- What happens when we're running a beta version? It should tell us about new _beta_ versions as well as stable.
- What happens when we're running the stable version? It should _not_ tell us about beta versions.
- What happens when we're running an alpha/dev version? It should not tell us about anything

src/stores/AppStore.ts
```diff
@@ -1303,12 +1303,40 @@ export class AppStore {
...
+        fetch("https://api.github.com/repos/CARTAvis/carta/releases", {headers: {Accept: "application/vnd.github+json"}})
+            .then(response => {
+                if (!response.ok) {
+                    throw new Error(response.statusText);
+                }
+                return response.json();
+            })
+            .then(async releases => {
+                const currentRelease = "v" + CARTA_INFO.version;
+                const latestRelease = releases[0].tag_name;
```
veggiesaurus: need to use `?.[0]?.tag_name` in case the releases JSON file is not the expected format

veggiesaurus (changes requested):

kswang1029:
> If we make the notification far less invasive (e.g. just a tiny icon next to the connection info that shows the alert when clicked), then this becomes less of an issue, I guess?

Wonder if we can repurpose the green circle and use yellow color and its tooltip to notify users about a new release? Ideally the yellow circle can blink but now we know it is a bad idea from performance point of view🙈. Or, if possible, having the green circle blinks for few seconds and if users hover over it, the tooltip shows the new release news.🤔️

ajm-ska:
> If we make the notification far less invasive (e.g. just a tiny icon next to the connection info that shows the alert when clicked), then this becomes less of an issue, I guess?

Just an idea: If we want the notification to be less invasive, could we have it appear on the loading splash screen instead? I know sometimes the splash screen may not show for very long, but then if there is an update notification, maybe display it for 3 seconds the first time, and then only very briefly every time after that?

YuHsuan-Hwang:
1.
> If we make the notification far less invasive (e.g. just a tiny icon next to the connection info that shows the alert when clicked), then this becomes less of an issue, I guess?

I think this is a good idea. Please check if this new design works:
* When the frontend is initialized (with preferences updated), check the release tags on GitHub.
* If the latest release tag is not the same as "v" + version info in package.json, we "send a message" to the users:
  
  image  image

* Click the message to show an alert or a popover (which one is better?):
  
  image

  image

* Click OK to close the notification; click Don't show again to hide the message icon and disable this feature.

However, I still prefer we skip this feature when the users can not upgrade the program, at least for read-only modes?

2.
> is it possible to add an option in the "Help" menu so users can use it to check for update manually?
> 
> Maybe above "About"?

I was wondering if the users should just check the official website in such cases. Maybe we can instead add a link to the about dialog with a hint, for example:

image

3.
> we need quite a bit more logic to determine this. For example:
> 
> * What happens when we're running a beta version? It should tell us about new _beta_ versions as well as stable.
> * What happens when we're running the stable version? It should _not_ tell us about beta versions.

I assume we also would like to inform the users when there's a beta release? We can show it's a beta release in the notification (see examples above)

> * What happens when we're running an alpha/dev version? It should not tell us about anything

We may try to skip this feature with env flags.

4.
> Wonder if we can repurpose the green circle and use yellow color and its tooltip to notify users about a new release? Ideally the yellow circle can blink but now we know it is a bad idea from performance point of view🙈. Or, if possible, having the green circle blinks for few seconds and if users hover over it, the tooltip shows the new release news.🤔️

> Just an idea: If we want the notification to be less invasive, could we have it appear on the loading splash screen instead? I know sometimes the splash screen may not show for very long, but then if there is an update notification, maybe display it for 3 seconds the first time, and then only very briefly every time after that?

I think these designs also work but may be hard to see the difference.

kswang1029:
l like this one (need to include current release version info)!
Screen Shot 2022-07-27 at 15 20 32


For manual check button, never mind. The new less invasive approach should be sufficient.

veggiesaurus:
image
 this one looks great

YuHsuan-Hwang:
with current release version info version:
image

kswang1029:
shall we just call "CARTA website" or "CARTA homepage"?

veggiesaurus:
@YuHsuan-Hwang if I press "Ok", does it still remind me of the same release the next time? I suggest we add a field to preferences for the previously dismissed version number, so that pressing Ok won't show anything until a different release.

e.g for a user on v3:

- new release: v4.0.0-beta.1. User presses ok. Next time they load the page, they won't see the notification.
- new release: v4.0.0-beta 2. User will see the notification. They press "Don't show again"
- new release: v4.0.0: User will not see notification because they've selected "don't show again"

YuHsuan-Hwang:
> shall we just call "CARTA website" or "CARTA homepage"?

"CARTA homepage" looks good to me.

> @YuHsuan-Hwang if I press "Ok", does it still remind me of the same release the next time? I suggest we add a field to preferences for the previously dismissed version number, so that pressing Ok won't show anything until a different release.
> 
> e.g for a user on v3:
> 
> * new release: v4.0.0-beta.1. User presses ok. Next time they load the page, they won't see the notification.
> * new release: v4.0.0-beta 2. User will see the notification. They press "Don't show again"
> * new release: v4.0.0: User will not see notification because they've selected "don't show again"

Sounds reasonable. We can apply this behavior.

ajm-ska:
I like that orange envelope with the popover too! It looks good!

src/stores/AppStore.ts
YuHsuan-Hwang: fixed

src/stores/AppStore.ts
YuHsuan-Hwang: changed to `axios`

src/stores/AppStore.ts
YuHsuan-Hwang: Currently for all versions, we show both new beta/stable version but with different description in the message. I think this shall be fine with the message popover design.
I think we can keep this feature in dev version for testing. Developers can dismiss the message icon easily.

YuHsuan-Hwang:
@kswang1029 @veggiesaurus @ajm-asiaa 
I have updated the branch to the "message popover" design. Please check the first comment for the details. Most of the changes are based on our discussion, except that I changed the `don't show` button to a toggle with clearer description.

ajm-ska:
@YuHsuan-Hwang That's a great idea to use the toggle switch!

Regarding the text, I wonder if it is not entirely necessary to say "now", particularly when already mentioning "new"? Perhaps just "A new release is available!".
Also, should we say "Visit the CARTA homepage for more details.", assuming it would fit in one line?
For the last line maybe "Do not show again." would be enough?

I would be interested to hear @confluence's opinion on this?

Screen Shot 2022-07-29 at 10 09 03 PM

YuHsuan-Hwang:
@ajm-asiaa 
> Regarding the text, I wonder if it is not entirely necessary to say "now", particularly when already mentioning "new"? Perhaps just "A new release is available!". Also, should we say "Visit the CARTA homepage for more details.", assuming it would fit in one line?

Good suggestions, thanks.

>For the last line maybe "Do not show again." would be enough?

I added the plural "new releases" to show that not only this message but also future releases will not be informed. I'm not sure if "Do not show again." is sufficient.

veggiesaurus:
@YuHsuan-Hwang how does one un-toggle the "Don't show again" thing? Surely if you toggle it once the popup will never come back?

YuHsuan-Hwang:
> how does one un-toggle the "Don't show again" thing? Surely if you toggle it once the popup will never come back?

currently the users have to toggle and click OK to disable the feature.

kswang1029:
@YuHsuan-Hwang I suggest we have a dynamic description. If the "new" release does not contain beta, we show something like "A new stable release is available." If the "new" release contains beta, then we show "A new beta release is available". 

Screen Shot 2022-07-31 at 19 11 16

kswang1029:
> @YuHsuan-Hwang I suggest we have a dynamic description. If the "new" release does not contain beta, we show something like "A new stable release is available." If the "new" release contains beta, then we show "A new beta release is available".
> 
> Screen Shot 2022-07-31 at 19 11 16

nevermind. I see it is implemented in [419ea43](https://github.com/CARTAvis/carta-frontend/pull/1927/commits/419ea438dce5c51ef5f53d6d024033508820c744)

kswang1029:
just as a note: all current e2e tests are passing. I will re-run the test once this draft becomes formal and handle the e2e checkbox.

YuHsuan-Hwang:
> Regarding the text, I wonder if it is not entirely necessary to say "now", particularly when already mentioning "new"? Perhaps just "A new release is available!". Also, should we say "Visit the CARTA homepage for more details.", assuming it would fit in one line?

This is adjusted:

image

src/stores/AppStore.ts
```diff
@@ -1303,12 +1315,30 @@ export class AppStore {
...
+                if (this.preferenceStore.checkNewRelease) {
+                    await this.checkNewRelease();
+                }
             } catch (err) {
                 console.error(err);
             }
         }
     };
 
+    private checkNewRelease = async () => {
```
veggiesaurus, line 1327: use `try/catch` and `await` if you're using an `async` function

veggiesaurus (changes requested):

kswang1029 (approved):
all current e2e tests are passing when the envelope is displayed and is not displayed. Manual tests are looking good.

ajm-ska:
I just tried the latest commit now. I didn't adjust any Snippets and I see the envelope is showing that the new release is v3.0.0-beta.3. I guess it should not show that the same release is a new release. Could this be a bug, or is it just the way we are testing it?

Screen Shot 2022-08-03 at 10 15 14 AM

Do you think it would be worth doing a quick real-life test at the very end? Perhaps just before the v3.0.0 release, after we have the CARTA packages ready with this feature, we could make a temporary fake tag on Github, say `v3.1.0`, then double check that it successfully detects the update?

kswang1029:
> I just tried the latest commit now. I didn't adjust any Snippets and I see the envelope is showing that the new release is v3.0.0-beta.3. I guess it should not show that the same release is a new release. Could this be a bug, or is it just the way we are testing it?
> 
> Screen Shot 2022-08-03 at 10 15 14 AM
> 
> Do you think it would be worth doing a quick real-life test at the very end? Perhaps just before the v3.0.0 release, after we have the CARTA packages ready with this feature, we could make a temporary fake tag on Github, say `v3.1.0`, then double check that it successfully detects the update?

what's in your preferences.json? (don't post telemetry UUID here)

ajm-ska:
> what's in your preferences.json? (don't post telemetry UUID here)

I see! I had:
`"latestRelease": "4.0",`
After deleting that line and restarting carta-backend, I no longer see the envelope.
I guess it didn't show `New release: 4.0` as it doesn't mix beta releases with full releases?

kswang1029:
> > what's in your preferences.json? (don't post telemetry UUID here)
> 
> I see! I had: `"latestRelease": "4.0",` After deleting that line and restarting carta-backend, I no longer see the envelope. I guess it didn't show `New release: 4.0` as it doesn't mix beta releases with full releases?

"latestRelease": "4.0" is retrieved from Github.

ajm-ska:
> "latestRelease": "4.0" is retrieved from Github.

When I was testing it a few days ago, I used the Snippets to set the new release to 4.0. `(app.updateNewRelease("4.0");` so that is why it was written to the `preferences.json` to start with.

YuHsuan-Hwang:
`"latestRelease": "4.0",` means that Carta compared v3b3 with 4.0 and shown a message, and the user dismissed the message. When a "new v3b3" is updated on Github, we shows that "new v3b3" because it's different from 4.0.

ajm-ska (approved):

veggiesaurus (approved):

PR #1929, opened 2022-07-26
carli/1880_cursor_info_when_profile_smoothing_enabled

crocka:
**Description**
The smoothed profile data is found inside the plotData object of the SpectralProfilerComponent. The data is sent to the genCursorInfoString method and went through binary search to find the corresponding y value based on the cursor x position on the plot. When both smoothing and overlay is enabled, the data of the main and smoothed profile is displayed in the SpectralProfilerInfoComponent. When only the smoothed profile is enabled, only the smoothed profile data would be displayed. When only smoothing is disabled, only the original profile data is shown.

**Checklist** 


kswang1029:
when we scan through the profile like the one in the video, we see the frequency of the original profile and the one of the smoothed profile change at the same time. It should not be the case in this example. 

https://user-images.githubusercontent.com/20819712/181194730-f12797f3-9b8e-49c8-bd84-4849c59d1253.mov

kswang1029:
@crocka good to be a formal PR. 👍

kswang1029 (approved):
this looks good. all current e2e tests are passing. Manual tests are fine. Note that this a fix only for the spectral profiler. We will need another fix for the spatial profiler later.

YuHsuan-Hwang:
@kswang1029 Do we also support smoothed data of the secondary axis after #1883 is merged?

kswang1029 (approved):
no regression is observed from the current e2e tests. Manual tests are fine as far as I can tell. 👍

kswang1029:
@crocka just spotted one minor formatting issue. It should be "Channel 123" instead of "123 channel".
Screen Shot 2022-09-05 at 13 44 50

YuHsuan-Hwang (approved):

kswang1029 (approved):

PR #1934, opened 2022-07-30
carli/1879_NaN_in_the_cursor_info_bar

### LINKED ISSUE #1879: NaN* pixel value in the cursor info bar of the image viewer when the image is 1x1 pixel
**Describe the bug**
When we load an image with 1 by 1 pixel (the sumwt image from CASA-tclean task), the pixel value is NaN* in the cursor info bar of the image viewer. This is likely due to how we trigger set_cursor as we use whole pixel to determine whether or not triggering a set_cursor.

**To Reproduce**
1. load the attached image
[image_1_1.fits.zip](https://github.com/CARTAvis/carta-frontend/files/8866601/image_1_1.fits.zip)

2. move the cursor around the single pixel image and observe the cursor info bar
3. the pixel value is NaN*

**Expected behavior**
The pixel value should not be NaN*. With the statistics widget, we can read the pixel value. 

 - Browser [e.g. chrome, safari, electron app]: chrome, electron app
 - Browser version [e.g. 22]: 102
 - Backend branch [e.g. dev, v3b2 release]: dev, v3b3 release
 - Frontend branch [e.g. dev, v3b2 release]: dev, v3b3 release

crocka:
**Description**
The (0, 0) index of the cursor position is set to the center of the bottom left pixel of any image. Therefore, the very bottom left corner of the image is at (-0.5, -0.5). The updateCursorRegion method of the FrameStore incorrectly identified the border of the image due to the issue I mentioned above. The method only triggers the updateCursorRegionPosition method in the RegionSetStore when the cursor is hovering on the image. Failed to identify the border of the image caused troubles with the 1px by 1px image because it is so small.

Another cause of this issue was the updateCursorRegionPosition method in the RegionSetStore. In order to reduce the number of times triggering setCenter method, an if-statement in this method checks for whether or not the previous pixel that the cursor hovered on is the same as the current pixel. However, the method is only triggered when cursor moves into the image, therefore, the rounded position of the cursor in the memory is always the same with the 1px by 1px picture. To get around this problem, an observable is added to the RegionSetStore to track whether or not the cursor is hovering on the image. When updateCursorRegion of the FrameStore calls the updateCursorRegionPosition in the RegionSetStore, an if statement checkout whether the isHover observable is true. If it is false, it makes a call to the backend to update the position of cursor.

**Checklist** 
- [x] changelog updated  / ~no changelog update needed~


kswang1029:
@crocka in short, the chance to trigger cursor value update in each dimension increases by a factor of 2, right?

crocka:
> @crocka in short, the chance to trigger cursor value update in each dimension increases by a factor of 2, right?

Not quite. The trigger zone is expanded 0.5 pixels around the perimeter of the image, which is not noticible for regular images. When the cursor is within the image, no change has been done.

src/stores/Frame/FrameStore.ts
```diff
@@ -1902,18 +1902,16 @@ export class FrameStore {
...
-                this.frameRegionSet.updateCursorRegionPosition(pos);
-            }
+        } else if (aroundImage) {
+            this.frameRegionSet.updateCursorRegionPosition(pos);
         }
-
         for (const frame of this.secondarySpatialImages) {
             const pointSecondaryImage = transformPoint(frame.spatialTransformAST, pos, false);
-            if (pointSecondaryImage.x >= 0 && pointSecondaryImage.x <= frame.frameInfo.fileInfoExtended.width - 1 && pointSecondaryImage.y >= 0 && pointSecondaryImage.y <= frame.frameInfo.fileInfoExtended.height - 1) {
+            if (pointSecondaryImage.x + 0.5 >= 0 && pointSecondaryImage.x + 0.5 <= frame.frameInfo.fileInfoExtended.width && pointSecondaryImage.y + 0.5 >= 0 && pointSecondaryImage.y + 0.5 <= frame.frameInfo.fileInfoExtended.height) {
```
YuHsuan-Hwang: Maybe we adjust this line to be consistent with `aroundImage` so that the trigger zone of the matched images are also enlarged.

src/stores/Frame/RegionSetStore.ts
```diff
@@ -39,7 +39,8 @@ export class RegionSetStore {
             const cursorRegion = this.regions[0];
             // Need to avoid redundant update (position not changed), backend may not reply to redundant requests.
             const roundedPos = {x: Math.round(pos.x), y: Math.round(pos.y)};
-            if (cursorRegion?.regionId === CURSOR_REGION_ID && (cursorRegion.center?.x !== roundedPos.x || cursorRegion.center?.y !== roundedPos.y)) {
+            const hoverImage = pos.x + 0.5 >= 0 && pos.x + 0.5 <= this.frame.frameInfo.fileInfoExtended.width && pos.y + 0.5 >= 0 && pos.y + 0.5 <= this.frame.frameInfo.fileInfoExtended.height;
+            if (cursorRegion?.regionId === CURSOR_REGION_ID && (cursorRegion.center?.x !== roundedPos.x || cursorRegion.center?.y !== roundedPos.y || !hoverImage)) {
```
YuHsuan-Hwang: We may remove `|| !hoverImage` because comparison of `cursorRegion.center` and `roundedPos` already triggers `setCenter`.

YuHsuan-Hwang (changes requested):
We need to avoid sending data request to the backend when the position is outside the image boundary. Perhaps we add a check in `RegionStore.updateRegion`.

image

src/stores/Frame/FrameStore.ts
```diff
@@ -1902,18 +1902,19 @@ export class FrameStore {
...
+            this.frameRegionSet.updateCursorRegionPosition(pos);
+        } else if (!isHoverImage) {
+            this.frameRegionSet.setIsHover(false);
         }
 
         for (const frame of this.secondarySpatialImages) {
             const pointSecondaryImage = transformPoint(frame.spatialTransformAST, pos, false);
-            if (pointSecondaryImage.x >= 0 && pointSecondaryImage.x <= frame.frameInfo.fileInfoExtended.width - 1 && pointSecondaryImage.y >= 0 && pointSecondaryImage.y <= frame.frameInfo.fileInfoExtended.height - 1) {
+            if (pointSecondaryImage.x + 0.5 >= 0 && pointSecondaryImage.x + 0.5 <= frame.frameInfo.fileInfoExtended.width && pointSecondaryImage.y + 0.5 >= 0 && pointSecondaryImage.y + 0.5 <= frame.frameInfo.fileInfoExtended.height) {
                 frame.frameRegionSet.updateCursorRegionPosition(pointSecondaryImage);
```
YuHsuan-Hwang, line 1922: We may need to adjust this part so that it is consistent with the above.

src/stores/Frame/RegionSetStore.ts
```diff
@@ -39,12 +40,17 @@ export class RegionSetStore {
...
             const roundedPos = {x: Math.round(pos.x), y: Math.round(pos.y)};
-            if (cursorRegion?.regionId === CURSOR_REGION_ID && (cursorRegion.center?.x !== roundedPos.x || cursorRegion.center?.y !== roundedPos.y)) {
+            if (cursorRegion?.regionId === CURSOR_REGION_ID && (!this.isHoverImage || cursorRegion.center?.x !== roundedPos.x || cursorRegion.center?.y !== roundedPos.y)) {
                 cursorRegion.setCenter(roundedPos);
+                this.setIsHover(pos.x + 0.5 >= 0 && pos.x + 0.5 <= this.frame.frameInfo.fileInfoExtended.width && pos.y + 0.5 >= 0 && pos.y + 0.5 <= this.frame.frameInfo.fileInfoExtended.height);
             }
         }
     };
 
+    public setIsHover = (bool: boolean) => {
```
YuHsuan-Hwang: Add `@action` to modify a observable.

src/stores/Frame/RegionSetStore.ts
```diff
@@ -39,12 +40,17 @@ export class RegionSetStore {
             const cursorRegion = this.regions[0];
             // Need to avoid redundant update (position not changed), backend may not reply to redundant requests.
             const roundedPos = {x: Math.round(pos.x), y: Math.round(pos.y)};
-            if (cursorRegion?.regionId === CURSOR_REGION_ID && (cursorRegion.center?.x !== roundedPos.x || cursorRegion.center?.y !== roundedPos.y)) {
+            if (cursorRegion?.regionId === CURSOR_REGION_ID && (!this.isHoverImage || cursorRegion.center?.x !== roundedPos.x || cursorRegion.center?.y !== roundedPos.y)) {
                 cursorRegion.setCenter(roundedPos);
+                this.setIsHover(pos.x + 0.5 >= 0 && pos.x + 0.5 <= this.frame.frameInfo.fileInfoExtended.width && pos.y + 0.5 >= 0 && pos.y + 0.5 <= this.frame.frameInfo.fileInfoExtended.height);
```
YuHsuan-Hwang: Perhaps we can change it to `this.setIsHover(true)` because we call this function only when the cursor is on the image.

YuHsuan-Hwang (changes requested):
Looks like the backend also does not send new data when we hover off the image and then hover on the same pixel. We may need to avoid continuously requesting the same position.
image

src/stores/Frame/FrameStore.ts
crocka: Yes, it is consistent, expect that pos is replaced with pointSecondaryImage, and this.frame is replaced with frame of this.secondarySpatialImages.

YuHsuan-Hwang (approved):
New changes are working nice without breaking the behavior of sending data requests.

kswang1029 (changes requested):
all e2e tests except some catalog related tests are passing. This might be due to the need to updating this branch with new commits in the dev branch since this branch is branched out from the dev. @crocka please update dev and I will run the e2e tests again to confirm the regression is gone.

Manual test of this fix looks good. No extra data request is seen as far as I can tell.

kswang1029 (approved):
all e2e tests are passing now.

PR #1940, opened 2022-08-03
carli/1796_show_hide_regions

crocka:
**Description**
Implemented show/hide region with 0, 0.1, and 1 opacity. 
Implemented lock all region functionality.
Moved export/import all region button to bottom left of th region list component as float button.

**Checklist**

### DIFF
--- a/src/components/ImageView/RegionView/LineSegmentRegionComponent.tsx
+++ b/src/components/ImageView/RegionView/LineSegmentRegionComponent.tsx
@@ -23,6 +23,7 @@ interface LineSegmentRegionComponentProps {
     stageRef: any;
     onSelect?: (region: RegionStore) => void;
     onDoubleClick?: (region: RegionStore) => void;
+    opacity: number;
 }
 
 const NEW_ANCHOR_MAX_DISTANCE = 16;
@@ -345,7 +346,7 @@ export class LineSegmentRegionComponent extends React.Component<LineSegmentRegio
                     y={centerPointCanvasSpace.y}
                     stroke={region.isSimplePolygon ? region.color : INVALID_POLYGON_COLOR}
                     strokeWidth={region.lineWidth}
-                    opacity={region.opacity === 2 ? (region.isTemporary ? 0.5 : region.locked ? 0.7 : 1) : region.opacity === 0 ? 0 : 0.3}
+                    opacity={this.props.opacity === 2 ? (region.isTemporary ? 0.5 : region.locked ? 0.7 : 1) : this.props.opacity === 0 ? 0 : 0.3}
                     dash={[region.dashLength]}
                     closed={!region.creating && region.regionType === CARTA.RegionType.POLYGON}
                     listening={this.props.listening && !region.locked}
--- a/src/components/ImageView/RegionView/LineSegmentRegionComponent.tsx
+++ b/src/components/ImageView/RegionView/LineSegmentRegionComponent.tsx
@@ -345,7 +345,7 @@ export class LineSegmentRegionComponent extends React.Component<LineSegmentRegio
                     y={centerPointCanvasSpace.y}
                     stroke={region.isSimplePolygon ? region.color : INVALID_POLYGON_COLOR}
                     strokeWidth={region.lineWidth}
-                    opacity={region.opacity === 2 ? (region.isTemporary ? 0.5 : region.locked ? 0.7 : 1) : region.opacity === 0 ? 0 : 0.1}
+                    opacity={region.opacity === 2 ? (region.isTemporary ? 0.5 : region.locked ? 0.7 : 1) : region.opacity === 0 ? 0 : 0.3}
                     dash={[region.dashLength]}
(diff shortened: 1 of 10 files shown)


kswang1029:
issues and suggestions:
1) when we lock all regions, we cannot unlock all region by clicking again
2) my initial guess for the 10% opacity seems not ideal. need a better one like 20% or 30% (need experiments).
3) please adopt the same style for the import and export buttons at the bottom-right corner as the ones for the spatial profiler for example. 
4) please set the opacity loop as 100%->20%->0%->100%->...
5) when it is 0% we should disable region selection via mouse clicking
6) I suggest we use unlocked icon as the default when all regions are not locked. (ie sychronize the locked/unlocked state)
7) need a special care to the point region when we change the states

kswang1029:
the state of the "(un)lock all" is affected by individual lock button.

https://user-images.githubusercontent.com/20819712/182980142-8ea51abf-5c98-41fa-afcb-566f413731f4.mov

kswang1029:
I also found a bug (not due to your new changes): when a point region is selected, we see the white box (as control point) around the point. When we lock it, we still see it. Like other regions, we should hide the white box (control point). Would you be able to fix it in this branch too?

crocka:
> the state of the "(un)lock all" is affected by individual lock button.
> 
>  Screen.Recording.2022-08-05.at.09.04.28.mov

Oh, this is intentional. When all the region are set to locked/unlocked, the lock/unlock all will be automatically set to locked/unlocked state, but I see there is some problem when you unlock all the regions the second time. It is supposed to stay unlocked. I will fix it soon.

kswang1029:
@crocka this looks good. feel free to turn this as a formal PR when you think proper. Performance-wise I confirm locking or changing opacity is almost instant for 20000 regions in my test cases.

kswang1029:
@crocka this "new" PR appears problematic in the logics of the lock button and the view button.

https://user-images.githubusercontent.com/20819712/188378457-b51dc74a-611f-4675-9ff7-44f2fe778465.mov

kswang1029:
@crocka yes, now the button logic appears fine. 👍 

One issue: if all region is locked and we select a region from the region list widget, the control points of that region should not be rendered in the image viewer as users may think the region can be modified via the control points.
Screen Shot 2022-09-06 at 08 50 57

kswang1029 (approved):
no regression is observed via the current e2e tests. Manual tests are looking good. This should improve UX significantly for users working with lots of regions. 👍

YuHsuan-Hwang (approved):
good work 👍

src/stores/Frame/RegionSetStore.ts
```diff
@@ -12,11 +12,19 @@ export enum RegionMode {
     CREATING
 }
 
+export enum RegionsOpacity {
+    Visible = 1,
+    Transparent = 0.5,
```
veggiesaurus: I think it would make sense to call this `SemiTransparent`. Transparent implies fully transparent.

src/stores/Frame/RegionSetStore.ts
```diff
@@ -240,4 +248,13 @@ export class RegionSetStore {
...
         }
     };
+
+    @action setOpacity(opacity: RegionsOpacity) {
+        this.opacity = opacity;
+    }
+
+    @action setLocked(locked?: boolean) {
+        this.locked = locked === undefined ? !this.locked : locked;
+        if (this.locked) this.selectRegionByIndex(0);
```
veggiesaurus: single-line branches also require braces according to our code style

veggiesaurus (changes requested):

src/components/ImageView/RegionView/LineSegmentRegionComponent.tsx
```diff
@@ -280,7 +279,7 @@ export class LineSegmentRegionComponent extends React.Component<LineSegmentRegio
             }
 
             // Construct anchors if region is selected
-            if (this.props.selected && this.props.listening && !region.locked) {
+            if (this.props.selected && this.props.listening && !region.locked && !AppStore.Instance.activeFrame?.regionSet.locked) {
```
veggiesaurus, line 282: makes sense to create a variable `const frame = AppStore.Instance.activeFrame` above, if you're going to reuse it a few times in the function

veggiesaurus: (or `regionSet`

veggiesaurus (changes requested):

src/components/ImageView/RegionView/LineSegmentRegionComponent.tsx
veggiesaurus: actually, don't worry. Not worth delaying this PR for such a minor thing 😄

veggiesaurus (approved):

YuHsuan-Hwang:
@crocka Please merge dev and update changelog, thanks.

PR #1963, opened 2022-08-05
Angus/1854 vector line width

### LINKED ISSUE #1854: inconsistent vector line width observed on spatially matched images
This is observed in v3-beta3 which I thought this issue was fixed during the implementation stage 🤔

Screen Shot 2022-05-17 at 10 20 07

@veggiesaurus could this be introduced from the colormap control map fix?

veggiesaurus:
**Description**
Closes #1854. It was actually the feather width, rather than the line thickness :facepalm: That's why it only showed up in extreme cases (where `featherWidth * relativeScale ~ lineThickness`)

(also reverts erroneous [commit](https://github.com/CARTAvis/carta-frontend/commit/a244aa7ccf3ab649c8c8e28fbed078adae3f0350)) 

**Checklist**
- [ ] e2e test passing / ~~added corresponding fix~~
- [ ] ~~protobuf updated to the latest dev commit~~ / no protobuf update needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed issue when the active frame changes while the region is being imported.
 * Fixed the displayed values in the cursor info of the histogram widget by adopting binary-searched data x and y values ([#1917](https://github.com/CARTAvis/carta-frontend/issues/1917)) 
 * Fixed missing regions when the image is matched or unmatched to the reference ([#1780](https://github.com/CARTAvis/carta-frontend/issues/1780)).
+* Fixed inconsistent vector line width on spatially matched images ([#1854](https://github.com/CARTAvis/carta-frontend/issues/1854)).
 
 
 ## [3.0.0-beta.3]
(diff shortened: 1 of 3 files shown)


kswang1029 (changes requested):
@veggiesaurus the issue persists. The black vectors are generated with the 1024 image as the data source, while the green vectors are with the 128 image.

Screen Shot 2022-08-05 at 17 36 01

veggiesaurus:
> @veggiesaurus the issue persists. The black vectors are generated with the 1024 image as the data source, while the green vectors are with the 128 image.
> 
> Screen Shot 2022-08-05 at 17 36 01

@kswang1029 Isn't this the expected behaviour? previously the vectors were _different_ thicknesses when rendered on different frames. In this case, they are the same.

The thickness being related to the pixel size seems like a different issue to me :thinking:

veggiesaurus:
@kswang1029 I've modified this, please re-review. I think this handles the other issue as well.
cluster_04096 fits-cluster_00128 fits-image-2022-08-05-12-08-01

kswang1029 (approved):
@veggiesaurus I have repeated my test with two images having very different pixel sizes. Tried to set each as a spatial matching reference and generated vector overlay using each image as data source. Now the line width of vector overlay is rendered corrected. I also tried block overlay and it works normally. 

All e2e tests are passing except the changes due to https://github.com/CARTAvis/carta-frontend/pull/1923 which has been merged to dev.

YuHsuan-Hwang (approved):

PR #1964, opened 2022-08-05
Thchang / implemented smoothed Mean and RMS in spatial and spectral profilers

### LINKED ISSUE #1838: mean and rms do not update after smoothing
The display of the mean and rms in a profile do not update when smoothing.  I set a profile do display the mean and rms.  Then I smoothed the data.  The mean should not change due to smoothing but the rms will.  However, the mean and rms do not update after smoothing.    Whenever a new smooth is implemented the mean and rms should update if they are turned on.

TienHao:
**Description**
closes #1838: mean and RMS do not update after profile smoothing.

updating smoothed mean and RMS in spatial and spatial profilers when profile smoothing, and displaying mean and RMS when there is only one profile.

https://user-images.githubusercontent.com/20379662/183055723-30e5b719-e3c2-4c3a-a02f-080a0cc3637c.mov

**Checklist**


YuHsuan-Hwang (approved):

kswang1029 (changes requested):
all e2e tests are passing for now.

When we show both smoothed and original profiles, we should hide mean/rms info (same reason when the spectral profiler is in the multiple-profile plot mode.
Screenshot from 2022-08-29 12-16-34

The mean/rms value is calculated based on the data in the current view of profile plot. If we zoom in to a subset of data, the mean/rms is recalcuated. This works for the original profile. When we enable profile smoothing, it seems to use _all_ data for the mean/rms calculations regardless the current profile zoom levels. 

[Screencast from 08-29-2022 12:17:23 PM.webm](https://user-images.githubusercontent.com/20819712/187122478-270bc3a5-33f2-4bca-ba92-54c1957dec2b.webm)

kswang1029:
@TienHao now in the spatial profiler if smoothing is applied, we see it adopts the current data in the view to compute mean and rms. :+1: 
in the spectral profiler, we need to support the same behavior.

kswang1029 (approved):
this looks good. no regression is observed from e2e tests.

YuHsuan-Hwang (approved):
@TienHao Please update the change log.

PR #1971, opened 2022-08-10
carli/1201_angular_distance_measuring_tool

crocka:
**Description**
Distance Measurement tool can be enabled in the image view toolbar by single clicking and the distance measurement setting dialog can be turned on by double clicking. The toolbar at the top of the page can also be used to enable distance measurement. In the top toolbar, single clicking would open both the distance measurement tool and distance measurement setting dialog.

There are two sections in the dialog, the line styles and properties sections. The line styles section sets the line width, font size, and the color of the line. The properties section includes wcs coordinate selection and a radio selection between pixel and wcs input. When pixel is selected, the corresponding wcs information is show on the right of the properties section, and vice versa. The last part of the dialog is an input for starting and finishing point in pixel values or wcs values.

**Checklist**
- [x] ~~e2e test passing~~ / added corresponding fix in a new branch


kswang1029 (changes requested):
Few comments so far (will provide more later if there is any):
1) the tooltip of the new "distance measurement" button is not consistent with the button in the image viewer and title of the new dialog. I suggest we switch to "distance measurement". @YuHsuan-Hwang is that okay?

2) in the new dialog, the default coordinate reference is "image". I suggest we set "world" as the default. If the image does not have valid world coordinates, set "image" as the fallback and disable "world".

3) released to 2), if we have an image without valid world coordinates (also pv images), switch to "world" will cause the frontend a hard crash.

4) when we manually enter a coordinate, hitting enter does not make it "accept" the value to trigger new calculations and render the new result.

5) please add a "help" button next to the "close" button at the top-right corner of the dialog and set up a placeholder for the help context. I will work on the help context later.

YuHsuan-Hwang:
> 1. the tooltip of the new "distance measurement" button is not consistent with the button in the image viewer and title of the new dialog. I suggest we switch to "distance measurement". @YuHsuan-Hwang is that okay?

ok with me.

kswang1029:
Few more issues for improvements:
1. when we switch from world to image coordinates, we still see the world coordinates in the text fields. We should show the converted image coordinates instead.

Screenshot from 2022-08-31 14-17-50

2. if the image does not have a valid WCS definition, we see empty image coordinates even after two clicks on the image.

Screenshot from 2022-08-31 14-19-42

3. If we have two images loaded in the multi-panel view mode and if we create a distance measurement in each panel, when we switch to a different coordinate system (eg to galactic) we will see in one of the panels the geodesic rendering is gone. We will need to zoom/pan a bit to trigger the re-rendering.

4. (probably not related to this branch) when there are multiple images loaded in the multi-panel view mode, enabling the distance measurement tool in one panel also enables the tool in all other panels. Can it be independent?

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
```diff
@@ -0,0 +1,241 @@
...
+                                        <td colSpan={2}>
+                                            <CoordinateComponent
+                                                onChange={(ev: React.FormEvent<HTMLInputElement>) => handleChangeWCSMode(ev)}
+                                                selectedValue={this.WCSMode && wcsInfo ? RegionCoordinate.World : RegionCoordinate.Image}
+                                                disableCoordinate={!wcsInfo}
+                                            />
+                                        </td>
+                                    </tr>
+                                    <tr className="distance-measuring-settings-table-input">
+                                        <td>Start{this.WCSMode && wcsInfo ? "" : " (px)"}</td>
```
YuHsuan-Hwang: We could set (px) as `labelInfo`

src/components/ImageView/Toolbar/ToolbarComponent.tsx
```diff
@@ -208,15 +213,22 @@ export class ToolbarComponent extends React.Component<ToolbarComponentProps> {
                             position={tooltipPosition}
                             content={
                                 <span>
-                                    Measure distance
+                                    Distance Measurement
                                     <br />
                                     <i>
                                         <small>Click to create geodesic curves.</small>
+                                        <br></br>
+                                        <small>Double Click to go to setting.</small>
```
YuHsuan-Hwang: Perhaps change it to "Double Click to open the settings."

src/stores/Frame/DistanceMeasuringStore.ts
```diff
@@ -62,4 +69,20 @@ export class DistanceMeasuringStore {
...
+    @action setLineWidth = (width: number) => {
+        this.lineWidth = width;
+    };
+
+    @action setFontSize = (size: number) => {
+        this.fontSize = size;
+    };
+
+    @action setStart = (x: number, y: number) => {
+        this.start = {x: x, y: y || null};
```
YuHsuan-Hwang: Shall we also check for x?

src/utilities/units.ts
```diff
@@ -36,7 +36,7 @@ export function toExponential(val: number, decimals: number = 0): string {
 
 // According to MDN, toFixed only works for up to 20 decimals
 export function toFixed(val: number, decimals: number = 0): string {
-    if (isFinite(val) && isFinite(decimals) && decimals >= 0 && decimals <= 20) {
+    if (val && isFinite(val) && isFinite(decimals) && decimals >= 0 && decimals <= 20) {
```
YuHsuan-Hwang: `isFinite(val)` only (without `val &&`) may be sufficient for checking `val`. When `val` is zero, `val &&` would cause `toFixed` to return `String(val)`, which is an unexpected behavior.

YuHsuan-Hwang (changes requested):

src/utilities/units.ts
crocka: I am worried it may cause a hard crash since isFinite(null) = true.

YuHsuan-Hwang: @crocka I see. I suggest we replace `null` with `NaN` in `DistanceMeasuringStore` to prevent such crash. Also change `x != null` and `y != null` to `isFinite(x)` and `isFinite(y)`.

crocka: Yup, please review the latest commit.

crocka: Yup, please check the lastest commit.

src/components/ImageView/RegionView/RegionViewComponent.tsx
```diff
@@ -430,10 +431,10 @@ export class RegionViewComponent extends React.Component<RegionViewComponentProp
...
                 if (!dist.isCreating && !dist.showCurve) {
-                    dist.start = imagePos;
+                    dist.setStart(imagePos.x, imagePos.y);
                     dist.setIsCreating(true);
                 } else if (dist.isCreating) {
-                    dist.finish = imagePos;
+                    dist.setFinish(imagePos.x, imagePos.y);
                     dist.updateTransformedPos(frame.spatialTransform);
                     dist.setIsCreating(false);
                 } else {
```
YuHsuan-Hwang, line 440: also change line 442 to `setStart`

src/stores/Frame/DistanceMeasuringStore.ts
```diff
@@ -6,36 +6,43 @@ import {getColorForTheme} from "utilities";
...
+        this.transformedFinish = {x: NaN, y: NaN};
         this.isCreating = false;
         this.color = DistanceMeasuringStore.DEFAULT_COLOR;
+        this.lineWidth = DistanceMeasuringStore.DEFAULT_WIDTH;
+        this.fontSize = DistanceMeasuringStore.DEFAULT_FONTSIZE;
     }
 
     @computed get showCurve(): boolean {
-        return this.transformedStart.x != null && this.transformedStart.y != null && this.transformedFinish.x != null && this.transformedFinish.y != null && AppStore.Instance.activeLayer === ImageViewLayer.DistanceMeasuring;
+        return isNaN(this.transformedStart.x) && isNaN(this.transformedStart.y) && isNaN(this.transformedFinish.x) && isNaN(this.transformedFinish.y) && AppStore.Instance.activeLayer === ImageViewLayer.DistanceMeasuring;
```
YuHsuan-Hwang: shall be either `!isNaN()` or `isFinite()`

src/stores/Frame/DistanceMeasuringStore.ts
```diff
@@ -6,36 +6,43 @@ import {getColorForTheme} from "utilities";
...
 
     static readonly DEFAULT_WIDTH = 1.5;
     static readonly DEFAULT_FONTSIZE = 14;
     static readonly DEFAULT_COLOR = "#62D96B";
 
     constructor() {
         makeObservable(this);
-        this.transformedStart = {x: null, y: null};
-        this.transformedFinish = {x: null, y: null};
+        this.start = {x: 0, y: 0};
```
YuHsuan-Hwang, line 25: shall we also initialize start/finish to NaNs?

src/stores/Frame/DistanceMeasuringStore.ts
```diff
@@ -44,22 +51,38 @@ export class DistanceMeasuringStore {
...
             this.transformedFinish = spatialTransform ? spatialTransform.transformCoordinate(this.finish) : this.finish;
         }
     };
 
     @action resetPos = () => {
-        this.start = {x: null, y: null};
-        this.finish = {x: null, y: null};
-        this.transformedStart = {x: null, y: null};
-        this.transformedFinish = {x: null, y: null};
+        this.start = {x: 0, y: 0};
```
YuHsuan-Hwang, line 61: same comment

YuHsuan-Hwang (changes requested):
Just a minor style issue, but I prefer we set the start/finish label width same with the region dialog.


also seeing a warning when opening the dialog
image

src/components/ImageView/RegionView/RegionViewComponent.tsx
crocka: The region dialog uses td to wrap the start/end label instead of using label and labelInfo props. Do you mind if I change back to how it was before and follow the code style of region dialogs?

src/stores/Frame/DistanceMeasuringStore.ts
crocka: That way, the values shown on the input section will be NaN when users first open the dialog, rather than 0.

src/components/ImageView/RegionView/RegionViewComponent.tsx
YuHsuan-Hwang: I think following the code style of region dialogs works.

src/stores/Frame/DistanceMeasuringStore.ts
YuHsuan-Hwang: I see... maybe both is ok

crocka: Do you mean keep the initialization value to 0?

kswang1029 (approved):
no regression is observed from the current e2e tests, except xpath changes due to UI changes. A branch is created to address the xpath issues in this PR. Once this PR is merged, I will update the e2e test too.

Manual test looks fine. Nice work @crocka

src/stores/Frame/DistanceMeasuringStore.ts
YuHsuan-Hwang: Yes we can keep it

YuHsuan-Hwang (approved):

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
```diff
@@ -0,0 +1,241 @@
...
+        const MissingFrame = (
+            <NonIdealState className={"distance-measuring-settings-nonIdealState"} icon={"error"} title={"Distance measurement tool is not enabled"} description={"Please enable distance measurement tool via the image view toolbar."} />
+        );
+
+        const handleChangeWCSMode = (formEvent: React.FormEvent<HTMLInputElement>) => {
+            const WCSMode = formEvent.currentTarget.value === RegionCoordinate.Image ? false : true;
+            this.setWCSMode(WCSMode);
+        };
+
+        const handleValueChange = (event: React.KeyboardEvent<HTMLInputElement> | React.FocusEvent<HTMLInputElement>, isX: boolean, finish?: boolean, pixel?: boolean) => {
```
veggiesaurus: I'd prefer to have this as a class function, so it doesn't get regenerated each render

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
```diff
@@ -0,0 +1,241 @@
...
+            ) : (
+                <>
+                    <td>
+                        <FormGroup inline={true}>
+                            <SafeNumericInput
+                                selectAllOnFocus
+                                buttonPosition="none"
+                                value={finish ? distanceMeasuringStore?.finish.x : distanceMeasuringStore?.start.x}
+                                onBlur={event => handleValueChange(event, true, finish, true)}
+                                onKeyDown={event => {
```
veggiesaurus: can you make this a class function

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
```diff
@@ -0,0 +1,241 @@
...
+                    }
+                } else {
+                    event.currentTarget.value = finish ? (isX ? WCSFinish.x : WCSFinish.y) : isX ? WCSStart.x : WCSStart.y;
+                }
+            }
+
+            distanceMeasuringStore.updateTransformedPos(frame.spatialTransform);
+        };
+
+        const input = (finish: boolean) => {
```
veggiesaurus: might be easier to read if this was a separate function

src/components/Dialogs/RegionDialog/EllipticalRegionForm/EllipticalRegionForm.tsx
```diff
@@ -305,7 +304,7 @@ export class EllipticalRegionForm extends React.Component<{region: RegionStore;
                             <tr>
                                 <td>Coordinate</td>
                                 <td colSpan={2}>
-                                    <CoordinateComponent region={region} disableCooridnate={!this.props.wcsInfo} />
+                                    <CoordinateComponent region={region} disableCoordinate={!this.props.wcsInfo} />
```
veggiesaurus, line 307: nice catch :+1:

veggiesaurus (changes requested):

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+@observer
+export class DistanceMeasuringDialog extends React.Component {
+    constructor(props: any) {
+        super(props);
+        makeObservable(this);
+    }
+
+    @observable WCSMode: boolean = true;
+
+    @action setWCSMode = (bool?: boolean) => {
```
veggiesaurus: rename to `val` or `value`. `bool` is confusing as a variable name

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+                distanceMeasuringStore?.setFinish(value, distanceMeasuringStore?.finish.y);
+            } else if (finish) {
+                distanceMeasuringStore?.setFinish(distanceMeasuringStore?.finish.x, value);
+            } else if (isX) {
+                distanceMeasuringStore?.setStart(value, distanceMeasuringStore?.start.y);
+            } else {
+                distanceMeasuringStore?.setStart(distanceMeasuringStore?.start.x, value);
+            }
+        } else if (wcsInfo) {
+            const value = target.value;
```
veggiesaurus: you only use `value` as a string, so rather do the `as string` conversion here

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+            <>
+                <td>
+                    <FormGroup inline={true}>
+                        <SafeNumericInput
+                            selectAllOnFocus
+                            buttonPosition="none"
+                            value={finish ? distanceMeasuringStore?.finish.x : distanceMeasuringStore?.start.x}
+                            onBlur={event => this.handleValueChange(event, distanceMeasuringStore, wcsInfo, WCSStart, WCSFinish, true, finish, true)}
+                            onKeyDown={event => {
+                                if (event.type === "keydown" && event.key === KEYCODE_ENTER) this.handleValueChange(event, distanceMeasuringStore, wcsInfo, WCSStart, WCSFinish, true, finish, true);
```
veggiesaurus: please move to separate function

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+                </td>
+                <td>
+                    <FormGroup inline={true}>
+                        <SafeNumericInput
+                            selectAllOnFocus
+                            buttonPosition="none"
+                            value={finish ? distanceMeasuringStore?.finish.y : distanceMeasuringStore?.start.y}
+                            onBlur={event => this.handleValueChange(event, distanceMeasuringStore, wcsInfo, WCSStart, WCSFinish, false, finish, true)}
+                            onKeyDown={event => {
+                                if (event.type === "keydown" && event.key === KEYCODE_ENTER) this.handleValueChange(event, distanceMeasuringStore, wcsInfo, WCSStart, WCSFinish, false, finish, true);
```
veggiesaurus: same as above

veggiesaurus (changes requested):
minor changes requested

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
crocka: resolved

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
crocka: resolved

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
crocka: resolved

src/components/Dialogs/DistanceMeasuringDialog/DistanceMeasuringDialog.tsx
crocka: resolved

veggiesaurus (approved):

YuHsuan-Hwang:
@crocka Please merge dev and update changelog, thanks.

Issue #1602, opened 2021-08-04
corrupted spatial profile when cursor is moving


kswang1029:
this is tested with a 128x128 image:

https://user-images.githubusercontent.com/20819712/128116384-8b8c413c-3827-41e1-9d3b-73a564ee9358.mov

also tested with different image sizes:
- 512: working fine
- 256: working fine
- 128: problem
- 64: problem
- 32: problem

YuHsuan-Hwang:
fixed via #1980

PR #1980, opened 2022-08-17
Icchen/1602 corrupted spatial profile when cursor is moving

I-Chenn:
**Description**
This PR is linked to issue #1602.
The definition of `maxMip` in [this line](https://github.com/CARTAvis/carta-frontend/blob/dev/src/stores/Frame/FrameStore.ts#L196) is somewhat incomplete. If an image with size smaller than 256*256 is applied, the value of `maxMip` would be less than **1**, and may cause the `mip` defined in [this line](https://github.com/CARTAvis/carta-frontend/blob/dev/src/stores/widgets/SpatialProfileWidgetStore.ts#L191) less than **1**, which is a meaningless value for `mip`.

To fulfill the incompleteness, an "**if**" condition is added into the definition of `maxMip`, that is:
**If the height or the width of an image is less than 256, the value of `maxMip` is set to 1**
The modification now fix the issue. The evidence is shown below:
[spatial_profile.webm](https://user-images.githubusercontent.com/106953609/185171614-2b382cd5-2763-4f35-8a9e-acf547d651b0.webm)

**Checklist**


src/stores/Frame/FrameStore.ts
```diff
@@ -193,7 +193,11 @@ export class FrameStore {
     }
 
     @computed get maxMip(): number {
-        return Math.pow(2, Math.ceil(Math.log2(this.frameInfo.fileInfoExtended.width / TILE_SIZE)));
+        if (this.frameInfo.fileInfoExtended.width < 256 || this.frameInfo.fileInfoExtended.height < 256) {
```
YuHsuan-Hwang: use `TILE_SIZE` instead of 256. We don't use height in the calculation, so maybe we don't need to check `height < 256`.

YuHsuan-Hwang (changes requested):

src/stores/Frame/FrameStore.ts
I-Chenn: The suggestion is practiced with the modification of [431aa93](https://github.com/CARTAvis/carta-frontend/pull/1980/commits/431aa938b8cc56909ea1f9f023bdd746d4eb56cf).

YuHsuan-Hwang (approved):

veggiesaurus (approved):

kswang1029 (approved):
all e2e tests are passing. Manual test looks good as far as I can tell.

PR #1939, opened 2022-08-02
Icchen/1826 condition of enabling catalog selection button

I-Chenn:
**Description**
Fix issue #1826. Now the condition of enabling "Catalog selection" button is:
1. **Catalog overlay is activated**
or
2. **There is at least one source is plotted on the viewer**

Result:
1. When "Catalog overlay" is activated, the "Catalog selection" button is still forbidden to click (note the mouse cursor):
buttonNotEnabled
(sorry for taking the picture with cellphone camera since print screen may let mouse cursor disappear)

2. When sources are plotted on the viewer, the  "Catalog selection" button is enabled
buttonEnabledWhenSourceIsPlotted

3. Nevertheless, this modification still doesn't fix the "**the first source in the catalog table is highlighted in red when click elsewhere on the image**" problem (even when the "Catalog selection" button is not enabled yet...):
image

**Checklist**
- [x] changelog updated / ~~no changelog update needed~~
- [x] e2e test passing / ~~added corresponding fix~~
- [x] ~~protobuf updated to the latest dev commit~~ / no protobuf update needed

PR #1985, opened 2022-08-23
Icchen/1826 condition of enabling catalog selection button

### LINKED ISSUE #1826: only enable catalog selection button when there is a layer of catalog overlay
**Describe the bug**
Once a catalog file is loaded, the catalog selection button in the image viewer is enabled and if we click it and click the image elsewhere, we will see the first source in the catalog table is selected. The catalog selection button should only be enabled when a catalog image overlay exists. 

**To Reproduce**
1. load an image and a catalog file
2. observe the state of the catalog selection button in the image viewer and it is enabled.
3. click the catalog selection button and click elsewhere in the image 
4. we will see the first source in the catalog table is highlighted in red.

**Expected behavior**
only enable the catalog selection button when a catalog overlay exists on top of the raster image.

**Screenshots or videos**
Screen Shot 2022-04-12 at 14 21 36

 - OS: macOS Monterey
 - Browser chrome
 - Version 100.0.4896.60 (Official Build) (x86_64)

I-Chenn:
**Description**
This PR is linked to the issue #1826 and the previous PR #1939.

PR #1939 causes a soft crush when matching two images, which is described in the video below:
https://user-images.githubusercontent.com/106953609/186116962-6153e001-cdca-4e34-a801-2f5f9697d467.mov

The reason for causing the crush in PR #1939 is that: I previously checked for `frame` and `frame.spatialReference` when defining `baseFrame` (by using nullish coalescing operator). Nevertheless, when `spatialReference` is not null (multiple images are matching) and is adopted and then imported into `visibleCatalogFiles` function, since the `visibleCatalogFiles` function adopted only `visibleFrames` when operating, which may cause the soft crush if the user switches to "single panel" mode (since the single panel mode makes the other images not "visible").

The issue is fixed by adopting only `frame` when defining `baseFrame`. Since the `visibleCatalogFiles` function ([here](https://github.com/CARTAvis/carta-frontend/blob/dev/src/stores/CatalogStore.ts#L233)) has already taken the matching images into consideration (`spatialSiblings`), thus the modification should be OK to close #1826 without problem. The result is shown below:
[crushWhenMatching.webm](https://user-images.githubusercontent.com/106953609/186135350-a2b44245-1d2e-41b8-9a30-a92fceb80bfa.webm)

**Checklist**


YuHsuan-Hwang (approved):

kswang1029 (approved):
all e2e tests are passing. manual tests appear fine.

Issue #666, opened 2020-01-08
edge pixels artifacts on the right side


kswang1029:
could this due to #623 ?
![IRAS_17150-3224_sci spw27 cube I manual image pbcor fits-image-2020-1-8-16-53-22](https://user-images.githubusercontent.com/20819712/71964260-32cbdd00-3238-11ea-96f8-b0f24c4f46a7.png)

The requested tiles (original image is 2800 pixel x 2800 pixel) are
![Screen Shot 2020-01-08 at 16 50 38](https://user-images.githubusercontent.com/20819712/71964071-dd8fcb80-3237-11ea-8d71-a91941bc8ad1.png)

Also a question: why it requested 9 tiles (0,0) to (2,2), instead of 16 tiles from (0,0) to (3,3)?

kswang1029:
weirdly I cannot reproduce it anymore...... Close for now and I will keep this in mind.

kswang1029:
OK, it shows up _when_ javascript console is enabled.... even more weird....

kswang1029:
It shows up and this time no js console is enabled!!!
 I suspect it is there for a while but with the default NaN color it is harder to see. 

Attached (generated with "export as png" button) is the case with transparent NaN.

![M83 LSB_L pbcor fits-image-2020-6-5-10-37-50](https://user-images.githubusercontent.com/20819712/83830997-0ea5f680-a719-11ea-97a3-6c641ada2611.png)

with different zoom levels, the artifacts change.
![Jun-05-2020 10-44-21](https://user-images.githubusercontent.com/20819712/83831333-c3401800-a719-11ea-8b28-3091090d2d97.gif)

kswang1029:
More examples:

Right edge:
<img width="958" alt="Screen Shot 2020-06-18 at 10 57 52 AM" src="https://user-images.githubusercontent.com/20819712/84973031-aca2b380-b152-11ea-9918-bf1a1745e14f.png">

Top edge:
<img width="960" alt="Screen Shot 2020-06-18 at 10 57 01 AM" src="https://user-images.githubusercontent.com/20819712/84973039-b0363a80-b152-11ea-873c-04483128bce1.png">

veggiesaurus:
@kswang1029 this was bugging me, so I tried to fix it. Can you see if `angus/666_edge_pixels` solves the problem?

PR #1990, opened 2022-08-29
adjust texCoordsPixel

### LINKED ISSUE #949: raster image rendering precision at high(er) zoom level
In the example below, initially we have a 3x3 px square region covering a 3x3 px area. When zooming in at the bottom-left corner, at some point we start to see visible offset between the raster image and the region. 

Jun-18-2020 11-06-29

top-right corner
Screen Shot 2020-06-18 at 11 10 59 AM

YuHsuan-Hwang:
**Description**
Closes #949. The root cause is that `texCoordsPixel` loses precision when adding with non zero `uTileTextureOffset`.

code snippet to reproduce the issue:
```javascript
await app.openFile(PATH, "HD163296_CO_2_1.fits"); // load another file to make uTileTextureOffset !== 0
await app.delay(1000); // delay to avoid missing tiles

const file = await app.openFile(PATH, "image_4_4.fits"); // an image with size 4 x 4
file.regionSet.addRectangularRegion({x: 1, y: 1}, 1, 1);
file.regionSet.selectRegionByIndex(1);
file.setZoom(4000000);
```

before fix
image
after fix
image

**Checklist**

### DIFF
--- a/src/services/GLSL/pixel_shader_raster.glsl
+++ b/src/services/GLSL/pixel_shader_raster.glsl
@@ -53,7 +53,7 @@ void main(void) {
     // Mimics texel fetch in WebGL1
     vec2 tileCoordsPixel = vUV * uTileTextureSize;
     // Prevent edge artefacts
-    vec2 texCoordsPixel = clamp(tileCoordsPixel, 0.5, uTileTextureSize - 0.5) + uTileTextureOffset;
+    vec2 texCoordsPixel = clamp(floor(tileCoordsPixel), 0.5, uTileTextureSize - 0.5) + uTileTextureOffset;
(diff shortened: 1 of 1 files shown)


veggiesaurus:
Code looks good  :+1: We just need to make sure it doesn't introduce border artefacts (e.g. like #666)

kswang1029 (approved):
all e2e tests are passing. don't see artifacts so far. 👍

YuHsuan-Hwang:
> Code looks good 👍 We just need to make sure it doesn't introduce border artefacts (e.g. like #666)

So far I don't see borders on both small (4x4) and large (4096x4096) images.

PR #2002, opened 2022-09-02
carli/1947_mirror_cursor_when_spatially_matched

crocka:
**Description**
This is an enhancement functionality. The cursor is mirrored on all other spatially matched images to allow users to see the current corresponding cursor position on other images. An boolean observable is added to the AppStore called cursorMirror to keep track of the functionality status, and a button is added to the cursorInfo widget to toggle on/off. Once the functionality is turned on, a cross cursor appears on the reference image and it follows the user's cursor. When the user turn on the spatial matching to the reference image, a cross cursor appears on the corresponding image. Users can hover on any of the images that are spatially matched to each other to see the corresponding cursor location on all other spatially matched frames.

**Checklist**
- [x] protobuf updated to the latest dev commit / ~no protobuf update needed~

### DIFF
--- a/src/components/ImageView/RegionView/CursorRegionComponent.tsx
+++ b/src/components/ImageView/RegionView/CursorRegionComponent.tsx
@@ -18,8 +18,8 @@ export class CursorRegionComponent extends React.Component<CursorRegionComponent
         const appStore = AppStore.Instance;
         const frame = this.props.frame;
         const posImageSpace = frame?.cursorInfo?.posImageSpace;
-        //check the current frame is the spatialReg
-        const isSpatialMatchingOn = appStore.frames.some(thisFrame => thisFrame.spatialReference === frame);
+        //check the current frame is the spatialRef
+        const isSpatialMatchingOn: boolean = frame.secondarySpatialImages?.length ? true : false;
 
         if ((appStore.cursorFrozen || (appStore.cursorMirror && (frame?.spatialReference || isSpatialMatchingOn || frame === appStore.activeFrame))) && posImageSpace) {
             const rotation = frame.spatialReference ? (frame.spatialTransform.rotation * 180.0) / Math.PI : 0.0;
--- a/src/components/ImageView/RegionView/CursorRegionComponent.tsx
+++ b/src/components/ImageView/RegionView/CursorRegionComponent.tsx
@@ -15,13 +15,14 @@ interface CursorRegionComponentProps {
 @observer
 export class CursorRegionComponent extends React.Component<CursorRegionComponentProps> {
     render() {
+        const appStore = AppStore.Instance;
         const frame = this.props.frame;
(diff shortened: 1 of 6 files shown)


kswang1029:
few initial comments:
1. we cannot use space as the hotkey as it can trigger image matching if we just click the matching button in the image list then press space key, for example. Maybe we can consider the G key as it is next to the F key, although G means nothing here (or "group"). Open to other ideas. 

https://user-images.githubusercontent.com/20819712/188296401-99f67169-6ecb-492b-88d5-7e273aaba315.mov

2. if we enable cursor mirroring _and_ trigger cursor freezing with the F key, we will see a cross marker at an unexpected location on the unmatched images. This is nothing to do with this PR and this behavior exists in the dev. I suggest we use a different color for the matched _and_ mirrored cross just to differentiate this case.

Screen Shot 2022-09-04 at 11 46 22

kswang1029:
@crocka feature-wise this is good for a PR. 👍

src/components/ImageView/RegionView/CursorRegionComponent.tsx
```diff
@@ -15,13 +15,16 @@ interface CursorRegionComponentProps {
 @observer
 export class CursorRegionComponent extends React.Component<CursorRegionComponentProps> {
     render() {
+        const appStore = AppStore.Instance;
         const frame = this.props.frame;
         const posImageSpace = frame?.cursorInfo?.posImageSpace;
+        //check the current frame is the spatialReg
```
YuHsuan-Hwang: typo for spatialRef?

src/components/ImageView/RegionView/CursorRegionComponent.tsx
```diff
@@ -15,13 +15,16 @@ interface CursorRegionComponentProps {
...
+        const isSpatialMatchingOn = appStore.frames.some(thisFrame => thisFrame.spatialReference === frame);
```
YuHsuan-Hwang: Perhaps we can reuse `frame.secondarySpatialImages?.length`

YuHsuan-Hwang (commented):
looks good, just minor comments

src/components/ImageView/RegionView/CursorRegionComponent.tsx
crocka: Yes, it is a typo. It is fixed now.

src/components/ImageView/RegionView/CursorRegionComponent.tsx
crocka: Yes, it works the same way. It is updated now.

kswang1029 (approved):
feature-wise this is good. no regression is observed via e2e tests.

YuHsuan-Hwang (approved):

YuHsuan-Hwang:
@crocka Please merge dev and update changelog.

PR #2008, opened 2022-09-08
Icchen/1797 scroll selected region into region list view

### LINKED ISSUE #1797: scroll the selected region into the region list view
part from https://github.com/CARTAvis/carta/issues/131

When there are lots of regions loaded, the list in the region list is long and due to space limitation usually there are only few visible in the view. If one selects a region on the image viewer, the region in the list is highlighted but may not be visible in the view. It is better to support the automatic scroll-into-view feature to enhance the usability.

I-Chenn:
**Description**
This PR is to fix #1797.
A method called "scrollToSelected" is added to the RegionListComponent.tsx to let the selected region auto-scroll into the view of the region list. The result is shown below:

[Screencast from 09-08-2022 04:59:35 PM.webm](https://user-images.githubusercontent.com/106953609/189081318-b96cd777-e418-4a01-9524-2ec95898e011.webm)

**Checklist**

### DIFF
--- a/src/components/RegionList/RegionListComponent.tsx
+++ b/src/components/RegionList/RegionListComponent.tsx
@@ -72,12 +72,9 @@ export class RegionListComponent extends React.Component<WidgetProps> {
             () => AppStore.Instance.activeFrame?.regionSet?.selectedRegion?.regionId,
             id => {
                 if (id > 0) {
-                    const validRegionId = [];
-                    for (let i = 0; i < this.validRegions.length; i++) {
-                        validRegionId.push(this.validRegions[i].regionId);
-                    }
+                    const validRegionId = this.validRegions.map(el => el.regionId);
                     this.scrollToSelected(validRegionId.findIndex(element => element === id));
-                } else return;
+                }
             }
         );
     }
--- a/src/components/RegionList/RegionListComponent.tsx
+++ b/src/components/RegionList/RegionListComponent.tsx
@@ -72,7 +72,11 @@ export class RegionListComponent extends React.Component<WidgetProps> {
             () => AppStore.Instance.activeFrame?.regionSet?.selectedRegion?.regionId,
             id => {
                 if (id > 0) {
-                    this.scrollToSelected(id);
+                    const validRegionId = [];
+                    for (let i = 0; i < this.validRegions.length; i++) {
+                        validRegionId.push(this.validRegions[i].regionId);
+                    }
+                    this.scrollToSelected(validRegionId.findIndex(element => element === id));
                 } else return;
             }
         );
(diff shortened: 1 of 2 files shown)


kswang1029 (approved):
this looks fine. no regression is observed from e2e tests.

YuHsuan-Hwang (changes requested):
Found an issue when there are multiple images, for example: create 10 regions (id 1-10) on the first image and then create 10 regions (id 11-20) on the second image. If we select region id 11 on the second image, we always jump to the end of the list because we only display 10 regions (id 11-20).

I-Chenn:
@YuHsuan-Hwang 
New modification has been committed. Should have solved the issue when multiple images opened.
A rough test is shown below:
[multipleFrame.webm](https://user-images.githubusercontent.com/106953609/190393641-f5d77d55-49fa-4560-af56-151f1a0cf5b3.webm)
p.s. For some unknown reason, the screen recording didn't record the process when I change to the second image, but the result of the second image is successfully recorded.

src/components/RegionList/RegionListComponent.tsx
```diff
@@ -54,9 +55,31 @@ export class RegionListComponent extends React.Component<WidgetProps> {
...
+
     constructor(props: any) {
         super(props);
         makeObservable(this);
+
+        reaction(
+            () => AppStore.Instance.activeFrame?.regionSet?.selectedRegion?.regionId,
+            id => {
+                if (id > 0) {
+                    const validRegionId = [];
```
YuHsuan-Hwang: can be simplified by `Array.map`

src/components/RegionList/RegionListComponent.tsx
```diff
@@ -54,9 +55,31 @@ export class RegionListComponent extends React.Component<WidgetProps> {
...
+                    for (let i = 0; i < this.validRegions.length; i++) {
+                        validRegionId.push(this.validRegions[i].regionId);
+                    }
+                    this.scrollToSelected(validRegionId.findIndex(element => element === id));
+                } else return;
```
YuHsuan-Hwang: `else return` can be removed

YuHsuan-Hwang (changes requested):
The new calculation looks good. Just some minor comments:

src/components/RegionList/RegionListComponent.tsx
I-Chenn: The code is simplified in 7c3abcd.

src/components/RegionList/RegionListComponent.tsx
I-Chenn: Removed in 7c3abcd

YuHsuan-Hwang (approved):

kswang1029 (approved):
tested with a few thousands regions on matched and unmatched images. All appear fine. no regression is observed from e2e tests.

PR #2009, opened 2022-09-12
Ajm/self hosted GitHub runner

### LINKED ISSUE #2006: Improvement: Use Github Actions for carta-frontend CI
Similarly to the carta-backend, our current carta-frotnend CI is handled by Jenkins. The Jenkins instance runs locally on a server at ASIAA and sends the jobs out to local nodes at ASIAA. We can not guarantee that the Jenkins service and nodes are online and functional 24 hours a day, 365 days a year.

An improvement would be to let Github Actions handle the CI instead.

Although it would not be practical to use the free [Github-hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) for building the carta-frontend due to limited resources (e.g. only 2 cores per VM), we could instead use [self-hosted runners](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners).

The self-hosted runners would be running locally at ASIAA so we could still not guarantee availability 24 hours a day 365 days a year, but perhaps we could have additional self-hosted runners available at another site i.e. at IDIA. Additionally, in the unlikely event that no self-hosted runners are available, a last resort fall-back could be the slow cloud-based Github-hosted runners.

ajm-ska:
**Description**

This PR removes the Jenkinsfile and uses Github Actions with local runners to perform the frontend CI. I am open to comments and suggestions to improve it further.

Previously, we used Jenkins to run the frontend CI on one server. The WebAssembly stage ran first. The files were 'stashed' and then used to run the 3 production builds (node v12, v14, and v16) in a parallel stage. As it was all occurring on the same server, the 'stash' and 'unstash' step was practically instant. The total build time took ~8 minutes (~4 minutes for web assembly, and ~4 minutes for each parallel npm run build). 

On Github Actions, I first tried the same way to save the intermediary files ('artifacts'), but it is not practical as it is required to upload and download the saved data to and from the Github servers. I can't see any other way to share data between 'jobs'. Anyway, the 'artifact' process adds several minutes to the build time. However, it is not really necessary now. We can use 3 local runners, and each one can run the WebAssembly _and_ the node production build steps. Although there are some duplicated computations, the overall run time will be basically the same.

I propose we continue to use local Linux runners with only Docker installed. To run the frontend CI, I have created a simple Docker container based on Ubuntu 22.04. It has Emscripten 2.0.14 installed, and 'n' with NodeJS v12, v14, and v16. It is hosted on Docker Hub at `carta/frontend-builder` and I also show the Dockerfile here:

```
FROM ubuntu:22.04

# Install the basic packages required
RUN \
  apt-get update && \
  apt-get -y upgrade && \
  apt-get -y install cmake git lbzip2 nodejs npm python3 wget xz-utils

# Install n and other the nodejs versions
RUN \
 npm -g install n && \
 n 12 && n 14 && n 16

# Install emscripten
RUN \
 git clone https://github.com/emscripten-core/emsdk.git && \
 cd emsdk && \
 ./emsdk install 2.0.14 && \
 ./emsdk activate 2.0.14

RUN echo 'source "/emsdk/emsdk_env.sh"' >> /root/.bashrc

ENV HOME /root
WORKDIR /root

CMD ["bash"]
```

It looks like Github Actions can also run directly from Dockerfiles, so an alternative would be to place the Dockerfile in the carta-frontend repo and run it something like:

```
runs:
  using: 'docker'
  image: 'Dockerfile-frontend'
```

A few other points:

It seems that running a Ubuntu 22.04 container in Docker on a Ubuntu 20.04 server gives problems. I found a workaround [here](https://askubuntu.com/questions/1405417/20-04-vs-22-04-inside-docker-with-a-16-04-host-thread-start-failures) (Despite the title not really being the same as our problem, the error message is similar, and the suggested solution works). The fix is adding Docker flags: `--security-opt seccomp=unconfined`. Another solution would be to use an Ubuntu 20.04 based Docker container.

Even though Emscripten is automatically sourced when the container starts, it seemed that each step could not find emcc for some reason (even for steps that I didn't think required emcc), so for now I begin each step with `source /emsdk/emsdk_env.sh`.

So far, I have set up 3 local runners at ASIAA. One of them is slower than the others and takes ~12 minutes. The other two take ~7 and ~9 minutes. Previously we were using a faster server for the frontend Jenkins CI where it took ~8 minutes. So at present, the slower server that takes ~12 minutes is a bottleneck, but I don't think it is too much of a problem. 

Previously we used the Github Actions cloud server to perform the format-check. We continue to do so here. We had added a workaround to a bug related to `cache: 'npm'`. I should check if the issue is gone.

**Checklist**
- [x] no changelog update needed
- [x] e2e test passing / added corresponding fix - will have no affect on the e2e tests
- [x] protobuf no protobuf update needed


ajm-ska:
Yes. It appears we can add `cache: 'npm'` back to the format-check job.

veggiesaurus:
@ajm-asiaa I will go over this in more detail tomorrow.

veggiesaurus (approved):

veggiesaurus:
@ajm-asiaa this looks fine to me. Sorry for taking so long to get to it. Feel free to merge it in when ready :+1:

Issue #1953, opened 2022-08-04
Support loading swapped-axes image cubes (eg RA-Channel-Dec) (frontend)


kswang1029:
user feedback from v3 beta test

This is to support rotated cubes such as RA-Channel-Dec.

The frontend needs to be able to render the coordinate grid properly based on the image header.

The companion backend issue is https://github.com/CARTAvis/carta-backend/issues/1178

markccchiang:
When a swapped-axes (spectral vs. direction) image is loaded, the channel axis labels and unit (conversion from channel to degree) in the animation widget or spectral profile widget need to be changed. I guess it would be better to be done in a separate issue.

kswang1029:
done via #2011

PR #2011, opened 2022-09-13
Mark/support loading swapped axes images

markccchiang:
**Description**
This is intended to solve #1953, plotting coordinates for swapped-axes image cubes. It has to work with the [backend PR 1194](https://github.com/CARTAvis/carta-backend/pull/1194). To test it we may need more sample images with swapped-axes cubes.

**Checklist**
- [x] ~e2e test passing~ / added corresponding fix
- [x] protobuf updated to the latest dev commit ~/ no protobuf update needed~


kswang1029:
@acdo2002 please help on testing this feature, especially for the AST grid accuracy. Thanks.

wasm_src/ast_wrapper/ast_wrapper.cc
```diff
@@ -587,4 +587,158 @@ EMSCRIPTEN_KEEPALIVE float* fillTransformGrid(AstFrameSet* wcsInfo, double xMin,
...
+        return nullptr;
+    }
+
+    if (dirAxis < 1 || dirAxis > axisCount || spectralAxis < 1 || spectralAxis > axisCount)
+    {
+        std::cerr << "Bad axis index!\n";
+        return nullptr;
+    }
+
+    AstMapping* originMap = static_cast<AstMapping*> astGetMapping(originFrameSet, AST__CURRENT, AST__BASE);
```
crocka: Just curious, if we invert the order of AST_CURRENT and AST_BASE here, are we able to skip the astInvert() step at line 621?

markccchiang: I think you are right. Thank you for the comment!

src/stores/Frame/FrameStore.ts
```diff
@@ -102,9 +102,10 @@ export class FrameStore {
 
     public requiredFrameViewForRegionRender: FrameView;
 
-    public readonly wcsInfo: AST.FrameSet;
+    public wcsInfo: AST.FrameSet;
     public readonly wcsInfoForTransformation: AST.FrameSet;
     public readonly wcsInfo3D: AST.FrameSet;
+    public readonly wcsInfoSpectralVsDirection: AST.FrameSet;
```
YuHsuan-Hwang: I would prefer we add "3D" in the naming (e.g. `wcsInfo3DSpecVsDir` or `wcsInfo3DSwappedZ`) so that we don't misunderstood this FrameSet is in 2D with the term `SpectralVsDirection`.
I wonder if we can use `wcsInfo3D` without adding a new variable. We will need additional handling in `setSpectralReference` in this case though.

src/stores/Frame/FrameStore.ts
```diff
@@ -585,6 +581,140 @@ export class FrameStore {
         return false;
     }
 
+    @computed get isSwappedZ(): boolean {
```
YuHsuan-Hwang: We may remove `@computed` for `isSwappedZ`, `channelType`, `getDirAxisInfo`, and `isSwappedXY` because there are no `@observable` variables. (We have same issue with other variables like `isPVImage`, but we can leave them for future code maintenance.)

src/stores/Frame/FrameStore.ts
```diff
@@ -585,6 +581,140 @@ export class FrameStore {
...
+    @computed get channelType(): string {
+        if (this.isSwappedZ) {
+            const dirX = this.frameInfo.fileInfoExtended.axesNumbers.dirX;
+            const dirY = this.frameInfo.fileInfoExtended.axesNumbers.dirY;
+            const entries = this.frameInfo.fileInfoExtended.headerEntries;
+            const axis3 = entries.find(entry => entry.name.includes("CTYPE3"));
+            const axis4 = entries.find(entry => entry.name.includes("CTYPE4"));
+            let dirName;
+            if (dirX === 3 || dirY === 3) {
+                dirName = axis3?.value ? axis3?.value : "Unknown";
```
YuHsuan-Hwang: can be simplified to `axis3?.value ?? "Unknown"`

src/stores/Frame/FrameStore.ts
```diff
@@ -585,6 +581,140 @@ export class FrameStore {
...
+            } else if (dirX === 4 || dirY === 4) {
+                dirName = axis4?.value ? axis4?.value : "Unknown";
```
YuHsuan-Hwang: can be simplified to `axis4?.value ?? "Unknown"`

src/stores/Frame/FrameStore.ts
```diff
@@ -585,6 +581,140 @@ export class FrameStore {
...
+        // For direction vs. spectral image, get rendered direction axis index and size
+        const dirX = this.frameInfo.fileInfoExtended.axesNumbers.dirX;
+        const dirY = this.frameInfo.fileInfoExtended.axesNumbers.dirY;
+        const dirAxis = dirX < dirY ? dirX : dirY;
+        const dirAxisSize = dirAxis === 1 ? this.frameInfo.fileInfoExtended.width : this.frameInfo.fileInfoExtended.height;
+
+        // Get rendered and hidden direction axes formats
+        const entries = this.frameInfo.fileInfoExtended.headerEntries;
+        const axisName = entries.find(entry => entry.name.includes(`CTYPE${dirAxis}`));
+        let axisValue = axisName?.value ? axisName?.value : "Unknown";
```
YuHsuan-Hwang: can be simplified to `axisName?.value ?? "Unknown"`

src/stores/Frame/FrameStore.ts
```diff
@@ -585,6 +581,140 @@ export class FrameStore {
...
+    private get isProjDistort(): boolean {
+        const dirAxis = this.getDirAxisInfo?.dirAxis;
+        const dirAxisSize = this.getDirAxisInfo?.dirAxisSize;
+        let checkPoints = [1, 2, Math.floor(dirAxisSize / 2), dirAxisSize];
+        let wcsValues = [0, 0, 0, 0];
+        let coord = [1, 1];
+        const requiredChannel = this.requiredChannel + 1;
+        for (let i = 0; i < checkPoints.length; i++) {
+            coord[dirAxis - 1] = checkPoints[i];
+            const value = AST.transform3DPoint(this.wcsInfoSpectralVsDirection, coord[0], coord[1], requiredChannel, true);
```
YuHsuan-Hwang: Is the extra 1 displacement required?

src/stores/Frame/FrameStore.ts
```diff
@@ -585,6 +581,140 @@ export class FrameStore {
...
+        // For the case depth axis is direction axis
+
+        // Check if the direction axis is projection distortion
+        if (this.isProjDistort) {
+            return "WCS: non-linear";
+        }
+
+        // Calculate the start of world coordinate
+        const depthAxisFormat = this.getDirAxisInfo?.depthAxisFormat;
+        AST.set(this.wcsInfoSpectralVsDirection, `Format(3)=${depthAxisFormat}`);
```
YuHsuan-Hwang: `depthAxisInfo` is recalculated when `requiredChannel` is changed, but it seems that `depthAxisFormat` does not change with channel. Perhaps move this line to constructor.

src/stores/Frame/FrameStore.ts
```diff
@@ -585,6 +581,140 @@ export class FrameStore {
...
+        const requiredChannel = this.requiredChannel + 1;
+
+        // Lambda function for WCS transformation
+        let wcs = (channel: number): string => {
+            const wcs = AST.transform3DPoint(this.wcsInfoSpectralVsDirection, 1, 1, channel, true);
```
YuHsuan-Hwang: Is the 1 displacement required?

src/stores/Frame/FrameStore.ts
```diff
@@ -1416,6 +1543,62 @@ export class FrameStore {
...
+        return AST.getFrameFromFitsChan(fitsChan, false);
+    };
+
+    public updateSpectralVsDirectionWcs = () => {
+        if (this.wcsInfoSpectralVsDirection) {
+            const spectralAxis = this.frameInfo.fileInfoExtended.axesNumbers.spectral;
+            const dirAxis = this.getDirAxisInfo?.dirAxis;
+            const dirAxisSize = this.getDirAxisInfo?.dirAxisSize;
+            const dirFormat = this.getDirAxisInfo?.dirAxisFormat;
+            const requiredChannel = this.requiredChannel + 1;
```
YuHsuan-Hwang: Is the 1 displacement required?

src/stores/CatalogOnlieQueryConfigStore.ts
```diff
@@ -317,9 +317,11 @@ export class CatalogOnlineQueryConfigStore {
             const wcsCopy = AST.copy(frame.wcsInfo);
             let astString = new ASTSettingsString();
             const sys = system ? system : overlay.global.explicitSystem ? overlay.global.explicitSystem : SystemType.ICRS;
+            const xAxis = frame.frameInfo.fileInfoExtended.axesNumbers.dirX;
```
YuHsuan-Hwang: I get AST errors in the console when opening a casa generated axes-swapped image. Looks like it's related to catalog online query store.
image
test image: https://drive.google.com/file/d/1xeMP-RoUvHhiLpNDQCyefO6f5lnsf0t_/view?usp=sharing

src/stores/Frame/FrameStore.ts
```diff
@@ -585,6 +581,140 @@ export class FrameStore {
...
+
+        // Calculate the start of world coordinate
+        const depthAxisFormat = this.getDirAxisInfo?.depthAxisFormat;
+        AST.set(this.wcsInfoSpectralVsDirection, `Format(3)=${depthAxisFormat}`);
+        const requiredChannel = this.requiredChannel + 1;
+
+        // Lambda function for WCS transformation
+        let wcs = (channel: number): string => {
+            const wcs = AST.transform3DPoint(this.wcsInfoSpectralVsDirection, 1, 1, channel, true);
+            const wcsVal = wcs.z < 0 ? wcs.z + 2 * Math.PI : wcs.z;
```
YuHsuan-Hwang: Maybe `+ 2 * PI` is not always needed because dec can be negative.

YuHsuan-Hwang (changes requested):
Overall the changes are good. I have some feedback, mostly related to `FrameStore`:

src/stores/Frame/FrameStore.ts
markccchiang: Here I just to let this angle value be consistent with the AST coordinate shown on the image panel for galactic files. Maybe @kswang1029 can comment on it which is the proper way.

src/stores/Frame/FrameStore.ts
markccchiang: I just changed the name to `wcsInfo3DSwappedZ`. Will try to find if it can be combined with the use of `wcsInfo3D`.

src/stores/CatalogOnlieQueryConfigStore.ts
markccchiang: These error messages are from the AST  function `astSet` in the AST wrapper function `AST.getFormattedCoordinates`. It seems the AST can not accept direction axis numbers that are not 1 or 2. I have added a condition check to avoid such error messages.

src/stores/Frame/FrameStore.ts
YuHsuan-Hwang: Ok

src/stores/Frame/FrameStore.ts
kswang1029: The range of galactic/ecliptic longitude is 0\~360 degrees.
The range of galactic/ecliptic latitude is -90\~+90 degrees.

For the equatorial system such as (ICRS, FK5, FK4), 
RA in degrees ranges from 0\~360 and Dec in degrees ranges from -90\~+90.

kswang1029:
spotted that with some image files, the AST spatial grid is not rendered. checking with @markccchiang via slack.

acdo2002 (changes requested):
Based on the previous discussion with Mark, following is our content:
Once we face the data cube that the axis sequence is RA-Stokes-Dec-Channel, the Stokes should be **skipped** and the frontend should display the RA-Dec as x-y plane, however the current frontend displays the RA-Stokes as x-y plan.
Should the Stokes be skipped in the x-y plan? @kswang1029 please confirm the user story, thanks!!

There are two testing images (casa format) [here](https://drive.google.com/drive/folders/1xJbRMcmq-6ka_WCgUMyiwMcodhQP45fg?usp=sharing): 
 HH211_IQU-swap-rsdf.image
 HH211_IQU-swap-dsrf.image

I found if opening the images with "Stokes" in the second axis, then click "Open image" or "Append image", then the frontend will crash, the backend is still alive without no crash.

kswang1029:
> Based on the previous discussion with Mark, following is our content: Once we face the data cube that the axis sequence is RA-Stokes-Dec-Channel, the Stokes should be **skipped** and the frontend should display the RA-Dec as x-y plane, however the current frontend displays the RA-Stokes as x-y plan. Should the Stokes be skipped in the x-y plan? @kswang1029 please confirm the user story, thanks!!
> 
> There are two testing images (casa format) [here](https://drive.google.com/drive/folders/1xJbRMcmq-6ka_WCgUMyiwMcodhQP45fg?usp=sharing): HH211_IQU-swap-rsdf.image HH211_IQU-swap-dsrf.image
> 
> I found if opening the images with "Stokes" in the second axis, then click "Open image" or "Append image", then the frontend will crash, the backend is still alive without no crash.

We should just keep the order as it is and prevent crash.

kswang1029:
@markccchiang please wait my test report before making any code changes. Thanks.

markccchiang:
> I found if opening the images with "Stokes" in the second axis, then click "Open image" or "Append image", then the frontend will crash, the backend is still alive without no crash.

In FrameStore the "channel" by default refers to the spectral axis. To deal with the swapped axes images, the related codes may need to be changed to avoid confusion.

acdo2002:
@markccchiang I see, unfortunately I think there are more frontend codes need to be change.
Please check the report [here](https://docs.google.com/document/d/1YnsNzq9RvtH4kPVor-8ZexfpzVviayGs6IQVZWeicvc/edit?usp=sharing), and please wait @kswang1029 's testing report.

@YuHsuan-Hwang mentioned add isSwappedZ in the FrameStore right, maybe can use it to disable the beamOverlay.

kswang1029:
to handle the general cases like
(0:ra, 1:dec, 2:stokes, 3:channel) ==> 24 combinations
(0:ra, 1:dec, 2:channel, 3:stokes) ==> 24 combinations
(0:ra, 1:dec, 2:channel) ==> 6 combinations
(0:ra, 1:dec, 2:stokes) ==> 6 combinations

we need to incorporate a general rule in the current PR to ignore stokes (if it exists) first, then from the rest, we pick up the 1st as x-axis and the 2nd as y-axis. The rest (if it exists) is z-axis (iterable via the animator). stokes (if it exists) remains as the polarization axis.

markccchiang:
> to handle the general cases like (0:ra, 1:dec, 2:stokes, 3:channel) ==> 24 combinations (0:ra, 1:dec, 2:channel, 3:stokes) ==> 24 combinations (0:ra, 1:dec, 2:channel) ==> 6 combinations (0:ra, 1:dec, 2:stokes) ==> 6 combinations
> 
> we need to incorporate a general rule in the current PR to ignore stokes (if it exists) first, then from the rest, we pick up the 1st as x-axis and the 2nd as y-axis. The rest (if it exists) is z-axis (iterable via the animator). stokes (if it exists) remains as the polarization axis.

@kswang1029 and @acdo2002 please check does the new commit, including the relative new backend commit solves the problem. Thank you! I will also try to refactor the codes, especially in the FramStore.

kswang1029:
We are almost there for the feature testing. The remaining issues are
1. for (RA, Stokes, Dec, FREQ), (Dec, Stokes, RA, FREQ), (Stokes, RA, Dec, FREQ), and (Stokes, Dec, RA, FREQ), the rendered pixel is not square.
2. for (RA, Stokes, FREQ, Dec), (Dec, Stokes, FREQ, Ra), (Stokes, RA, FREQ, Dec), (Stokes, Dec, FREQ, RA), (Stokes, FREQ, RA, Dec), and (Stokes, FREQ, Dec, RA), the spectral axis is displayed as Frequency (Hz). Need to display this as ($SPECSYS) Frequency (GHz) like other cubes do. The spectral axis might be in velocity so we need to refer to the CTYPE header.
3. "Right ascension" -> "Right Ascension" 

As for gridline accuracy, it looks good as far as I can tell. 

In the extreme case with a wide field image, if the directional axis does not vary monotonically, AST would not render the grid line. This is fine for now.

kswang1029:
FYI. Working on fixing and add basic e2e tests for this new feature now. Once done and the above issues are fixed I will approve this PR.

kswang1029:
@markccchiang the above issues are fixed 👍 The PV image needs a fix when we switch to x:spectral, y:spatial mode. The AST grid becomes linear index axes in this case.
Screen Shot 2023-03-01 at 08 15 15

acdo2002:
One suggestion (tested images are in [here](https://drive.google.com/drive/u/1/folders/1xJbRMcmq-6ka_WCgUMyiwMcodhQP45fg))

Canvas information does not update if loading “ngc3079-swap-rfd.fits” or “ngc3079-swap-fdr.fits”, the upper information of freq/vel is not changing if the cursor is moving along the y-axis (now is freq) and x-axis (now is freq), respectively.

Screen Shot 2023-03-01 at 6 20 57 PM

The Frequency(LSRK) and Velocity in the upper panel of the canvas should change as the cursor moving along the y-axis(Freq in GHz), now is the fixed value (changed with z-axis i.e. Dec or RA).

After this suggestion has been fixed, and the ICD-RxJS tests passed, I will approve this PR

markccchiang:
Since this PR is growing large, I will try to refactor the code if there are no problems remaining.

acdo2002 (approved):
Based on my [report](https://docs.google.com/document/d/1YnsNzq9RvtH4kPVor-8ZexfpzVviayGs6IQVZWeicvc/edit)

Almost the problems are all solved. About PV diagram for the swapped image.
Here I have a test, if I set a line region (x coordinate starts from 400.5 and ends to -0.5; y coordinate starts from 20 to ends to 20) in "ngc3079-swap-rfd.fits" and making a PV diagram (e.g. ngc3079-swap-rfd_pv-ConvertToChannel20Image.fits), then the generated PV diagram should be a 2D image in "ngc3079-swap-rdf.fits" of channel of 20 (e.g. ngc3079-swap-rdf-Channel20Image-drop.fits).
And the generated PV is consistent to 2D image of channel 20 (two images subtracted and all pixels are 0).
This is because the line region length is odd number (if line region length is not odd number then, the generated PV may not 100% match to the 2D image), this part requires more investigation (my feeling is more related to PV rather than supporting swapped image)

I have tested this backend to the current ICD-RxJS, there is one test needs update, once the corresponding backend has been merged, then I can update the test (FILEINFO_FITS_MULTIHDU.test.ts) code very soon. 

Please go ahead for the code refactoring and code reviewing.

kswang1029:
> Based on my [report](https://docs.google.com/document/d/1YnsNzq9RvtH4kPVor-8ZexfpzVviayGs6IQVZWeicvc/edit)
> 
> Almost the problems are all solved. About PV diagram for the swapped image. Here I have a test, if I set a line region (x coordinate starts from 400.5 and ends to -0.5; y coordinate starts from 20 to ends to 20) in "ngc3079-swap-rfd.fits" and making a PV diagram (e.g. ngc3079-swap-rfd_pv-ConvertToChannel20Image.fits), then the generated PV diagram should be a 2D image in "ngc3079-swap-rdf.fits" of channel of 20 (e.g. ngc3079-swap-rdf-Channel20Image-drop.fits). And the generated PV is consistent to 2D image of channel 20 (two images subtracted and all pixels are 0). This is because the line region length is odd number (if line region length is not odd number then, the generated PV may not 100% match to the 2D image), this part requires more investigation (my feeling is more related to PV rather than supporting swapped image)
> 
> I have tested this backend to the current ICD-RxJS, there is one test needs update, once the corresponding backend has been merged, then I can update the test (FILEINFO_FITS_MULTIHDU.test.ts) code very soon.
> 
> Please go ahead for the code refactoring and code reviewing.

Again, for the current purpose, we don’t focus on any region related analyses. It is not the scope here, whether the results are correct or not.  This is a doc capturing all affected analyses https://docs.google.com/document/d/1eNLZt-VxSrzeYaOFjajJqbKXzIOo-u29xL7g3Cstz2c/edit

kswang1029:
@markccchiang when a swapped cube is loaded, we will still need to display `wcs: ( , )` in the cursor info bar according to the displayed axes. If it is RA vs freq, then we display something like `wcs: (12:30:30, 1400.0)`, for example. This is to be consistent with what we do for PV images. 

Screen Shot 2023-03-07 at 11 46 37

kswang1029 (approved):
feature-wise this is looking good for the current scope - supporting image visualization of a swapped-axes cube. 👍 Please proceed code review.

src/stores/Frame/FrameStore.ts
```diff
@@ -1449,6 +1639,80 @@ export class FrameStore {
...
+    };
+
+    public updateSpectralVsDirectionWcs = () => {
+        if (this.wcsInfo3D) {
+            this.wcsInfo = AST.makeSwappedFrameSet(this.wcsInfo3D, this.dirAxis, this.spectral, this.requiredChannel, this.dirAxisSize);
+            AST.set(this.wcsInfo, `Format(${this.dirAxis})=${this.dirAxisFormat}, Unit(${this.dirAxis})=""`);
+        }
+    };
+
+    private getDirAxisInfo() {
```
YuHsuan-Hwang: I would prefer we name it `updateDirAxisInfo`. also prefer using an arrow function.

src/stores/Frame/FrameStore.ts
```diff
@@ -1536,11 +1800,14 @@ export class FrameStore {
 
             const normalizedNeighbourhood = cursorNeighbourhood.map(pos => AST.normalizeCoordinates(this.wcsInfo, pos.x, pos.y));
 
+            let precisionX = 0;
```
YuHsuan-Hwang: `getCursorInfo.precision` is returning zeros. Shall we remove these two variables?

YuHsuan-Hwang (commented):
The changes are looking good! I only have two minor comments.

YuHsuan-Hwang (approved):

kswang1029:
for some reason [e49334d](https://github.com/CARTAvis/carta-frontend/pull/2011/commits/e49334dec827e98980e812bd86b55fc61b844703) introduces test failures for regions. world coordinates of the region are not updating. There are errors from AST.

Screen Shot 2023-03-14 at 14 50 59

kswang1029 (approved):
looking good. new e2e tests are implemented.

PR #2016, opened 2022-09-19
OIDC serverside

daikema:
**Description**

Supersedes / replaces #1861, with the OIDC login moved to the serverside to eliminate a dependence on public clients which aren't accepted by all identity providers.  Note that the scope of changes required in the frontend has been significantly reduced due to the resulting similarity with the local authentication providers supported by the server such that the changes herein are expected to be minimal.  (The bulk of the changes to support OIDC login can be found in [this pull request in the carta-controller repo](https://github.com/CARTAvis/carta-controller/pull/129))

**Checklist**
- [?] changelog updated / no changelog update needed

Given the short nature of the change here I'm not sure that a changelog update is needed beyond what might be needed for the carta-controller.  (That said, I don't see a corresponding changelog in the carta-controller repo).

- [?] e2e test passing / added corresponding fix

Not sure how to run these tests but I suspect that the changes here are small enough in scope that it's unlikely to trigger new test failures.

- [X] protobuf updated to the latest dev commit / no protobuf update needed

### DIFF
--- a/src/services/ApiService.ts
+++ b/src/services/ApiService.ts
@@ -189,7 +189,7 @@ export class ApiService {
             try {
                 await this.axiosInstance.post(ApiService.RuntimeConfig.logoutAddress);
             } catch (err) {
-                if (err.response.status == 404) {
+                if (err.response.status === 404) {
                     // OIDC logout requires GET for logout vs POST for other mechanisms
                     window.open(ApiService.RuntimeConfig.logoutAddress, "_self");
                     return; // avoid later potential dashboard redirect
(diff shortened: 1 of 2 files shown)


src/services/ApiService.ts
```diff
@@ -184,10 +184,18 @@ export class ApiService {
         if (ApiService.RuntimeConfig.googleClientId) {
             this.authInstance?.signOut();
         } else if (ApiService.RuntimeConfig.logoutAddress) {
+            // The controller will assume an existing login session exists if this exists
+            localStorage.removeItem("authenticationType");
             try {
                 await this.axiosInstance.post(ApiService.RuntimeConfig.logoutAddress);
             } catch (err) {
-                console.log(err);
+                if (err.response.status == 404) {
```
veggiesaurus: replace with `===`

veggiesaurus (changes requested):

src/services/ApiService.ts
daikema: Done

veggiesaurus (approved):

PR #2017, opened 2022-09-23
Yuhsuan/1802 catalog overlay refactor

### LINKED ISSUE #1662: The size of catalog overlay on Apple M1 and with Retina display

### LINKED ISSUE #1802: Migrate from point sprites to manually drawn rectangles for catalogue rendering
We need to move from using `GL.POINTS` to draw aliased points when rendering the catalogue overlay. This has some severe limiations, as the maximum point size is heavily browser and GPU dependent. Instead, we should use a vertex-pulling approach to draw two triangles per data point.

YuHsuan-Hwang:
**Description**

Closes #1802 and closes #1662: migrated from `GL2.POINTS` to `GL2.TRIANGLES` for catalog overlay rendering

changes:
* change `drawArrays` function
* use texture instead of attribute for source position data in vertex shader
* add uniform `uZoomLevel` for calculating point size in px unit in vertex shader
* use varying `v_pointCoord` for `gl_PointCoord` in pixel shader
* move `getOffsetFromId` from vector overlay vertex shader to utilities

**Checklist**
- [ ] changelog updated / ~no changelog update needed~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed NaN pixel value in the cursor info bar of the image viewer when the image is 1x1 pixel ([#1879](https://github.com/CARTAvis/carta-frontend/issues/1879)).
 * Fixed issue to show cursor info of smoothed profiles in the spatial and spectral profilers ([#1880](https://github.com/CARTAvis/carta-frontend/issues/1880), [#1938](https://github.com/CARTAvis/carta-frontend/pull/1938)).
 * Fixed mean and RMS not updating when smoothing in the spatial and spectral profilers ([#1838](https://github.com/CARTAvis/carta-frontend/issues/1838)).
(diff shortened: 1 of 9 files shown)


veggiesaurus:
This is looking good from a code change perspective @YuHsuan-Hwang 👍

YuHsuan-Hwang:
Please note that seldom there are subtle color changes. I think the difference is minor and can be ignored.
For example, this branch / dev branch:
image

kswang1029 (approved):
no regression is observed from the e2e tests. Manual tests with non-retina and Retina display,  different markers, and size/orientation/color mappings appear fine. 👍

veggiesaurus (approved):

PR #2019, opened 2022-09-27
Carli/267 support image annotation

crocka:
**Description**
- PreferenceDialogComponent.tsx
Added annotationSettingsPanel in the annotation tab. 
getPointShape function that return the Icon component for point annotation shape selection.

- AppearanceForm.tsx
Added font size, font, and font style selection if the region is compass, text, or ruler. The textOffsetForm() returns the form that sets x and y offset

- CompassRulerRegionForm.tsx
This is the region setting form opened by double clicking compass or ruler annotation. The textOffsetForm() returns the form that sets x and y offset for texts in compass or ruler annotations. The coordinateInput() returns the form that sets the controlpoints. The compassArrowTipForm() returns the form that sets the arrow length, arrow width, and toggle for showing or hiding the arrow tips of the north and east arrow. RulerAuxiliaryLineForm() sets the dash length and toggle of the ruler auxiliary lines. handleValueChange() calls the setControlPoint() from regionStore to set the controlPoints.

- LineRegionForm.tsx
Added vectorArrowTipForm() that returns the form that sets the arrow length and width of the vector arrow tip.

- PointRegionForm.tsx
Added Selection for point annotation shape and point width. The point shape selection is achieve by renderShapePopOver, which calls getPointShape(), which call Icon component from blueprintjs or IconWrapper() that returns a svg component. 

- RectangularRegionForm.tsx
Added textAnnotationPositionSelect that sets the relative text position in the textbox.

- CompassAndRulerAnnotationComponent.tsx
  - Compass Annotation
Added the compass and ruler annotation React functional component. The CompassAnnotation is comprised of 2 arrow or line konva components, and 2 text konva components. There are 3 anchors for the compass. 2 anchors are at the tips of the arrow/line that controls the length of the arrow/line, and the 3rd anchor is at the origin of the arrow where the 2 arrow/line join that moves the compass location. The structure of the controlpoints of the compass is [{x: x-position, y: y-position}, {x: length, y: length}], where x-position and y-position are the x and y coordinate of the origin anchor, and length is the length of the arrow/line in pixel coordinate. The length of the north and east arrow/line stay the same. The length, pointerWidth, pointerLength, textOffset, and fontSize are relative to the imageViewing area instead of the image. Therefore, the values in image coordinate are multiplied by a factor imageRatio / zoomLevel to go to the coordinate relative to screen. 
getOffset() is called when the compass annotation is first rendered using React.useEffect, with empty array dependency. The getOffset() applies offset to text konva component that translates the text labels in the direction where the north/east arrows are pointing by half of the text width and text height.
The lists of points on the arrow/line curves are obtained from getRegionApproximation() from CompassAnnotationStore. Please view below for me detail. The lists of points are transformed to canvas position using transformedImageToCanvasPos(). Then, since the getRegionApproximation() outputs the lists of points longer than the length, for-loops goes through each of the points for north arrow/line and east arrow/line and compare the distance from the point to the origin. If the distance is shorter than the set length of the particular compass annotation, the point is saved in an other array that is used by the arrow/line konva component to render the compass. Once that the distance of the points are longer than the compass length, the for-loop breaks. 
     - Dragging
The handleDragStart(), handleDrag(), and handleDragEnd() handle the dragging of the compass annotation. The handleDragStart() calls te beginEditing() from regionStore and current position of the mouse using React.useRef. The handleDrag() convert the mousePosition ref (old position at the start of the dragging action) and current mouse position from canvas to image position. The difference between the two position is calculated using subtract2D and the controlPoints[0] of the compass is set by adding on to the difference (controlPoints[1] of the compass is the length). * At the beginning, we planned to drag the compass by dragging anywhere on the compass, that's why the mousePosition ref is needed to store where is the initial location, but with further discussion, we decided to drag the compass only by dragging the origin anchor. The step to save the starting postion of the cursor is not really needed now, but it is kept at this moment if there are further changes.
    - Anchor action
The handleAnchorMouseEnter() and handleAnchorMouseOut() are used to change the cursor style when hovering on the anchor points.
The handleAnchorDragStart(), handleAnchorDrag(), and handleAnchorDragEnd() are similar to handling dragging actions. The handleAnchorDragStart() calls the beginEditing() from regionStore. In handleAnchorDrag(), the new position is transformed to image position and the difference between the new position and the origin anchor is calculated as the length, the length is subtracted by the frame.zoomLevel / imageRatio to go from screen coordinate to image coordinate.
  - Ruler Annotation
Ruler Annotation is comprised of line and text konva components, more lines are added if the auxiliary line toggle is on. An invisible closed shape is drawn inside the ruler annotation (if auxiliary lines are shown) to facilite clicking, dragging, and double clicking. 
    - getDistanceText() is used to calculate the distance between the starting and finishing points. It simply calls the AST geodesicDistance to calculate the distance. Then, it is formatted to the desired unit in string.
    - Other parts of the ruler annotation is similar to the implementation of compass annotation. It uses the getRegionApproximation from RulerAnnotationStore to generate the list of points on the ruler curves  to render the curved lines under different WCS. To keep the center of the text align with the distance curve, the text konva component must be offseted by half of the width of the text, and this is achieved using React.useEffect that listens to changes in zoomLevel and user-defined text offset to keep the text centered. Dragging and anchor dragging works the same as compass annotation.

- InvariantShapes.tsx
  - Changed handleSquareDraw() to handlePointShapeDraw() which used shape konva component to draw custom shapes: circle, circle_lined, diamond, diamond_lined, cross, x, and rectangle. 

- LineSegmentRegionComponent.tsx
  - Line and arrow konva components require similar props other than pointer width and length. Otherwise, if the region is vector, arrow component is used rather than the line component.

- RegionViewComponent.tsx
  - Extended the existing code to accomodate regions that are annotations, nothing new.

- SimpleShapeRegionComponent.tsx
  - getTextProps() generates the props for text annotation. Since the text annotation is put inside a rectangular textbox, the text can be positioned anywhere inside the box using the align and verticalAlign text konva component props and a switch statement to choose the values to the props. Other code changes in this file are just extending current implementations to text annotations.

- ToolbarComponent.tsx
  - Added an annotationMenu to select annotation to draw, similar to regionMenu. The annoationMenu is opened as a submenu inside the regionMenu, using MenuItem component from blueprintjs. Also, the popover for region selection button now is changed to "Create x region" or "Create x annotation", where x is the name of the region/annotation. If the active region (region/annotation that the user has selected to draw) is an annotation, the icon of the anchorButton on the toolbar is shown as a pen, instead of rectangle, line, ellipse, etc icons.

- AppStore.ts
  - regionStyles in protobuf
    - An additional key-value pair is added to the regionStyles enum in protobuf called annotationStyle, which is an enum that contains the list of properties such as compass text label, font, fontsize, point annotation width, shape, etc. 
  - In *exportRegions(), a switch statement is used to identify the required fields in annotationStyles, and add them to a JavaScript object called annotationStyles, which will be saved in regionStyles and sent as a message to the backend for region/annotation export.

- RegionSetStore.ts
  - The changes in this file are mostly extending current code to support annotations. Below are some of the notable differences.
  - migrateRegionsFromExistingSet()
    - This function is called when the images are spatially matched. It searches regions that are different in all images and copies regions that are unique to each image to all other images. The way that the function compares whether the regions are unique is through timestamp of when the region is last modified. For instance, setControlPoints() modifies the timestamp where as appearance changes such as setColor() do not trigger changes in timestamp. However, after some discussion, the team decided that the appearance changes such as font, fontSize, fontStyle, text compass text labels, pointShape, vector pointer width and height, etc are important for annotations and should modify timestamp. Therefore, when spatial matching is enabled, the annotationStyles are saved and unique annotations are copied to the new images and annotationStyles are applied to the newly cloned annotations.
    - addRegion()
      - This function instantiates regions/annotations. Since some annotations have their own annotationStore class, a switch statement is used to instantiate the required region/annotation. 

- RegionStore.ts
  - the modifiedTimestamp property is modified if the region is an annotation inside of setColor(), setLineWidth(), and setDashLength(). Otherwise, other code changes are extension to support annotations.

- PreferenceStore.ts
  - PreferenceKeys enum and the DEFAULTS object are extend to include fields for annotation preference settings. Other code changes follow the same style are the original code to account for annotation preference settings.

- ast_wrapper.cc
  - Added pointList() that takes in 2 inputs points and outputs the list of points on the geodesic curve between the two input points. This function is used to draw ruler annotation. First, the geodesic distance between the start and end point is calculated. The distance is then divided by npoint, which is defined by programmer to be the number of points in the output point list. The larger npoint is, the more calculation is needed, but the curve would look smoother. The distance interval is obtained and the x y coordinates of the points on the geodesic curve is obtained using the astOffset. A for-loop loops through 1 * distance interval, 2 * distance interval... and store the point in the x and y output array. The output points are transformed to pixel coordinate and a for-loop is used to format the x y output array to a single output array.
  - Added axPointList() that takes in 2 points and outputs the list of points on the curve along the direction of the axis incrementation. The implementation is identical to pointList() but instead of astOffset() astAxOffset() is used.

- post.ts
  - Added transformPointList() used to wrap pointList() from ast_wrapper.cc. It first allocate memory in the heap, then calls pointList ast function and then extract the array from the heap to a regular Javascript array.
  - Added transformAsPointList() used to wrap axPointList() ast function. The implementation is identical to transformPointList().

**Checklist**
- [ ] e2e test passing / added corresponding fix


kswang1029:
@crocka There is a small issue with the polygon annotation type. If it is polygon region, there is a feature to warn users if the being created polygon is complex or simple by changing temporarily its color to magenta as demonstrated below:

https://user-images.githubusercontent.com/20819712/215680133-6f95e7df-a250-4148-b50c-bf3d8776e9b8.mov

If we _modify_ a complex polygon region back to a simple one, the warning color disappears.

However, if it is polygon annotation, once the color becomes magenta, it stays as magenta no matter what new color is set by users.

https://user-images.githubusercontent.com/20819712/215680446-52691af2-d6cb-48cc-8151-012655026763.mov

I suggest we disable this "warning" feature for the polygon annotation. So when creating or modifying a polygon annotation, the color stays the same as its default (orange color).

kswang1029:
@crocka for the compass annotation, the layout of the settings dialog becomes messy if we switch the coordinate reference from world to image. It is due to the `wcs: ....` string wrapping. 

https://user-images.githubusercontent.com/20819712/215685471-2944170d-63bd-44ff-9e4b-ab339d7d52f2.mov

kswang1029:
feature testing looks good as far as I can tell. Please switch to formal PR. Code review may proceed.

kswang1029:
@crocka if we create a text annotation in the crtf/ds9 format manually without specifying a font size, the rendered font size on screen changes with zoom levels. The annotation dialog says the font size is 0.5 but changing it does not change the rendered size. For this use case, could you assign a default size (20?) if the font size is not defined in the region file?

```
#CRTFv0 CASA Region Text Format version 0
text[[355.94pix, 202.05pix], 'CO_2-1']
```

Once this is fixed, please turn this draft as a formal PR. Code review may start now.

kswang1029:
> @crocka if we create a text annotation in the crtf/ds9 format manually without specifying a font size, the rendered font size on screen changes with zoom levels. The annotation dialog says the font size is 0.5 but changing it does not change the rendered size. For this use case, could you assign a default size (20?) if the font size is not defined in the region file?
> 
> ```
> #CRTFv0 CASA Region Text Format version 0
> text[[355.94pix, 202.05pix], 'CO_2-1']
> ```
> 
> Once this is fixed, please turn this draft as a formal PR. Code review may start now.

@crocka now I see the font size is set to 20. But when I change it, the rendered text size stays the same. In addition, when I zoom the image, the rendered text size on screen scales with zoom actions. Could you have another look please?

https://user-images.githubusercontent.com/20819712/220835485-9c4f84c4-f818-4ba7-af07-c50641e02184.mov

src/components/Dialogs/RegionDialog/CompassRulerRegionForm/CompassRulerRegionForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+                region?.setControlPoint(0, {x: value, y: region?.controlPoints[0].y});
+            } else {
+                region?.setControlPoint(0, {x: region?.controlPoints[0].x, y: value});
+            }
+        } else if (wcsInfo) {
+            const value = target.value as string;
+            if (isX && isWCSStringFormatValid(value, AppStore.Instance.overlayStore.numbers.formatTypeX)) {
+                if (finish) {
+                    const finishPixelFromWCS = getPixelValueFromWCS(wcsInfo, {...WCSFinish, x: value});
+                    region?.setControlPoint(1, {x: finishPixelFromWCS.x, y: finishPixelFromWCS.y});
```
YuHsuan-Hwang: Could we simplify to `region?.setControlPoint(1, finishPixelFromWCS);`?

src/components/Dialogs/RegionDialog/CompassRulerRegionForm/CompassRulerRegionForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+                    </FormGroup>
+                </td>
+            </>
+        );
+    };
+
+    private compassTextOffsetForm = (isNorth: boolean) => {
+        return (
+            <>
+                {this.props.region.regionType === CARTA.RegionType.ANNCOMPASS && (
```
YuHsuan-Hwang: We can move it to the front to reduce nesting: `if (this.props.region.regionType !== CARTA.RegionType.ANNCOMPASS) { return null; }`

src/components/Dialogs/RegionDialog/CompassRulerRegionForm/CompassRulerRegionForm.tsx
```diff
@@ -0,0 +1,288 @@
...
+        WCSFinish: WCSPoint2D,
+        isX: boolean,
+        finish?: boolean,
+        pixel?: boolean
+    ) => {
+        const target = event.target as HTMLInputElement;
+
+        if (pixel) {
+            const value = parseFloat(target.value);
+            if (!isFinite(value)) return;
```
YuHsuan-Hwang: add new line for `return`

src/components/Dialogs/RegionDialog/LineRegionForm/LineRegionForm.tsx
```diff
@@ -531,71 +531,73 @@ export class LineRegionForm extends React.Component<{region: RegionStore; frame:
...
-                                        </td>
-                                        <td>
-                                            <SafeNumericInput selectAllOnFocus={true} buttonPosition="none" placeholder="P.A." value={region.rotation} onBlur={this.handleRotationChange} onKeyDown={this.handleRotationChange} />
-                                        </td>
-                                    </tr>
-                                </React.Fragment>
-                            ) : null}
-                        </tbody>
-                    </table>
+            <>
```
YuHsuan-Hwang: Could we remove the fragment?

src/components/Dialogs/RegionDialog/RectangularRegionForm/RectangularRegionForm.tsx
```diff
@@ -597,75 +612,88 @@ export class RectangularRegionForm extends React.Component<{region: RegionStore;
...
-                                        placeholder="P.A."
-                                        value={region.rotation}
-                                        onBlur={this.handleRotationChange}
-                                        onKeyDown={this.handleRotationChange}
-                                    />
-                                </td>
-                            </tr>
-                        </tbody>
-                    </table>
+            <>
```
YuHsuan-Hwang: Could we remove the fragment?

src/components/ImageView/RegionView/CompassAndRulerAnnotationComponent.tsx
```diff
@@ -0,0 +1,562 @@
...
+export const CompassAnnotation = observer((props: CompassRulerAnnotationProps) => {
+    const shapeRef = React.useRef();
+    const northLabelRef = React.useRef<Konva.Text>();
+    const eastLabelRef = React.useRef<Konva.Text>();
+    const frame = props.frame;
+    const region = props.region as CompassAnnotationStore;
+    const mousePoint = React.useRef({x: 0, y: 0});
+
+    const handleClick = (event: Konva.KonvaEventObject<MouseEvent>) => {
+        console.log("selecting");
```
YuHsuan-Hwang: remove console

src/components/ImageView/RegionView/CompassAndRulerAnnotationComponent.tsx
```diff
@@ -0,0 +1,562 @@
...
+        props.onSelect(region);
+    };
+    const handleDoubleClick = (event: Konva.KonvaEventObject<MouseEvent>) => {
+        console.log("double clicking");
```
YuHsuan-Hwang: remove console

src/components/ImageView/RegionView/CompassAndRulerAnnotationComponent.tsx
```diff
@@ -0,0 +1,562 @@
...
+    }
+
+    // Dummy variables for triggering re-render
+    /* eslint-disable no-unused-vars, @typescript-eslint/no-unused-vars */
+    const system = AppStore.Instance.overlayStore.global.explicitSystem;
+    const darktheme = AppStore.Instance.darkTheme;
+    const title = frame.titleCustomText;
+    /* eslint-enable no-unused-vars, @typescript-eslint/no-unused-vars */
+
+    const getOffset = () => {
```
YuHsuan-Hwang: I would suggest we rename it to `updateOffset`

src/components/ImageView/RegionView/CompassAndRulerAnnotationComponent.tsx
```diff
@@ -0,0 +1,562 @@
...
+
+        region.setNorthTextOffset((northXOffset * zoomLevel) / imageRatio, true, true);
+        region.setNorthTextOffset((northYOffset * zoomLevel) / imageRatio, false, true);
+        region.setEastTextOffset((eastXOffset * zoomLevel) / imageRatio, true, true);
+        region.setEastTextOffset((eastYOffset * zoomLevel) / imageRatio, false, true);
+    };
+
+    React.useEffect(() => {
+        getOffset();
+        // eslint-disable-next-line
```
YuHsuan-Hwang: remove line

src/components/ImageView/RegionView/CompassAndRulerAnnotationComponent.tsx
```diff
@@ -0,0 +1,562 @@
...
+            hitStrokeWidth: NEW_ANCHOR_MAX_DISTANCE * 2
+        };
+    };
+
+    return (
+        <>
+            <Group ref={shapeRef} listening={!region.locked} onClick={handleClick} onDblClick={handleDoubleClick} onDragStart={handleDragStart} onDragEnd={handleDragEnd} onDragMove={handleDrag}>
+                {region.eastArrowhead ? <Arrow {...generateProps(false)} /> : <Line {...generateProps(false)} />}
+                {region.northArrowhead ? <Arrow {...generateProps(true)} /> : <Line {...generateProps(true)} />}
+                <Text
```
YuHsuan-Hwang, line 241: We can use `...commonProps` for `Text` and `Anchor`.

src/components/ImageView/RegionView/InvariantShapes.tsx
```diff
@@ -1,9 +1,11 @@
 import {Group, Shape} from "react-konva";
+import {CARTA} from "carta-protobuf";
+import Konva from "konva";
 
 import {AppStore} from "stores";
 
-const POINT_WIDTH = 6;
-const POINT_DRAG_WIDTH = 13;
+// const POINT_WIDTH = 6;
+// const POINT_DRAG_WIDTH = 13;
```
YuHsuan-Hwang: remove comment

src/components/ImageView/RegionView/InvariantShapes.tsx
```diff
@@ -35,20 +68,25 @@ interface PointProps {
...
     onClick: (ev) => void;
     onDblClick: (ev) => void;
+    pointShape?: CARTA.PointAnnotationShape;
+    pointWidth?: number;
 }
 
 export const Point = (props: PointProps) => {
-    const handlePointDraw = (ctx, shape) => {
-        HandleSquareDraw(ctx, shape, POINT_WIDTH);
+    const pointWidth = props.pointWidth && props.pointWidth !== 0 ? props.pointWidth : 6;
```
YuHsuan-Hwang: We can define a const for 6, ex `DEFAULT_POINT_WIDTH`.

src/components/ImageView/RegionView/InvariantShapes.tsx
```diff
@@ -35,20 +68,25 @@ interface PointProps {
...
+    const handlePointDraw = (ctx: Konva.Context, shape: Konva.Shape) => {
+        HandlePointShapeDraw(ctx, shape, pointWidth, props.pointShape);
     };
 
-    const handlePointBoundDraw = (ctx, shape) => {
-        HandleSquareDraw(ctx, shape, POINT_DRAG_WIDTH);
+    const handlePointBoundDraw = (ctx: Konva.Context, shape: Konva.Shape) => {
+        HandlePointShapeDraw(ctx, shape, 7 + pointWidth);
```
YuHsuan-Hwang: We can define a const for 7, ex `POINT_HOVER_DIST`.

src/components/Shared/PointShapeSelectComponent/PointShapeSelectComponent.tsx
```diff
@@ -0,0 +1,71 @@
...
+import {CARTA} from "carta-protobuf";
+import {observer} from "mobx-react";
+
+import {AppStore} from "stores";
+
+export const PointShapeSelectComponent = observer((props: {handleChange: (pointShape: CARTA.PointAnnotationShape) => void; pointShape: CARTA.PointAnnotationShape}) => {
+    const appStore = AppStore.Instance;
+    const preference = appStore.preferenceStore;
+
+    const IconWrapper = (path: React.SVGProps<SVGPathElement>, color: string, fill: boolean, strokeWidth = 2, viewboxDefault = 16) => {
```
YuHsuan-Hwang: Shall we use camelCase `iconWrapper`?

src/stores/AppStore/AppStore.ts
```diff
@@ -2336,7 +2366,7 @@ export class AppStore {
             this.setIsExportingImage(true);
             this.setImageRatio(imageRatio);
             this.waitForImageData().then(() => {
-                const backgroundColor = this.preferenceStore.transparentImageBackground ? "rgba(255, 255, 255, 0)" : this.darkTheme ? "rgba(0, 0, 0, 1)" : Colors.WHITE;
+                const backgroundColor = this.preferenceStore.transparentImageBackground ? "rgba(255, 255, 255, 0)" : this.darkTheme ? Colors.DARK_GRAY3 : Colors.LIGHT_GRAY5;
```
YuHsuan-Hwang: Changes from new dev commit is reverted.

src/stores/Frame/AnnotationStore.ts
```diff
@@ -0,0 +1,556 @@
...
+        this.northArrowhead = northArrowhead;
+        this.modifiedTimestamp = performance.now();
+    };
+
+    @action setEastArrowhead = (eastArrowhead: boolean) => {
+        this.eastArrowhead = eastArrowhead;
+        this.modifiedTimestamp = performance.now();
+    };
+
+    public getRegionApproximation(astTransform: AST.FrameSet, spatiallyMatched?: boolean, spatialTransform?: AST.FrameSet): any {
```
YuHsuan-Hwang: I prefer we check the type instead of using `any`.

src/stores/Frame/AnnotationStore.ts
```diff
@@ -0,0 +1,556 @@
...
+        font: Font;
+        pointerWidth: number;
+        pointerLength: number;
+        length: number;
+        northTextOffset: Point2D;
+        eastTextOffset: Point2D;
+        isNorthArrow: boolean;
+        isEastArrow: boolean;
+    }) => {
+        console.log(annotationStyles);
```
YuHsuan-Hwang: remove console.log

src/stores/Frame/AnnotationStore.ts
```diff
@@ -0,0 +1,556 @@
...
+    @action setTextOffset = (offset: number, isX: boolean) => {
+        if (isX) {
+            this.textOffset = {...this.textOffset, x: offset};
+        } else {
+            this.textOffset = {...this.textOffset, y: offset};
+        }
+        this.modifiedTimestamp = performance.now();
+    };
+
+    public getRegionApproximation(astTransform: AST.FrameSet, spatiallyMatched?: boolean): any {
```
YuHsuan-Hwang: I prefer we check the type instead of using `any`.

src/stores/Frame/Region/RegionSetStore.ts
```diff
@@ -110,31 +155,165 @@ export class RegionSetStore {
         if (dashes?.length) {
             region.dashLength = dashes[0];
         }
+
+        if (annotationStyles) {
+            switch (regionType) {
+                case CARTA.RegionType.ANNPOINT:
+                    const pointAnnotation = region as PointAnnotationStore;
```
YuHsuan-Hwang: We can avoid creating a copy: `(region as PointAnnotationStore).initializeStyles(annotationStyles);`

src/stores/Frame/Region/RegionSetStore.ts
```diff
@@ -110,31 +155,165 @@ export class RegionSetStore {
...
+                    this.preference.regionLineWidth,
+                    this.preference.regionDashLength,
+                    rotation,
+                    regionName
+                );
+                break;
+        }
+
         this.regions.push(region);
+        //Need to be removed
```
YuHsuan-Hwang: Is this a ToDo?

src/stores/PreferenceStore/PreferenceStore.ts
```diff
@@ -195,6 +204,17 @@ const DEFAULTS = {
         regionCreationMode: RegionCreationMode.CENTER,
         regionSize: 30
     },
+    ANNOTATION: {
+        annotationColor: "#FFBA01",
+        annotationLineWidth: 2,
+        annotationDashLength: 0,
+        annotationType: CARTA.RegionType.ANNRECTANGLE,
+        annotationCreationMode: RegionCreationMode.CENTER,
```
YuHsuan-Hwang: Looks like annotation creation mode and annotation size are not used. We use the settings in the region tab. Shall we remove these two settings?

wasm_src/ast_wrapper/ast_wrapper.cc
```diff
@@ -420,6 +419,105 @@ EMSCRIPTEN_KEEPALIVE int transform(AstFrameSet* wcsinfo, int npoint, const doubl
...
+    double* yOut = new double[npoint];
+    
+    for(int i = 0; i < npoint; i++) {
+        double distance = discreteDist * i;
+        astOffset(wcsinfo, start, finish, distance, output);
+        xout[i] = output[0];
+        yout[i] = output[1];
+    }
+
+    astTran2(wcsinfo, 201, xout, yout, 0, xOut, yOut);
```
YuHsuan-Hwang: Could we use `npoint` for 201?

wasm_src/ast_wrapper/post.ts
```diff
@@ -194,6 +196,54 @@ Module.transformPointArrays = function (wcsInfo: number, xIn: Float64Array, yIn:
     return result;
 };
 
+Module.getGeodesicPointArray = function (wcsInfo: number, npoint: number, start: {x: number, y: number}, finish: {x: number, y: number}) {
+    // Return empty array if arguments are invalid
+    const yIn = new Float64Array([start.y, finish.y]);
+    const xIn = new Float64Array([start.x, finish.x]);
+
+    if (!(xIn instanceof Float64Array) || !(yIn instanceof Float64Array) || xIn.length !== yIn.length) {
+        return {x: new Float64Array(1), y: new Float64Array(1)};
```
YuHsuan-Hwang: `return new Float64Array(1);` because we return one array.

wasm_src/ast_wrapper/post.ts
```diff
@@ -194,6 +196,54 @@ Module.transformPointArrays = function (wcsInfo: number, xIn: Float64Array, yIn:
...
+    }
+
+    // Allocate and assign WASM memory
+    const N = xIn.length;
+    const xInPtr = Module._malloc(N * 8);
+    const yInPtr = Module._malloc(N * 8);
+    const outPtr = Module._malloc(201 * 2 * 8);
```
YuHsuan-Hwang: Could we use `npoint` for `201`? Same for all `201` in `getGeodesicPointArray` and `getAxisPointArray`.

src/stores/Frame/Region/RegionSetStore.ts
```diff
@@ -110,31 +155,165 @@ export class RegionSetStore {
...
-            this.preference.regionDashLength,
-            rotation,
-            regionName
-        );
+    private addRegion(points: Point2D[], rotation: number, regionType: CARTA.RegionType, temporary: boolean = false, isAnnotation: boolean = false, regionId: number = this.getTempRegionId(), regionName: string = "") {
+        let region: RegionStore;
+
+        switch (regionType) {
+            case CARTA.RegionType.ANNCOMPASS:
+                region = new CompassAnnotationStore(
```
YuHsuan-Hwang: Can we avoid repeating the input parameters?

YuHsuan-Hwang (changes requested):
Some suggestions about the code changes. I think most of them will not largely affect the feature.

src/components/ImageView/RegionView/CompassAndRulerAnnotationComponent.tsx
crocka: If I remove the line, it will cause warning, and wouldn't able to compile.

YuHsuan-Hwang: Oh I see. We can disable just one rule: `// eslint-disable-next-line react-hooks/exhaustive-deps`

kswang1029 (approved):
feature-wise this is looking good. A set of annotation e2e tests is prepared. No regression is seen from e2e tests.

YuHsuan-Hwang (approved):

src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
```diff
@@ -98,14 +132,126 @@ export class RegionDialogComponent extends React.Component {
...
+                                {region.regionType === CARTA.RegionType.ANNCOMPASS && (
+                                    <div className="form-contents">
+                                        <FormGroup inline={true} label="Show North Arrowhead">
+                                            <Switch
+                                                checked={(region as CompassAnnotationStore).northArrowhead}
+                                                onChange={(ev: React.ChangeEvent<HTMLInputElement>) => (region as CompassAnnotationStore).setNorthArrowhead(ev.target.checked)}
+                                            />
+                                        </FormGroup>
+                                        <FormGroup inline={true} label="Show East Arrowhead">
+                                            <Switch checked={(region as CompassAnnotationStore).eastArrowhead} onChange={(ev: React.ChangeEvent<HTMLInputElement>) => (region as CompassAnnotationStore).setEastArrowhead(ev.target.checked)} />
```
veggiesaurus: Only comment I have is that I think we should rather use a dropdown here for arrowheads (north, east, both)

image

That would stop the line from wrapping at the default dialog width. @kswang1029 what do you think?

veggiesaurus (changes requested):
One tiny issue: it looks like there's an extra `;` floating around somewhere:

image

Other than that (and the other comment that I'd like @kswang1029 to weigh in on), this is great! :rocket:

src/components/Dialogs/RegionDialog/RegionDialogComponent.tsx
kswang1029: We plan to have a new region/annotation dialog which would have two tabs, one for region props and the other for styling. So I guess we can leave it as it is for now. As for using a dropdown, that’s a good idea.

crocka: I updated the selection to drop down. Please take a look at the updated code and ui.

kswang1029: ui side looks good

PR #2022, opened 2022-09-29
add variable for gradient visibility

### LINKED ISSUE #1908: Scripting support: allow hiding of colorbar gradient element
In the `OverlayColorbarSettings`, every element of the colorbar can be shown or hidden independently, except for the colour gradient itself.

This is inconvenient for the scripting use case of fetching different layers of the image view separately and vectorising some of them. While the border, numbers, etc., are drawn elements which are vectorisable, the gradient is not, and should be kept as a raster image.

It is currently possible to save just the gradient to its own layer (by hiding everything else), but it is not possible to save everything _except_ the gradient to its own layer. An additional boolean flag for gradient visibility would solve this problem.

YuHsuan-Hwang:
**Description**
Closes #1908: the colorbar gradient can be hidden by `app.overlayStore.colorbar.setGradientVisible(false)`.

**Checklist**
- [X] ~changelog updated~ / no changelog update needed

### DIFF
--- a/src/components/ImageView/Colorbar/ColorbarComponent.tsx
+++ b/src/components/ImageView/Colorbar/ColorbarComponent.tsx
@@ -163,7 +163,7 @@ export class ColorbarComponent extends React.Component<ColorbarComponentProps> {
                 height={rectHeight}
                 fillLinearGradientStartPoint={rectGradientStart}
                 fillLinearGradientEndPoint={rectGradientEnd}
-                fillLinearGradientColorStops={frame.renderConfig.colorscaleArray}
+                fillLinearGradientColorStops={colorbarSettings.gradientVisible ? frame.renderConfig.colorscaleArray : null}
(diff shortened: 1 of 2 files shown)


veggiesaurus (approved):

confluence (approved):
I have confirmed that this works in the scripting wrapper (but it's `setGradientVisible` with a small s in both interfaces, right?).

confluence:
Steps to test in the Python wrapper:
```python
from carta.session import Session
from carta.constants import Overlay
session = Session.interact("TOKEN_URL", SESSION_ID)
# Open any image
# Hide
session.call_overlay_action(Overlay.COLORBAR, "setGradientVisible", False)
# Show
session.call_overlay_action(Overlay.COLORBAR, "setGradientVisible", True)
```

YuHsuan-Hwang:
> I have confirmed that this works in the scripting wrapper (but it's `setGradientVisible` with a small s in both interfaces, right?).

@confluence Right, it should be a small s. I have revised the description.

kswang1029 (approved):

PR #2031, opened 2022-10-13
Icchen/1948 display long filename properly

### LINKED ISSUE #1948: display long filename properly
user feedback from v3 beta test

Images from the ALMA Science Archive usually have long file names such as 

> member.uid___A001_X879_X8dc.J1924-2914_pol.spw0_1_2_3.mfs.IQU.manual.image.pbcor.fits

with the v3 release, mostly when the space is not sufficient, only the first few characters are displayed so sometimes we see identical file name from different images.

The request is to have a better way to handle long file names so that users can easily read the full file names to know which image is which. 

We may consider to use tooltip to show the full file name. For example:

https://user-images.githubusercontent.com/20819712/182782205-4a104f4d-194a-4a6d-8027-a19d7ae7d9d7.mov

We may also consider to show first part and last part of the long file name with ... in between like what macOS does.

I-Chenn:
**Description**
This PR is to resolve #1948.

Added tooltips to the **Image** column for **Image List** and **Cursor Info** overlay. The results are shown below:
image
image

Although the modification should be OK for the requested issue, what concerns me is that:

Since the table of **Cursor Info** is generated with [SimpleTableComponent](https://github.com/CARTAvis/carta-frontend/blob/dev/src/components/Shared/Tables/SimpleTableComponent.tsx) method and all the columns are rendered as a whole, it is hence hard to put in tooltip for a specific column. Therefore I added a condition to the [SimpleTableComponent](https://github.com/CARTAvis/carta-frontend/blob/dev/src/components/Shared/Tables/SimpleTableComponent.tsx) to present tooltip only when column name is "Image", like this:
image

This does perfectly add the tooltip that we want, but to add a specific condition into a universal method is somewhat... inappropriate I would say.

Nevertheless, I haven't yet come up with a better idea, please let me know if you have any better one about it.

**Checklist**

For the pull request:


kswang1029:
initial test appears fine.

kswang1029:
manual tests appear fine. no regression is observed from e2e tests. Please proceed code review and let me know if user test is needed after code revision (if any).

kswang1029 (approved):

I-Chenn:
@kswang1029 @YuHsuan-Hwang @veggiesaurus @crocka 
Sorry for a bit delay. The latest commit (82aa62a) of this branch is ready for review now.
In dcdb102 and 82aa62a, I added two props: `isTooltip` and `tooltipIndex` and one method: `getTooltip` to provide customized tooltip to `SimpleTableComponent.tsx`, which is adopted by `CursorInfoComponent.tsx` to create table. The `Image` column of `Cursor Info` overlay is now tooltipped:
image

kswang1029 (approved):
this works fine as far as I can tell. no regression is observed from e2e tests.

src/components/Shared/Tables/SimpleTableComponent.tsx
```diff
@@ -15,6 +15,8 @@ export class SimpleTableComponentProps {
     isIndexZero?: boolean;
     boldIndex?: number[];
     updateTableRef?: (ref: Table) => void;
+    isTooltip?: boolean;
```
YuHsuan-Hwang: Just a minor suggestion, but perhaps we can use just `tooltipIndex` and check `typeof this.props.tooltipIndex !== "undefined"` so that we don't need to set an additional `isTooltip` boolean.

I-Chenn: Agree.
The suggestion is accomplished with the latest commit 283d4cd.

crocka (approved):

YuHsuan-Hwang (approved):

PR #2035, opened 2022-10-26
thchang / 458 rename layout

### LINKED ISSUE #458: rename saved layouts
I made a mistake naming a layout and want to rename it.  There doesn't seem to be a way to rename layouts, only to delete or save.  This is a minor issue as one can always delete and re save.  But it would be slick to be able to rename saved layout.

TienHao:
**Description**

closes #458. Added the rename button in the save layout dialog which users can rename layout name directly.

**Checklist**

For the pull request:


kswang1029:
@TienHao overall this looks good. 👍 There are few things to be considered:
1) if we rename a layout with a name same as other existing layouts, a popup will show up as a warning. Then it disappeared. I suggest once the user clicked the ok button, we show the rename pop up again so users can try another name.
2) please add a check for unix/linux reserved characters and words and prevent users using those as part of the layout name. They are (https://www.cyberciti.biz/faq/linuxunix-rules-for-naming-file-and-directory-names/)
Screen Shot 2022-10-28 at 11 37 20

Please help me to check if there are other forbidden characters.

I tried `>` and terrible things happened if I try to open it with vi in the terminal. 🤦🏻‍♂️ Also tried a space and it works but it is weird to have a space as the layout's name/json file name.

TienHao:
@kswang1029

 I have fixed the first item.

About the space character in second item, I think it is OK to have that since we have space characters in several preset layout like "Cube View", "Cube Analysis". Or, we replace all space characters in preset layouts once a space is not allowed in a layout name. either way is ok for me.

I added a alert tooltips to be shown when a forbidden character has be entered.截圖 2022-11-30 下午3 01 50

kswang1029:
> @kswang1029
> 
> I have fixed the first item.
> 
> About the space character in second item, I think it is OK to have that since we have space characters in several preset layout like "Cube View", "Cube Analysis". Or, we replace all space characters in preset layouts once a space is not allowed in a layout name. either way is ok for me.
> 
> I added a alert tooltips to be shown when a forbidden character has be entered.截圖 2022-11-30 下午3 01 50

ok to have space in the names but we should not allow a name with a single space only.

kswang1029:
@TienHao I tried `\` so the saved file is `\.json`. Odd thing happens when I try to use vi to open it. So I think we will need to include `\` as forbidden char.  `;` needs to be included too as if we do `vi ;.json` the file is not loaded.

kswang1029:
@TienHao on a related issue: in fact when we try to "save" a layout, we should also check special char. Could you also fix it too in the same PR?

TienHao:
@kswang1029 

I have added `\` and `;` into the forbidden char set. The special char check in the "save" layout has already been implemented.

kswang1029:
> @kswang1029
> 
> I have added `\` and `;` into the forbidden char set. The special char check in the "save" layout has already been implemented.

@TienHao I tried all possible special char and loaded the file with `vi` and compiled the following list to see which should be forbidden (due to zsh/bash semantics). Please include those with "no" to the forbidden list. 

```
Char	canUse?
========================
~	no
`	no
!	no
@	yes
#	yes
$	yes
%	yes
^	yes
&	no
*	no
(	no
)	no
-	no
_       yes
+	no
=	no
[	no	
]	yes
{	yes
}	yes
\	no
|	no
;	no
:	no
'	no
<	no
>	no
.	no
,	yes
?	no
/	no 
```

src/stores/LayoutStore.ts
```diff
@@ -185,6 +195,64 @@ export class LayoutStore {
...
 
+    @flow.bound *renameLayout(oldName: string, newName: string) {
+        const appStore = AppStore.Instance;
+
+        if (!this.layouts || !newName || !this.dockedLayout) {
+            appStore.alertStore.showAlert("Save layout failed! Empty layouts or name.");
+            return;
+        }
+
+        if (appStore.layoutStore.layoutExists(newName)) {
```
YuHsuan-Hwang: can be simplified to `this.layoutExists`

src/stores/LayoutStore.ts
```diff
@@ -185,6 +195,64 @@ export class LayoutStore {
...
+            appStore.alertStore.showAlert("Layout name already exists.");
+            return;
+        }
+
```
YuHsuan-Hwang: Shall we also check for preset names as in `saveLayout`?

src/stores/LayoutStore.ts
```diff
@@ -185,6 +195,64 @@ export class LayoutStore {
...
+        appStore.dialogStore.hideSaveLayoutDialog();
+
+        // save layout to layouts[] & server/local storage
+        const configToSave = this.layouts[oldName];
+        this.layouts[newName] = configToSave;
+        if (!PresetLayout.isPreset(this.layoutNameToBeSaved)) {
+            try {
+                const success = yield appStore.apiService.setLayout(newName, configToSave);
+                if (success) {
+                    appStore.apiService.clearLayout(oldName).then(
```
YuHsuan-Hwang: Could we use `yield` here? Maybe also adjust `deleteLayout` to use `@flow.bound` and `yield`.

YuHsuan-Hwang (changes requested):

src/stores/LayoutStore.ts
YuHsuan-Hwang: @TienHao Maybe we can remove `then()` by
```
try {
    const success = yield appStore.apiService.setLayout(newName, configToSave);
    if (success) {
        const success = yield appStore.apiService.clearLayout(oldName);
        if (success) {
            delete this.layouts[oldName];
        }
        this.handleRenameResult(oldName, newName, success);
    }
} catch (err) {
    console.log(err);
    this.handleRenameResult(oldName, newName, false);
}
```
We can also try to remove `then()` in `deleteLayout` similarly.

src/components/Dialogs/LayoutDialog/SaveLayoutDialogComponent.tsx
```diff
@@ -54,13 +54,24 @@ export class SaveLayoutDialogComponent extends React.Component {
...
+        this.clearInput();
+    };
+
     @computed get isEmpty(): boolean {
-        return !this.layoutName;
+        return !this.layoutName?.trim();
+    }
+
+    @computed get validName(): boolean {
+        return this.layoutName.match(/^[^~`!*()-+=[.'?<>/|\\:;&]+$/)?.length > 0;
```
kswang1029: @TienHao could you check why the minus sign "-" is _not_ detected as invalid char?

src/stores/LayoutStore/LayoutStore.ts
```diff
@@ -186,30 +196,81 @@ export class LayoutStore {
...
+        const appStore = AppStore.Instance;
+        if (!layoutName || !this.layoutExists(layoutName)) {
+            appStore.alertStore.showAlert(`Cannot delete layout ${layoutName}! It does not exist.`);
+            return;
+        }
+
+        try {
+            const success = yield appStore.apiService.clearLayout(layoutName);
+            if (success) {
+                delete this.layouts[layoutName];
```
YuHsuan-Hwang, line 265: The new changes look good! Just one minor question: do we need to preserve the lines `if (layoutName === this.currentLayoutName) {  this.currentLayoutName = ""; }` here?

TienHao: I am thinking that these lines still need to be preserved. If a user delete a current layout (eg. _layout_A_) without clear `this.currentLayoutName`, and then rename another layout(eg. _layout_B_) or add a new layout with the deleted name (_layout_A_), the current layout hint will be misleading.

kswang1029 (approved):
looks good. please proceed code review

src/stores/LayoutStore/LayoutStore.ts
YuHsuan-Hwang: Oh, I see we also have them in `handleDeleteResult`. I missed that.

YuHsuan-Hwang (approved):

PR #2070, opened 2022-12-21
Icchen/1930 persistent connectivity popup

### LINKED ISSUE #1930: Hovering popup no longer useful for obtaining session ID for interactive scripting
The current way to get the session ID from the frontend is a popup which appears when you mouse over the green light in the top-right corner and disappears when you mouse away. This was fine when IDs were small integers which the user could copy from memory, but now they are long UUIDs which need to be copied and pasted. It's impossible to do this using the current popup. The only practical way to copy the ID is to look at the backend terminal output, which is not possible with SDM, which we now support in the scripting wrapper.

Suggestion: change the mouseover element to a persistent popup / menu. Possibly also add a button inside the popup to copy the ID with one click.

I-Chenn:
**Description**
This draft-PR is to resolve #1930. The modification includes:
1. Change the `connectivityTooltip` from `Tooltip2` to `Popover2` to make it persistent.
2. Add a button to copy the session ID to clipboard for later use.

The result is shown below:
https://user-images.githubusercontent.com/106953609/208884219-5bb6ce50-c7a7-4be1-a153-338ebd7cc545.mov

World like to discuss with @kswang1029  and @YuHsuan-Hwang:
Is it OK to change the `connectivityTooltip` from `Tooltip2` to `Popover2`?

And would like to ask @confluence:
Is this is the effect you want?

**Checklist**

For the pull request:


kswang1029:
@I-Chenn this looks good. Just few suggestions:
1) we need to beautify it with css (eg, add padding to boarder, use blueprint button, etc).
2) we also need another button as Copy session URL (handy for the python scripting interface setup)
3) after clicking the copy button, we can use "toast" (please refer to blueprintjs) to show a confirmation such as "Session ID copied to clipboard".

YuHsuan-Hwang:
> Is it OK to change the connectivityTooltip from Tooltip2 to Popover2?

`isOpen` prop in `Tooltip2` could be another solution, but I'm ok with both `Tooltip2` and `Popover2`

kswang1029:
> > Is it OK to change the connectivityTooltip from Tooltip2 to Popover2?
> 
> `isOpen` prop in `Tooltip2` could be another solution, but I'm ok with both `Tooltip2` and `Popover2`

It would be “better” to have less clicks. So if the small box can be launched by just hovering and can be closed by clicking elsewhere (except those copy buttons) it feels better in my point of view.

I-Chenn:
OK, I will:
1. Try `isOpen` prop of `ToolTip2` and see if I can maintain the original `Tooltip2` (hover, not click) version.
2. Beautify the button & add a feedback like "Session ID copied!" to it.
3. Another button for session URL.

in my next commit.

Nevertheless, as for the "Session URL button", is that related to Adriana's reply in #1930? I did some search for "Session URL" in the front-end code, but still have no idea what that is. May I have some indication of what "Session URL" is?

kswang1029:
> OK, I will:
> 
> 1. Try `isOpen` prop of `ToolTip2` and see if I can maintain the original `Tooltip2` (hover, not click) version.
> 2. Beautify the button & add a feedback like "Session ID copied!" to it.
> 3. Another button for session URL.
> 
> in my next commit.
> 
> Nevertheless, as for the "Session URL button", is that related to Adriana's reply in #1930? I did some search for "Session URL" in the front-end code, but still have no idea what that is. May I have some indication of what "Session URL" is?

It is the browser url which we used to launch the gui

I-Chenn:
@kswang1029 @YuHsuan-Hwang

I still find it hard to get a method of "maintaining" Tooltip2 until we can click on the button within it. The major problem is that I am not able to set up a proper scenario (until the next click?) to set `isOpen` prop to false. Please enlighten me if you have any good idea.

An alternative way is adopted temporarily in 219ae51, that is to set `hoverCloseDelay` prop to **3 seconds**, which I think should be long enough for user to click on the button. I know it's not perfect, but it resolve the problem to some extent.

To be honest, I personally don't think it's a good idea to add an interactive element within tooltip, since tooltip is supposed to display additional information. A popup is instead more reasonable to put in interactive elements by comparison.

By the way, I still don't quite understand what **Session URL** is. It's definitely not `http://localhost:3000/` and probably not:
截圖 2023-01-04 下午11 35 10
Can you show me how to get Session URL?

YuHsuan-Hwang:
> @kswang1029 @YuHsuan-Hwang
> 
> I still find it hard to get a method of "maintaining" Tooltip2 until we can click on the button within it. The major problem is that I am not able to set up a proper scenario (until the next click?) to set `isOpen` prop to false. Please enlighten me if you have any good idea.
> 
> An alternative way is adopted temporarily in [219ae51](https://github.com/CARTAvis/carta-frontend/commit/219ae5109fb1b1422bf368d41f39b52c406fa4d2), that is to set `hoverCloseDelay` prop to **3 seconds**, which I think should be long enough for user to click on the button. I know it's not perfect, but it resolve the problem to some extent.
> 
> To be honest, I personally don't think it's a good idea to add an interactive element within tooltip, since tooltip is supposed to display additional information. A popup is instead more reasonable to put in interactive elements by comparison.

I see, `Tooltip2` does not support clicks well. We can use `Popover2` instead.
`hoverCloseDelay` design also looks good to me because users don't need any extra clicks in this way, but I don't have a strong preference.

kswang1029:
we should use https://blueprintjs.com/docs/#core/components/toast to notify a user that a session id is copied.

kswang1029:
@I-Chenn @YuHsuan-Hwang I have one simple idea on adding the copy session id/URL feature. We can simply add two more options as "Copy session ID" and "Copy session URL" to the server sub-menu.
Screen Shot 2023-01-09 at 10 48 40

That "restart service" option should be displayed only when CARTA is deployed with the CARTA controller.

I-Chenn:
@kswang1029 
OK, roger that.

Nevertheless, I would like to know how to reveal that "Service" sub-menu? I did some survey through the source code:
image
image
image
and it seems that I need to be "authenticated" and with an "access token"? I would like to know how to get the "access token" and be "authenticated"? Does that relate to "Carta Controller" you mentioned? 

> That "restart service" option should be displayed only when CARTA is deployed with the CARTA controller.

kswang1029:
> @kswang1029 OK, roger that.
> 
> Nevertheless, I would like to know how to reveal that "Service" sub-menu? I did some survey through the source code: image image image and it seems that I need to be "authenticated" and with an "access token"? I would like to know how to get the "access token" and be "authenticated"? Does that relate to "Carta Controller" you mentioned?

No, even without the CARTA controller, when we start a carta_backend process, be default it will generate a token as part of the URL for users to connect to.

veggiesaurus:
> @I-Chenn @YuHsuan-Hwang I have one simple idea on adding the copy session id/URL feature. We can simply add two more options as "Copy session ID" and "Copy session URL" to the server sub-menu.

How do we adapt this to the electron app? Hide the session URL entry?

kswang1029:
> > @I-Chenn @YuHsuan-Hwang I have one simple idea on adding the copy session id/URL feature. We can simply add two more options as "Copy session ID" and "Copy session URL" to the server sub-menu.
> 
> How do we adapt this to the electron app? Hide the session URL entry?

Can we keep it? Then users can still use it if they want to have a new session in a browser or use the URL in a python script to control the GUI programmatically?

Guess we can still consider the electron app as a (local)server application.

I-Chenn:
Blueprintjs' toaster is not yet applied to the commit of 9cb4069. Will do it tomorrow.

I-Chenn:
@kswang1029 @YuHsuan-Hwang 
An "InfoToast" function is added to toast.ts and is then used to present "XXX is copied!" information after clicking the menus mentioned above. The result is shown below:
[Screencast from 01-10-2023 12:19:29 PM.webm](https://user-images.githubusercontent.com/106953609/211463090-198577f1-4da1-4746-ad32-e70a6e8a2eb1.webm)

Another concern is that: most of the content in the **Server** sub-menu don't usually show up. They need some condition to show up. Take "**Restart Service**" for example, it needs authentication to show up (and that's why when I tried to run CARTA with `--debug_no_auth` flag when developing, the whole **Server** sub-menu doesn't show up at all). I haven't set any condition for the two new sub sub-menus in 9cb4069 and 07035f4, do I need to add any condition for them to show up?

kswang1029:
> @kswang1029 @YuHsuan-Hwang An "InfoToast" function is added to toast.ts and is then used to present "XXX is copied!" information after clicking the menus mentioned above. The result is shown below: [Screencast from 01-10-2023 12:19:29 PM.webm](https://user-images.githubusercontent.com/106953609/211463090-198577f1-4da1-4746-ad32-e70a6e8a2eb1.webm)
> 
> Another concern is that: most of the content in the **Server** sub-menu don't usually show up. They need some condition to show up. Take "**Restart Service**" for example, it needs authentication to show up (and that's why that when I tried to run CARTA with `--debug_no_auth` flag when developing, the whole **Server** sub-menu doesn't show up at all). I haven't set any condition for the two new sub sub-menus in [07035f4](https://github.com/CARTAvis/carta-frontend/commit/07035f4deafc906454174689fb92059ade0069ff), do I need to add any condition for them to show up?

Regarding the "Server" menu and it context, we need to refine the current logic. 
1. We should always show the Server menu from now.
2. We always show the two copy functions
3. For "Restart Service", "Dashboard", and "Logout", we only need to show them when a carta controller is used for deployment.
Screen Shot 2023-01-10 at 13 01 19

I-Chenn:
@kswang1029 @YuHsuan-Hwang 
Sorry for making a further commit again.
Though we decided to open a new issue for the "appearance of the **Restart Service**" issue during our discussion yesterday, I asked @veggiesaurus about this issue and it seems that we only need a minor modification to fix it, I thus made the fix in 321ad04 and the result is shown below (Note: in the image, CARTA is run with production build to be authenticated):
image
The modification does fix #2086

kswang1029:
@ajm-asiaa is there a way to test this with the carta controller?

ajm-ska:
@kswang1029 I have put the 321ad049 frontend commit on our cartaviewer server, used the `"frontendPath"` configuration option in `/etc/carta/config.json` to point towards it, and restarted the carta-controller service. You can try it out on https://cartaviewer.asiaa.sinica.edu.tw/ (only if inside the IAA network).

The menu options show up correctly:
Screenshot 2023-01-11 at 6 29 53 PM

I notice there is no token in the copied URL, so I guess for the carta-controller version only the first part of the URL and session ID are needed for carta-python to connect to it?

kswang1029:
> @kswang1029 I have put the [321ad04](https://github.com/CARTAvis/carta-frontend/commit/321ad049336d4bfcccf3f1b821aaa6fb4834e2c9) frontend commit on our cartaviewer server, used the `"frontendPath"` configuration option in `/etc/carta/config.json` to point towards it, and restarted the carta-controller service. You can try it out on https://cartaviewer.asiaa.sinica.edu.tw/ (only if inside the IAA network).
> 
> The menu options show up correctly: Screenshot 2023-01-11 at 6 29 53 PM
> 
> I notice there is no token in the copied URL, so I guess for the carta-controller version only the first part of the URL and session ID are needed for carta-python to connect to it?

@ajm-asiaa thanks. This looks great (@I-Chenn). As for using carta-python via the controller, I do not have answer for now. @confluence @veggiesaurus help...

confluence:
@ajm-asiaa yes, there's no token in the URL if you use Python scripting with the controller. The token is only used when you connect directly to a backend.

confluence (approved):
I have tested this with a local backend, and it looks like it's doing the correct thing.

kswang1029 (changes requested):
It appears that if we just launch CARTA and try to copy session ID/URL, there is no toast info showing up. 

https://user-images.githubusercontent.com/20819712/211956880-bbe4bfeb-8ca1-4895-8620-8d3e636f4cca.mov

I-Chenn:
@kswang1029: 
That's weird, since I am able to show the toast right after launching Carta on my desktop (Ubuntu 22.04):
[Screencast from 01-12-2023 10:44:26 AM.webm](https://user-images.githubusercontent.com/106953609/211963956-0f10ab53-6546-4881-8c3e-6bb6fd8b0e8e.webm)
I will try it on my MacBook later.

src/components/Menu/RootMenuComponent.tsx
```diff
@@ -178,6 +178,26 @@ export class RootMenuComponent extends React.Component {
...
+                        AppToaster.show(InfoToast("Session ID copied!"));
+                    });
+                }}
+            />
+        );
+        serverMenu.push(
+            <Menu.Item
+                text="Copy session URL to clipboard"
+                onClick={() => {
+                    navigator.clipboard.writeText(window.location.href).then(() => {
```
YuHsuan-Hwang: We can use `async/await` syntax, as in `PreferenceDialogComponent.handleUserIdCopied`

src/components/Menu/RootMenuComponent.tsx
```diff
@@ -178,6 +178,26 @@ export class RootMenuComponent extends React.Component {
...
+                        AppToaster.show(InfoToast("Session URL copied!"));
```
YuHsuan-Hwang: How about we just reuse `SuccessToast("clipboard", ...`?

YuHsuan-Hwang (changes requested):

src/components/Menu/RootMenuComponent.tsx
I-Chenn: @YuHsuan-Hwang 
image
OK, I did mean to use `SuccessToast` at first, but I didn't see the similar clipboard example in `PreferenceDialogComponent.handleUserIdCopied` thus I don't know what icon should be applied for the situation, thus I made `InfoToast` as a simple version of `SuccessToast`. I will take your comment, anyway.
By the way, is `async/await` syntax method used to solve @kswang1029's issue mentioned in:
https://github.com/CARTAvis/carta-frontend/pull/2070#pullrequestreview-1244855995
? I still can't reproduce Kuo-Song's situation yet......

YuHsuan-Hwang: >By the way, is async/await syntax method used to solve @kswang1029's issue mentioned in:
https://github.com/CARTAvis/carta-frontend/pull/2070#pullrequestreview-1244855995
? I still can't reproduce Kuo-Song's situation yet......

No, I think they're not related

kswang1029:
if we "click" the Server menu first, there is a chance that the toast will not be displayed. This might be something else in general so please do not spend time on this now.

kswang1029:
Uh, in fact, if we don't see the toast, copy is not successful.

kswang1029:
> Uh, in fact, if we don't see the toast, copy is not successful.

This weird issue happens to other menus. So please ignore it for now.

I-Chenn:
@YuHsuan-Hwang 
`SuccessToast` and `async/await` syntax is adopted in 13c76fb.
The `InfoToaster` is thus removed from `toaster.ts` in the same commit.

YuHsuan-Hwang (approved):
The new changes look good.

kswang1029 (approved):
this looks good. 👍 no regression is seen from e2e tests.

PR #1870, opened 2022-05-27
Yuhsuan/1843 control map border

YuHsuan-Hwang:
**Description**

Closes #1843 and #1842: control map border issue. Please see the document for the details of the fix:
https://docs.google.com/document/d/1GOrcs4t2_wLh7hkWS-IUBLU5-bLbx4CY1dULiEzWf2Q/edit

changes:
* adjusted the min/max point of the control map to avoid using data at the same index for bicubic interpolation
* excluded filled zeros when finding min of catalog sources
* removed unused functions including `getTransformedCoordinate`, `IsWidthValid`, and `astTransform` calculation in `CatalogControlMap` constructor
* refactored `ControlMap` and `CatalogControlMap` to remove duplicate code

tested:
* no extra displacement for projected contours (set smoothing factor to 1)
cosmos_spitzer24micron fits-cosmos_spitzer24micron fits-image-2022-05-27-15-56-29

* no extra displacement for projected catalog overlays
cosmos_spitzer24micron fits-cosmos_spitzer24micron fits-image-2022-05-27-16-00-01
model fits-model_imregrid_galactic fits-image-2022-05-27-17-21-48

* tested the above two cases without clamping the index for `texelFetch` (disable `clampPixel` in `utilities.glsl`). makes sure that no index out of range is used for interpolation
* catalog control map is updated when the source number changes

**Checklist**
- [x] changelog updated / ~no changelog update needed~
- [x] e2e test passing / ~added corresponding fix~
- [x] ~protobuf updated to the latest dev commit~ / no protobuf update needed

PR #2073, opened 2022-12-24
carli/2034_single_source_catalog_file_not_rendering

### LINKED ISSUE #2034: missing catalog image overlay when there is only one source in the catalog file
**Describe the bug**
It appears that if a catalog file has only one source, if we have two images matched and enable the catalog image overlay, we will only see one image has the catalog overlay. The other is empty. If we load a catalog with two sources instead and repeat the exercise, we will see two sources on both images as expected.

**To Reproduce**
1. download the test images and test catalog files
https://drive.google.com/file/d/164C3NjSyE30T2BHXi7Y9FaTFm7AzfDc7/view?usp=sharing

2. load the two images and match them
3. load catalog.vot which contains only one source and enable catalog overlay. We will see the issue.
4. repeat the test with catalog2.vot which contains two sources. We will see the expected behavior.

**Expected behavior**
with the single source catalog file, we should see the catalog overlay on all matched images.

 - Browser [e.g. chrome, safari, electron app]: electron app
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: v3-stable
 - Frontend branch [e.g. dev, v3b2 release]: v3-stable

crocka:
**Description**

The problem is from `controlMapLookUp` in `utilities.glsl`. The `index` is calculated with `range`, but when there is only one source, `range` is 0, which results in the wrong `index`. The resolution is to add an if-statement to account for when `range` is 0. This PR fixes #2034.

For the pull request:

### DIFF
--- a/src/services/GLSL/utilities.glsl
+++ b/src/services/GLSL/utilities.glsl
@@ -75,6 +75,9 @@ vec2 controlMapLookup(sampler2D controlMapTexture, vec2 pos, vec2 controlMapSize
     vec2 range = controlMapMax - controlMapMin;
     vec2 shiftedPoint = pos - controlMapMin;
     vec2 index = shiftedPoint / range * controlMapSize;
+    if (controlMapMax == controlMapMin) {
+        return bicubicFilter(controlMapTexture, vec2(0, 0), controlMapSize).rg;
+    }
     return bicubicFilter(controlMapTexture, index, controlMapSize).rg;
 }
 
--- a/src/services/GLSL/utilities.glsl
+++ b/src/services/GLSL/utilities.glsl
@@ -75,7 +75,7 @@ vec2 controlMapLookup(sampler2D controlMapTexture, vec2 pos, vec2 controlMapSize
     vec2 range = controlMapMax - controlMapMin;
     vec2 shiftedPoint = pos - controlMapMin;
     vec2 index = shiftedPoint / range * controlMapSize;
-    if(range.x == 0.0 && range.y == 0.0) {
+    if (range.x == 0.0 && range.y == 0.0) {
         index = vec2(0, 0);
     }
     return bicubicFilter(controlMapTexture, index, controlMapSize).rg;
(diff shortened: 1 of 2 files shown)


src/services/GLSL/utilities.glsl
```diff
@@ -75,6 +75,9 @@ vec2 controlMapLookup(sampler2D controlMapTexture, vec2 pos, vec2 controlMapSize
     vec2 range = controlMapMax - controlMapMin;
     vec2 shiftedPoint = pos - controlMapMin;
     vec2 index = shiftedPoint / range * controlMapSize;
+    if (controlMapMax == controlMapMin) {
```
crocka: I tried
```if (range.x == 0 && range.y == 0) { index = vec2(0, 0) };```
but it didn't seem to work. I am not very familiar with the syntax of GLSL, so please advice if there is a better solution.

kswang1029: The current solution works fine. No regression is observed from e2e tests. If there are better solutions from @YuHsuan-Hwang , I will re-do the test.

YuHsuan-Hwang: @crocka Does `if (range.x == 0.0 && range.y == 0.0) { index = vec2(0, 0) };` works?
`controlMapMax == controlMapMin` is also fine.

crocka: This is actually the first thing I tried. It didn't work and it is giving me the following error.
Screenshot from 2023-01-03 20-07-20

YuHsuan-Hwang: Did you use `0.0` or `0`? You can check the console for error messages.

crocka: I see the problem. We should use range.x == 0.0 && range.y == 0.0. I tried `index = vec2(0.0, 0.0) but ignored the if-statement.

YuHsuan-Hwang (approved):
need a space after `if`. Other than that this looks good.
I think this is a regression from my control map border fix PR #1870. Thanks for fixing again 👍  Please also update the change log.

@kswang1029 We are ready for a final feature test, thanks.

kswang1029 (approved):

Issue #1561, opened 2021-07-12
support pv interactive preview (frontend)


kswang1029:
related to [support preview of a pv image from a downsampled cube](https://app.zenhub.com/workspaces/carta-5c7fc74483f19a07c7bcacaf/issues/cartavis/carta-backend/795)

PR #2100, opened 2023-02-14
Carli/1561_support_pv_preview

crocka:
**Description**
This PR is in response to feature request #1561. This pull request introduces a new feature that allows users to view and interact with PV image preview before requesting the full PV image. This PR closes issue #2123.

The changes include modifications to several existing files:

- PreferenceDialogComponent.tsx + PreferenceStore
Added preference setting to set PV preview cube size limit. Corresponding default values are adde to the PreferenceStore.

- Floating widget
Added preview widget under PV generator, so that when the PV generator widget closes, the preview widget is closing with it. In the WidgetStore.ts, a block of code was added to removeAssociatedFloatingSetting to close the docked preview widget.

- ToolbarComponent
Restricted preview frame to accesss only a limited function on the toolbar.

- ImagePanel + ImageView
The preview widget reuses the ImagePanel component, so widget width and height is different from the ImageViewer width and height, which are calculated based on how many images are open.

- RasterViewComponent
Instead of using tile based rendering like a regular image, the preview images use raster data that are rendered as a whole. In the ``renderTile`` function, a texture is created if the image is preview, and added to the GPU as a whole.

  - Shaders
The uTextureSize and uTileTextureSize is changed to a vec2 rather than float, because the entire preview image is now considered one tile that might not necessarily be 256 by 256.

- PvGeneratorComponent
Added preview region and rebin input. Added ``estimatedCubeSize`` to calculate preview region based on the rectangular region and rebin, as well as frame info data. The estimated size is shown on the PVGenerator component and the ``start preview`` button is disabled if the estimated size is larger than the limit set in preference setting.

- PvPreviewComponent
A component that reuses the ImagePanelComponent along with the ReactResizeDetector.

- OverlayComponent + OverlayStore
Added ``previewRenderWidth`` and ``previewRenderHeight`` to replace the default render width and height of the imageViewer in the OverlayStore. renderWidth and renderHeight are added to FrameStore so that when the frame is preview, correct width and height is used.

- PvGeneratorWidgetStore
The original ``requestPV`` function is modified to allow for pv preview request, marked by a boolean argument ``preview``. A ``previewFrame`` observable is added to store the previewFrame that currently belongs to the PV generator widget.

- RegionWidgetStore
The original ``RegionId`` is modified to include a NONE region for PV cut selection in the PV generator. The ``effectiveRegionId`` returns null if RegionId.NONE is selected as the PV cut.

- RenderConfigStore
When the user starts dragging the PV cut, new histogram bounds are needed to update the render config. PreviewHistogramMax and PreviewHistogramMin are added to temporarily store the bounds value. In the ``updatePreviewData`` function of FrameStore, the PreviewHistogramMax and Min are set to null if the full histogram data is sent, when the user drops or stops editing the PV cut, otherwise, when the full histogram data is not being sent, because the user is still editing the PV cut, only the histogram bounds are sent to reduce the amount of total data transfer.

- RegionStore
The line region must send setRegion message to the backend while the PV cut is being edited. The ``updateRegion`` is modified to send message if the region is a line and if the line region is not creating.

- FrameStore
FrameStore is modified to include previewFrame. FrameInfo is changed to an observable value since previewFrame requires constant update of FrameInfo. isPreview boolean is added to distinguish a previewFrame. rasterData of previewFrame is stored in the FrameStore. ``previewCursorValue`` uses the rasterData to find the cursor value. ``updatePreviewData`` is a generator function that gets called when new preview data is received, which sets new rasterData, sets histogram data for render config store, sets the new frameInfo. Since the aspectRatio may be different when user modifies the PV cut, the function also sets new center calculated by how much has the aspectRatio changed with respect to the previous one. The zoom level is also modified by how much has the height changed (needed is the axis are swapped). Finally in the ``updatePreviewdata``, the new wcsInfo is initialized with new FrameInfo.
  - Decompression
   ``updatePreviewData`` function uses the 0th indexed ZFPWorker to decompress the raster data. When the decompressed data is sent back to the main thread, the frontend saves the raster data in the frameStore, and calls ``updataPreviewData.next()`` to update frameInfo and relating values. The purpose of the use of generator function is to allow the raster data to be updated first before rerendering. Otherwise, the mismatch between raster data and values such as width and height of the image may cause image distortion.

- AppStore + BackendService
A Rxjs pvPreviewStream is added in the backendService to handle incoming preview data. The AppStore is subscribed to pvPreviewStream with the ``handlePvPreviewStream`` function, which calls the ``updataPreviewData`` in FrameStore and pass in the new preview data.

**Checklist**

For the pull request:
- [x] ~~`BackendService` unchanged~~ / `BackendService` changed and corresponding ICD test fix added


src/stores/PreferenceStore/PreferenceStore.ts
```diff
@@ -888,7 +902,8 @@ export class PreferenceStore {
                 PreferenceKeys.REGION_CREATION_MODE,
                 PreferenceKeys.WCS_OVERLAY_AST_COLOR,
                 PreferenceKeys.CATALOG_TABLE_SEPARATOR_POSITION,
-                PreferenceKeys.PIXEL_GRID_COLOR
+                PreferenceKeys.PIXEL_GRID_COLOR,
+                PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT_UNIT
```
YuHsuan-Hwang: We don't need to change `upgradePreferences` because it's for old keys.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -748,6 +770,34 @@ export class PreferenceDialogComponent extends React.Component {
                         </option>
                     </HTMLSelect>
                 </FormGroup>
+                <FormGroup inline={true} label="PV Preview Cube Size Limit">
```
YuHsuan-Hwang: The label and placeholder need to be in sentence case.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -748,6 +770,34 @@ export class PreferenceDialogComponent extends React.Component {
...
+                    <div className="pv-preview-cube-size-limit">
+                        <SafeNumericInput
+                            placeholder="PV Preview Cube Size Limit"
+                            min={1e-12}
+                            max={this.pvPreviewCubeSizeMaxValue}
+                            value={preference.pvPreivewCubeSizeLimit}
+                            majorStepSize={1}
+                            stepSize={1}
+                            onValueChange={value => PreferenceStore.Instance.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT, value)}
```
YuHsuan-Hwang: can be simplified to `preference.setPreference`.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -52,14 +52,31 @@ enum PreferenceDialogTabs {
 
 const PercentileSelect = Select.ofType<string>();
 
+const PV_PREVIEW_CUBE_SIZE_LIMIT = 1000000000; //need to be removed and replaced by backend limit
+
 @observer
 export class PreferenceDialogComponent extends React.Component {
     @observable selectedTab: PreferenceDialogTabs = PreferenceDialogTabs.GLOBAL;
+    @observable pvPreviewCubeSizeLimitUnit: string;
```
YuHsuan-Hwang: This needs to be initialized otherwise `pvPreviewCubeSizeMaxValue` will be `PV_PREVIEW_CUBE_SIZE_LIMIT` before `handlePvPreviewCubeSizeUnitChange` is triggered.
How about we directly use`preferenceStore.pvPreivewCubeSizeLimitUnit`?

src/stores/PreferenceStore/PreferenceStore.ts
```diff
@@ -223,7 +225,9 @@ const DEFAULTS = {
         streamContoursWhileZooming: false,
         lowBandwidthMode: false,
         stopAnimationPlaybackMinutes: 5,
-        limitOverlayRedraw: true
+        limitOverlayRedraw: true,
+        pvPreviewCubeSizeLimit: 1,
+        pvPreviewCubeSizeLimitUnit: "GB"
```
YuHsuan-Hwang: Perhaps we could create a enum `MemoryUnit` to specify the type.

src/stores/OverlayStore/OverlayStore.ts
```diff
@@ -899,13 +898,13 @@ export class OverlayColorbarSettings {
         return this.position === "right" ? padding?.top : padding?.left;
     }
 
-    @computed get height(): number {
+    @action height(frame?: FrameStore): number {
```
YuHsuan-Hwang: Looks like the colorbar tick positions are not responsive to the height, both in the main viewer and the preview viewer:
image
Maybe we need to change `height` to a `@computed` variable returning a function (`(frame?: FrameStore) => number`) so that `colorbarStore.positions` responses with `height`.
Another way would be moving `height` under `colorbarStore`, which is frame dependent.

src/stores/OverlayStore/OverlayStore.ts
```diff
@@ -899,13 +898,13 @@ export class OverlayColorbarSettings {
...
         const overlayStore = AppStore.Instance?.overlayStore;
-        return this.position === "right" ? overlayStore?.renderHeight : overlayStore?.renderWidth;
+        return this.position === "right" ? frame?.renderHeight || overlayStore?.renderHeight : frame?.renderWidth || overlayStore?.renderWidth;
     }
 
     @computed get tickNum(): number {
-        const tickNum = Math.round((this.height / 100.0) * this.tickDensity);
+        const tickNum = Math.round((this.height() / 100.0) * this.tickDensity);
```
YuHsuan-Hwang: I think the tick number is also dependent to the preview frame height.

src/stores/OverlayStore/OverlayStore.ts
```diff
@@ -1213,4 +1208,20 @@ export class OverlayStore {
         const renderHeight = this.viewHeight - this.paddingTop - this.paddingBottom;
         return renderHeight > 1 ? renderHeight : 1; // return value > 1 to prevent crashing
     }
+
+    @action previewRenderWidth(viewWidth: number) {
```
YuHsuan-Hwang: Looks like the preview image panel is not responsive to the new paddings when we change the colorbar position from right to bottom/top:
image
Maybe we need to change preview render width/height to a `@computed` variable returning a function so that the image panel responses with paddings.

src/components/ImageView/ImageViewComponent.tsx
```diff
@@ -33,8 +33,10 @@ export function getImageViewCanvas(padding: Padding, colorbarPosition: string, b
         const column = index % appStore.numImageColumns;
         const row = Math.floor(index / appStore.numImageColumns);
         const panelCanvas = getPanelCanvas(column, row, padding, colorbarPosition, backgroundColor);
+        const viewHeight = frame.previewViewHeight || appStore.overlayStore.viewHeight;
```
YuHsuan-Hwang: It's for exporting png files, and we may remove changes if we don't support it. Shall we also disable Export image in the File menu?

src/stores/Frame/FrameStore.ts
```diff
@@ -192,6 +192,11 @@ export class FrameStore {
 
     @observable stokesFiles: CARTA.StokesFile[];
 
+    @observable isPreview: boolean;
+    @observable previewViewWidth: number;
+    @observable previewViewHeight: number;
+    @observable rasterData: Float32Array;
```
YuHsuan-Hwang: I suggest we rename it to `previewPVRasterData` to show that it's not used for all images.

src/stores/Frame/Region/RegionStore.ts
```diff
@@ -654,22 +660,28 @@ export class RegionStore {
...
     };
 
+    @action setIsPreviewCut = (isPreviewCut: boolean) => {
+        this.isPreviewCut = this.regionType === CARTA.RegionType.LINE && isPreviewCut;
+    };
+
     // Update the region with the backend
     // TODO: Determine whether we should await when calling this function from above
-    private updateRegion = async () => {
+    private updateRegion = async (isRequestingPreview?: boolean) => {
```
YuHsuan-Hwang: We can assign a default value `false` to `isRequestingPreview`.

src/components/PvGenerator/PvGeneratorComponent.tsx
```diff
@@ -291,28 +382,63 @@ export class PvGeneratorComponent extends React.Component<WidgetProps> {
...
+                                onValueChange={value => this.widgetStore.setZRebin(value)}
+                            />
+                        </FormGroup>
+                    </div>
+                </FormGroup>
+                <div className="cube-size-button-group">
+                    <FormGroup className="cube-size-group" inline label="Preview cube size" labelInfo={this.estimatedCubeSize ? `(${this.estimatedCubeSize.unit})` : ""} disabled={!this.estimatedCubeSize}>
+                        <label className="cube-size" style={{color: this.isCubeSizeBelowLimit ? "" : "red"}}>{`${this.estimatedCubeSize?.value ?? ""}`}</label>
+                    </FormGroup>
+                    <div className="generate-button">
```
YuHsuan-Hwang: I prefer we move the buttons to a new line.

src/components/PvGenerator/PvGeneratorComponent.scss
```diff
@@ -65,8 +81,30 @@
...
+                    width: 2em;
+                }
+            }
+        }
+
         .generate-button {
             float: right;
+            display: flex;
+            div {
+                margin: 1em;
```
YuHsuan-Hwang: I prefer we just add left margin.

src/components/PvGenerator/PvGeneratorComponent.tsx
```diff
@@ -291,28 +382,63 @@ export class PvGeneratorComponent extends React.Component<WidgetProps> {
...
         );
 
         return (
             <div className="pv-generator-widget">
-                <div className="pv-generator-panel">
-                    <Tabs id="pvGeneratorTabs" selectedTabId={this.selectedTabId} onChange={this.setSelectedTab} animate={false}>
-                        <Tab id={PvGeneratorComponentTabs.PV_IMAGE} title="Generate PV Image" panel={pvImagePanel} />
-                    </Tabs>
-                </div>
+                <div className="pv-generator-panel">{pvImagePanel}</div>
```
YuHsuan-Hwang, line 434: It would be useful if we can scroll in y direction when the widget is small.

YuHsuan-Hwang (changes requested):
Overall the implementation is good 👍 Here are my initial comments:

src/stores/OverlayStore/OverlayStore.ts
crocka: I added dummy variables: paddingLeft/top/right/bottom in frame.renderWidth and frame.renderHeight. It works. What do you think of this solution?

YuHsuan-Hwang: I would prefer we change `previewRenderWidth/Height` so that we won't have the same problem when reusing it in other places.

src/stores/Frame/FrameStore.ts
```diff
@@ -2821,4 +2837,85 @@ export class FrameStore {
...
+            nanEncodings: nanEncodings32,
+            tileCoordinate: 0,
+            layer: 0,
+            requestId: 0,
+            previewId: previewData.previewId
+        };
+
+        // Using the 'yield' keyword of generator functions to wait for decompressed raster data from other WebWorker thread.
+        // next() will be called in setPreviewPVRasterData, which will be called in the onmessage() function after receiving the decompressed data from other worker thread.
+        yield TileService.Instance.decompressPreviewRasterData(compressedView, eventArgs, previewData, nanEncodings32);
```
YuHsuan-Hwang: I suggest we adjust the function to `decompressPreviewRasterData(fileId: number, previewData: CARTA.PvPreviewData)` and move lines 2860-2878 into `TileService`.

src/stores/Frame/FrameStore.ts
```diff
@@ -2821,4 +2837,85 @@ export class FrameStore {
...
     };
+
+    @action setIsPreview = (isPreview: boolean) => {
+        this.isPreview = isPreview;
+    };
+
+    @action setPreviewPVRasterData = (previewPVRasterData: Float32Array, skipUpdatePreviewData: boolean = false) => {
+        this.previewPVRasterData = previewPVRasterData;
+        // if skipUpdatePreviewData is false, the code after the yield keyword in the updatePreviewData function will be executed by calling the next() function.
+        !skipUpdatePreviewData && this.updatePreviewDataGenerator.next();
```
YuHsuan-Hwang: I think an if-statement is clearer: `if (!skipUpdatePreviewData) {...}`

src/services/PreviewWebGLService.ts
```diff
@@ -0,0 +1,174 @@
...
+    TileTextureSize: WebGLUniformLocation;
+    TextureSize: WebGLUniformLocation;
+    TileBorder: WebGLUniformLocation;
+    PixelGridCutoff: WebGLUniformLocation;
+    PixelGridColor: WebGLUniformLocation;
+    PixelGridOpacity: WebGLUniformLocation;
+    PixelAspectRatio: WebGLUniformLocation;
+}
+
+export class PreviewWebGLService {
```
YuHsuan-Hwang: Is it possible to inherit `TileWebGLService` and override `Instance()`?

YuHsuan-Hwang (changes requested):

YuHsuan-Hwang (approved):
New changes are looking good.

kswang1029 (approved):

YuHsuan-Hwang:
This PR is ready for merge once the protobuf merge conflict is resolved and changelog is updated.

markccchiang (approved):

Issue #2102, opened 2023-02-17
"Copy Session URL to clipboard" modification needed for macOS Electron version


ajm-ska:
**Describe the bug**

For the macOS Electron version of CARTA, the `File > Server > Copy Session URL to clipboard` feature will need to be modified. 
At the moment it copies something like this: 
```
file:///Applications/CARTA.app/Contents/Resources/app/index.html?socketUrl=ws://localhost:55051&token=c533b90f-5bf1-4814-b81c-a889a35e79d8/
```

It should instead be:
```
http://localhost:55051?token=c533b90f-5bf1-4814-b81c-a889a35e79d8/
```

I imagine it should be easy to fix. I guess an extra loop can be added to the appropriate section in `src/components/Menu/RootMenuComponent.tsx` to check for the presence of "socketUrl" in the string, and if present it could take the value of socketUrl, as well as substitute "ws" with "http", and "&" with "?" (Replacing "&" with "?" is necessary as the python scripting interface does not seem to be able to parse it "&").

**To Reproduce**
1. Run our CARTA-python-test-17-2-23.dmg macOS Electron version.
2. Click on File > Server > Copy Session URL to clipboard
3. Paste it somewhere.
4. See the extended URL.

**Expected behavior**
For easier python scripting interface use, the URL should be shorter. There should be 3 changes:
1. Only copy the text after socketUrl=
2. Replace "ws" with "http"
3. Replace "&" with "?"

**Platform info (please complete the following information):**
 - OS [e.g. macOS Monterey]:  macOS Monterey
 - Browser [e.g. chrome, safari, electron app]: Electron app  CARTA-python-test-17-2-23.dmg
 - Browser version [e.g. 22]: Electron 23.0.0 (Chrome 108)
 - Backend branch [e.g. dev, v3b2 release]: dev 28152ab
 - Frontend branch [e.g. dev, v3b2 release]: dev 2978e72

ajm-ska:
Maybe I can tackle this issue? I have a working solution now. It can be solved by replacing lines 197 to 209 of `src/components/Menu/RootMenuComponent.tsx` with the following: 
```
serverMenu.push(
  <Menu.Item
    text="Copy session URL to clipboard"
    onClick={async () => {
      try {
        const url = new URL(window.location.href);
        if (!url.searchParams.has("socketUrl")) {
          await navigator.clipboard?.writeText(window.location.href);
        } else {
          const socketUrl = url.searchParams.get("socketUrl");
[...]

PR #2103, opened 2023-02-17
To read session URL correctly from the Electron version

ajm-ska:
**Description**

This addresses issue #2102. 

The problem is if using the Electron version of CARTA, the URL is not as expected e.g.: 
```
file:///Applications/CARTA.app/Contents/Resources/app/index.html?socketUrl=ws://localhost:55051&token=c533b90f-5bf1-4814-b81c-a889a35e79d8/
```
We only want http://localhost, the port number, and the token information.

This fix assumes that if the word "socketUrl" is present in the string, then you must be using the Electron version.
If "socketUrl" is not present, it simply produces the URL as before using "window.location.href"
If "socketUrl" is present, it works on assembling the correct shorter URL. It also switches "ws" with "http" and uses "?" instead of "&".

**Checklist**

- [ ] assignee and label added
- [ ] ZenHub issue connection, board status, and estimate updated

For the pull request:
- [ ] ZenHub estimate, milestone, and release (if needed) added
- [ ] e2e test passing / ~corresponding fix added~
- [ ] `BackendService` unchanged / `BackendService` changed and corresponding ICD test fix added

### DIFF
--- a/src/components/Menu/RootMenuComponent.tsx
+++ b/src/components/Menu/RootMenuComponent.tsx
@@ -200,7 +200,7 @@ export class RootMenuComponent extends React.Component {
                 onClick={async () => {
                     try {
                         const url = new URL(window.location.href);
-                        if (!url.searchParams.has("socketUrl")) {
+                        if (url.protocol.startsWith("file")) {
                             await navigator.clipboard?.writeText(window.location.href);
                         } else {
                             const socketUrl = url.searchParams.get("socketUrl");
--- a/src/components/Menu/RootMenuComponent.tsx
+++ b/src/components/Menu/RootMenuComponent.tsx
@@ -200,7 +200,7 @@ export class RootMenuComponent extends React.Component {
                 onClick={async () => {
                     try {
                         const url = new URL(window.location.href);
-                        if (url.protocol.startsWith("file")) {
+                        if (!url.protocol.startsWith("file")) {
                             await navigator.clipboard?.writeText(window.location.href);
                         } else {
                             const socketUrl = url.searchParams.get("socketUrl");
--- a/src/components/Menu/RootMenuComponent.tsx
+++ b/src/components/Menu/RootMenuComponent.tsx
@@ -200,7 +200,7 @@ export class RootMenuComponent extends React.Component {
                 onClick={async () => {
                     try {
                         const url = new URL(window.location.href);
-                        if (!url.protocol.startsWith("file://")) {
+                        if (!url.protocol.startsWith("file")) {
                             await navigator.clipboard?.writeText(window.location.href);
                         } else {
                             const socketUrl = url.searchParams.get("socketUrl");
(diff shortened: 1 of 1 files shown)


kswang1029:
Wonder if we can use file:// as the keyword to check if it is the electron instance?

kswang1029:
this works fine in dev and in "CARTA 3.3.5" (via slack). No regression is seen from e2e test.

kswang1029 (approved):

ajm-ska:
This is still a Draft PR.

Doing `file://` didn't work, so I had to change it back to `file`. (The test version 3.3.5 uses `file`).

Also, I wonder if anyone would be using a simple shared setup like our Mix 'N Match service? 
For example, the form of the URL would be as follows: 
```
https://carta.asiaa.sinica.edu.tw/frontend/0024edf3/socketUrl=wss://carta.asiaa.sinica.edu.tw/socketdev&skipTelemetry=1
```
That is the reason I used `socketUrl` originally. Although I know that commit `da02236` does not work on Mix 'N Match because it would need to be further modified to replace `wss` with http. But commit `da02236` was only a proof-of-concept intended to work with the Electron version that uses `ws`.

kswang1029:
> This is still a Draft PR.
> 
> Doing `file://` didn't work, so I had to change it back to `file`. (The test version 3.3.5 uses `file`).
> 
> Also, I wonder if anyone would be using a simple shared setup like our Mix 'N Match service? For example, the form of the URL would be as follows:
> 
> ```
> https://carta.asiaa.sinica.edu.tw/frontend/0024edf3/socketUrl=wss://carta.asiaa.sinica.edu.tw/socketdev&skipTelemetry=1
> ```
> 
> That is the reason I used `socketUrl` originally. Although I know that commit `da02236` does not work on Mix 'N Match because it would need to be further modified to replace `wss` with http. But commit `da02236` was only a proof-of-concept intended to work with the Electron version that uses `ws`.

guessing we need to escape `/`?

ajm-ska:
> guessing we need to escape `/`?

Maybe, but if using `file` I still think it is simpler to use "startsWith"? 
`if (url.protocol.startsWith("file")) {`

If someone is using the "file" in the URL to directly open an image, it would not be appearing at the beginning of the URL.

ajm-ska:
It is strange that attempting to escape the forward slashes does not work either. i.e The following does not work:
```if (!url.protocol.startsWith("file:\\/\\/")) {```
So let's stick with `file`.

I also want to check what URL the Linux AppImage version reports, and if this solution works there too.

ajm-ska:
I feel this PR is ready as a fix to the original issue - To correctly copy the Session URL in the macOS Electron version.
I know it has to lead to the discovery of issue #2108 so this part of the code will need further modification. 
I'm not sure if it is worth merging the current fix now, or combining it with an eventual fix to #2108?

kswang1029:
> I feel this PR is ready as a fix to the original issue - To correctly copy the Session URL in the macOS Electron version.
> I know it has to lead to the discovery of issue #2108 so this part of the code will need further modification.
> I'm not sure if it is worth merging the current fix now, or combining it with an eventual fix to #2108?

I suggest we integrate the current solution for #2108 (from a changelog point of view)

ajm-ska:
This PR has been superseded by PR #2109

PR #2126, opened 2023-03-23
Yuhsuan/2114 splatalogue backend query

### LINKED ISSUE #2114: switch to Splatalogue advanced backend query
as v3-stable and dev, the spectral line query widget makes a query to the Splatalogue frontend to get line data back. NRAO has been implementing a new AngularJS-based frontend and suggesting us to use the Splatalogue backend for the query instead as it is more straightforward without unnecessary routing. 

A doc is prepared for some details.
https://docs.google.com/document/d/1xEZ4_IcyJUo5CermETjOtO96jp4oYC7fJhy7iJlB25I/edit#

Let's try to get this in v4-stable.

### LINKED ISSUE #2119: Spectral line query crashes when no network connection
**Describe the bug**
Spectral line query crashes when no network connection. The root cause is `alertStore.showAlert` does not accept `Error` objects.

**To Reproduce**
1. Enable network connection and open spectral line query widget.
2. Disable network connection.
3. Input parameters and send a query.

**Expected behavior**
See an alert.

**Screenshots or videos**
image

 - OS [e.g. macOS Monterey]: macOS Big Sur
 - Browser version [e.g. 22]: 108
 - Backend branch [e.g. dev, v3b2 release]: dev
 - Frontend branch [e.g. dev, v3b2 release]: dev

YuHsuan-Hwang:
**Description**

1. Closes #2114

- Implemented Splatalogue advanced backend query based on the [document](https://docs.google.com/document/d/1xEZ4_IcyJUo5CermETjOtO96jp4oYC7fJhy7iJlB25I/edit#)
  - See "Implementation -> mode 1 and 2" in the document for the details of the query parameters
  - See "Implementation -> Headers" in the document for the details of the received data format and the corresponding headers from the current approach. I preserved the displayed columns in the table, with some minor changes:
    1. "CDMS/JPL Intensity": `0.00000` becomes empty
    image --> image

    2. "Resolved QNs" and "Species": additional spaces
    image --> image

    3. almost all columns: trailing zeros are removed
    image --> image
    
    4. "Chemical Name": html tags are removed
    image --> image

    5. "Species", "Resolved QNs", "Unresolved QNs": convert html entities to string
    image --> image

    6. One additional line, could be due to different database versions
    image

- Removed the connection check when opening the widget since it does not guarantee the connection is available when the users click `Query`
- Added simple unit tests, focusing on query parameters and parsing strings

2. Fixes #2119 by changing the Error object to string

ToDo:
- ~update the final url~ will be updated in the stable release

**Checklist**

For the pull request:
- [x] ~e2e test passing~ / corresponding fix added


YuHsuan-Hwang:
New changes:
* filter received data instead of removing the last data (which could be storing only the query string)
* improve parsing string

kswang1029 (changes requested):
@YuHsuan-Hwang in general this looks good based on some manual tests and e2e tests. A few issues spotted:
1. In the resolved QN column, there are some html tags such as <sup></sup> or <sub></sub> left, as well as others (eg <font ...>)
Screen Shot 2023-03-28 at 11 36 17
Screen Shot 2023-03-28 at 11 34 58

2. If we switch to "Center" mode to define a center and a range in mm or cm, the query process seems taking forever. This is picked up by the e2e tests.

YuHsuan-Hwang:
> @YuHsuan-Hwang in general this looks good based on some manual tests and e2e tests. A few issues spotted:
> 
> 1. In the resolved QN column, there are some html tags such as  or  left, as well as others (eg <font ...>)

@kswang1029 Do you mean  only the resolved QN column or also other columns?

kswang1029:
> > @YuHsuan-Hwang in general this looks good based on some manual tests and e2e tests. A few issues spotted:
> > 
> > 1. In the resolved QN column, there are some html tags such as  or  left, as well as others (eg <font ...>)
> 
> @kswang1029 Do you mean only the resolved QN column or also other columns?

so far I only spotted the issue in the resolved QN column.

YuHsuan-Hwang:
> so far I only spotted the issue in the resolved QN column.

ok

src/services/SplatalogueService.ts
```diff
@@ -67,139 +70,136 @@ export class SplatalogueService {
...
-        }
-
-        const numColumns = headerEntries.length;
-
-        if (numColumns !== SplatalogueService.SplatalogueHeaders.length) {
-            throw new Error("Unexpected header data received from Splatalogue");
-        }
+        // remove data with invalid species id and invalid species name
+        // this will remove data which only stores the query string
+        data = data.filter(line => line?.["species_id"] !== null || line?.["name"]);
```
crocka: If line is undefined, then undefined !== null would equal true. Is that what you want? Also, should the || be && instead according to your comment on line 84?

YuHsuan-Hwang: > If line is undefined, then undefined !== null would equal true. Is that what you want?

You're right. I adjusted to first check if `line` is valid.

> Also, should the || be && instead according to your comment on line 84?

I would like to exclude lines with both species id and species name invalid, so I think `||` works fine.

YuHsuan-Hwang:
@kswang1029 
> 1. In the resolved QN column, there are some html tags such as  or  left, as well as others (eg <font ...>)

fixed. I found html tags in chemical name column (and they exist in dev branch), so I also removed them.

image --> image

Another detail is that there are also additional spaces in species column, but I think this is very minor. These are updated in the description.

> 2. If we switch to "Center" mode to define a center and a range in mm or cm, the query process seems taking forever. This is picked up by the e2e tests.

fixed

kswang1029:
@YuHsuan-Hwang the regression spotted by the e2e test is fixed. 👍 Please convert this to formal PR and please proceed _code review_ to save time. I will perform extra manual test to know more about this new query backend meanwhile.

kswang1029:
Spotted some minor parsing issues:
1. in the Species column, `&#150;` appears. It should be "-" so in the following example, the species is C6H-.
Screen Shot 2023-04-14 at 13 32 38
There is also `&le;` as "<=".

2. in the Species column, Resolved QNs column, and Unresolved QNs column, possibly there are greek letters, such as `&alpha;`, `&gamma;`, `&epsilon;` (recombination lines), `&Omega;` (as state in QN). I suggest we show them as `alpha`, `epsilon`, `Omega`, etc.. so that users can still easily use the filter.
Screen Shot 2023-04-14 at 13 42 18

veggiesaurus (approved):
Code looks fine to me :+1:

crocka (approved):

YuHsuan-Hwang:
@kswang1029 I think it's not safe to convert `&xxx;` to `xxx` because there might be strings similar to `&le;`, unless we list all the possibilities.
Alternatively, I converted the symbols with `DOMParser`. Please check if this works fine, thanks.
Actually this exists in the dev branch, so I have noted it in the description.

kswang1029:
> @kswang1029 I think it's not safe to convert `&xxx;` to `xxx` because there might be strings similar to `&le;`, unless we list all the possibilities.
> Alternatively, I converted the symbols with `DOMParser`. Please check if this works fine, thanks.
> Actually this exists in the dev branch, so I have noted it in the description.

yes, certainly this is a much better solution. Then shall we enable "select" in the table, so users can do copy-and-paster for filtering purposes?

kswang1029:
this might work https://github.com/palantir/blueprint/blob/b0f1aff2b5a70433778e3b6661db90bdf755bfa2/packages/docs-app/src/examples/table-examples/tableSortableExample.tsx#L206

or demo https://blueprintjs.com/docs/versions/3/#table/features.sorting

(this table behaves not the way I imagined 🤦🏻‍♂️)

I think we need selection and keyboard copy. The right-click context menu is not needed.

kswang1029:
> this might work https://github.com/palantir/blueprint/blob/b0f1aff2b5a70433778e3b6661db90bdf755bfa2/packages/docs-app/src/examples/table-examples/tableSortableExample.tsx#L206
> 
> or demo https://blueprintjs.com/docs/versions/3/#table/features.sorting
> 
> (this table behaves not the way I imagined 🤦🏻‍♂️)
> 
> I think we need selection and keyboard copy. The right-click context menu is not needed.

Postponing this as a future work due to its complexities released to a shared component.

kswang1029 (approved):
e2e tests are revised accordingly. No regression is observed from other e2e tests. Manual tests are fine as far as I can tell.

There is a potential performance issue: it seems with this new advanced backend query, performance degradation is observed comparing to the v3-stable spatalogue-frontend query.  This requires future investigation but it is not a showstopper of merging.

UPDATE: the speed is not stable. it might be due to internet.

Issue #2142, opened 2023-04-14
Improve CI robustness


ajm-ska:
**Describe the bug**
Our carta-frontend CI occasionally fails on the "Production build stage". The error being that GSL header files are missing:

> gsl_wrapper.cc:2:10: fatal error: 'gsl/gsl_math.h' file not found
> #include "gsl/gsl_math.h"
>          ^~~~~~~~~~~~~~~~
> 1 error generated.

The issue actually originates during the previous "Build libs" stage, specifically in the `build_gsl.sh` script. But that stage is issued as a success even though the GSL libraries failed to build.

The problem is that our self-hosted Github Runners fail to download the GSL source code from http://ftpmirror.gnu.org/gsl/gsl-2.6.tar.gz

The download problem only happens with GSL and not the other library builds (ast, zfp, zstd), so it suggests it is not a network problem at our end. It typically occurs during a PR. Before #2139 the CI performed builds on 3 node versions in parallel (Node v12, v14, and v16). However, when there is a PR, there are actually two sets of builds; one with the current branch, and one of what it would be if merged into the dev branch. Therefore would be 6 parallel builds. Not to mention the possibility of other builds from other branches and PRs. Therefore, I suspect there could be a lot of requests during a short period of time to the GSL download server which could be considered suspicious and access from our IP address is likely blocked by them for a short amount of time, perhaps through a service such as fail2ban.

**To Reproduce**
1. Make several PRs and commits.
2. Wait for the CI to fail on the Production build stage.
3. See error

**Expected behavior**
The CI should not fail due to a missing "gsl/gsl_math.h", or if it fails due to that, the CI should have reported the issue and already exited at that stage.

I will improve the `build_gsl.sh` script to address these issues.

PR #2139, opened 2023-04-11
ajm/ci node update

ajm-ska:
**Description**

This PR will address #2132. Please see the issue for more information. In summary, as Node v12 is end-of-life (EOL) and Node v14 will be EOL by the end of the month, this PR removes the Node v12 and v14 builds. It adds a Node v18 build instead, and also runs the format-check in v16 instead of v14. It does:

- carta-frontend format-check using Node v16 LTS
- carta-frontend production build using Node v16 LTS
- carta-frontend production build using Node v18 LTS

If you disagree with this plan, please let me know.

Now that #1974 is merged with the 'dev' branch, the  `--openssl-legacy-provider` flag is no longer needed for the Node v18 build. But it seems the `--legacy-peer-deps` flag is still required.

By reducing the number of simultaneous builds from 3 (v12, v14, v16) to 2 (v16, v18), it might be less likely to run into the problem of downloading GSL during the "Build libs" section but I will try to address that issue further in another PR.

**Checklist**

For linked issues (if there are):
- [ ] assignee and label added
- [ ] ZenHub issue connection, board status, and estimate updated

For the pull request:
- [x] reviewers and assignee added
- [ ] ZenHub estimate, milestone, and release (if needed) added
- [x] e2e test passing / ~corresponding fix added~
- [x] no changelog update needed
- [x]  no protobuf update needed
- [x] `BackendService` unchanged / `BackendService` changed and corresponding ICD test fix added

PR #2143, opened 2023-04-14
ajm/build libs update

ajm-ska:
**Description**

This should address issue #2142 

Although the issue only happens for the GSL library build, I have made the same modification to the AST, ZFP, and ZSTD build scripts for consistency. 

These changes are only necessary for the carta-frontend Github Actions CI. I doubt there would ever be such an issue for individual developers, particularly because if a library source code has been previously downloaded, the script will not download it again. 

I attempted to have a local copy of the library file source code already downloaded on our self-hosted Github Runner servers, and have it simply copy over e.g. To the GitHub Actions workflow file I added:

> if [[ -f /opt/ast-9.1.1.tar.gz ]]; then cp /opt/ast-9.1.1.tar.gz ./wasm_libs/; fi

Unfortunately, this simple solution does not appear to work at all. It seems Github Actions on the self-hosted runners has no access outside its own Github Action build directory, which is probably good security-wise (Well, I suppose another option would be to have our GitHub Actions workflow file 'wget' the source code bundle from one of our local servers instead. But I did not do so and am trying to address the issue in another way).

An explanation of the changes are as follows:

- If a source code fails to download, the script will wait 10 seconds and try one more time. If it fails the 2nd time, an `exit 1` will be issued and the CI will stop and not continue.
- If a library fails to build, an `exit 1` has been added after the `echo "Not found!"` which will also stop the CI from continuing to the next stage. 

I know a 10 seconds wait and one retry may not be enough if the GSL server has already temporarily blocked our access. However, I think the issue is less likely to occur after PR #2139 where the number of parallel CI builds has been reduced from 3 to 2. Also, at least the changes in this PR will stop the CI if an error occurs during the `Build Libs` stage, and the one extra download retry may help add some robustness to the build.

**Checklist**

For the pull request:
- [x]  no protobuf update needed
- [x] `BackendService` unchanged

### DIFF
--- a/wasm_libs/build_gsl.sh
+++ b/wasm_libs/build_gsl.sh
@@ -15,9 +15,15 @@ if ! [[ $(find gsl-2.6.tar.gz -type f 2>/dev/null && md5sum -c gsl.md5 &>/dev/nu
         echo "Failed to fetch GSL 2.6 from http://ftpmirror.gnu.org"
         if ! wget https://mirror.ossplanet.net/gnu/gsl/gsl-2.6.tar.gz; then
             echo "Failed to fetch GSL 2.6 from https://mirror.ossplanet.net"
-            if ! wget https://mirrors.ocf.berkeley.edu/gnu/gsl/gsl-2.6.tar.gz; then
-                echo "Failed to fetch GSL 2.6 from https://mirrors.ocf.berkeley.edu"
-                exit 1
+            if ! wget https://ftp.jaist.ac.jp/pub/GNU/gsl/gsl-2.6.tar.gz; then
+                echo "Failed to fetch GSL 2.6 from https://ftp.jaist.ac.jp" 
+                if ! wget https://mirrors.ocf.berkeley.edu/gnu/gsl/gsl-2.6.tar.gz; then
+                    echo "Failed to fetch GSL 2.6 from https://mirrors.ocf.berkeley.edu"
+                     if ! wget https://mirror.dogado.de/gnu/gsl/gsl-2.6.tar.gz; then
+                         echo "Failed to fetch GSL 2.6 from https://mirror.dogado.de"
+                         exit 1
+                     fi
+                fi
             fi
         fi
     fi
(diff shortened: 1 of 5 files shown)


wasm_libs/build_gsl.sh
```diff
@@ -3,7 +3,18 @@ command -v emcc >/dev/null 2>&1 || { echo "Script requires emcc but it's not ins
 cd "${0%/*}"
 if ! [[ $(find gsl-2.6.tar.gz -type f 2>/dev/null && md5sum -c gsl.md5 &>/dev/null) ]]; then
     echo "Fetching GSL 2.6"
-    wget http://ftpmirror.gnu.org/gsl/gsl-2.6.tar.gz
+    retry_count=0
+    max_retries=2
+    while (( retry_count < max_retries )); do
+        wget http://ftpmirror.gnu.org/gsl/gsl-2.6.tar.gz && break
```
kswang1029, line 9: if I visit gnu.org and let it redirect me to the nearest mirror to get a copy of gsl, it is https://mirror.ossplanet.net/gnu/gsl/. So maybe if http://ftpmirror.gnu.org/gsl/gsl-2.6.tar.gz is not accessible, we can try this mirror?

kswang1029: or this is not just a wget from Taiwan?

ajm-ska: @kswang1029 Trying different mirrors is an excellent idea! I know it says [here](https://www.gnu.org/prep/ftp.html) that by going to http://ftpmirror.gnu.org, it automatically chooses a nearby up-to-date mirror. I didn't know that before. I have just checked with `wget http://ftpmirror.gnu.org/gsl/gsl-2.6.tar.gz` and it used `mirror.ossplanet.net`.
So maybe to be even safer, we should explicitly try a completely different mirror too.

How about the following change: I have kept the retry part in all four of the download scripts. However, to the GSL script, I have added the fallback to try to download from ossplanet once, plus an extra fallback to download from a mirror in Berkeley.

I have double checked the md5sum of the gsl-2.6.tar.gz from ossplanet and Berkeley and they agree with the md5sum we have on file.

kswang1029: yes, having a few explicit mirror sites to try is a good idea. We may consider mirrors in Japan and US perhaps?

ajm-ska: OK. I have added two more explicit mirror sites. Now it will try Taiwan > Japan > US (Berkeley) > Europe (Germany). It should definitely be able to succeed every time now.

kswang1029 (approved):

YuHsuan-Hwang (approved):

PR #2148, opened 2023-04-28
Yuhsuan/2145 line table copy

### LINKED ISSUE #2145: enable cell copy in the spectral line query result table
https://github.com/CARTAvis/carta-frontend/pull/2126 contains an improvement of showing greek letters and other mathematical symbols properly in the table. It would be a good enhancement if we enable cell copy function so that users can just copy a cell from the Species column and perform filtering in a straightforward way. Without the ability of copy, when there is a special symbol, we need to use OS feature for inputting (unless users' keyboard has native support such as Greek keyboard). 

Note that the table in the spectral line query widget is a shared component from the implementation point of view. The catalog widget also uses the same shared table component. So careful implementation and testing are required.

YuHsuan-Hwang:
**Description**

Simple CSS fixes.

1. Closes #2145: enabled user-select for the spectral line query table. Although the behavior of multiple cell selection may be a bit weird, this is still useful for copying special symbols for filtering.

    image

2. Changed the path input and the arithmetic input in the file browser dialog to `monospace`. The placeholders remain unchanged.

    image
    image

**Checklist**

For the pull request:
- [ ] e2e test passing / corresponding fix added

### DIFF
--- a/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.scss
+++ b/src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.scss
@@ -55,4 +55,14 @@
 .directory-path-input {
     width: 100%;
     margin-right: 25px;
+
+    .bp3-input:not(:placeholder-shown) {
+        font-family: monospace;
+    }
+}
+
+.arithmetic-input {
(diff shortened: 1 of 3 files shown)


kswang1029 (approved):
looks good.

PR #2172, opened 2023-05-18
carli/2171_two_image_one_pv_widget_bug

### LINKED ISSUE #2171: [v4-beta user feedback]
**Describe the bug**
This is a user feedback.

If we have two cube loaded in carta, the PV preview function works only on the first image. On the 2nd image, when we drag the pv cut, we don't see live update of the pv preview.

@crocka could you have a check please? (I am not sure which side causing the issue for now)

**To Reproduce**
1. load two cubes in carta (can have them matched or unmatched)
2. try pv preview on the first image (it should work expectedly)
3. try pv preview on the 2nd image 

**Expected behavior**
pv preview should work when there are multiple image cubes loaded

**Screenshots or videos**
If applicable, add screenshots/videos to help explain your problem.

 - OS [e.g. macOS Monterey]: macOS ventura
 - Browser [e.g. chrome, safari, electron app]: electron app
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: v4-beta
 - Frontend branch [e.g. dev, v3b2 release]: v4-beta

crocka:
**Description**

When user requests a preview, the code first checks if there is an existing preview frame, if there is, then the code skips the steps of creating a FrameStore and updating the regionId in the PvGeneratorWidgetStore. The bug is caused by this implementation, because if the same Pv widget is first used to create a preview in imageA, then when we try to create preview for a pv cut on imageB, the code would fail to update the regionId in the widget store for that particular Pv generator widget, resulting in failing to recognize that the pv cut on image B is a preview cut, which allows the code to skip live update. The fix would be simply updating the regionId every time the user requests a preview. The PR resolves #2171.

**Checklist**

For the pull request:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 * Fixed bug where line region computation width cannot be changed in spatial profile setting widget ([#2000](https://github.com/CARTAvis/carta-frontend/issues/2000)).
+* Fixed when multiple images are open, PV generator can only produce PV preview with live update for one of the images ([#2171](https://github.com/CARTAvis/carta-frontend/issues/2171)).
 
 ## [4.0.0-beta.1]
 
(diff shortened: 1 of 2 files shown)


kswang1029 (approved):
thanks for fixing this quickly. Tested and the issue is gone. no regression from e2e tests. Will add tests for ipv asap.

YuHsuan-Hwang (approved):
@crocka Please update the change log, thanks.

PR #2173, opened 2023-05-19
Yuhsuan/2140 support undefined in scripting

### LINKED ISSUE #2140: Scripting: support for distinct "undefined" and "null" parameters
It's impossible to serialize `undefined` to JSON. The Python wrapper serializes `None` as `null` in JSON and deserializes `null` as `None`. The frontend serializes *both* `null` and `undefined` to `null`, and deserializes `null` as `null`. This means that it is currently impossible to set an action parameter to `undefined` from the wrapper. Additionally, it seems that the scripting service is rejecting parameters with a value of `null`, and that the macro expansion evaluates `undefined` parameters as `null`.

The specific use case which triggered this is `setIntensityRange` in the vector overlay config. If a min or max option is undefined, a calculated default is used. It's possible for the user to set only one of the two options, but there's only one action to set both at the same time.

In the GUI widget, when one option is changed, the current value of the other option is passed in as the other parameter (if the current value is `undefined`, that is preserved). I *thought* that we could replicate this from the wrapper by using a macro to refer to the other value -- because that avoids the serialization / deserialization round trip, and evaluates the value entirely in the frontend -- but after careful testing using both the wrapper and the snippet interface, I have realised that if the value referenced by the macro is `undefined`, the macro expands it to `null`. This is a problem, because null and undefined are treated differently by the function using these options (`undefined` causes the default value to be used, but `null` is treated as zero).

The user can also explicitly clear/unset one or both of the options, which would require us to send an explicit "undefined" value from the wrapper, and this is currently impossible.

The current workaround for this function is to use a non-empty string (it has to be non-empty because `""` is zero in JS), but this is a very fragile hack -- I have confirmed by inspecting the values through the snippet interface that this actually causes the options to be set to these nonsense string values, but downstream these invalid values have the same effect as unset values. If I set the values to `null` through the snippet interface, they're treated as a zero values, so in this case `null` and `undefined` are not interchangeable.

Using `None` (which the frontend deserializes to `null`, so it wouldn't be the correct value in any case) causes an error. No error message is sent back to the wrapper (I'm not sure why that is; I can see one when I inspect the scripting response in the console -- it may be a backend issue). The error in the message is `TypeError: e is null`. It's referencing minified code so I'm not entirely sure what the issue is, but we are not applying strict type checks to `null` and `undefined`, and I have confirmed that I *can* call this function with `null` parameters from the snippets UI, so it looks like this is happening in the scripting service.

```
# This works to set both values:
img.call_action('vectorOverlayConfig.setIntensityRange', 2, 3)

# This works, but not if the other value is undefined (it's converted to null)
img.call_action('vectorOverlayConfig.setIntensityRange', img.macro("vectorOverlayConfig", "intensityMin"), 3)

# This causes an error -- trying to use None for the undefined value
img.call_action('vectorOverlayConfig.setIntensityRange', None, 3)

# Horrible workaround for unsetting: non-empty string
img.call_action('vectorOverlayConfig.setIntensityRange', "zzz", 3)
```

So there are three related issues here:

1. It's impossible to send an explicit `undefined` value (perhaps we could use a macro-like placeholder for this).
2. The macro substitution converts `undefined` to `null` (`undefined` needs to be preserved).
3. Null parameters are rejected (and they shouldn't be; they could be valid values for some action parameters).

The first two issues are a problem for https://github.com/CARTAvis/carta-python/pull/83 -- I previously thought that we could technically make everything work using the string workaround, but I have now realised that `undefined` values are not preserved by the macro expansion. The last issue is not directly related to this functionality (since we don't actually need to use `null` parameters here), but it's probable that we will need to use them in future.

YuHsuan-Hwang:
**Description**

Closes #2140.

Changes (in the order of the commits):

1.
> No error message is sent back to the wrapper (I'm not sure why that is; I can see one when I inspect the scripting response in the console -- it may be a backend issue).

This is because an Error object was given to the message instead of a string in `ScriptingService.handleScriptingRequest`. Fixed by converting the Error object to a string.

2.
> Null parameters are rejected (and they shouldn't be; they could be valid values for some action parameters).

This is because `typeof null === "object"` [returns true](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null) in `ExecutionEntry.mapMacro`. `parameter.macroVariable` is not defined for null, resulting in error. Fixed by adding error handlings for `parameter`.

Now `img.call_action('vectorOverlayConfig.setIntensityRange', None, 3)` should be able to set min to 0 and max to 3.

3.
>The macro substitution converts undefined to null (undefined needs to be preserved).

This is because `undefined == null` [returns true](https://javascript.info/comparison#comparison-with-null-and-undefined) in `ExecutionEntry.GetTargetObject` and target is set to null instead of undefined. Fixed by changing to strict equality check `undefined === null`.

Now `img.call_action('vectorOverlayConfig.setIntensityRange', img.macro("vectorOverlayConfig", "intensityMin"), 3)` should be able to set min to undefined and max to 3.

4.
>It's impossible to send an explicit undefined value (perhaps we could use a macro-like placeholder for this).

I adjusted `ExecutionEntry.mapMacro` to return `undefined` when macroVariable is string `"undefined"` (I assume there won't be variables called "undefined").

Now both `img.call_action('vectorOverlayConfig.setIntensityRange', img.macro("", "undefined"), 3)` and `img.call_action('vectorOverlayConfig.setIntensityRange', Macro("", "undefined"), 3)` should be able to set min to undefined and max to 3.

@confluence If you prefer other interface for `undefined` values please let me know, and I can adjust the code.

**Checklist**

For the pull request:

### DIFF
--- a/src/services/ScriptingService.ts
+++ b/src/services/ScriptingService.ts
@@ -112,7 +112,7 @@ export class ExecutionEntry {
             } else {
                 target = target[targetEntry];
             }
-            if (target == null) {
+            if (target === null) {
                 return null;
             }
         }
--- a/src/services/ScriptingService.ts
+++ b/src/services/ScriptingService.ts
@@ -173,7 +173,7 @@ export class ScriptingService {
                 response: JSON.stringify(toJS(response))
             };
         } catch (err) {
-            console.error(err)
+            console.error(err);
             return {
                 scriptingRequestId: requestMessage.scriptingRequestId,
                 success: false,
--- a/src/services/ScriptingService.ts
+++ b/src/services/ScriptingService.ts
@@ -121,6 +121,9 @@ export class ExecutionEntry {
 
     private mapMacro = (parameter: any) => {
         if (typeof parameter === "object" && parameter?.macroVariable) {
+            if (parameter.macroVariable === "undefined") {
+                return undefined;
+            }
             const targetString = parameter?.macroTarget ? `${parameter.macroTarget}.${parameter.macroVariable}` : parameter.macroVariable;
             return ExecutionEntry.GetTargetObject(AppStore.Instance, targetString);
         }
--- a/src/services/ScriptingService.ts
+++ b/src/services/ScriptingService.ts
@@ -174,10 +174,11 @@ export class ScriptingService {
                 response: JSON.stringify(toJS(response))
             };
         } catch (err) {
+            console.error(err)
             return {
                 scriptingRequestId: requestMessage.scriptingRequestId,
                 success: false,
-                message: err
+                message: err?.toString()
             };
         }
     };
--- a/src/services/ScriptingService.ts
+++ b/src/services/ScriptingService.ts
@@ -54,7 +54,7 @@ export class ExecutionEntry {
             }
             this.parameters = JSON.parse(substitutedParameterString);
         } catch (e) {
-            console.log(e);
+            console.error(e);
             return false;
         }
         return true;
@@ -88,15 +88,11 @@ export class ExecutionEntry {
         let target = baseObject;
         const targetNameArray = targetString.split(".");
 
-        console.log(targetNameArray);
-
         for (const targetEntry of targetNameArray) {
             const arrayRegex = /(\w+)(?:\[(\d+)])?/gm;
             const matches = arrayRegex.exec(targetEntry);
             // Check if there's an array index in this parameter
             if (matches && matches.length === 3 && matches[2] !== undefined) {
-                console.log(target);
-                console.log(matches);
                 target = target[matches[1]];
                 if (target == null) {
                     return null;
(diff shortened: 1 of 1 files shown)


confluence (approved):
I have confirmed that all the issues have been resolved. I think that using a macro with a target of `"undefined"` makes sense -- we can add a special macro subclass for this to the wrapper.

I-Chenn (approved):
Confirmed that the use of macro to the fronend's `undefined` does work.

kswang1029 (approved):
no regression from e2e test. 👍 @crocka please proceed code review.

crocka (approved):

kswang1029:
@YuHsuan-Hwang once a "merge dev" is done, this PR is ready for merge.

kswang1029:
@confluence @I-Chenn this is in.

PR #2184, opened 2023-06-05
Yuhsuan/2015 scripting set channels

### LINKED ISSUE #2015: Scripting support: separate, consistent actions for changing the channel and polarization
Currently there is only an action for changing both the channel and polarization at once, so if you want to change only one of them you have to explicitly set the other to its current value after retrieving it. This does not map well conceptually to the behaviour of the GUI, where you always change one at a time, since you can only click on one selector at once.

Additionally, polarizations are selected by Stokes index *unless* they're computed, in which case they're selected by value, which is inconsistent and confusing.

The Python wrapper can hide this from the user in a high-level function, but snippets don't have a layer like this. It would be useful to add two actions to the frame store which change *only* the channel and *only* the polarization (always by value). They could both call the current action, if that is efficient.

YuHsuan-Hwang:
**Description**

Closes #2015 by adding new methods `setChannel`, `setStokes`, and `setStokesByIndex`.

* The usage is documented in the codebase. I tried to clarify the usage with two terms:
    Polarization: corresponds to the options in the animator UI.
    image
    Stokes: defined as polarization index for standard polarization and polarization value for computed polarization. Used in protobuf messages.
* Refactored code to use the new added methods.
* `setStokes` is not used in frontend, so I noted “required for carta-python” to prevent deleting it.

**Checklist**

For the pull request:

### DIFF
--- a/src/components/Animator/AnimatorComponent.tsx
+++ b/src/components/Animator/AnimatorComponent.tsx
@@ -60,7 +60,7 @@ export class AnimatorComponent extends React.Component<WidgetProps> {
             if (val >= frame.frameInfo.fileInfoExtended.depth) {
                 val = 0;
             }
-            frame.setChannels(val, frame.requiredStokes, true);
+            frame.setChannel(val);
         }
     };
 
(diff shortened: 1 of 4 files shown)


confluence (approved):
This looks sensible to me. I will make a carta-python issue for updating the wrapper code to use this once it is merged.

kswang1029 (approved):
Manual tests are fine as far as I can tell. No regression from e2e tests.

crocka (approved):

kswang1029:
@confluence i have merged in this for the python scripting interface.

PR #2188, opened 2023-06-09
Yuhsuan/2156 scripting starting directory

### LINKED ISSUE #2156: Scripting feature request: optional flag to suppress changing the file browser directory when opening a file
It makes sense to save the last directory where a file was opened through the graphical file browser, but this behaviour makes less sense when the command to open a file is triggered by a script, bypassing the file browser. In the original implementation of the scripting wrapper, this set of commands would fail:

```
session.open_file("path/to/subdir/one.fits")
session.open_file("path/to/subdir/two.fits")
```

... because after the first command, the file browser's current directory would be changed to `path/to/subdir`, and the subsequent relative path would be resolved relative to that path which would either fail or (worse!) quietly succeed with the wrong file. This is very unexpected behaviour -- I don't think that a user would expect this side effect when they are not using the file browser to navigate to the file.

Saving and restoring the previous directory leads to a mess of race conditions, so at the moment the wrapper deals with this issue by maintaining its own separate `pwd`, which does not change unless the user explicitly changes it, and rewriting all relative file paths to absolute paths relative to this `pwd`. But while this solution is fine for non-interactive scripts, it can be confusing for someone using an interactive scripting session if there are two different base directories in the GUI and in the wrapper. Intuitively I think the user would expect that if they change the directory by opening a file in the GUI, that directory will be the new base directory in the wrapper as well.

The cleanest way to solve this would be an additional boolean flag in `addFrame`, which can be true by default, and which determines whether the file browser directory is updated with the frame's directory. This flag would also need to be added to `loadFile`, `openFile` and `appendFile`, so that the wrapper can set it to false when using these functions. It probably doesn't need to be added to `loadConcatStokes`, because the directory parameter is independent of the actual file paths and can be set to the current directory, or to any of the functions for opening generated images, for the same reason.

If you open a single file in the GUI using a LEL expression with a path to a different directory, that doesn't change the directory, so technically we could use that as a hack to achieve this behaviour when using `loadFile`, but this is not an ideal solution because the file title (and possibly other metadata) is different.

YuHsuan-Hwang:
**Description**

Closes #2156 by adding flags to `openFile` and `appendFile`:

```
// open files without changing the starting directory
const file = await app.openFile(PATH, FILENAME, HDU, IS_LEL, false);

// append files without changing the starting directory
const file2 = await app.appendFile(PATH, FILENAME, HDU, IS_LEL, SET_ACTIVE, false);
```

Documentation of the methods is added (or revised from jsdoc to tsdoc) in the codebase.

**Checklist**

For the pull request:

### DIFF
--- a/src/stores/AppStore/AppStore.ts
+++ b/src/stores/AppStore/AppStore.ts
@@ -479,7 +479,7 @@ export class AppStore {
         return frameMap;
     }
 
-    @action addFrame = (ack: CARTA.IOpenFileAck, directory: string, lelExpr: boolean, hdu: string, generated: boolean = false, setAsActive: boolean = true): boolean => {
+    @action addFrame = (ack: CARTA.IOpenFileAck, directory: string, lelExpr: boolean, hdu: string, generated: boolean = false, setAsActive: boolean = true, updateStartingDirectory: boolean = true): boolean => {
         if (!ack) {
             return false;
         }
@@ -551,7 +551,10 @@ export class AppStore {
                 this.setSpectralMatchingEnabled(newFrame, true);
             }
         }
-        this.fileBrowserStore.saveStartingDirectory(newFrame.frameInfo.directory);
+
+        if (updateStartingDirectory) {
+            this.fileBrowserStore.saveStartingDirectory(newFrame.frameInfo.directory);
+        }
 
         return true;
     };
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
looks good. Tested manually with GUI. No regression from e2e tests.

crocka (approved):

confluence (approved):
I can confirm that this makes it possible to open images from the wrapper without changing the starting directory. I have made a wrapper issue for updating the wrapper once this is merged.

PR #2194, opened 2023-06-23
Carli/1327 spectral line off precision

### LINKED ISSUE #1327: spectral line label off by one channel?
In the following example, if we set the velocity shift to be 1540 km/s, it is expected to see the label is plotted at 1540 km/s but it appears there is one channel offset. Changing to other velocity shift (eg 1600 km/s) has the same issue. @duidae do you think this is due to the 0-index/1-index issue?

Screen Shot 2021-02-23 at 1 58 05 PM

crocka:
**Description**

Closes #1327. The redshift factor is calculated incorrectly.

**Checklist**

For the pull request:
- [x] ~e2e test passing~ / corresponding fix added
- [ ] changelog updated / ~~no changelog update needed~~
- [ ] protobuf updated to the latest dev commit / ~~no protobuf update needed~~
- [ ] `BackendService` unchanged / ~~`BackendService` changed and corresponding ICD test fix added~~


kswang1029 (approved):
this looks good with some high-z sources. a fix to e2e test is applied. please proceed code review.

kswang1029:
@crocka @YuHsuan-Hwang a quick question: are we using AST to perform frequency to radio velocity conversion (ie calling an AST function to do the conversion, instead of making our own conversion function)? If so, are we able to tell the formula of the AST conversion? I would like to figure our this detail and probably this is the key to explain why our relativistic formula does not work.

YuHsuan-Hwang:
Yes, I think we are using AST. Will need to check the user guide.

kswang1029:
> Yes, I think we are using AST. Will need to check the user guide.

ok thanks. My speculation is that AST does not use relativistic formula to convert frequency to radio velocity. But in CARTA before @crocka's fix, we use relativistic formula to compute line ID label shift. Now @crocka uses non-relativistic formula and we don't see the inconsistency.

kswang1029 (approved):
conventionally frequency to radio velocity conversion is non-relativistic. The original line ID label shift formula we adopted is relativistic. So this is the root cause of the offset we observed when source has high(er) radial velocity. The current fix switches back to the non-relativistic formula for consistency.

YuHsuan-Hwang (approved):

YuHsuan-Hwang:
Ready for merge. @crocka Please update the change log, thanks.

YuHsuan-Hwang:
@crocka Please also update to the latest dev branch, thanks.

PR #2206, opened 2023-07-18
Yuhsuan/inactive frame channel

### LINKED ISSUE #103: raster image is not re-rendered
with the following snippet there are two image re-rendering issues

```
img0 = session.open_image(image_file_7)
img1 = session.append_image(image_file_1)

img0.make_active() # issue1
img0.set_channel(20)
img0.set_polarization(Polarization.Q)

time.sleep(5) # issue2

img1.make_active()
img1.set_channel(20)
```

Issue1
In this case, img1 will be the active image. Then if I do img0.set_channel(20) and/or img0.set_polarization(Polarization.Q) without img0.make_active(), we will see indeed in the image list widget, the channel index and polarization string are updated. However, the rendered image is not. We need to do img0.make_active() first to re-render the raster image. I suggest in set_channel and set_polarization we internally make the target image as "active" before setting channel/polarization. In the scripting interface when there are multiple image loaded, it is non-trivial which one is currently "active" I suppose.

Issue2
In the above snippet, if I comment out time.sleep(5), even if we do img0.make_active() first before setting channel and polarization, we do not see img0 got re-rendered. I suspect this is an sync/async issue.

YuHsuan-Hwang:
**Description**

Closes https://github.com/CARTAvis/carta-python/issues/103: setting the channel/stokes of inactive frames. Fixed by adjusting the criteria for requesting tiles when channel/stokes is manually changed.

Instead of only checking the active frame and its spectrally matched frames, we now check all visible frames and their spectrally matched frames. When checking the spectrally matched frames, we skip the visible frames to avoid duplicate requests.

Currently, users cannot set the channel/stokes of inactive frames through UI, but it is possible to do so using the code snippet feature or the console. For example:
```javascript
// in multi-panel mode showing >= 2 frames
const file = await app.openFile("[path]", "[filename1]");
const file2 = await app.appendFile("[path]", "[filename2]");

// set inactive frame channel
file.setChannel(1);
// set inactive frame stokes
file.setStokes(1);
```
In this branch, the above code snippet will immediately request new tiles for the inactive frame. In the dev branch, it will request new tiles after the frame becomes active.

@acdo2002 There will be additional tile request messages with this change, so you might want to check it.

**Checklist**

For the pull request:

### DIFF
--- a/src/stores/AppStore/AppStore.ts
+++ b/src/stores/AppStore/AppStore.ts
@@ -1812,10 +1812,10 @@ export class AppStore {
                         const isVisible = this.visibleFrames.includes(frame);
                         const siblingUpdateRequired = frame.requiredChannel !== frame.channel || frame.requiredStokes !== frame.stokes;
                         if (!isVisible && siblingUpdateRequired) {
-                            updates.push({ frame, channel: frame.requiredChannel, stokes: frame.requiredStokes });
+                            updates.push({frame, channel: frame.requiredChannel, stokes: frame.requiredStokes});
                         }
                     });
-                }   
+                }
             }
 
             if (updates.length) {
--- a/src/stores/AppStore/AppStore.ts
+++ b/src/stores/AppStore/AppStore.ts
@@ -1796,26 +1796,30 @@ export class AppStore {
         // TODO: Move setChannels actions to AppStore and remove this autorun
         // Update channels when manually changed
         autorun(() => {
-            if (this.activeFrame) {
-                const updates: ChannelUpdate[] = [];
-                // Calculate if new data is required for the active channel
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
This should be fine. No regression from e2e tests.

confluence (approved):
I can confirm that the inactive image is now re-rendered immediately when the channel is changed.

crocka (approved):

crocka (approved):

acdo2002 (approved):
Currently the ICD-RxJS are all one file to test tiles' response.
I have confirmed this PR add additional ICD message when open two frames. ICD-RxJS need add one more tile test.

I-Chenn (approved):
Confirm that the inactive image is rendered right after the channel is changed with my test.

PR #2209, opened 2023-07-21
Thchang/ hide the blinking toolbar when animating images

### LINKED ISSUE #2163: blinking toolbar when animating images
When I animate to blink between two images, if the cursor is on the image so that the that the tool bar at bottom right appears it blinks with the animation.  Not sure if we want the tool bar to even appear during animation.  But if it does, it should not blink.

TienHao:
**Description**

closes #2163. The toolbar blinked during animating images.

The blinking was caused by property `imageToolbarVisible`, which controlled the visibility of the toolbar in ImagePanelComponent, being reset to `false` when the next image is displayed during animating images. After we have grid view function, the ImageViewComponent computes list of new ImagePanelComponent when displayed images change. 

This part could be fixed by passing properties from ImageViewComponent. But another problem raised, that the toolbar might still be remained after the cursor leaving the image view, since the `onMouseLeave` won't be triggered when the node of images gets detached. (refer the [issue](https://github.com/facebook/react/issues/6807) from React)

As a result, hiding the toolbar when animating images to fix this issue.

**Checklist**

For the pull request:


YuHsuan-Hwang (changes requested):
Currently we can click the invisible buttons during animation. We might need to avoid rendering the toolbar.

YuHsuan-Hwang (approved):

kswang1029 (approved):

crocka (approved):

PR #2231, opened 2023-08-15
Yuhsuan/region scripting

### LINKED ISSUE #2213: Scripting feature request: access to region list / region lookup by ID from Python wrapper
Could we please add computed properties to the region set store which 1) list the region IDs in a serializable structure, and 2) expose the region store objects in a map with region IDs as keys? Rationale given below:

The wrapper can easily access the list of images in the session, and to use an image ID to refer to a frame object in the frontend. This is because
1. the app store has a `frameNames` property (a simple, serializable structure which contains the indices, filenames and IDs of all images), which the wrapper can fetch and parse to get a list of all image IDs
2. the app store has a `frameMap` property with IDs as keys and frame stores as values, which the wrapper can reference using a macro which resolves to the frame object.

However, I cannot perform analogous operations with images and regions.

First, if I want to get a list of all the regions of an image, I can only access the `regions` property of the frame store's `regionSet`. But this is a list with region stores as values. It's not serializable, so I can't fetch the whole thing. To solve this problem for a single object, I can use the `returnPath` option to select a serializable portion of the object to return. This uses the lodash `get` function on the frontend side. Unfortunately it does not appear that this can be applied elementwise to a list or map.

(It looks like the lodash `map` function *can* be used exactly for this, so perhaps we could also extend the scripting service to offer this option. If we autodetected the type of the object, and used `map` on an array and `get` otherwise, we could reuse the existing `returnPath` option without having to modify the protocol.)

I can temporarily work around this by fetching the length of the array and then iterating in the wrapper to fetch each region ID individually by index, but this is inefficient, and would probably scale poorly for large numbers of regions.

The next problem is attempting to refer to a region store by its ID (which we need in order to call actions which take a region store as a parameter). I can create a region store macro if I know the index of the region, but I can't create a macro if I know the ID and want to look up the index (which may be important if e.g. some regions have been deleted and the indices have changed).

There is a `getRegion` function on the frame store, but I can't use this in a macro, because it's not a property. I can't call `regions.indexOf` from the wrapper, because that requires a region store object, which requires a macro, which is what I'm trying to get.

I could also technically work around this by fetching all the region IDs and performing the lookup in the wrapper, but that would also be very inefficient.

### LINKED ISSUE #2215: Scripting feature request: specify any frame when importing / exporting regions
`importRegion` and `exportRegions` in the app store always use the active image (or the spatial reference of the active image). The wrapper should be able to specify a different image -- this could be achieved with an optional `frame` parameter in these functions (defaulting to the current behaviour).

YuHsuan-Hwang:
**Description**

This PR is for scripting enhancements related to regions.

* Changes:

1. Closes #2213:
    `regionList` and `regionMap` are added to `RegionSetStore` to provide flexible ways to access the regions.

2. Closes #2215:
    added optional parameter `targetFrame` to `importRegion` and `exportRegions`.

3. Supports creating regions in scripting.
    * When we use the existing functions to create regions, the functions return the RegionStore object without waiting for the backend update. For example:
    
        ```javascript
        const region = file.regionSet.addLineRegion([{x: [start x], y: [start y]}, {x: [end x], y: [end y]}]);
        console.log(region.regionId); // -1
        ```

        This will cause error if we immediately use the region ID, for example, generating a PV image from a line region. Therefore, a new async function `addRegionAsync` is added for creating regions in scripting:

        ```javascript
        const region = await file.regionSet.addRegionAsync(1, [{x: [start x], y: [start y]}, {x: [end x], y: [end y]}]); // add line region
        console.log(region.regionId) // 1
        ```
        This ensures that the region is updated before executing the remaining script.

    * In order to create regions in WCS coordinates, a new function `getImagePosFromWCS` is added for position conversion. For example:

        ```javascript
        const start = file.getImagePosFromWCS([{x: "x in wcs", y: "y in wcs"}])[0];
        const end = file.getImagePosFromWCS([{x: "x in wcs", y: "y in wcs"}])[0];
        const region = await file.regionSet.addRegionAsync(1, [start, end]); // add line region
        ```

        Note that size conversion is already implemented:
        ```javascript
        const width = file.getImageXValueFromArcsec([width in arcsec]);
        const height = file.getImageYValueFromArcsec([height in arcsec]);
        const region = await file.regionSet.addRegionAsync(3, [{x: [center x], y: [center y]}, {x: width, y: height}]); // add rectangle region
        ```
    * A new function `getWCSFromImagePos` is also added for converting positions from image to wcs coordinates.

    Please check the added TSDoc for the details of the usages for the functions.
       
* Code maintenance:
    * Removed unused parameter `isAnnotation` in `addRegion`.
    * Refactored `addRegion`.
    * Refactored `createRegions` in image fitting and use the new function `addRegionAsync`.

* ToDo
     - [x] adjust the snippet tutorial in this PR or the documentation PR, depending on which one is merged first

**Checklist**

For the pull request:


kswang1029 (approved):
no regression from e2e tests (including region related tests)

src/stores/Frame/Region/RegionSetStore.ts
```diff
@@ -181,97 +203,78 @@ export class RegionSetStore {
...
-                    regionName
-                ]))();
-                break;
+                return new RulerAnnotationStore(...commonInputs, ...annotationStyles);
             case CARTA.RegionType.ANNTEXT:
-                region = new (TextAnnotationStore.bind.apply(TextAnnotationStore, [
-                    null,
-                    ...commonInputParameters,
-                    this.preference.annotationColor,
-                    this.preference.textAnnotationLineWidth,
```
crocka, line 217: The text annotation has unique textAnnotationLineWidth. Has this been overwritten by annotationLineWidth?

YuHsuan-Hwang:
New changes: convert an array of positions instead of one position in `getImagePosFromWCS` as @confluence suggested. (PR description is also updated.)

Note that the function applies `AST.transformPoint` for each position. If we frequently call the function with a large amount of positions (which is not the current use case), we might consider apply `AST.transformPointArrays` instead to improve the performance.

src/stores/Frame/Region/RegionSetStore.ts
```diff
@@ -71,6 +71,28 @@ export class RegionSetStore {
...
+     * Returns a map associating region IDs with their corresponding RegionStore instances.
+     */
+    @computed get regionMap(): Map<number, RegionStore> {
+        const reigonMap = new Map<number, RegionStore>();
+
+        for (const region of this.regions) {
+            reigonMap.set(region.regionId, region);
+        }
+
+        return reigonMap;
```
confluence: Minor issue: there's a typo in the local variable here.

kswang1029:
update: GUI e2e tests are passing. Manual tests of other use cases are in progress. please proceed code review.

src/stores/Frame/Region/RegionSetStore.ts
YuHsuan-Hwang: Fixed.

src/stores/Frame/Region/RegionSetStore.ts
YuHsuan-Hwang: Yes, this is fixed. Thanks for spotting this!

confluence:
I'm currently trying to integrate WCS support, and I think that I will also need functions to convert pixels back into WCS or angular sizes. Would that be possible?

Edit: I see that `getWcsSizeInArcsec` already exists, so I think this would only need to be added for world coordinates.

YuHsuan-Hwang:
> I'm currently trying to integrate WCS support, and I think that I will also need functions to convert pixels back into WCS or angular sizes. Would that be possible?

@confluence Yes, I can implement `getFormattedWCSPoint` in `FrameStore`.

kswang1029 (approved):
👍

YuHsuan-Hwang:
New change: added new function `getWCSFromImagePos`

crocka (approved):

confluence (approved):
I have done some basic tests with the wrapper branch, and I think everything here works as expected.

PR #2232, opened 2023-08-15
Yuhsuan/2222 stokes guess

### LINKED ISSUE #2222: Scripting feature request: make polarization guessing function accessible
The functions `getStokeType`, `getTypeFromHeader` and `getTypeFromName` are currently private functions in `StokesDialogComponent.tsx`. It would be helpful if these functions were accessible through one of the stores, so that the wrapper could reuse them.

YuHsuan-Hwang:
**Description**

Closes #2222 make polarization guessing function accessible ~by moving `getStokesType` from `StokesDialogComponent` to `FileBrowserStore`.~

Changes:

* Added function `getStokesFile` and move the code from component to stores.

    Example code snippet:
    ```javascript
    const currentDirectory = app.fileBrowserStore.fileList.directory;
    const filename = "example.fits";
    const hdu = "";
    const stokesFile = await app.fileBrowserStore.getStokesFile(currentDirectory, filename, hdu);
    console.log(stokesFile.polarizationType);
    ```
* Enhanced `GetTypeFromName` by adjusting splitting and converting to uppercase.
* Moved `GetStokesType`, `GetTypeFromHeader`, and `GetTypeFromName` to stores and set them to static.
* Set `getConcatFilesHeader` to private.

~ToDo:~
 - [x] ~add TSDoc for `getStokesType`~

**Checklist**

For the pull request:


src/stores/FileBrowserStore/FileBrowserStore.ts
```diff
@@ -317,6 +317,50 @@ export class FileBrowserStore {
         return {file: res.fileInfo.name, info: res.fileInfoExtended};
     };
 
+    getStokesType = (fileInfoExtended: CARTA.IFileInfoExtended, file: string): CARTA.PolarizationType => {
```
crocka: this can be made as static as well?

kswang1029:
GUI e2e tests are passing. Manual tests for other use cases under way. please proceed code review.

src/stores/FileBrowserStore/FileBrowserStore.ts
YuHsuan-Hwang: Static functions are not accessible to scripting, so it can not be changed to static.

crocka: Oh, I see. Thank you!

kswang1029 (approved):

crocka (approved):

src/stores/FileBrowserStore/FileBrowserStore.ts
```diff
@@ -317,6 +317,57 @@ export class FileBrowserStore {
...
+        return type;
+    };
+
+    private static GetTypeFromName = (fileName: string): CARTA.PolarizationType => {
+        let type = CARTA.PolarizationType.POLARIZATION_TYPE_NONE;
+        const separators = [".", "_"];
+        separators.forEach(separator => {
+            const words = fileName.split(separator);
+            words.forEach(word => {
+                const matchedType = CARTA.PolarizationType[word];
```
confluence: I would suggest uppercasing `word` here, so that `i`, `q`, etc., are not ignored. Unless we think that lowercase letters are likely to have a different meaning in filenames.

confluence (approved):
I have tested this with the wrapper PR, and it's working as expected.

src/stores/FileBrowserStore/FileBrowserStore.ts
YuHsuan-Hwang: This is added.

src/stores/FileBrowserStore/FileBrowserStore.ts
```diff
@@ -317,6 +317,57 @@ export class FileBrowserStore {
...
+        }
+
+        return type;
+    };
+
+    private static GetTypeFromName = (fileName: string): CARTA.PolarizationType => {
+        let type = CARTA.PolarizationType.POLARIZATION_TYPE_NONE;
+        const separators = [".", "_"];
+        separators.forEach(separator => {
+            const words = fileName.split(separator);
```
confluence: While testing this, I spotted another logic error here. The way these separators are being applied makes no sense: they're being used one at a time instead of together (e.g. if a file is called `foo_i.fits`, first this code will first split it into `foo_i` and `fits` and then into `foo` and `i.fits`, and never into `foo`, `i` and `fits`, which is what we actually want). We need a single split on multiple separators (so, using a regex as the parameter to `split`).

confluence (changes requested):

src/stores/FileBrowserStore/FileBrowserStore.ts
YuHsuan-Hwang: I see, will fix this.

YuHsuan-Hwang: fixed.

kswang1029 (approved):
retested and all good. no regression from e2e tests.

confluence:
There's an issue [here](https://github.com/CARTAvis/carta-python/pull/106#pullrequestreview-1391647521) which I think is being triggered by files which 1) contain header entries with no values (e.g. `HISTORY` entries) and 2) don't contain a `STOKES` entry (so the entire header is searched, and the line in `GetTypeFromHeader` which checks if the value is `STOKES` causes an error because it's trying to call `toUpperCase` on an undefined object).

I was trying to figure out why the same file was *not* causing a crash in the GUI, and I think that it may be because the original header entry objects have some kind of default accessor for the value field which returns an empty string -- but the corresponding objects passed to the frontend by the wrapper (passing through a serialization and deserialization round trip) do not. I have confirmed that if I edit the header entry dictionaries I'm passing from the wrapper to add a `value` key with an empty string value if it does not exist fixes the issue.

However, I think that it would be much simpler to avoid this round trip entirely. Currently the wrapper has to fetch the header for the file and then send it back to the frontend, because `getStokesType` takes the header as a parameter. The Stokes dialog component fetches the header with `getConcatFilesHeader`, gets the first extended info entry, and then calls `getStokesType` with that and the filename. Would it be possible to move that logic _inside_ `getStokesType`, and have `getStokesType` take the directory and filename as parameters? That would make it much simpler to call it from the wrapper, and would eliminate the issue with the header serialization.

YuHsuan-Hwang:
@confluence Please check if the added function `getStokesFile` works for you, thanks.

Example code snippet:
```javascript
const currentDirectory = app.fileBrowserStore.fileList.directory;
const filename = "example.fits";
const hdu = "";
const stokesFile = await app.fileBrowserStore.getStokesFile(currentDirectory, filename, hdu);
console.log(stokesFile.polarizationType);
```

I also assumed `getConcatFilesHeader` and `getStokesType` are no longer needed in scripting and set them to private.

~ToDos: add doc for `getStokesFile`; set `getStokesType` as static~

confluence:
Thanks! I'm testing it out now.

confluence (approved):
I have confirmed that the wrapper code works with the simplified function API.

crocka (approved):

PR #2277, opened 2023-10-20
Yuhsuan/1652 large catalog

### LINKED ISSUE #1652: regression: cannot load large catalog
In my test env, I tried to load a catalog with 100M sources.
with
- latest dev : the frontend soft crashed after streaming ~5.5M sources
- v3b1 Mac electron build: frontend hard crash at ~5.5M
- v2.0: it seems to load fine at the time of filing this issue (~24M) and keep streaming. UPDATE: I killed the process at ~33M as my OS is swapping a lot...

with dev I see
Screen Shot 2021-09-10 at 1 56 51 PM

YuHsuan-Hwang:
**Description**

Fixed the error `Invalid array length` in #1652 by setting the array size to the current received source number instead of the entire source number.

However, loading 100M source catalogs still crashes when ~30M sources are loaded (possibly caused by running out of memory). Do we need to fully support 100M source catalogs?

Update: closes #1652.

**Checklist**

For the pull request:
- [ ] e2e test passing / ~corresponding fix added~ / ~new e2e test created~
- [x] ~protobuf version bumped~ / protobuf version not bumped

### DIFF
--- a/src/stores/Catalog/CatalogProfileStore.ts
+++ b/src/stores/Catalog/CatalogProfileStore.ts
@@ -116,7 +116,6 @@ export class CatalogProfileStore extends AbstractCatalogProfileStore {
         const subsetEndIndex = catalogFilter.subsetEndIndex;
         const startIndex = subsetEndIndex - subsetDataSize;
 
-        const totalDataSize = catalogFilter.requestEndIndex;
         this.filterDataSize = catalogFilter.filterDataSize;
 
         if (this.subsetEndIndex <= this.filterDataSize) {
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
@YuHsuan-Hwang The fix looks good. With macOS-32GB ram, I can load ~43M sources before the frontend crashing. With ubuntu-64GB ram, I can also load ~43M sources before the frontend frozen (the browser keep notifying me the browser becomes unresponsive). I think it is okay for now. We can revisit this issue if we receive request from users.

No regression from e2e tests. 👍

YuHsuan-Hwang:
> The fix looks good. With macOS-32GB ram, I can load ~43M sources before the frontend crashing. With ubuntu-64GB ram, I can also load ~43M sources before the frontend frozen (the browser keep notifying me the browser becomes unresponsive). I think it is okay for now. We can revisit this issue if we receive request from users.
> 
> No regression from e2e tests. 👍

Great, then let's close the issue with this PR.

YuHsuan-Hwang:
I realized that we reallocate the arrays more frequently with the fix. Might need better solution.

PR #2280, opened 2023-10-27
carli/2246_scripting_ruler_setSize

### LINKED ISSUE #2246: Scripting: correctly support setSize for a ruler annotation
This is a minor bug / inconsistency in the region class: the `size` property correctly handles the ruler annotation in the same way as a line or vector, by returning the x and y distances between the endpoints. However, the `setSize` function does *not* include the ruler in the line and vector case (which calculates and sets the endpoints using the center point and the new size), and instead falls back to the default case, which sets the second control point to the values provided (which is not the correct interpretation of the control points for this class).

I don't think that this has any effect in the GUI, since this function is not exposed to the user in the edit dialog for this region class, but a scripting user can access this function on the parent class. I have added a workaround to the wrapper (implementing the correct method for this subclass), but a snippet user could run into the same problem, so it would be a good idea to fix this in the frontend.

Edit: `setCenter` has the same issue, and this class also doesn't store a rotation.

crocka:
**Description**

Closes #2246. This PR added ANNRULER to appropriate if-statement in regionStore. Also fixed the failure to update control points when the starting and ending points of a line region have the same x or y value.

**Checklist**

For the pull request:
- [x] ~e2e test passing~ / corresponding fix added / ~new e2e test created~
- [x] ~~protobuf version bumped~~ / protobuf version not bumped

### DIFF
--- a/src/stores/Frame/Region/RegionStore.ts
+++ b/src/stores/Frame/Region/RegionStore.ts
@@ -374,9 +374,7 @@ export class RegionStore {
         }
     };
 
-    public getRegionApproximation(
-        astTransform: AST.Mapping
-    ): Point2D[] | {northApproximatePoints: number[]; eastApproximatePoints: number[]} | {xApproximatePoints: number[]; yApproximatePoints: number[]; hypotenuseApproximatePoints: number[]} {
+    public getRegionApproximation(astTransform: AST.Mapping): Point2D[] {
         let approximatePoints = this.regionApproximationMap.get(astTransform);
         if (!approximatePoints) {
             if (this.regionType === CARTA.RegionType.POINT) {
@@ -455,7 +453,7 @@ export class RegionStore {
         if (this.regionType === CARTA.RegionType.POLYGON || this.regionType === CARTA.RegionType.ANNPOLYGON) {
             this.simplePolygonTest();
         }
-        if ((this.regionType === CARTA.RegionType.LINE || this.regionType === CARTA.RegionType.ANNLINE || this.regionType === CARTA.RegionType.ANNVECTOR) && controlPoints.length === 2) {
+        if ((this.regionType === CARTA.RegionType.LINE || this.regionType === CARTA.RegionType.ANNLINE || this.regionType === CARTA.RegionType.ANNVECTOR || this.regionType === CARTA.RegionType.ANNRULER) && controlPoints.length === 2) {
             this.rotation = this.controlPoints.length === 2 ? this.getLineAngle(this.controlPoints[0], this.controlPoints[1]) : 0;
         }
         this.modifiedTimestamp = performance.now();
@@ -506,7 +504,7 @@ export class RegionStore {
                 this.simplePolygonTest(index);
             }
 
-            if (this.regionType === CARTA.RegionType.LINE || this.regionType === CARTA.RegionType.ANNLINE || this.regionType === CARTA.RegionType.ANNVECTOR) {
+            if (this.regionType === CARTA.RegionType.LINE || this.regionType === CARTA.RegionType.ANNLINE || this.regionType === CARTA.RegionType.ANNVECTOR || this.regionType === CARTA.RegionType.ANNRULER) {
                 this.rotation = this.controlPoints.length === 2 ? this.getLineAngle(this.controlPoints[0], this.controlPoints[1]) : 0;
             }
         }
@@ -531,7 +529,7 @@ export class RegionStore {
             this.simplePolygonTest();
         }
 
-        if (this.regionType === CARTA.RegionType.LINE || this.regionType === CARTA.RegionType.ANNLINE || this.regionType === CARTA.RegionType.ANNVECTOR) {
+        if (this.regionType === CARTA.RegionType.LINE || this.regionType === CARTA.RegionType.ANNLINE || this.regionType === CARTA.RegionType.ANNVECTOR || this.regionType === CARTA.RegionType.ANNRULER) {
             this.rotation = points.length === 2 ? this.getLineAngle(points[0], points[1]) : 0;
         }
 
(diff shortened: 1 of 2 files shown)


YuHsuan-Hwang (approved):
The changes looks good. 👍

confluence:
Are we making a decision for the 2D size of line-shaped regions to have positive components regardless of the direction of the line? This will lead to some unexpected behaviour if the user interacts with region sizes using scripting:

* If the user sets a size with one or more negative components, the size returned with `size` will not be the same as the size set with `setSize`.
* Passing the size returned by `size` back into `setSize` (or another size with positive components) may transform the shape of the region, depending on what the original signs are:
  * If both raw values calculated from the endpoints have been negated, the endpoints will be switched. (Are there any calculations based on line regions where the direction of the endpoints is significant?)
  * If only one of the values has been negated, the region will be rotated by 90°!
* It is no longer possible to specify an orientation for the line by setting the size.

My suggestion would be to keep (half of) the current behaviour, where the 2D size of a line can have negative components, but to flip the endpoints in the calculation -- that is, to subtract endpoint 0 from endpoint 1, so that a component pointing in a positive direction will be positive. Then the orientation of the line will determine the signs of the components, an unambiguous orientation can be selected when the size is set, and the value returned by `size` will preserve the value that was set with `setSize`.

confluence:
Revised idea (as discussed in today's dev meeting): only allow positive sizes, but take the current rotation into account when setting the size, so that the current orientation of the line is preserved.

src/stores/Frame/Region/RegionStore.ts
```diff
@@ -483,8 +482,13 @@ export class RegionStore {
 
     @action setSize = (p: Point2D, skipUpdate = false) => {
         if (this.regionType === CARTA.RegionType.LINE || this.regionType === CARTA.RegionType.ANNLINE || this.regionType === CARTA.RegionType.ANNVECTOR || this.regionType === CARTA.RegionType.ANNRULER) {
-            const newStart = {x: this.center.x - p.x / 2, y: this.center.y - p.y / 2};
-            const newEnd = {x: this.center.x + p.x / 2, y: this.center.y + p.y / 2};
+            const rotation = (this.rotation * Math.PI) / 180.0;
+            const x = Math.abs(p.x);
+            const y = Math.abs(p.y);
+            const _x = x * Math.sign(Math.sin(rotation));
```
YuHsuan-Hwang: Perhaps rename `_x` to `newX`.

src/stores/Frame/Region/RegionStore.ts
```diff
@@ -483,8 +482,13 @@ export class RegionStore {
 
     @action setSize = (p: Point2D, skipUpdate = false) => {
```
YuHsuan-Hwang, line 494: It would be nice if we can add TSDoc to `setSize()` and `size`.

YuHsuan-Hwang (changes requested):
The changes are good.

When the rotation is 0, we fail to set x size. To reproduce with JS snippet:
```js
const file = await app.openFile([path], [any file]);
const region = await file.regionSet.addRegionAsync(1, [{x: 1.5, y: 2}, {x: 1.5, y: 1}]);
//region.setSize({x: 2, y: 3}); // incorrect
//region.setSize({x: 0, y: 3}); // correct
//region.setSize({x: 2, y: 0}); // incorrect
```

src/stores/Frame/Region/RegionStore.ts
```diff
@@ -198,6 +198,10 @@ export class RegionStore {
         }
     }
 
+    /**
+     * Getter for the sizes of regions and annotations
```
YuHsuan-Hwang: I suggest we revise it to "Returns the sizes of regions and annotations. For line regions and annotations, vector annotations, and ruler annotations, returns x and y absolute displacements."

src/stores/Frame/Region/RegionStore.ts
```diff
@@ -198,6 +198,10 @@ export class RegionStore {
...
+     * @returns A Point2D object with x and y size components.
```
YuHsuan-Hwang: I suggest we revise it to "The x and y size components." (We don't need to include types.)

src/stores/Frame/Region/RegionStore.ts
```diff
@@ -481,11 +484,23 @@ export class RegionStore {
         }
     };
 
+    /**
+     * Sets the size for regions and annotations
+     *
+     * @param p - Specifies the x and y size components.
+     *            For line region and annotation, vector annotation, and ruler annotation, the function sets the individual new x and y components of the linear segment.
```
YuHsuan-Hwang: Perhaps revise it to "For line regions and annotations, vector annotations, and ruler annotations, the function sets the new start and end positions while keeping the rotation within the same quadrant."

YuHsuan-Hwang (changes requested):
The new changes looks good to me. I have some comments for the documentation.

Currently, `setSize` updates start/end while keeping the rotation within the same quadrant (with a rotation change <= 90 degrees). @confluence Could you check if this works for scripting?

confluence:
> @confluence Could you check if this works for scripting?

I will check it now, and make the current region PR behave consistently (and adjust the scripting code in dev after the merge / branch).

confluence (approved):
I'm happy with the current behaviour -- I tested with 4 x 4 regions (one per line-like type x one per quadrant), and confirmed that all the types behave consistently and that orientation is preserved (unless one of the dimensions is set to zero, in which case the orientation is reset to consistent defaults).

I will update the patched functionality in the region PR to match this.

YuHsuan-Hwang (approved):

YuHsuan-Hwang:
Requires final checks for e2e tests.

kswang1029 (approved):
tested. a few ruler-related e2e tests have been fixed in a branch.

PR #2306, opened 2023-12-01
Yuhsuan/unit test ci

YuHsuan-Hwang:
**Description**

Changes:

1. Adjusted the PR template check list

    - Added check boxes for unit tests, API documentation, and the user manual.
    - Rearranged and added a section for dependencies.
    - Please check the updated documentation for the details.

2. Updated documentation

    - Updated the description for the PR template check list.
    - Added the instruction of applying custom PR template.

3. Adjusted CI

    - Added unit test CI for node 16 and 18. 
    - Removed Github Action for `pull_request` to remove duplicated CI.

4. Adjusted unit tests

    - Moved the test from `SpectralProfileWidgetStore.test.ts` to `FrameStore.test.ts` because the logic was moved to `FrameStore`.
    - Adjusted class field initialization in `AppStore` to avoid Mobx warnings in tests.
    - Applied the same TS module.

**Checklist**

- [x] ~ZenHub issue connection, board status, and estimate updated~

For the pull request:
- [x] ~protobuf version bumped~ / protobuf version not bumped


ajm-ska (approved):
Looks good!

At first I was wondering if it would be better to have the Unit Tests as a separate job so that they would appear as a separate entry on the jobs list and as a separate box on the flow diagram outside of the "node-v16" and "node-v18" boxes. But I see that is not practical as the Unit Tests appear to require a built frontend in order to run. Although it could be possible to move them to a separate job and use "upload-artifact/download-artifact" in order to transfer the built frontend files from one job to another, it would be very slow, particularly as we use self-hosted Github runners with Docker. I think it would upload the files to Github servers, and then download them again. So now I agree with the way you have it with "Run unit tests" as a stage of the "node-v16" and "node-v18" jobs.

YuHsuan-Hwang:
@ajm-asiaa Yes, the unit tests require building protobuf and wasm wrappers, so it's more efficient to run them in the same job. We could try to create a unit test report, but maybe not worth doing since currently there are only a few tests.

loveluthien (approved):
Good.

kswang1029 (changes requested):
@YuHsuan-Hwang shall we now switch to GitHub project and de-emphasize zenhub now? Documentation needs additional revision about this. Now each "issue" will have the following status which can be configured directly on GitHub (following as an example).

Screenshot 2023-12-11 at 11 54 33

YuHsuan-Hwang:
> @YuHsuan-Hwang shall we now switch to GitHub project and de-emphasize zenhub now? Documentation needs additional revision about this. Now each "issue" will have the following status which can be configured directly on GitHub (following as an example).

Ok. I will change it.

YuHsuan-Hwang:
Looks like we need more investigation for GitHub Projects setups, so we will keep the Zenhub checkbox for now.

kswang1029 (approved):

PR #2190, opened 2023-06-14
Yuhsuan/1734 submilliarcsecond scale match

YuHsuan-Hwang:
**Description**

Closes #1734 spatial matching in sub-milliarcsecond scale. Changed the approach of spatial matching by following AST library programmer's guide section 6.5 "Example—Transforming Between Two Calibrated Images."

Instead of converting between two image frame sets, a compound mapping is created from the frame sets. I'm not sure why the former approach does not work in sub-milliarcsecond scale. It's also suggested in the programmer's guide (section 14.3).

**Checklist**

For linked issues (if there are):
- [x] assignee and label added
- [x] ZenHub issue connection, board status, and estimate updated

For the pull request:
- [x] reviewers and assignee added
- [x] ZenHub estimate, milestone, and release (if needed) added
- [x] e2e test passing / ~corresponding fix added~
- [x] changelog updated / ~no changelog update needed~
- [x] ~protobuf updated to the latest dev commit~ / no protobuf update needed
- [x] `BackendService` unchanged / ~`BackendService` changed and corresponding ICD test fix added~

Issue #1619, opened 2021-08-10
Overlay's coordinate system should be independent between images


veggiesaurus:
At the moment, loading two images side-by-side forces their overlay to use the same coordinate system. There's no way to load two images and show different coordinate systems in each panel.

![image](https://user-images.githubusercontent.com/592504/128841441-edfe7e0d-457a-4f16-b55c-ca9bf1cac33f.png)

kswang1029:
As a note here: we should be careful about making the changes here as region wcs is sync’d with the viewer too. 

In my point view having all labels showing same wcs reference system is not a bad idea. But I am open to hear the feedback from the science team.

veggiesaurus:
@kswang1029 are there situations where this will break things? is it always possible to translate between systems?

PR #2322, opened 2023-12-21
Yuhsuan/2293 vector on matched frame

YuHsuan-Hwang:
**Description**

Fixed #2293. The issue is that the image does not show any vectors if the user did not apply vector overlay to the base frame (the raster data frame).

Fixed by updating the vector overlay when all the required vectors are ready instead of when the base frame vectors are ready.

Tested that:
* When applying vectors on the first image and then matching a second image, vectors from the first image is shown on the second image.
* When matching two images and applying vectors on the first image, vectors from the first image is shown on the second image.

Please note that this changes the timing of updating the vector overlay. The vector overlay will be updated only when all the required vectors from the frames are ready.

**Checklist**

For linked issues (if there are):
- [x] assignee and label added
- [x] ZenHub issue connection, board status, and estimate updated

For the pull request:
- [x] reviewers and assignee added
- [x] ZenHub estimate, milestone, and release (if needed) added
- [x] changelog updated / ~no changelog update needed~
- [x] ~unit test added (for functions with no dependenies)~
- [x] ~API documentation added (for public variables and methods in stores)~

For dependencies:
- [x] e2e test passing / ~corresponding fix added~ / new e2e test created
- [x] ~protobuf version bumped~ / no protobuf version bumped needed
- [x] ~protobuf updated to the latest dev commit~ / no protobuf update needed
- [x] ~corresponding ICD test fix added (`BackendService` changed)~ / no ICD test fix needed (`BackendService` unchanged)
- [x] ~user manual prepared (for large new features)~

Issue #2293, opened 2023-11-22
vector overlay generated from secondary image is not rendered on the reference image


kswang1029:
**Describe the bug**
When there are two Stokes cubes matched spatially and spectrally, if we try to generate a vector overlay from the 2nd image cube, we should see the vector overlay shows up in both images but we only see one on the secondary image, not the reference image. However, if we try to use the reference image to generate a vector overlay, we do see both images have the vector overlay rendered. If now we try again using the secondary image, we now see both images have the vector overlay rendered. 

**To Reproduce**
1. load two stokes cubes and match them spatially and spectrally
2. set the second image as active
3. enable vector overlay dialog to plot vector overlay
4. we will only see the vector overlay on the 2nd image, not the 1st image
5. try to do over again but with the first image as active. now we see what it should behave.

**Expected behavior**
In the step4 above, we should see the vector overlay got rendered on both images as they are spatially and spectrally matched.

**Platform info (please complete the following information):**
 - OS [e.g. macOS Monterey]: macOS Ventura
 - Browser [e.g. chrome, safari, electron app]: chrome, electron
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: dev, v4-stable
 - Frontend branch [e.g. dev, v3b2 release]: dev, v4-stable

PR #2324, opened 2023-12-28
Yuhsuan/2252 fix spatial match

### LINKED ISSUE #2252: failed to match images spatially
**Describe the bug**

In some special cases (probably trying to match images with different coordinate systems, such as FK5 <-> GALACTIC), CARTA refuses to match images. The console log shows

Screenshot 2023-09-11 at 21 00 57

Once we start seeing the error, we will not be able to match any images. Sometimes, if we change the active image and redo matching, it may work or may not.

See the steps below to reproduce the issue.

This might be a regression between the v3 and v4 releases. With v3, the issue cannot be reproduced.

**To Reproduce**
1. download the test images at https://drive.google.com/file/d/1ZVJxTFF_K56cTlnB4vXYuAIen-xKr5dy/view?usp=share_link
2. load the folder using filename to sort
3. load all the images _at once_ in the order like below
Screenshot 2023-09-11 at 20 50 54

4. the image viewer should stay as 2x2 grid
5. without touching anything else, try to match images in order (or try to match the 8th image directly)
6. until the 8th image we will see matching failure and cannot match any other images anymore
7. try to change the active image and repeat the test, sometimes we can match images but it may fail again

If we preconfigure the grid layout to 4x4, for example, all images can be matched successfully. Maybe this is something to do with AST memory management/leakage or racing condition?

**Expected behavior**
images should match

 - OS [e.g. macOS Monterey]: macOS Ventura
 - Browser [e.g. chrome, safari, electron app]: electron
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: dev, v4-stable
 - Frontend branch [e.g. dev, v3b2 release]: dev, v4-stable

YuHsuan-Hwang:
**Description**

Fixed #2252. The root cause is that the coordinate system is updated only for rendered images. Fixed by updating all the images instead of only rendered images when the coordinate system is changed.

This is a regression from #2190. Before the PR, we create the transformation by combining two AST FrameSet. The images can still be matched when their coordinate systems are different. After the PR, we create it simply by combining the conversion of image coordinate <-> WCS coordinate of the two images. Therefore, it fails when their WCS values are different.

Please note that if we implement independent coordinate systems (#1619) in the future, for spatial matching we might need to create a temporary copy of the matched frame, set it to the same coordinate system, and then create the transformation.

Changes:
* Added coordinate system update to overlay store `autorun` function.
* Refactoring
    * Removed unused variables in overlay store `autorun` functions.
    * Removed unnecessary update in `setSystem`.

Tested that:
* All the image overlays update with the global coordinate system and the number custom XY formats.
* Pan and zoom and region inputs update with the global coordinate system and the number custom XY formats as usual.
* Distance measuring on all the images updates with the global coordinate system as usual.

**Checklist**

For the pull request:

For dependencies:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed the performance issue when panning images ([#2291](https://github.com/CARTAvis/carta-frontend/issues/2291)).
 * Improved the dragging performance of 'Export Region' widget when the list of region/annotation to be exported is large. ([#1867](https://github.com/CARTAvis/carta-frontend/issues/1867)).
 * Fixed the flashing contour rendering during animation ([#579](https://github.com/CARTAvis/carta-frontend/issues/579)).
+* Fixed failing to match images spatially ([2252](https://github.com/CARTAvis/carta-frontend/issues/2252)).
 
 ## [4.0.0]
 
--- a/src/components/ImageView/Overlay/OverlayComponent.tsx
+++ b/src/components/ImageView/Overlay/OverlayComponent.tsx
@@ -178,8 +178,6 @@ export class OverlayComponent extends React.Component<OverlayComponentProps> {
         const axesStyleString = this.props.overlaySettings.axes.styleString;
         const numbersStyleString = this.props.overlaySettings.numbers.styleString;
         const labelsStyleString = this.props.overlaySettings.labels.styleString;
-        const formatStringX = this.props.overlaySettings.numbers.formatStringX;
-        const formatStyingY = this.props.overlaySettings.numbers.formatStringY;
 
         if (frame.isSwappedZ) {
             const requiredChannel = frame.requiredChannel;
--- a/src/components/ImageView/Overlay/OverlayComponent.tsx
+++ b/src/components/ImageView/Overlay/OverlayComponent.tsx
@@ -201,7 +201,12 @@ export class OverlayComponent extends React.Component<OverlayComponentProps> {
         } else if (frame.isSwappedZ && frame.spectralAxis?.valid) {
             AST.set(frame.wcsInfo, spectralAxisSetting + dirAxesSetting);
         } else {
-            AST.set(frame.wcsInfo, dirAxesSetting);
+            const formatStringX = this.props.overlaySettings.numbers.formatStringX;
+            const formatStyingY = this.props.overlaySettings.numbers.formatStringY;
+            const explicitSystem = this.props.overlaySettings.global.explicitSystem;
+            if (formatStringX !== undefined && formatStyingY !== undefined && explicitSystem !== undefined) {
+                AST.set(frame.wcsInfo, `Format(${frame.dirX})=${formatStringX}, Format(${frame.dirY})=${formatStyingY}, System=${explicitSystem},` + dirAxesSetting);
+            }
         }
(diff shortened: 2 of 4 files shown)


kswang1029 (changes requested):
@YuHsuan-Hwang The matching part appear fine. However, if we use one of the secondary image as the data source for contours, it seems that other images may not have the contours rendered _initially_ (in my test case it is with a 2x2 layout). We may need to either set to next image as active or pan/zoom a bit to trigger re-rendering. 

https://github.com/CARTAvis/carta-frontend/assets/20819712/bc0220b1-f3d9-437a-add4-0a919bd8579a

UPDATE: would it be fixed with https://github.com/CARTAvis/carta-frontend/pull/2322 ? Guess not?

YuHsuan-Hwang:
> However, if we use one of the secondary image as the data source for contours, it seems that other images may not have the contours rendered _initially_ (in my test case it is with a 2x2 layout). We may need to either set to next image as active or pan/zoom a bit to trigger re-rendering.

Looks like this also exists in dev?

kswang1029:
> > However, if we use one of the secondary image as the data source for contours, it seems that other images may not have the contours rendered _initially_ (in my test case it is with a 2x2 layout). We may need to either set to next image as active or pan/zoom a bit to trigger re-rendering.
> 
> Looks like this also exists in dev?

Right, it is in dev (but not v4-stable). So it is another regression. Would you prefer to fix it in https://github.com/CARTAvis/carta-frontend/pull/2322 as well? Or as a separate PR (plus a new issue)?

YuHsuan-Hwang:
Separate PR would be better for preparing v4.1.

kswang1029:
> Separate PR would be better for preparing v4.1.

ok. issue filed as https://github.com/CARTAvis/carta-frontend/issues/2330

I would need to re-test this once #2293 and #2230 are fixed.

loveluthien:
Is this under expectation? Using test data provided in #2252, there are four images that can not be matched. However, if I manually click the toggle, they still can do the spatial match.
截圖 2024-01-09 下午2 25 37
截圖 2024-01-09 下午2 37 08
截圖 2024-01-09 下午2 37 37

YuHsuan-Hwang:
> Is this under expectation? Using test data provided in #2252, there are four images that can not be matched. However, if I manually click the toggle, they still can do the spatial match.

I see. I will fix this.

YuHsuan-Hwang:
Updates:
* Fixed incorrect precision of the overlay numbers (regression in this branch) by adding back the AST updates in the overlay component.
* Fixed failing to match images on append by updating images when setting AST defaults of a new image.

loveluthien (approved):
The fix is good.

YuHsuan-Hwang:
Updates:
* Update the frames in FrameStore instead of OverlayStore.
* When matching images on append, update images only when the system is set as auto.

kswang1029 (approved):
tested manually with the test images mentioned in the issue in some other orders, with some other test images (different target sources), and with images that cannot be matched. all looking fine so far. No other regression from e2e tests. 

Note:
1. contour rendering issue https://github.com/CARTAvis/carta-frontend/issues/2330 exists in this PR. 
2. vector rendering issue https://github.com/CARTAvis/carta-frontend/issues/2293 has been fixed in this PR. 👍

crocka (approved):

Issue #2343, opened 2024-02-01
Enable strictNullChecks compiler option


YuHsuan-Hwang:
Enable the TypeScript strictNullChecks compiler option to prevent JS runtime errors such as “Uncaught TypeError: Cannot read property ‘...’ of undefined.”

[details of the plan](https://docs.google.com/document/d/17uP7VpT9MTl0xtwfBIu0Dvv4V-sy6auIXPCNPZC_6gA/edit?usp=sharing)

The migration could be separated into several parts:

- [x] utilities, models, services
- [x] stores
- [x] components and icons (require blueprint upgrade)

PR #2344, opened 2024-02-01
Yuhsuan/null check tsc silent

YuHsuan-Hwang:
**Description**

This PR is for the first part of #2343. Files in `/utilities`, `/models`, and `/services` are migrated with the strictNullChecks option enabled.

A new script is added to trigger a separate ts compiler with the strcitNullChecks option enabled. `tsc-silent` is applied to suppress errors from files that are not migrated yet.

* To run the null checks, run `npm run strict-null-checks`. This is also triggered before development and production builds so that developers and the CI are aware of the checks.
* To include files for the null checks, add the files in the `migratedFiles` list in `tsc-silent.config.js`.

**Checklist**

- [ ] ZenHub issue connection, board status, and estimate updated

For the pull request:
- [ ] ZenHub estimate, milestone, and release (if needed) added

For dependencies:
- [x] corresponding ICD test fix added (`BackendService` changed) / ~no ICD test fix needed (`BackendService` unchanged)~


YuHsuan-Hwang:
@kswang1029 Currently the branch could cause regression for all features that is connected to the changed files, but I don't have a good way to test them all. Perhaps the most efficient way is to increase our e2e tests in the future.

YuHsuan-Hwang:
Revised some fixes by changing the variable types instead of the values. Hope this reduces the possibility of regressions.

src/models/AbstractCatalogProfile/AbstractCatalogProfile.ts
```diff
@@ -295,22 +295,28 @@ export abstract class AbstractCatalogProfileStore {
...
+            columnWidth: current?.columnWidth ?? null
         };
         this.catalogControlHeader.set(columnName, newHeader);
         this.updateTableStatus(true);
     };
 
     @action.bound setTableColumnWidth(width: number, columnName: string) {
-        this.catalogControlHeader.get(columnName).columnWidth = width;
+        const header = this.catalogControlHeader.get(columnName);
+        if (header) {
```
crocka, line 310: Can we use has() function? If(Map.has(key)) then Map.get(key) = width

src/models/AbstractCatalogProfile/AbstractCatalogProfile.ts
```diff
@@ -295,22 +295,28 @@ export abstract class AbstractCatalogProfileStore {
...
+            header.columnWidth = width;
+        }
     }
 
     @action setHeaderDisplay(val: boolean, columnName: string) {
-        this.catalogControlHeader.get(columnName).display = val;
+        const header = this.catalogControlHeader.get(columnName);
```
crocka, line 316: Same comment as above. If the key is present by the value is undefined, then the has() function should not be used.

src/services/SplatalogueService.ts
```diff
@@ -125,7 +125,9 @@ export class SplatalogueService {
                     entry = entry?.match(/^(^[\d.]+)/)?.[1] ?? ""; // match the string before the first char that is not a number or "."
                 }
 
-                column.stringData[i] = entry;
+                if (column.stringData !== null && column.stringData !== undefined) {
```
crocka, line 128: We can also use isFinite?

src/utilities/CatalogApiProcessed/CatalogApiProcessed.ts
```diff
@@ -158,14 +160,16 @@ export class CatalogApiProcessing {
                             tableElement: tableElement
                         }
                     };
-                    resources.set(name, res);
+                    if (name !== null) {
```
crocka, line 163: Maybe we can use isFinite?

src/utilities/parsing/parsing.ts
```diff
@@ -43,7 +43,9 @@ export function trimFitsComment(val: string): string {
 export function mapToObject<K, T>(map: Map<K, T>) {
     const obj: {[k: string]: T} = {};
     map.forEach((value, key) => {
-        obj[key.toString()] = value;
+        if (key !== null && key !== undefined) {
```
crocka, line 46: Maybe we can use isFinite?

src/utilities/table/table.ts
```diff
@@ -74,7 +74,7 @@ export function numericFiltering(columnData: Array<number>, dataIndexes: number[
         return [];
     }
 
-    let compareFunction = undefined;
+    let compareFunction: (data: number) => boolean;
```
crocka: If the function is undefined originally, does it mean we should add | undefined to prevent regression?

src/services/TileService.ts
```diff
@@ -389,7 +391,7 @@ export class TileService {
 
     private clearTile = (tile: RasterTile, _key: any) => {
         if (tile.data) {
-            delete tile.data;
+            tile.data = new Float32Array();
```
crocka: If we assign the tile data to empty array, will it allocate memory space to it?

src/models/AbstractCatalogProfile/AbstractCatalogProfile.ts
YuHsuan-Hwang: I agree it's more natural to use `Map.has` here, but unfortunately TS doesn't support type narrowing in this way. Looks like it's a known [issue](https://github.com/microsoft/TypeScript/issues/13086).
Error message:
image

src/models/AbstractCatalogProfile/AbstractCatalogProfile.ts
YuHsuan-Hwang: Same comment.

src/services/SplatalogueService.ts
YuHsuan-Hwang: It's `string[]`, so we can't use isFinite.

src/utilities/CatalogApiProcessed/CatalogApiProcessed.ts
YuHsuan-Hwang: It's a string, so we can't use isFinite.

src/utilities/parsing/parsing.ts
YuHsuan-Hwang: isFinite also doesn't accept type `K`.

src/utilities/table/table.ts
YuHsuan-Hwang: Yes, this looks safer. 👍

src/services/TileService.ts
YuHsuan-Hwang: Good point... I revised to fix it in a safer way.

YuHsuan-Hwang:
@crocka Thanks for reviewing the changes! Please check my replies and revision.

kswang1029 (approved):
No regression from e2e tests. With some other manual tests based on the changed files in the codebase, so far I do not see odd behaviors. Certainly my manual test coverage is not complete, but let's just proceed and see if there are issues from the tests in the future.

acdo2002 (approved):
The BackendService modification has been applied to the MessangeController.ts in the ICD-RxJS repository. (ICD-RxJS branch mylin/modify_MessageController_IBackendResponse)
The ICD-RxJS normal tests and performances tests are all passed without any problem.

crocka (approved):

PR #2352, opened 2024-02-23
Duplicate CI steps

YuHsuan-Hwang:
**Description**

Looks like we have duplicate steps in node-v16 and node-v18 runs. The steps take about 1min, which is smaller than the fluctuation of the total used time, so the total used time does not change much.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -75,18 +75,6 @@ jobs:
           n exec 16 node -v
           n exec 16 npm install --legacy-peer-deps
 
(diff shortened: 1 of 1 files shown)


kswang1029 (approved):
Right, as defined in https://github.com/CARTAvis/carta-frontend/blob/a9c54fd4bfb1eae6fc8a0dc7a19790a859cd4933/package.json#L94 build_wrappers.sh and build_proto.sh will be run already. Good catch. 👍 

As expected, e2e tests are passing.

acdo2002 (approved):

PR #2353, opened 2024-02-23
Normalize directory and filename in loadFile using a path library

### LINKED ISSUE #1357: Casa image paths saved incorrectly in workspaces if images added via snippets with a trailing slash

### LINKED ISSUE #1816: Opening a CASA image with a trailing / using the URL parameter doesn't work properly
**Describe the bug**
@Jordatious found this issue. If you open a CASA image using the `?file=<CASA image path>/` (note trailing slash), then odd things happen:

- The current folder gets set to the _inside_ of the CASA image, rather than the containing directory
- Resuming sessions seems to break

**To Reproduce**
1. Go to a CARTA browser instance and add `?file=<CASA image path>/`
2. Open the file browser. The starting folder will be incorrect.
3. Resume the session after disconnecting. This will fail.

**Expected behavior**
Resuming should work, and the file browser should start in the containing folder, rather than the image folder.

confluence:
**Description**

This fixes [backend bug #1357](https://github.com/CARTAvis/carta-backend/issues/1357) and #1816 by using (the browser port of) node's `path` module to normalize the directory and filename in the `loadFile` function.

The library correctly handles edge cases such as the one causing the bug. Since it's already available to us as a dependency, I would recommend making use of it instead of hand-coding path parsing (which is error-prone!). The browser version only implements posix paths, which I think is what we always need.

Step by step instructions for reproducing the backend bug can be found [in the bug's description](https://github.com/CARTAvis/carta-backend/issues/1357#issue-2107173165). ~I have not yet tested the issue in #1816, but I believe that it has the same underlying cause.~ I have confirmed that this also fixes #1816 (or at least I can see that it prevents the initial incorrect image request which later causes errors).

Outstanding questions:

1. I'm not sure if I'm importing `path-browserify` in the optimal way. It's [intended to be imported as `path`](https://github.com/browserify/path-browserify), because it's the browser port of node's `path` module, and webpack is supposed to provide it as a fallback transparently. I know that webpack 5 no longer sets this up automatically, and it looks like we have a fix for this in a patch -- but when I attempt to import `path`, I get an error about `path-browserify` being outside the `src` tree. This doesn't happen when I import `path-browserify` directly, and doesn't happen with other dependencies in `node_modules`. I'm not sure if this is some weird interaction with the alias, and I haven't been able to find a solution. @veggiesaurus, is there a piece missing somewhere?
2. This code always performs the same normalization step, regardless of what the original directory and filename parameters are. This includes cases where the directory is empty and the filename does not contain a slash (which are skipped in the previous code). The effect of this is to change the directory to `"."`. I did some tests in the previous code, and from what I understand, a directory which is `""` or `"."` is evaluated as the top-level directory. Is that correct? If that's the case, then changing `""` to `"."` shouldn't make a difference.

**Checklist**

For the pull request:

For dependencies:
- [X] ~protobuf version bumped~ / no protobuf version bumped needed
- [x] corresponding ICD test fix added (`BackendService` changed) / no ICD test fix needed (`BackendService` unchanged)
- [x] user manual prepared (for large new features)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fixed incorrect range values in the save image dialog with non-convertible spectral axis images by supporting native/channel options in the spectral coordinate selector for non-convertible spectral axis images ([#2225](https://github.com/CARTAvis/carta-frontend/issues/2225)).
 * Fixed a bug when applying a filter to the shifted frequency column in the spectral line query widget ([#2326](https://github.com/CARTAvis/carta-frontend/issues/2326)).
 * Improved the cursor interaction area in the image view widget ([#1794](https://github.com/CARTAvis/carta-frontend/issues/1794)).
+* Fixed saving or opening CASA image with a trailing slash using the URL parameter or the snippets ([#1816](https://github.com/CARTAvis/carta-frontend/issues/1816) and [#1357](https://github.com/CARTAvis/carta-backend/issues/1357)).
 
 ## [4.1.0]
 
--- a/src/stores/AppStore/AppStore.ts
+++ b/src/stores/AppStore/AppStore.ts
@@ -7,6 +7,7 @@ import {CARTA} from "carta-protobuf";
 import * as _ from "lodash";
 import * as Long from "long";
 import {action, autorun, computed, flow, makeObservable, observable, ObservableMap, when} from "mobx";
+import * as Path from "path-browserify";
 import * as Semver from "semver";
 
 import {getImageViewCanvas, ImageViewLayer, PvGeneratorComponent} from "components";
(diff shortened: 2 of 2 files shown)


YuHsuan-Hwang (approved):
Looks good to me.

confluence:
I'm moving this out of draft -- I think it's up to frontend devs whether the import needs to be changed or is fine as-is. Either way this should probably be tested.

kswang1029 (approved):
Tested the URL cases
http://localhost:3002/?file=/Users/kswang/set_QA_e2e_v2/M17_SWex.image
http://localhost:3002/?file=/Users/kswang/set_QA_e2e_v2/M17_SWex.image/
and both work fine.

Tested the snippet case and workspace and it works fine with the fix. 

No regression from e2e tests.

YuHsuan-Hwang:
We install `path-browserify` directly without adding alias after upgrading to react-script v5, so the import looks fine to me.

PR #2356, opened 2024-03-06
Yuhsuan/2064 e2e testid

### LINKED ISSUE #2064: Initial test of using unique IDs to simplify e2e tests
@kswang1029 currently has to update e2e tests every time the UI changes. If we add unique identifiers to the elements, we should be able to simplify the e2e tests considerably. 

We should try this with a few simple input elements (e.g. filename input or filter string input)

YuHsuan-Hwang:
**Description**

Closes #2064. All the changes are basically adding `data-testid` attributes.

Advantages of adding data-* attributes for tests:
* We create a system separated from classes and ids, which could prevent changing current behaviors, e.g. css styling.
* React testing library has an API `getByTestId` as a shortcut for querying elements like `<div data-testid="custom-element" />`, so we can reuse the same system if we want to add unit tests for components in the future.

Disadvantages:
* We might need frontend companion PRs when adding e2e tests.
* Some components do not support adding data-* attributes:
    * Blueprintjs v3 `<Dialog/>`: in `DraggableDialog`, used the ref of the parent div to get inner components (by bp class) and added testids in `componentDidUpdate`.
    * Blueprintjs v3 `<Slider/>`, `<Text/>`, and `<TagInput/>`: (work-around in e2e) query from the parent div in e2e tests (and use bp class for getting inner components in `<Slider/>`).
    * Blueprintjs v3 `<Table/>`: (work-around in e2e) query from class.
    * react-window, e.g. `<List/>`, `<FixedSizeList/>`: (work-around in e2e) query from the parent div in e2e tests.
* Some components do not add the data-* attributes to the elements we need:
    * Blueprintjs v3 spin box buttons in `<NumericInput/>` (only `<input>` is added): in `SafeNumericInput`, used the ref of the html input to get the parent div. Used the parent div to get inner components (by bp class) and added testids in `componentDidMount`.
    * Blueprintjs v3 `<Table/> <Column/>` headers (creates two duplicated tables are created, and only the second one is interactable): (work-around in e2e) query the second element.
    * Blueprintjs v3 `<Radio/>`, `<Switch/>`, and `<Checkbox/>` (need to click `<span>` or `<label>` instead of `<input>`): (work-around in e2e) query the span from the input element in e2e tests.

Notes:
* Frontend work-arounds are only applied to `<Dialog/>` and `<NumericInput/>` because they are frequently used in e2e tests and there already exists wrapper components. Applying similar workarounds for other components might be possible, but I would prefer we set it as future work if it's necessary.
* Tested that Blueprintjs v5 `<Text/>` supports adding data-* attributes. We can remove the workaround in another PR after upgrading to v5. Issues of other components remain.

Companion e2e PR: https://github.com/CARTAvis/carta_robot_test/pull/63

**Checklist**

For the pull request:

For dependencies:
- [x] corresponding ICD test fix added (`BackendService` changed) / ~no ICD test fix needed (`BackendService` unchanged)~
- [x] user manual prepared (for large new features)


kswang1029 (approved):
thank you for sorting this tedious work out!

veggiesaurus (approved):

PR #2370, opened 2024-04-21
Kuanchou/681 relative coordinates

### LINKED ISSUE #681: relative coordinates to a given point
A user request: add an option for a drop pin to which all coordinates are referenced to, so the viewer panel show relative and absolute coordinates.

loveluthien:
**Description**

This PR closes #681 adding an offset coordinates. The offset coordinates consider the projection. We can turn on the offset coordinates using overlay coordinates in the toolbar or the image view setting. The default offset center is the image center and can be customized by setting it to the view center or a manual input. The unit of delta angles changes with the field of view size. 

Due to the offset mapping using AST which changes the projection center, the distance measurement went wrong when applying the offset coordinates. Therefore, we replace the distance measurement with the ruler annotation which is not affected by the overlay coordinates.

**Checklist**
- coordinates show delta angles. 
- UI: toolbar and image view setting panel.
- distance measurement is replaced with ruler annotation.

For the pull request:

For dependencies:
- [x] e2e test passing / corresponding fix added / ~new e2e test created~
- [ ] ~protobuf version bumped~ / no protobuf version bumped needed
- [ ] ~protobuf updated to the latest dev commit~ / no protobuf update needed


loveluthien:
@kswang1029 @crocka There is a discussion item. The distance measurement fails when using a relative coordinate, more precisely after applying an AST shifted map. The easiest way is to replace the distance measurement with the ruler annotation. @YuHsuan-Hwang has some concerns about doing that. What's your opinion?

kswang1029:
Do you mean we deprecate the distance measurement tool and ask users to use the rule annotation instead? If so, I am fine with the idea which actually enables multiple measurements and users can see all the results at the same time.

loveluthien:
Yes. I feel the ruler is more useful than the distance measurement.

YuHsuan-Hwang:
I'm ok with deprecating the distance measuring tool.

kswang1029 (approved):
looks very nice! A corresponding fix is applied to an e2e test branch. New tests will be added asap.

src/components/ImageView/ImageViewSettingsPanel/ImageViewSettingsPanelComponent.tsx
```diff
@@ -203,8 +203,42 @@ export class ImageViewSettingsPanelComponent extends React.Component<WidgetProps
                         wcsDisabled={isPVImage}
                         customPlaceholder="Height"
                     />
-                    <span className="info-string">{getFovInfoString(frame?.fovSize?.y, frame?.fovSizeWCS?.y)}</span>
-                </FormGroup>
+                    <span className="info-string">{getInfoString(frame?.fovSize?.y, frame?.fovSizeWCS?.y)}</span>
+                </FormGroup>
+                <FormGroup inline={true} label="Offset Coordinates">
```
YuHsuan-Hwang: Set the second word to lowercase for consistency. Same for "Offset Center (X/Y)".

src/components/ImageView/Toolbar/ToolbarComponent.tsx
```diff
@@ -244,7 +251,7 @@ export class ToolbarComponent extends React.Component<ToolbarComponentProps> {
                                     position={tooltipPosition}
                                     content={
                                         <span>
-                                            Distance measurement
+                                            Ruler annotation
                                             <br />
                                             <i>
                                                 <small>Click to create geodesic curves.</small>
```
YuHsuan-Hwang: Perhaps update it to "Click-and-drag to create geodesic curves." Maybe we should remove the double-click shortcut because the active region might not be a ruler annotation.

src/stores/Frame/FrameStore.ts
```diff
@@ -2139,6 +2145,109 @@ export class FrameStore {
...
+                    for (const frame of this.secondarySpatialImages) {
+                        const frameCenterInRad = getUnformattedWCSPoint(frame.wcsInfo, frame.offsetCenter);
+                        frame.wcsInfoShifted = AST.createShiftmapFrameset(frame.wcsInfo, frameCenterInRad.x, frameCenterInRad.y);
+                    }
+                }
+            }
+        }
+    };
+
+    @action updateOffseCenter = () => {
```
YuHsuan-Hwang: Typo: "offset".

wasm_src/ast_wrapper/ast_wrapper.cc
```diff
@@ -278,32 +294,6 @@ EMSCRIPTEN_KEEPALIVE int plotGrid(AstFrameSet* wcsinfo, double imageX1, double i
...
-        double in[2][4] = {{xtran[0], xtran[1], xtran[1], xtran[0]}, {ytran[0], ytran[1], ytran[0], ytran[0]}};
-        const double* inPtr = in[0];
-        astPolyCurve(plot, 4, 2, 4, inPtr);
-
-        double start[] = {xtran[0], ytran[0]};
-        double finish[] = {xtran[1], ytran[1]};
-        if (isPVImage)
-        {
-            double corner[] = {xtran[1], ytran[0]};
-            plotDistText(wcsinfo, plot, start, corner);
```
YuHsuan-Hwang, line 298: Looks like function `plotDistText` could be removed.

src/stores/Frame/FrameStore.ts
```diff
@@ -2139,6 +2145,109 @@ export class FrameStore {
...
+        } else {
+            this.isOffsetCoord = isoffset;
+            for (const frame of this.secondarySpatialImages) {
+                frame.isOffsetCoord = isoffset;
+            }
+        }
+    }
+
+    /**
+     * Toggle of the offset coordinates.
```
YuHsuan-Hwang: Would be good to mention that the function sets the current center (instead of the image center) as the offset center.

src/stores/Frame/FrameStore.ts
```diff
@@ -2139,6 +2145,109 @@ export class FrameStore {
...
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const system = AppStore.Instance.overlayStore.global.explicitSystem;
+        if (!this.wcsInfoShifted) {
+            return null;
+        }
+        return getFormattedWCSPoint(this.wcsInfoForTransformation, this.offsetCenter);
+    }
+
+    /**
+     * Set the offset center in the pixel coordinates.
```
YuHsuan-Hwang: Would be good to mention that the function not only set the new offset center but also updates the displayed coordinates.

YuHsuan-Hwang (changes requested):
The implementation looks nice. The removal of distance measuring is also good.
Minor comments only:

kswang1029:
UPDATE: Added a corresponding e2e test which checks AST grid rendering including the offset cases.

wasm_src/ast_wrapper/ast_wrapper.cc
loveluthien: This might old version. I had removed entire if (showCurve) section.

YuHsuan-Hwang: Sorry I meant the function `plotDistText` itself in this file could be removed.

loveluthien: okay, I got it.

YuHsuan-Hwang (approved):

kswang1029 (approved):
looks good 👍

src/stores/Frame/FrameStore.ts
```diff
@@ -2704,6 +2832,14 @@ export class FrameStore {
         this.spatialReference = frame;
         console.log(`Setting spatial reference for file ${this.frameInfo.fileId} to ${frame.frameInfo.fileId}`);
 
+        this.isOffsetCoord = frame.isOffsetCoord;
+
+        // initialize wcsInfoShifted if it is not existed
+        if (this.isOffsetCoord && !this.wcsInfoShifted) {
+            const centerInRad = getUnformattedWCSPoint(this.wcsInfo, this.center);
+            this.wcsInfoShifted = AST.createShiftmapFrameset(this.wcsInfo, centerInRad.x, centerInRad.y);
```
YuHsuan-Hwang: Please check if `centerInRad` is null.

src/stores/Frame/FrameStore.ts
```diff
@@ -2149,6 +2158,110 @@ export class FrameStore {
...
+            this.spatialReference.createWcsInfoShifted();
+        } else {
+            if (this.wcsInfo && this.offsetCenter) {
+                const centerInRad = getUnformattedWCSPoint(this.wcsInfo, this.offsetCenter);
+
+                if (centerInRad) {
+                    this.wcsInfoShifted = AST.createShiftmapFrameset(this.wcsInfo, centerInRad.x, centerInRad.y);
+                    for (const frame of this.secondarySpatialImages) {
+                        const frameCenterInRad = getUnformattedWCSPoint(frame.wcsInfo, frame.offsetCenter);
+                        frame.wcsInfoShifted = AST.createShiftmapFrameset(frame.wcsInfo, frameCenterInRad.x, frameCenterInRad.y);
```
YuHsuan-Hwang: Same comment.

src/stores/Frame/FrameStore.ts
```diff
@@ -2739,6 +2875,7 @@ export class FrameStore {
 
         // udpate center position for setting inputs
         this.center = this.spatialTransform.transformCoordinate(this.spatialReference.center, false);
+        this.setOffsetCenter(this.center.x, this.center.y);
```
YuHsuan-Hwang: Shall we update the offset center only when coord offset is enabled?

YuHsuan-Hwang (changes requested):

src/stores/Frame/FrameStore.ts
loveluthien: yes, it is better.

YuHsuan-Hwang (approved):

PR #2388, opened 2024-06-06
Thchang/2366 support image coordinates

### LINKED ISSUE #2366: [WCS grid] support image coordinate grid in the image viewer
With v4.1 / dev, the WCS list in the image viewer contains the regular world coordinate systems. When we load an image without valid WCS headers, the coordinate grid will be rendered in the image coordinate, which is nice. However, for images with valid WCS headers, there is no way to render the coordinate grid in the image coordinate. It would be helpful if we include this option in the list. Once we have this, handling exceptions in the multi-panel view would be more straightforward.

### LINKED ISSUE #2449: Wrong XY coordinate if no wcs
**Describe the bug**
The image panel shows wrong X and Y coordinates.
Snipaste_2024-12-12_15-51-32
It is a simple image with 8 pixels. If I show is using matplotlib:
image
It seems a `X-1, Y-1` somewhere.

**To Reproduce**
1. Mock a simple fits image using python:
```python
import numpy as np
from astropy.io import fits
data = np.array([[1,2,3,4],
                 [5,6,7,8]])
hdu = fits.PrimaryHDU(data)
hdu.writeto('test.fits', overwrite=True)
```
3. Open  `test.fits` using CARTA.
4. Check the numbers of the grid.

**Expected behavior**
It should be something like the matplotlib screenshot above.

 - macOS

TienHao:
**Description**

This PR closes #2366. The image(pixel) coordinates can be displayed for images with valid WCS headers.

The switching between image and WCS coordinates is implemented by changing the "Current" frame in the **wcsInfo**(AST.FrameSet), since all the WCS images has a base frame using pixel coordinates in the **wcsInfo**.

There is a similar approach for the relative coordinates. By adding a third AST.Frame for relative image coordinates in the **wcsInfoShifted**(AST.FrameSet) and updating both the relative image and WCS AST.Frame at the same time, the switching between a relative IMG/WCS can also be implemented by changing the "Current" frame in the **wcsInfoShifted**.

For the consistence of spatial matching, the spatial matching will be cleared once the user change the system between WCS/IMG. 

This PR also removed a no longer existing fix for the offsets of the dummy frame.

(Also closes #2449.)

**Checklist**

For the pull request:

For dependencies:


loveluthien:
@TienHao spatially matching will be wrong under the pixel coordinates.

YuHsuan-Hwang (changes requested):
Some suggestions:
1. Perhaps we disable wcs inputs when the system coordinates is IMG:
    
    
2. We can set the labels to "X/Y Coordinate" when the system coordinates is IMG (to be consistent with opening images without wcs info in dev branch).
3. (regression) Looks like the overlay disappears when opening images without wcs info.

src/stores/AppStore/AppStore.ts
```diff
@@ -1987,6 +1987,8 @@ export class AppStore {
                             this.contourDataSource = frame;
                         }
                     }
+
+                    this.overlayStore.setDefaultsFromFrame(this.activeFrame);
```
YuHsuan-Hwang: I think we need to remove this line to avoid changing the coord system when changing the active frame.

src/components/ImageView/Overlay/OverlayComponent.tsx
```diff
@@ -70,11 +70,11 @@ export class OverlayComponent extends React.Component<OverlayComponentProps> {
                 const scaleMapping = AST.scaleMap2D(1.0, 1.0 / frame.aspectRatio);
                 const newFrame = AST.frame(2, "Domain=PIXEL");
                 AST.addFrame(tempWcsInfo, 1, scaleMapping, newFrame);
-                AST.setI(tempWcsInfo, "Base", 3);
-                AST.setI(tempWcsInfo, "Current", 2);
+                AST.setI(tempWcsInfo, "Base", frame.validWcs ? 3 : 2);
+                AST.setI(tempWcsInfo, "Current", frame.validWcs ? 2 : 1);
```
YuHsuan-Hwang: Looks like this causes incorrect labels for special images like PV images.

src/stores/OverlayStore/OverlayStore.ts
```diff
@@ -32,7 +32,8 @@ export enum SystemType {
     FK4 = "FK4",
     FK5 = "FK5",
     Galactic = "GALACTIC",
-    ICRS = "ICRS"
+    ICRS = "ICRS",
+    Image = "Cartesian"
```
YuHsuan-Hwang: Change to upper case for consistency.

YuHsuan-Hwang (changes requested):
The approach of using AST base frame looks good. 👍

kswang1029 (changes requested):
When it is in the image coordinate mode, if we enable the offset coordinate mode, we will see the reference image changes correctly of the offset coordinate but the secondary image changes to RA-offset and Dec-offset.


https://github.com/user-attachments/assets/fdddec1e-f89f-428d-871d-2cb913b95b99

kswang1029:
@TienHao When we have two images matched and displayed in world coordinate grid, if we then switch to the image coordinate grid, both images will have image coordinate grids as expected. However, if we further enable the offset mode and click the recenter button, only the spatial reference image will have the offset image coordinate grid, not the secondary image. Any idea?
Screenshot 2024-11-18 at 14 01 56

src/components/Dialogs/OnlineDataQueryDialog/CatalogOnlineQueryComponent.tsx
```diff
@@ -236,7 +238,9 @@ export class CatalogQueryComponent extends React.Component {
                     <div className={Classes.DIALOG_FOOTER_ACTIONS}>
                         <AnchorButton intent={Intent.WARNING} disabled={!configStore.isQuerying} onClick={() => CatalogApiService.Instance.cancelQuery(configStore.catalogDB)} text={"Cancel"} />
                         {configStore.enableLoadVizier ? <AnchorButton intent={Intent.PRIMARY} disabled={disable} onClick={() => this.loadVizierCatalogs()} text={"Load selected"} /> : null}
-                        <AnchorButton intent={Intent.SUCCESS} disabled={disable} onClick={() => this.query()} text={"Query"} />
+                        <Tooltip content={"Pleace select WCS coordinates"} disabled={!appStore.overlayStore.isImgCoordinates} position={Position.BOTTOM} hoverOpenDelay={300}>
```
YuHsuan-Hwang: Typo: Pleace -> Please.

src/components/Dialogs/OnlineDataQueryDialog/CatalogOnlineQueryComponent.tsx
```diff
@@ -236,7 +238,9 @@ export class CatalogQueryComponent extends React.Component {
...
+                            <AnchorButton intent={Intent.SUCCESS} disabled={disable || !appStore.overlayStore.isImgCoordinates} onClick={() => this.query()} text={"Query"} />
```
YuHsuan-Hwang: Shall we disable when `appStore.overlayStore.isImgCoordinates`?

src/components/ImageView/Overlay/OverlayComponent.tsx
```diff
@@ -214,7 +214,7 @@ export class OverlayComponent extends React.Component<OverlayComponentProps> {
             const formatStringX = this.props.overlaySettings.numbers.formatStringX;
             const formatStyingY = this.props.overlaySettings.numbers.formatStringY;
             const explicitSystem = this.props.overlaySettings.global.explicitSystem;
-            if (formatStringX !== undefined && formatStyingY !== undefined && explicitSystem !== undefined) {
+            if (formatStringX !== undefined && formatStyingY !== undefined && explicitSystem !== undefined && explicitSystem !== SystemType.Image && frame.validWcs) {
```
YuHsuan-Hwang: Perhaps reuse `OverlayStore.Instance.isWcsCoordinates`.

src/stores/Frame/FrameStore.ts
```diff
@@ -1920,7 +1932,7 @@ export class FrameStore {
         let cursorPosWCS, cursorPosFormatted;
         let precisionX = 0;
         let precisionY = 0;
-        if (this.validWcs || this.isYX || this.isPVImage || this.isUVImage || this.isSwappedZ) {
+        if ((this.validWcs || this.isYX || this.isPVImage || this.isUVImage || this.isSwappedZ) && this.overlayStore.isWcsCoordinates) {
```
YuHsuan-Hwang: Shall we allow special images as in `dev`: `(((this.validWcs || this.isYX) && this.overlayStore.isWcsCoordinates) || this.isPVImage || this.isUVImage || this.isSwappedZ)`?

YuHsuan-Hwang (changes requested):
The new approach looks good to me.

TienHao:
> @TienHao When we have two images matched and displayed in world coordinate grid, if we then switch to the image coordinate grid, both images will have image coordinate grids as expected. However, if we further enable the offset mode, only the spatial reference image will have the offset image coordinate grid, not the secondary image. Any idea? 

@kswang1029: I omitted to adjust the offset of relative image coordinates for secondary image when spatial matching. That caused a fixed shift between the grids of the spatial reference image and the secondary image. It should be fixed now.

src/components/Dialogs/OnlineDataQueryDialog/CatalogOnlineQueryComponent.tsx
TienHao: fixed

src/stores/Frame/FrameStore.ts
TienHao: fixed

src/components/Dialogs/OnlineDataQueryDialog/CatalogOnlineQueryComponent.tsx
TienHao: Although I have removed the option of IMG in this component, the user can still select IMG from other places(like the toolbar in Image viewer). So, I am thinking it is still needed.

YuHsuan-Hwang: @TienHao I mean currently we are enabling this button only when it is image coordinate. Shall it be the opposite?

TienHao: Yes, I have made that other way around. Thanks for spotting that. Just fixed!

YuHsuan-Hwang (approved):

kswang1029:
@TienHao when image coordinate grid is enabled, we see the x label becomes "X coordinate". When we then enable the offset mode, it is still "X coordinate". I suggest we display it as "X offset coordinate (pixel)" and change "X coordinate" to "X coordinate (pixel)".

kswang1029 (approved):
this looks fine now. Thanks for making this fundamental and essential feature possible! 👍 New e2e tests are added.

kswang1029:
@loveluthien please have a look as well.

loveluthien:
@TienHao Overall looks nice. I found a bug. 
1) Open one image and switch to image coordinates. 
2) Appending another image that cannot match the first image in WCS. But currently, they can matched due to image coordinates. 
3) Zoom out to some level, the image viewer (or overlay?) will crash. 

There is no problem if we open two WCS matching unavailable images and switch to image coordinates, then do the matching.

There may be a more fundamental problem. When we append an image under the image coordinates and spatial match, the append image's coordinate looks wrong.

kswang1029:
@loveluthien 
for 

> Appending another image that cannot match the first image in WCS. But currently, they can matched due to image coordinates.

it is okay to have them matched in the image coordinates (matching (0, 0)). (actually this may enables some user cases). But I am open to block spatial matching if images cannot be matched in world coordinates.

As for zooming out crashing, is it due to overflow of tick values?

kswang1029:
> wrong.

Do you have video records or screenshots to show the issues?

loveluthien:
Uploading Screen Recording 2024-12-03 at 5.01.25 PM.mov…

loveluthien:
https://github.com/user-attachments/assets/65c700ce-19b8-4b50-a98e-ce4f978ea3f6

TienHao:
@loveluthien: Thanks for spotting these. All these happened because the appending image was using the wrong AST.Frame(the wcs one) to match the AST.Frame (the pixel one) of spatial reference, when both the spatial matching on append and image coordinates were on.

TienHao:
@kswang1029 : I found another bug of this two WCS-unmatchable images matching in image coordinates in the workspace. Since the current workspace does not store system type, there will be a failure to match when we open a workspace which contains such matched images. Shall we just block spatial matching if images cannot be matched in world coordinates? or extend the store of workspace in this PR (or open a new issue for it)?

https://github.com/user-attachments/assets/21fcddc2-873a-4913-8845-ae84606a77f6

kswang1029:
> @kswang1029 : I found another bug of this two WCS-unmatchable images matching in image coordinates in the workspace. Since the current workspace does not store system type, there will be a failure to match when we open a workspace which contains such matched images. Shall we just block spatial matching if images cannot be matched in world coordinates? or extend the store of workspace in this PR (or open a new issue for it)?
> 
>  2024-12-06.8.46.49.mov

Maybe we should forbid spatial matching in image coordinates _if_ images cannot be matched in world coordinates. Is it doable?

TienHao:
> Maybe we should forbid spatial matching in image coordinates _if_ images cannot be matched in world coordinates. Is it doable?

Yes, that should cause the least side effects.

loveluthien:
Spatial matching is still available in the image coordinates for images that WCS cannot be matched. Since there is no problem with matching in the image coordinates,  why don't we just turn off spatial matching when switching back from the image to WCS coordinates? That is a bit odd for disabling matching in image coordinates according to whether the world coordinates can match if that is not a technical consideration. 

I feel we should support storing the system type later rather than put a weird constraint on matching.

YuHsuan-Hwang:
> Spatial matching is still available in the image coordinates for images that WCS cannot be matched. Since there is no problem with matching in the image coordinates, why don't we just turn off spatial matching when switching back from the image to WCS coordinates? That is a bit odd for disabling matching in image coordinates according to whether the world coordinates can match if that is not a technical consideration.
> 
> I feel we should support storing the system type later rather than put a weird constraint on matching.

In this case, I think we need to either show a warning confirm dialog or disable changing the wcs system with a hint tooltip shown.

TienHao:
Here is the adjustments for spatial matching in image coordinates:
1. Disable spatial matching after switching between image and world coordinates, to make the spatial matching consistent in different coordinates. The image coordinates match images at (0, 0) pixel, and the world coordinates match images by  spatialTransformAST. A confirm alert was added.  
2. Stop save workspace if any spatial matching is enabled in the image coordinates. This constraint could be removed once the SystemType is added in the workspace storing.

src/components/Dialogs/OnlineDataQueryDialog/CatalogOnlineQueryComponent.tsx
```diff
@@ -144,7 +144,9 @@ export class CatalogQueryComponent extends React.Component {
                 </FormGroup>
                 <FormGroup inline={false} label="Center coordinates" disabled={disable}>
                     <Select
-                        items={Object.keys(SystemType).map(key => SystemType[key])}
+                        items={Object.keys(SystemType)
+                            .map(key => SystemType[key])
+                            .filter(sys => sys !== SystemType.Image)}
```
loveluthien: maybe we can replace it with `items={Object.values(SystemType).filter(sys => sys !== SystemType.Image)}`

loveluthien:
@TienHao, you blocked the online catalog search when the system type is in image coordinates. Maybe we should also block the catalog source plot under the same conditions. If we don't, the catalog source will be in the wrong places.

loveluthien (approved):
well done!!

src/stores/AppStore/AppStore.ts
```diff
@@ -603,6 +603,9 @@ export class AppStore {
                 this.setSpectralMatchingEnabled(newFrame, true);
             }
             if (this.preferenceStore.autoWCSMatching & WCSMatchingType.SPATIAL && this.spatialReference !== newFrame) {
+                if (this.overlayStore.isImgCoordinates) {
+                    AST.setI(newFrame.wcsInfo, "Current", 1);
```
YuHsuan-Hwang: I suggest we instead call `FrameStore.updateWcsSystem` in `FrameStore` `constructor` to make sure that `wcsInfo` is correctly updated after we create a frame.

src/components/Dialogs/WorkspaceDialog/WorkspaceDialogComponent.tsx
```diff
@@ -67,6 +67,12 @@ export const WorkspaceDialogComponent = observer(() => {
                 return;
             }
 
+            // TODO: to be removed after storing SystemType in workspace
+            if (appStore.overlayStore.isImgCoordinates && appStore.frames.map(frame => frame.spatialReference !== null).includes(true)) {
+                AlertStore.Instance.showAlert("Not supporting save workspace with spatial matching in image cooordinates");
```
YuHsuan-Hwang: Might be more explicit if we add: "Saving workspace failed: ..." to show that the workspace is not saved (instead of saved but with matching disabled).

YuHsuan-Hwang (changes requested):
The new fixes look good. Some minor comments:

src/stores/AppStore/AppStore.ts
TienHao: yes, that is more thoughtful!

src/components/Dialogs/WorkspaceDialog/WorkspaceDialogComponent.tsx
TienHao: has changed!

src/stores/AppStore/AppStore.ts
```diff
@@ -560,8 +560,9 @@ export class AppStore {
         };
         this.telemetryService.addFileOpenEntry(ack.fileId, ack.fileInfo.type, ack.fileInfoExtended.width, ack.fileInfoExtended.height, ack.fileInfoExtended.depth, ack.fileInfoExtended.stokes, generated);
 
+        console.log("before constructor");
```
YuHsuan-Hwang: Please remove logs.

YuHsuan-Hwang (approved):

src/stores/AppStore/AppStore.ts
TienHao: Sorry for that

YuHsuan-Hwang:
Closes https://github.com/CARTAvis/carta-frontend/issues/2449.

kswang1029 (approved):
looks good! Coordinate grid e2e tests are adjusted accordingly.

loveluthien (approved):

PR #2389, opened 2024-06-06
Yuhsuan/save image

YuHsuan-Hwang:
**Description**

Companion PR for https://github.com/CARTAvis/carta-backend/pull/1378 and https://github.com/CARTAvis/carta-protobuf/pull/97.

Changes:
* ~Send the save file/export region message with flag `overwrite = true` when the user confirmed to replace a file.~ The existing file checks in frontend is removed. ~When save image/export region is failed with message related to existing files,~ Frontend checks the `overwriteConfirmationRequired` flag when receiving `SaveFileAck` and `ExportRegionAck`. It shows the alert asking for overwrite confirmation when necessary and then resends the request with the `overwrite` flag = true.
* For saving images, disable Save button when the filename is empty.

**Checklist**

~For linked issues (if there are):~
- [x] ~GitHub Project estimate added~

For the pull request:

For dependencies:
- [x] protobuf version bumped / ~no protobuf version bumped needed~
- [x] protobuf updated to the latest dev commit / ~no protobuf update needed~


kswang1029 (approved):
looks good with https://github.com/CARTAvis/carta-backend/pull/1378. no other regression from e2e.

pford:
In CARTAvis/carta-backend/pull/1378, we discussed adding an overwrite flag to ExportRegion to protect existing files (region or not).  The frontend only confirms existing region files. The backend currently does not overwrite directories but returns success=true, so I will fix that in the backend.

- ExportRegion fails if file does not exist and cannot create (backend only)
- ExportRegion fails if path exists and not a file (backend only)
- ExportRegion fails if path is a file and overwrite is false (backend + frontend)

YuHsuan-Hwang:
Require changes for saving region files.

pford:
> Require changes for saving region files.

I have not changed protobuf, waiting for input from @kswang1029.  Currently, the frontend confirms overwrite of image files/directories for SaveFile, and overwrite of region files for ExportRegion, in the current directory.  In other directories (user types in path), the frontend does not know if the path exists or if it is an image or region file, so does not pop up overwrite dialog.

- SaveFile never overwrites a non-image directory, or a non-image file and any file/directory in another directory (overwrite=false, backend returns error).
- ExportRegion never overwrites a directory (backend returns error) but overwrites (1) a region file where the user confirmed overwrite and (2) a non-region file or a file in another directory.

YuHsuan-Hwang:
Update: send the export region message with flag overwrite = true when the user confirmed to replace a file.

kswang1029 (approved):
looks good. 👍

confluence:
I've looked at the diff in this PR, and at the moment it doesn't look like the frontend is checking for a specific error message from the backend and then retrying the save (or region export) with the overwrite flag set. I can only see the old check for existing files in `handleSaveFileClicked`, unchanged.

This frontend-side check should be removed -- the backend is now checking whether the path resolves to an existing file or directory. The frontend should be checking the text of the backend error message, and prompting the user for an overwrite confirmation, I believe *only* in the case of these error messages:

* "Cannot overwrite existing file or symlink." (images)
* "Cannot overwrite existing image." (images)
* "Export region failed: cannot overwrite existing file." (regions)

(The backend should now always refuse to overwrite existing directories, even with the overwrite flag set -- apart from the special case of images which are directories, which should be overwriteable with confirmation like a file by the image save, but *not* by the region export.)

I would suggest testing that this works appropriately for each different type of existing object, but with test objects in a subdirectory from the current directory (since that case is definitely not handled by the current frontend-side code).

confluence:
@pford @YuHsuan-Hwang Now that I'm looking at this, I think it may be a lot cleaner to add a boolean field (maybe called `overwrite_confirmation_required`, although that's a bit long) to both `SaveFileAck` and `ExportRegionAck`, to communicate this state unambiguously. Then we won't have to parse message text, and we won't break everything if we change the text later,

YuHsuan-Hwang:
Update:
Modified the flow design. The existing file checks in frontend is removed. When save image/export region is failed with message related to existing files, it shows the alert asking for overwrite confirmation and then resends the request with the overwrite flag = true.
Currently we are relying on the arbitrary error message received from backend and only specific strings will trigger the confirmation, so we might want to improve this or at least note it in the protobuf documentation.

YuHsuan-Hwang:
> I think it may be a lot cleaner to add a boolean field (maybe called `overwrite_confirmation_required`, although that's a bit long) to both `SaveFileAck` and `ExportRegionAck`, to communicate this state unambiguously.

Currently frontend only checks the `message` field when `success = false`. We need to either change the structure, or return `success = true` and `overwrite_confirmation_required = false` which could be a bit confusing.

YuHsuan-Hwang:
Update: frontend checks the `overwriteConfirmationRequired` flag when receiving `SaveFileAck` and `ExportRegionAck` and shows the alert asking for overwrite confirmation and then resends the request with the `overwrite` flag = true.

kswang1029 (approved):
looks good. no regression from e2e.

loveluthien (approved):
It looks good.

PR #2457, opened 2025-01-08
Yuhsuan/948 edge pixel

### LINKED ISSUE #948: missing raster image when zooming in edge pixels
It appears that if we zoom in the edge pixels (part of image view is empty while zooming in), at some point the raster image will disappear. We can zoom much further in and still see the raster image if there is no empty area in the image view while zooming in. 

Jun-18-2020 10-46-00

update:
Image pan has similar issue
Jun-18-2020 10-53-12

YuHsuan-Hwang:
**Description**

Closes #948. I could only reproduce the issue when panning the image to the top edge and the right edge.

The root cause is the calculation of the required tiles. `frame.requiredFrameView` is defined such that the pixel centers are at [0, 1, ..., size - 1], and the raster image borders are at [-0.5, size - 0.5]. However, `GetRequiredTiles` assumes the borders are at [0, size], so it returns an empty array for the range [-0.5, 0].

Tested that: when the image is at the top edge or right edge (with only half of pixels visible),
- the image does not disappear.
- the image does not disappear after we click next channel.
- the image does not disappear after we start animation.

**Checklist**

For the pull request:
- [x] ~unit test added (for functions with no dependencies)~

For dependencies:


loveluthien (approved):
nice fix.

kswang1029 (approved):
looks good. no regression from e2e.

PR #2512, opened 2025-04-16
Yuhsuan/2507 channel map pv preview

YuHsuan-Hwang:
**Description**

1.

Fixed #2507. This PR changes the behavior of the image view when a PV preview becomes active. In v5-beta when a preview becomes active, the image view shows the first page in non channel map mode and remains unchanged in channel map mode. Adjusted the behavior so that the source image is showed instead.

Tested that

When the preview becomes active, the image view shows
* the source image in single panel mode.
* the page of the source image in multi panel mode.
* the source image in channel map mode (with the image dropdown updated).

Code changes
* Added preview source image file id in `FrameStore.frameInfo`.
* Adjusted `ImageViewConfigStore.currentImagePage` to show the preview source image.
* Added `ChannelMapStore.displayedImage` which reuses `ImageViewConfigStore.currentImagePage`.
* Applied `ChannelMapStore.displayedImage` in `ChannelMapStore`, `ChannelMapControlComponent`, and `ChannelMapViewComponent`.
* Set the image as active when clicking the raster canvas in channel map mode.

2.

Fixed #2523 and fixed #2524.

Code changes
* Hide the image slider when the PV preview is active (when the active frame id is -1).
* Added state `AnimatorStore.startAnimationDisabled`. Disabled the play button based on the state and checked the state when calling the start animation function.
* Added `WidgetsStore.imageViewWidgetTitle` which changes with the active image and the image view visible images. Called `WidgetsStore.updateImageWidgetTitle` when initialization and when `WidgetsStore.imageViewWidgetTitle` is changed.

**Checklist**

For linked issues (if there are):
- [x] assignee and label added
- [x] GitHub Project estimate added

For the pull request:
- [x] reviewers and assignee added
- [ ] GitHub Project estimate added
- [x] changelog updated / ~no changelog update needed~
- [x] ~unit test added (for functions with no dependenies)~
- [x] API documentation added (for public variables and methods in stores)

For dependencies:
- [x] e2e test passing / corresponding fix added / new e2e test created
- [x] ~protobuf version bumped~ / no protobuf version bumped needed
- [x] ~protobuf updated to the latest dev commit~ / no protobuf update needed
- [x] ~corresponding ICD test fix added (`BackendService` changed)~ / no ICD test fix needed (`BackendService` unchanged)
- [x] ~user manual prepared (for large new features)~

Issue #2492, opened 2025-03-14
Channel map view enhancements


YuHsuan-Hwang:
Some possible enhancements for the channel map view feature:
- [ ] tile related enhancements
  - [ ] investigate if `BigInt` could be replaced
  - [ ] investigate if the usage of tile cache in non channel map mode is optimized
  - [x] https://github.com/CARTAvis/carta-frontend/pull/2348#discussion_r1938842339
- [x] coordinate overlay related enhancements
  - [x] investigate if overlayComponentRef+throttling+useEffect could be simplified (https://github.com/CARTAvis/carta-frontend/pull/2348#discussion_r1928358711)
  - [x] consistent gaps comparing to non channel map mode
- [x] UI design enhancement
  - [x] clip min/max of the active channel
  - [x] active channel related design (animator widget, render config widget)
- [ ] additional features
  - [ ] cursor info widget
  - [x] channel/spectral/velocity string
  - [x] PNG export
  - [ ] beam per channel
  - [ ] color blending related

PR #2519, opened 2025-05-05
Yuhsuan/2492 channel map overlay

YuHsuan-Hwang:
**Description**

This PR is for #2492: channel map coordinate overlay related enhancements. The overlay in channel map mode are adjusted so that they are consistent with the overlays in non-channel map mode.

Code changes

* Created a new store `OverlaySettings` for the shared settings in the image view settings widget. `OverlayStore` becomes frame-dependent.
* Unified variable naming in `OverlayStore` and `ChannelMapViewComponent`, with detailed documentation added in `OverlayStore`.
* Created `OverlayStore` for PV previews and channel map outer and inner part.
* Adjusted the redraw algorithm in `ChannelMapInnerOverlayComponent`: a canvas is created for each channel, and the refs to all canvases are stored in a map ([ref](https://18.react.dev/learn/manipulating-the-dom-with-refs#how-to-manage-a-list-of-refs-using-a-ref-callback)). A draw function is called in `OverlayComponent` to redraw all canvases.
* Moved the image view widget width/height from `OverlayStore` to `AppStore`.
* Moved the PV preview widget width/height from `FrameStore` to `PvGeneratorWidgetStore`.
* Moved frame-dependent colorbar variables `yOffset`, `height`, `tickNum` from `OverlayStore` to `ColorbarStore`.
* Removed optional props in `ColorbarComponent`, `OverlayComponent`, and `ToolbarComponent`.
* Removed unused prop `overlaySettings` in `RegionViewComponent`.
* Changed `OverlaySettings.base` and `OverlaySettings.defaultGap` from computed getter functions to regular fields.

Tested that (with devicePixelRatio = 1 and 2)

* The wcs overlay does not appear blurry in channel map mode.
* The wcs overlay remains correct when the number of visible channels is less than row × column.
* Labels, title, cursor info, and (right/top/bottom) colorbar remain unchanged when switching between channel map and non-channel map modes.
* The outermost borders (top, bottom, left, right) remain unchanged with different row num and column num in channel map mode (there may be a 1-pixel difference in gaps).
* There is no gap between the raster canvas and the wcs overlay border in channel map mode.
* The active channel highlight border is roughly symmetric to the wcs overlay border (there may be a 1-pixel difference with difference row num and column num). There is no gap between the active channel highlight border and the wcs overlay border.
* The regions and channel labels are placed correctly in channel map mode.
* The wcs overlay remains correct in single/multi-panel mode and PV previews.

~ToDo: update to latest dev after #2512 is merged.~

**Checklist**

For the pull request:

For dependencies:
- [x] ~protobuf updated to the latest dev commit ~/ no protobuf update needed


kswang1029 (commented):
@YuHsuan-Hwang Mostly this PR looks good. Thank you for cleaning this up. 

Just one issue I spotted so far: the axis labeling may be switched to interior mode. When this happens, we need a separate rule for masking the AST layer at different parts of the channel map grid, otherwise we may see labels are out of the axes box. If this is not straightforward to fix, we can leave it as an open issue for the future. 

Screenshot 2025-05-26 at 10 52 35

src/components/ImageView/RasterView/RasterViewComponent.tsx
```diff
@@ -67,17 +66,21 @@ export class RasterViewComponent extends React.Component<RasterViewComponentProp
             return;
         }
 
+        const pixelRatio = devicePixelRatio * AppStore.Instance.imageRatio;
```
loveluthien: we could use AppStore.Instance.pixelRatio

YuHsuan-Hwang:
> Just one issue I spotted so far: the axis labeling may be switched to interior mode. When this happens, we need a separate rule for masking the AST layer at different parts of the channel map grid, otherwise we may see labels are out of the axes box. If this is not straightforward to fix, we can leave it as an open issue for the future.

@kswang1029 I'm not sure if we should slice that part since we will remove some information. I think we will also get labels out of the box with small image view panels in non channel map mode, so I prefer we file a new issue if necessary.

src/components/ImageView/RasterView/RasterViewComponent.tsx
YuHsuan-Hwang: @loveluthien Adjusted, thanks.

loveluthien (approved):
😎

kswang1029 (approved):
looks good!

PR #2563, opened 2025-07-09
Fix preference schema bugs

### LINKED ISSUE #2561: Update schemas to correct permitted ranges and enums in preference definitions; add conversion from incorrect wcsType values
This is a companion issue for [backend #1485](https://github.com/CARTAvis/carta-backend/issues/1485).

Once [the schema has been corrected](https://github.com/CARTAvis/schemas/issues/8), the submodule here should be updated.

I previously intended to capitalise the `wcsType` values in the schema, which would have required no other changes in the frontend, but this would be inconsistent with other string enums in the schema. Instead, as @loveluthien suggested, I think we should change to the lowercase values in the frontend. This requires these changes:
1. Updating the string values in the `WCSType` class
2. Updating the preference dialog component so that the display names of the options are still capitalised
3. The function for upgrading the preferences should normalise the case of the `wcsType` value (lowercasing it), because previous versions of the backend (not the controller) allowed the invalid capitalised values to be saved, and they may be present in existing preference files.

loveluthien:
**Description**

Close #2561. Fix the bug that invalid values in preferences.json cause all preferences to be reset. 

This PR needs backend: https://github.com/CARTAvis/carta-backend/pull/1486 and schemas: https://github.com/CARTAvis/schemas/pull/10.

**Checklist**

For the pull request:

For dependencies:
- [x] corresponding ICD test fix added (`BackendService` changed) / no ICD test fix needed (`BackendService` unchanged)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [5.0.2]
-* Fixed invalid values in preferences.json causes all preferences to be reset ([#2561](https://github.com/CARTAvis/carta-frontend/issues/2561)).
+* Fixed invalid values in preferences.json cause all preferences to be reset ([#2561](https://github.com/CARTAvis/carta-frontend/issues/2561)).
 
 ## [5.0.1]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [5.0.2]
+
+### Fixed
 * Fixed invalid values in preferences.json cause all preferences to be reset ([#2561](https://github.com/CARTAvis/carta-frontend/issues/2561)).
 
 ## [5.0.1]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [5.0.2]
+* Fixed invalid values in preferences.json causes all preferences to be reset ([#2561](https://github.com/CARTAvis/carta-frontend/issues/2561)).
+
 ## [5.0.1]
 
 ### Fixed
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -610,7 +610,14 @@ export class PreferenceDialogComponent extends React.Component {
                     </HTMLSelect>
                 </FormGroup>
                 <FormGroup inline={true} label="Region size" labelInfo="(px)">
-                    <SafeNumericInput placeholder="Region size" min={1} value={preference.regionSize} stepSize={1} onValueChange={(value: number) => preference.setPreference(PreferenceKeys.REGION_SIZE, Math.max(1, value))} />
+                    <SafeNumericInput
+                        placeholder="Region size"
+                        min={10}
+                        max={100}
+                        value={preference.regionSize}
+                        stepSize={1}
+                        onValueChange={(value: number) => preference.setPreference(PreferenceKeys.REGION_SIZE, Math.max(10, Math.min(100, value)))}
+                    />
                 </FormGroup>
                 <FormGroup inline={true} label="Creation mode">
                     <RadioGroup selectedValue={preference.regionCreationMode} onChange={ev => preference.setPreference(PreferenceKeys.REGION_CREATION_MODE, ev.currentTarget.value)}>
@@ -668,9 +675,10 @@ export class PreferenceDialogComponent extends React.Component {
                     <SafeNumericInput
                         placeholder="Point size"
                         min={1}
+                        max={100}
                         value={preference.pointAnnotationWidth}
                         stepSize={1}
-                        onValueChange={(value: number) => preference.setPreference(PreferenceKeys.POINT_ANNOTATION_WIDTH, Math.max(1, value))}
+                        onValueChange={(value: number) => preference.setPreference(PreferenceKeys.POINT_ANNOTATION_WIDTH, Math.max(1, Math.min(100, value)))}
                     />
                 </FormGroup>
             </React.Fragment>
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -55,13 +55,13 @@ export class PreferenceDialogComponent extends React.Component {
     };
 
     @computed get pvPreviewCubeSizeMaxValue(): number {
-        if (PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit === MemoryUnit.TB) {
+        if (PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit === MemoryUnit.TB) {
             return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e12;
-        } else if (PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit === MemoryUnit.GB) {
+        } else if (PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB) {
             return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e9;
-        } else if (PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit === MemoryUnit.MB) {
+        } else if (PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit === MemoryUnit.MB) {
             return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e6;
-        } else if (PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit === MemoryUnit.kB) {
+        } else if (PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit === MemoryUnit.kB) {
             return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e3;
         } else {
             return PV_PREVIEW_CUBE_SIZE_LIMIT;
@@ -822,12 +822,12 @@ export class PreferenceDialogComponent extends React.Component {
                             placeholder="PV preview cube size limit"
                             min={1e-12}
                             max={this.pvPreviewCubeSizeMaxValue}
-                            value={preference.pvPreivewCubeSizeLimit}
+                            value={preference.pvPreviewCubeSizeLimit}
                             majorStepSize={1}
                             stepSize={1}
                             onValueChange={value => preference.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT, value)}
                         />
-                        <HTMLSelect value={preference.pvPreivewCubeSizeLimitUnit} onChange={ev => this.handlePvPreviewCubeSizeUnitChange(ev.target.value)}>
+                        <HTMLSelect value={preference.pvPreviewCubeSizeLimitUnit} onChange={ev => this.handlePvPreviewCubeSizeUnitChange(ev.target.value)}>
                             <option key={0} value={"MB"}>
                                 MB
                             </option>
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -45,7 +45,7 @@ export enum MemoryUnit {
 
 const PercentileSelect = Select<string>;
 
-const PV_PREVIEW_CUBE_SIZE_LIMIT = 1000000000; //need to be removed and replaced by backend limit
+const PV_PREVIEW_CUBE_SIZE_LIMIT = 2000000000; //need to be removed and replaced by backend limit
 
 @observer
 export class PreferenceDialogComponent extends React.Component {
@@ -55,17 +55,22 @@ export class PreferenceDialogComponent extends React.Component {
     };
 
     @computed get pvPreviewCubeSizeMaxValue(): number {
-        if (PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit === MemoryUnit.TB) {
+        if (this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.TB) {
             return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e12;
-        } else if (PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB) {
+        } else if (this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.GB) {
             return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e9;
-        } else if (PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit === MemoryUnit.MB) {
+        } else if (this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.MB) {
             return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e6;
-        } else if (PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit === MemoryUnit.kB) {
+        } else if (this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.kB) {
             return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e3;
         } else {
             return PV_PREVIEW_CUBE_SIZE_LIMIT;
         }
+    }
+
+    @computed get showedPvPreviewCubeSizeLimit(): number {
+        const preference = PreferenceStore.Instance;
+        return this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? preference.pvPreviewCubeSizeLimit : preference.pvPreviewCubeSizeLimit * 1000;
     }
 
     constructor(props: any) {
@@ -99,8 +104,18 @@ export class PreferenceDialogComponent extends React.Component {
     }, 100);
 
     @action private handlePvPreviewCubeSizeUnitChange = _.throttle(unit => {
-        PreferenceStore.Instance.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT_UNIT, unit);
+        this.showedPvPreviewCubeSizeLimitUnit = unit;
+        // always store value in GB
+        PreferenceStore.Instance.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT_UNIT, MemoryUnit.GB);
     }, 100);
+
+    @action private handlePvPreviewCubeSizeChange = _.throttle((size, unit) => {
+        const storedSize = unit === MemoryUnit.GB ? size : size / 1000; // Convert to GB if necessary
+        PreferenceStore.Instance.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT, storedSize);
+    }, 100);
+
+    // variable for showing preview cube size unit in the dialog
+    @observable private showedPvPreviewCubeSizeLimitUnit = PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit;
 
     private reset = () => {
         const preference = PreferenceStore.Instance;
@@ -822,12 +837,12 @@ export class PreferenceDialogComponent extends React.Component {
                             placeholder="PV preview cube size limit"
                             min={1e-12}
                             max={this.pvPreviewCubeSizeMaxValue}
-                            value={preference.pvPreviewCubeSizeLimit}
-                            majorStepSize={1}
-                            stepSize={preference.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 0.1 : 100}
-                            onValueChange={value => preference.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT, value)}
+                            value={this.showedPvPreviewCubeSizeLimit}
+                            majorStepSize={2}
+                            stepSize={this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 0.1 : 100}
+                            onValueChange={value => this.handlePvPreviewCubeSizeChange(value, this.showedPvPreviewCubeSizeLimitUnit)}
                         />
-                        <HTMLSelect value={preference.pvPreviewCubeSizeLimitUnit} onChange={ev => this.handlePvPreviewCubeSizeUnitChange(ev.target.value)}>
+                        <HTMLSelect value={this.showedPvPreviewCubeSizeLimitUnit} onChange={ev => this.handlePvPreviewCubeSizeUnitChange(ev.target.value)}>
                             <option key={0} value={"MB"}>
                                 MB
                             </option>
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -451,11 +451,11 @@ export class PreferenceDialogComponent extends React.Component {
                     </HTMLSelect>
                 </FormGroup>
                 <FormGroup inline={true} label="WCS format">
-                    <HTMLSelect
-                        options={[WCSType.AUTOMATIC, WCSType.DEGREES, WCSType.SEXAGESIMAL]}
-                        value={preference.wcsType}
-                        onChange={(event: React.FormEvent<HTMLSelectElement>) => preference.setPreference(PreferenceKeys.WCS_OVERLAY_WCS_TYPE, event.currentTarget.value)}
-                    />
+                    <HTMLSelect value={preference.wcsType} onChange={(event: React.FormEvent<HTMLSelectElement>) => preference.setPreference(PreferenceKeys.WCS_OVERLAY_WCS_TYPE, event.currentTarget.value)}>
+                        <option value={WCSType.AUTOMATIC}>Automatic</option>
+                        <option value={WCSType.DEGREES}>Degrees</option>
+                        <option value={WCSType.SEXAGESIMAL}>Sexagesimal</option>
+                    </HTMLSelect>
                 </FormGroup>
                 <FormGroup inline={true} label="Colorbar visible">
                     <Switch checked={preference.colorbarVisible} onChange={ev => preference.setPreference(PreferenceKeys.WCS_OVERLAY_COLORBAR_VISIBLE, ev.currentTarget.checked)} />
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -43,9 +43,17 @@ export enum MemoryUnit {
     B = "B"
 }
 
+export enum ConvertToGB {
+    TB = 1e3,
+    GB = 1,
+    MB = 1e-3,
+    kB = 1e-6,
+    B = 1e-9
+}
+
 const PercentileSelect = Select<string>;
 
-const PV_PREVIEW_CUBE_SIZE_LIMIT = 2000000000; //need to be removed and replaced by backend limit
+const PV_PREVIEW_CUBE_SIZE_LIMIT = 2; //in unit of GB
 
 @observer
 export class PreferenceDialogComponent extends React.Component {
@@ -55,22 +63,16 @@ export class PreferenceDialogComponent extends React.Component {
     };
 
     @computed get pvPreviewCubeSizeMaxValue(): number {
-        if (this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.TB) {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e12;
-        } else if (this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.GB) {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e9;
-        } else if (this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.MB) {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e6;
-        } else if (this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.kB) {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e3;
-        } else {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT;
-        }
+        return this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? PV_PREVIEW_CUBE_SIZE_LIMIT : PV_PREVIEW_CUBE_SIZE_LIMIT / ConvertToGB.MB;
+    }
+
+    @computed get pvPreviewCubeSizeMinValue(): number {
+        return this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 0.1 : 128;
     }
 
     @computed get showedPvPreviewCubeSizeLimit(): number {
         const preference = PreferenceStore.Instance;
-        return this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? preference.pvPreviewCubeSizeLimit : preference.pvPreviewCubeSizeLimit * 1000;
+        return this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? preference.pvPreviewCubeSizeLimit : preference.pvPreviewCubeSizeLimit / ConvertToGB.MB;
     }
 
     constructor(props: any) {
@@ -104,18 +106,16 @@ export class PreferenceDialogComponent extends React.Component {
     }, 100);
 
     @action private handlePvPreviewCubeSizeUnitChange = _.throttle(unit => {
-        this.showedPvPreviewCubeSizeLimitUnit = unit;
-        // always store value in GB
-        PreferenceStore.Instance.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT_UNIT, MemoryUnit.GB);
+        this.pvPreviewCubeSizeLimitUnit = unit;
     }, 100);
 
     @action private handlePvPreviewCubeSizeChange = _.throttle((size, unit) => {
-        const storedSize = unit === MemoryUnit.GB ? size : size / 1000; // Convert to GB if necessary
+        const storedSize = unit === MemoryUnit.GB ? size : size * ConvertToGB.MB; // Convert to GB if necessary
         PreferenceStore.Instance.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT, storedSize);
     }, 100);
 
     // variable for showing preview cube size unit in the dialog
-    @observable private showedPvPreviewCubeSizeLimitUnit = PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit;
+    @observable private pvPreviewCubeSizeLimitUnit = PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit;
 
     private reset = () => {
         const preference = PreferenceStore.Instance;
@@ -835,18 +835,18 @@ export class PreferenceDialogComponent extends React.Component {
                     <div className="pv-preview-cube-size-limit">
                         <SafeNumericInput
                             placeholder="PV preview cube size limit"
-                            min={1e-12}
+                            min={this.pvPreviewCubeSizeMinValue}
                             max={this.pvPreviewCubeSizeMaxValue}
                             value={this.showedPvPreviewCubeSizeLimit}
-                            majorStepSize={this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 2 : 2000}
-                            stepSize={this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 0.1 : 100}
-                            onValueChange={value => this.handlePvPreviewCubeSizeChange(value, this.showedPvPreviewCubeSizeLimitUnit)}
+                            majorStepSize={this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 2 : 2048}
+                            stepSize={this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 0.1 : 128}
+                            onValueChange={value => this.handlePvPreviewCubeSizeChange(value, this.pvPreviewCubeSizeLimitUnit)}
                         />
-                        <HTMLSelect value={this.showedPvPreviewCubeSizeLimitUnit} onChange={ev => this.handlePvPreviewCubeSizeUnitChange(ev.target.value)}>
-                            <option key={0} value={"MB"}>
+                        <HTMLSelect value={this.pvPreviewCubeSizeLimitUnit} onChange={ev => this.handlePvPreviewCubeSizeUnitChange(ev.target.value)}>
+                            <option key={0} value={MemoryUnit.MB}>
                                 MB
                             </option>
-                            <option key={1} value={"GB"}>
+                            <option key={1} value={MemoryUnit.GB}>
                                 GB
                             </option>
                         </HTMLSelect>
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -824,7 +824,7 @@ export class PreferenceDialogComponent extends React.Component {
                             max={this.pvPreviewCubeSizeMaxValue}
                             value={preference.pvPreviewCubeSizeLimit}
                             majorStepSize={1}
-                            stepSize={1}
+                            stepSize={preference.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 0.1 : 100}
                             onValueChange={value => preference.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT, value)}
                         />
                         <HTMLSelect value={preference.pvPreviewCubeSizeLimitUnit} onChange={ev => this.handlePvPreviewCubeSizeUnitChange(ev.target.value)}>
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -265,7 +265,13 @@ export class PreferenceDialogComponent extends React.Component {
                 </FormGroup>
                 {(preference.scaling === FrameScaling.LOG || preference.scaling === FrameScaling.POWER) && (
                     <FormGroup label={"Alpha"} inline={true}>
-                        <SafeNumericInput buttonPosition={"none"} value={preference.scalingAlpha} onValueChange={value => preference.setPreference(PreferenceKeys.RENDER_CONFIG_SCALING_ALPHA, value)} />
+                        <SafeNumericInput
+                            min={RenderConfigStore.ALPHA_MIN}
+                            max={RenderConfigStore.ALPHA_MAX}
+                            buttonPosition={"none"}
+                            value={preference.scalingAlpha}
+                            onValueChange={value => preference.setPreference(PreferenceKeys.RENDER_CONFIG_SCALING_ALPHA, value)}
+                        />
                     </FormGroup>
                 )}
                 {preference.scaling === FrameScaling.GAMMA && (
--- a/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
+++ b/src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
@@ -35,14 +35,6 @@ enum PreferenceDialogTabs {
     COMPATIBILITY
 }
 
-export enum MemoryUnit {
-    TB = "TB",
-    GB = "GB",
-    MB = "MB",
-    kB = "kB",
-    B = "B"
-}
-
 export enum ConvertToGB {
     TB = 1e3,
     GB = 1,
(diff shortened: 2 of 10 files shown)


src/services/ApiService.ts
```diff
@@ -191,6 +192,7 @@ export class ApiService {
                         console.log(`Removing invalid preference ${error.instancePath}`);
                         // Trim the leading "." from the path
                         delete preferences[error.instancePath.substring(1)];
+                        this.clearPreferences([error.instancePath.substring(1)]);
```
confluence: Do we actually want to remove the invalid preference from the config? 

I originally thought that we did this, but we currently only remove it from the runtime preferences without modifying the preferences in the database, and I can see why we made this decision.

I have no strong opinion, but if we do, I think that we should signal the change to the user in a more prominent way (by displaying a warning in the actual UI, not just the console).

src/services/ApiService.ts
```diff
@@ -200,44 +202,27 @@ export class ApiService {
...
-                    case 8:
-                        preferences.astColor = "auto-gray";
-                        break;
-                    default:
-                        preferences.astColor = "auto-blue";
-                        break;
-                }
-            }
-            preferences.version = 2;
+        if (preferences["version"] && preferences["version"] === 1) {
```
confluence: The first `preferences["version"]` here is redundant. `preferences["version"] === 1` correctly evaluates to false whether the key is not present (in which case `preferences["version"]` is undefined) or the key is present and not equal to 1.

In the original conditional, `preferences?.version` is also checking whether `preferences` exists. Since there's an explicit test for this before this function is called, that's not necessary either.

src/services/ApiService.ts
```diff
@@ -200,44 +202,27 @@ export class ApiService {
...
+            // Convert preferences[PreferenceKeys.WCS_OVERLAY_AST_COLOR] from a number in version 1 to a string in version 2
+            // default to "auto-blue" if the value is not in the AST_COLORS map
+            const astColorKey = PreferenceKeys.WCS_OVERLAY_AST_COLOR;
+            if (preferences[astColorKey] || preferences[astColorKey] === 0) {
```
confluence: This is better written as `if (astColorKey in preferences)`.

src/services/ApiService.ts
```diff
@@ -200,44 +202,27 @@ export class ApiService {
...
+                preferences[astColorKey] = astColorArrays.includes(preferences[astColorKey]) ? AST_COLORS.get(preferences[astColorKey]) : "auto-blue";
+            }
+            preferences["version"] = 2;
             this.setPreferences(preferences);
         }
+
+        // Upgrade to schema version 2
+        const schemaVersionRegex = /schema_(\d+)\.json/;
+        const schemaKey = "$schema";
+        if (preferences[schemaKey] && preferences[schemaKey].match(schemaVersionRegex)[1] === "1") {
```
confluence: This is not correct or necessary. The incorrect `wcsType` values may also be present in preferences which have the v2 schema. Lowercasing will produce the correct result regardless of what they are, but if you want to check if the conversion is necessary (to avoid making an unnecessary request) you can directly test the values (variable added for readability):
```
const key = PreferenceKeys.WCS_OVERLAY_WCS_TYPE;
if (/[A-Z]/.test(preferences[key])) {
    preferences[key] = preferences[key].toLowerCase();
}
```

src/services/ApiService.ts
```diff
@@ -200,44 +202,27 @@ export class ApiService {
...
+            if (preferences[PreferenceKeys.WCS_OVERLAY_WCS_TYPE]) {
+                preferences[PreferenceKeys.WCS_OVERLAY_WCS_TYPE] = preferences[PreferenceKeys.WCS_OVERLAY_WCS_TYPE].toLowerCase();
+                this.setPreferences(preferences);
```
confluence: As I mentioned in the issue, this (and the other conditional `setPreferences` in this function) will always fail if there are other invalid preferences.

I think that the cleanest way to fix this is to make these requests with the changed value *only* (rather than the entire preferences object). There's already a function for this, `setPreference`, which can be used in both places.

(Both the backend and the controller add a schema version before validation, which is how the function works despite not explicitly including one. We should probably revise that at some point, but it's not necessary right now.)

confluence (changes requested):

src/utilities/color/color.ts
```diff
@@ -29,6 +29,19 @@ export const SWATCH_COLORS = [
...
+    [1, "auto-white"],
+    [2, "auto-red"],
+    [3, "auto-forest"],
+    [4, "auto-blue"],
+    [5, "auto-turquoise"],
+    [6, "auto-violet"],
+    [7, "auto-gold"],
+    [8, "auto-gray"]
+]);
+
```
confluence: 1. This can just be an array, since the "keys" are sequential integers starting from zero.
2. Since this is *only* used for conversion of a legacy preferences format, it may be better to define this directly in the upgrade function.

confluence (changes requested):

src/services/ApiService.ts
kswang1029: with a warning toast (non blocking) I suppose? Should not need a popup (blocking UI) for confirmation.

loveluthien: I found that `ApiService.PreferenceValidator.errors` can only catch one invalid value. If we have two invalid preferences, we need to reload preference and trigger the validation again.

confluence:
A minor thing: this PR's name should be something like "Fix preference schema bugs" -- the reset issue specifically is 100% a backend problem and doesn't have anything to do with the frontend changes.

src/services/ApiService.ts
loveluthien: Found the solution, we need `allErrors: true` for `Ajv`. @confluence, all comments are done. thanks.

confluence:
@loveluthien I'm still preparing a detailed list of the schema changes and possible impact on existing preferences, but I'd like to flag some more obsolete preferences that can be corrected (I found these when testing the latest schemas against existing prefs on the Ilifu production CARTA server):

* `imagePanelMode`: some old prefs have the disallowed value `"static"` -- I assume that this is supposed to be `"fixed"`.
* `logEventList`: some old prefs have a list of strings corresponding to the event enum names, e.g. `"CLOSE_FILE"`. We should be able to look these up and translate them to numbers.

@kswang1029 Should we add these translations to the current version of the upgrade function and remove them in the next (possibly including the AST colour translation)? I think that this would make sense for gradual deprecation.

I don't think that we should remove the AST translation in the current version, because (as a result of the way the function currently calls `setPreferences`) this corrected preference would never have been saved by the controller if other invalid preferences were present. That should be fixed in this PR, so I think it will be safe to deprecate the translation now and remove it in the next version.

src/services/ApiService.ts
```diff
@@ -185,58 +185,54 @@ export class ApiService {
...
-                        break;
-                    default:
-                        preferences.astColor = "auto-blue";
-                        break;
+        if (preferences["version"] === 1) {
+            // Convert preferences[PreferenceKeys.WCS_OVERLAY_AST_COLOR] from a number in version 1 to a string in version 2
+            // default to "auto-blue" if the value is not in the AST_COLORS map
+            const astColors = ["auto-black", "auto-white", "auto-red", "auto-forest", "auto-blue", "auto-turquoise", "auto-violet", "auto-gold", "auto-gray"];
+            const astColorKey = PreferenceKeys.WCS_OVERLAY_AST_COLOR;
+            if (astColorKey in preferences || preferences[astColorKey] === 0) {
```
confluence: You don't need the `|| preferences[astColorKey] === 0`; that was only necessary when `preferences[astColorKey]` was being evaluated for truthiness (because zero is false). `astColorKey in preferences` replaces both of these with one check.

confluence:
> @kswang1029 Should we add these translations to the current version of the upgrade function and remove them in the next (possibly including the AST colour translation)? I think that this would make sense for gradual deprecation.

Alternatively, since we have decided to delete unhandled invalid preferences, we could keep things simple and let these be deleted.

src/services/ApiService.ts
```diff
@@ -185,58 +185,54 @@ export class ApiService {
...
+        }
+
+        // Upgrade to schema version 2
+        const schemaVersionRegex = /schema_(\d+)\.json/;
+        const schemaKey = "$schema";
+        if (preferences[schemaKey]) {
+            const matchResult = preferences[schemaKey].match(schemaVersionRegex);
+            if (matchResult && matchResult[1] === "1") {
+                if (preferences[PreferenceKeys.WCS_OVERLAY_WCS_TYPE]) {
+                    preferences[PreferenceKeys.WCS_OVERLAY_WCS_TYPE] = preferences[PreferenceKeys.WCS_OVERLAY_WCS_TYPE].toLowerCase();
```
confluence: My [previous comment about this](https://github.com/CARTAvis/carta-frontend/pull/2563#discussion_r2194255642) is still unresolved.

src/services/ApiService.ts
```diff
@@ -185,58 +185,54 @@ export class ApiService {
...
+                    this.setPreferences(preferences);
```
confluence: And my [previous comment about this](https://github.com/CARTAvis/carta-frontend/pull/2563#discussion_r2194284710) is also still unresolved.

kswang1029:
> > @kswang1029 Should we add these translations to the current version of the upgrade function and remove them in the next (possibly including the AST colour translation)? I think that this would make sense for gradual deprecation.
> 
> Alternatively, since we have decided to delete unhandled invalid preferences, we could keep things simple and let these be deleted.

Let’s keep to simple as long as users are not affected at all.

loveluthien:
> > > @kswang1029 Should we add these translations to the current version of the upgrade function and remove them in the next (possibly including the AST colour translation)? I think that this would make sense for gradual deprecation.
> > 
> > 
> > Alternatively, since we have decided to delete unhandled invalid preferences, we could keep things simple and let these be deleted.
> 
> Let’s keep to simple as long as users are not affected at all.

Does that mean we can remove the code related to upgrading old preferences to the new form?

src/services/ApiService.ts
```diff
@@ -185,58 +185,54 @@ export class ApiService {
...
-                    default:
-                        preferences.astColor = "auto-blue";
-                        break;
+        if (preferences["version"] === 1) {
+            // Convert preferences[PreferenceKeys.WCS_OVERLAY_AST_COLOR] from a number in version 1 to a string in version 2
+            // default to "auto-blue" if the value is not in the AST_COLORS map
+            const astColors = ["auto-black", "auto-white", "auto-red", "auto-forest", "auto-blue", "auto-turquoise", "auto-violet", "auto-gold", "auto-gray"];
+            const astColorKey = PreferenceKeys.WCS_OVERLAY_AST_COLOR;
+            if (astColorKey in preferences || preferences[astColorKey] === 0) {
+                preferences[astColorKey] = astColors.includes(preferences[astColorKey]) ? astColors[preferences[astColorKey]] : "auto-blue";
```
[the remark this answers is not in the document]
confluence: @loveluthien I have an updated, cleaner suggestion: Use a [numeric enum](https://www.typescriptlang.org/docs/handbook/enums.html) -- no explicit numbers, since they start at 0, and leave off the `auto-` (and put it back later).
```
enum ASTColor {black, white, red (...)};

// then later
const color = preferences[astColorKey] in ASTColor ? AstColor[preferences[astColorKey]] : "blue";
preferences[astColorKey] = `auto-${color}`;
```

confluence (changes requested):

confluence:
> Does that mean we can remove the code related to upgrading old preferences to the new form?

@loveluthien @kswang1029 My understanding is:
* `upgradePreferences` in `PreferenceStore`: yes, remove (read from very old localstorage format -- no longer needed)
* `upgradePreferences` in `ApiService`: NO -- we should keep the `astColor` fix in this version, and add the `wcsType` fix, and only remove them in the next stable release. But *don't* add the other old format issues I mentioned (let them be deleted instead).

src/services/ApiService.ts
```diff
@@ -4,10 +4,11 @@ import {action, computed, makeObservable, observable} from "mobx";
 
 import {AppToaster} from "components/Shared";
 import {LayoutConfig, Snippet, Workspace, WorkspaceListItem} from "models";
+import {AppStore, PreferenceKeys} from "stores";
+import {Pre} from "@blueprintjs/core";
```
confluence: This is unused and causing CI to fail.

src/services/ApiService.ts
```diff
@@ -185,58 +186,65 @@ export class ApiService {
...
+            let deletedKeys: string[] = [];
             if (!valid) {
                 for (const error of ApiService.PreferenceValidator.errors ?? []) {
                     if (error.instancePath) {
                         console.log(`Removing invalid preference ${error.instancePath}`);
                         // Trim the leading "." from the path
                         delete preferences[error.instancePath.substring(1)];
+
+                        this.clearPreferences([error.instancePath.substring(1)]);
+                        deletedKeys.push(error.instancePath.substring(1));
```
confluence: Maybe use an intermediate variable for `error.instancePath.substring(1)`?

src/services/ApiService.ts
```diff
@@ -185,58 +186,65 @@ export class ApiService {
...
+                violet,
+                gold,
+                gray
+            }
+            const astColorKey = PreferenceKeys.WCS_OVERLAY_AST_COLOR;
+            const color = preferences[astColorKey] in ASTColors ? ASTColors[preferences[astColorKey]] : "blue";
+            preferences[astColorKey] = `auto-${color}`;
+            this.setPreference(astColorKey, preferences[astColorKey]);
+
+            // preferences["version"] = 2;
```
confluence: This should be restored; setting the preference on the server doesn't change the value in the preferences object. This may not be used currently, but could cause an unexpected issue.

src/services/ApiService.ts
```diff
@@ -185,58 +186,65 @@ export class ApiService {
...
+            this.setPreference("version", 2);
+        }
+
+        // Upgrade to schema version 2
```
confluence: This comment is inaccurate and should be removed (maybe add a comment briefly explaining that this is fixing an inconsistency with the schema).

src/services/ApiService.ts
```diff
@@ -185,58 +186,65 @@ export class ApiService {
...
+        let key = PreferenceKeys.WCS_OVERLAY_WCS_TYPE;
+        if (preferences[key]) {
+            preferences[key] = preferences[key].toLowerCase();
+            this.setPreference(key, preferences[key]);
+        }
+
+        key = PreferenceKeys.IMAGE_PANEL_MODE;
+        if (preferences[key] === "static") {
+            preferences[key] = "fixed";
+            this.setPreference(key, preferences[key]);
```
confluence: I think we agreed to leave this out (the bad pref will just be deleted in the validation step).

src/services/ApiService.ts
```diff
@@ -185,58 +186,65 @@ export class ApiService {
...
+            preferences[astColorKey] = `auto-${color}`;
+            this.setPreference(astColorKey, preferences[astColorKey]);
+
+            // preferences["version"] = 2;
+            this.setPreference("version", 2);
+        }
+
+        // Upgrade to schema version 2
+        let key = PreferenceKeys.WCS_OVERLAY_WCS_TYPE;
+        if (preferences[key]) {
```
confluence: Please include the regex test here (`if (/[A-Z]/.test(preferences[key]))`), otherwise you will unnecessarily send this request if the correct value is present.

confluence (changes requested):

loveluthien:
@confluence, thanks the comments.

src/services/ApiService.ts
```diff
@@ -185,58 +185,60 @@ export class ApiService {
...
+            const astColorKey = PreferenceKeys.WCS_OVERLAY_AST_COLOR;
+            const color = preferences[astColorKey] in ASTColors ? ASTColors[preferences[astColorKey]] : "blue";
+            preferences[astColorKey] = `auto-${color}`;
+            this.setPreference(astColorKey, preferences[astColorKey]);
+
+            preferences["version"] = 2;
+            this.setPreference("version", 2);
+        }
+
+        // Fix the inconsistent `wcsType` value between schema v1 and v2
```
confluence: This is not a v1/v2 issue; this was always lowercase in the schema. I suggest: `Normalise case of wcsType value, which may have been saved incorrectly in existing preferences`

confluence (changes requested):
Only a minor remark about a comment. Before I approve, I would like to list the "unreachable" preferences, which I think should also be deprecated in some way (I will create this list ASAP).

loveluthien:
@confluence @kswang1029, we bump the version to 5.1.0, right?

kswang1029:
> @confluence @kswang1029, we bump the version to 5.1.0, right?

5.0.2 I think

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [5.0.2]
+* Fixed invalid values in preferences.json cause all preferences to be reset ([#2561](https://github.com/CARTAvis/carta-frontend/issues/2561)).
+
```
confluence: This is missing the `Fixed` subheading.

confluence (changes requested):

CHANGELOG.md
loveluthien: Thanks. I added.

confluence:
I have made a small change to the schema (I missed a minimum value).

@loveluthien @kswang1029 @veggiesaurus I think that this is a relatively complete summary of the impact of the changes in the schemas PR, plus other outstanding schema / frontend issues that I found while making these changes:

**Impact of changes to the schema**

Probably no existing illegal values; some legal values unblocked:
```
percentile
vectorOverlayPixelAveraging
imageCompressionQuality
animationCompressionQuality
```

Existing illegal values possible; some legal values unblocked:
```
scalingAlpha (frontend input fixed in this PR)
```

Existing illegal values possible; no legal values unblocked:
```
regionSize (new frontend constraint added in this PR)
pointAnnotationWidth (new frontend constraint added in this PR)
```

Probably no effect (unless constraints have changed from an older frontend version):
```
scalingGamma
contourSmoothingFactor
contourNumLevels
contourThickness
vectorOverlayThickness
beamWidth
regionLineWidth
regionDashLength
regionType
stopAnimationPlaybackMinutes (also removed duplicate schema preference with different name)
annotationLineWidth
annotationDashLength
catalogDisplayedColumnSize
```

Probably no effect, but these are data sizes which have changed before and are likely to change again frequently. **Question:** should the schema remain more lax than the frontend here, so that we are not unnecessarily constrained? Should we specify a type and step size only? I think that would be sensible, but we should then add manual validation for these in the upgrade function to ensure that values outside the current frontend limits are reset.
```
GPUTileCache
systemTileCache
contourChunkSize
contourControlMapWidth
pvPreviewCubeSizeLimit (and pvPreviewCubeSizeLimitUnit)
```

**Other issues:**

Typos in internal functions (which make it harder to search the code) -- it should be straightforward to search & replace these:
```
pvPreivewCubeSizeLimit
pvPreivewCubeSizeLimitUnit
```

Obsolete preferences (completely unused). **Question**: should these also be removed from the schema? They won't cause validation errors if they are present in old files.
```
username (very old?)
wcsMatchingType (never-used incorrect name for autoWCSMatching, which wasn't in the frontend schema before the cleanup)
```

Unreachable preferences (valid, read from the database and applied if they exist, but cannot be changed through the GUI unless the whole section is reset, or through snippets). **Question:** should these be removed and replaced by constant values? Or are they supposed to be modifiable?
```
textAnnotationLineWidth
colormapHexStart
```

Unreachable preferences which are probably for developer use and should be left alone:
```
statsPanelEnabled
statsPanelMode
```

loveluthien:
I think `username` and `wcsMatchingType` can be removed. Maybe we can keep `textAnnotationLineWidth` and `colormapHexStart`, in case some crazy users set values for them.

confluence:
@loveluthien @kswang1029 @veggiesaurus here's a summary of what I'd like to resolve before this is merged:

* Completely unused properties: remove from the schema
* Unreachable properties: leave alone for now (there are workarounds for changing / resetting them)
* Data sizes: if you don't object, I would like to relax the constraints in the schema (set the type only) and handle the minimum and maximum in the frontend only. That mostly means reverting my schema changes for those properties, but there should also be something in the upgrade function for making sure existing values conform to the current min/max. I suggest clipping the values instead of resetting them if they exceed the bounds; I think that's reasonable UX (maybe alert the user if this happens?).

The other outstanding change I wanted to make was making the backend server error responses include a body with the error message (to match the controller). The frontend should also check if the error response contains data, and if yes, retrieve the message and log it to the console (currently you have to dig into the message object to find the data). But this isn't critical, and I would also like to revise some of the controller messages to make them more specific (e.g. a different message for malformed JSON and invalid JSON), so I don't want to delay the release for it -- I will make separate issues instead.

confluence:
@loveluthien I have made the following changes in the schema PR:
* removed `username` and `wcsMatchingType`
* `percentile` can now be any valid percentage
* `GPUTileCache` and `systemTileCache` must be multiples of 256 and be at least 1024; *no maximum in schema*
* `contourChunkSize` and `contourControlMapWidth` reverted to previous integer ranges and step sizes
* `stopAnimationPlaybackMinutes` can be any number from 5 to 30
* `textAnnotationLineWidth` has the same min and max as other line widths for consistency
* `pvPreviewCubeSizeLimit` can be any positive number; *no maximum in schema*
* `pvPreviewCubeSizeLimitUnit` restored to previous options (including units not currently in use)

This relaxes some schema constraints which were based on options currently available in the UI which may change in the future or are limited for simplicity. In most cases, intermediate values which aren't selectable in the UI but are valid shouldn't cause a problem, and if they're invalid they will be reset.

The question is whether it's necessary to impose a maximum on `GPUTileCache`, `systemTileCache` and `pvPreviewCubeSizeLimit` for security rather than performance reasons when reading these properties from the database (I will discuss this with @veggiesaurus tomorrow). Since the preference can be set at runtime through scripting anyway, maybe not.

Currently it's complicated to set the bounds on `pvPreviewCubeSizeLimit` because the unit is set separately. In future I would suggest eliminating the separate unit property and performing unit conversion only in the dialog.

loveluthien:
Looks good. It's nice to have a max number for `GPUTileCache`, `systemTileCache`, and `pvPreviewCubeSizeLimit` for security reasons. But it seems not trivial to deal with the variation between PCs or servers.

confluence:
@loveluthien I discussed this with @veggiesaurus and we've decided that it's unnecessary to correct large data limits on load (and agreed on some further relaxation of the schema constraints). So I will make one final set of changes to the schema, and I don't think you will need to make any further changes to the frontend code.

I will make a separate issue for my unrelated schema suggestions.

confluence:
@loveluthien the latest schema and backend PRs are now ready for testing.

src/services/ApiService.ts
```diff
@@ -185,58 +185,60 @@ export class ApiService {
...
+            const color = preferences[astColorKey] in ASTColors ? ASTColors[preferences[astColorKey]] : "blue";
+            preferences[astColorKey] = `auto-${color}`;
+            this.setPreference(astColorKey, preferences[astColorKey]);
+
+            preferences["version"] = 2;
+            this.setPreference("version", 2);
+        }
+
+        // Normalize case of wcsType value, which may have been saved incorrectly in existing preferences
+        let key = PreferenceKeys.WCS_OVERLAY_WCS_TYPE;
```
veggiesaurus: shouldn't this be const?

src/services/ApiService.ts
```diff
@@ -185,58 +185,60 @@ export class ApiService {
...
+                        // Remove the invalid preference from the preferences object
+                        delete preferences[errorKey];
+                        this.clearPreferences([errorKey]);
+                        deletedKeys.push(errorKey);
                     }
                 }
+
+                if (deletedKeys.length > 0) {
+                    // Show an alert to the user about the deleted preferences
+                    const appStore = AppStore.Instance;
```
veggiesaurus: if we're just using AppStore once we don't need to create a temp variable

veggiesaurus (changes requested):

src/services/ApiService.ts
loveluthien: Thanks. They are fixed.

veggiesaurus (approved):

confluence:
@loveluthien please update the schemas submodule in this PR to the latest version of the branch.

loveluthien:
@confluence. done.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -55,19 +55,24 @@ export class PreferenceDialogComponent extends React.Component {
     };
 
     @computed get pvPreviewCubeSizeMaxValue(): number {
-        if (PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit === MemoryUnit.TB) {
+        if (this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.TB) {
```
confluence: I suggest redefining the `MemoryUnit` enum to make this all much simpler:
```
export enum MemoryUnit {
    TB = 1e12,
    GB = 1e9,
    MB = 1e6,
    kB = 1e3,
    B = 1,
}
```

Those values are aligned with the conversion factors used in `PvGeneratorComponent.tsx` in `formatBitValue` and `getBitValueFromFormatted`. In the input UI you can use `value / (MemoryUnit[string_unit_here] / 1e9` for the conversion from GB to the selectable units.

However, the only reason those functions exist in the pv component is to convert units for the cube size comparison. If the limit is always going to be stored in GB, you can completely remove all of that, and have `estimatedCubeSize` return the size in GB, which you can compare directly to the preference value. You can then express the MemoryUnit values in terms of GB for convenience:
```
enum MemoryUnit {
    TB = 1e3,
    GB = 1,
    MB = 1e-3,
    kB = 1e-6,
    B = 1e-9,
}
```

...and then you can use `value / MemoryUnit[string_unit_here]` for the conversion to GB.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -55,19 +55,24 @@ export class PreferenceDialogComponent extends React.Component {
...
+        } else if (this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.kB) {
             return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e3;
         } else {
             return PV_PREVIEW_CUBE_SIZE_LIMIT;
         }
     }
 
+    @computed get showedPvPreviewCubeSizeLimit(): number {
+        const preference = PreferenceStore.Instance;
+        return this.showedPvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? preference.pvPreviewCubeSizeLimit : preference.pvPreviewCubeSizeLimit * 1000;
```
confluence: I recommend normalizing the preferences in the upgrade function: if the unit pref exists, if it isn't GB, convert the limit value to GB and save, and also clear the unit pref. Don't save or use the unit pref anywhere else.

We can't remove it from the preferences store yet, because `clearPreferences` expects valid `PreferenceKeys`, so I suggest that we just mark it as deprecated in the schema and stop using it, and only remove it fully in a few releases' time.

confluence (changes requested):

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
loveluthien: It is a good idea, but there are other files using `MemoryUnit`, so I would rather not touch it at this moment.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
loveluthien: okay, I did it in ApiService.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -55,17 +63,16 @@ export class PreferenceDialogComponent extends React.Component {
...
-        } else if (PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit === MemoryUnit.GB) {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e9;
-        } else if (PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit === MemoryUnit.MB) {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e6;
-        } else if (PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit === MemoryUnit.kB) {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e3;
-        } else {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT;
-        }
+        return this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? PV_PREVIEW_CUBE_SIZE_LIMIT : PV_PREVIEW_CUBE_SIZE_LIMIT / ConvertToGB.MB;
```
confluence: You don't need a conditional; you can use `PV_PREVIEW_CUBE_SIZE_LIMIT / ConvertToGB[this.pvPreviewCubeSizeLimitUnit]`. This was the point of refactoring the enum.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -43,9 +43,17 @@ export enum MemoryUnit {
     B = "B"
 }
 
+export enum ConvertToGB {
```
confluence: You don't need the original MemoryUnit enum: it just maps string names to the same names. All the code that uses that enum can either be removed entirely or refactored to use the new ConvertToGB enum.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -55,17 +63,16 @@ export class PreferenceDialogComponent extends React.Component {
...
-        } else if (PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit === MemoryUnit.kB) {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT / 1e3;
-        } else {
-            return PV_PREVIEW_CUBE_SIZE_LIMIT;
-        }
+        return this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? PV_PREVIEW_CUBE_SIZE_LIMIT : PV_PREVIEW_CUBE_SIZE_LIMIT / ConvertToGB.MB;
+    }
+
+    @computed get pvPreviewCubeSizeMinValue(): number {
+        return this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 0.1 : 128;
```
confluence: 0.1GB is not the same as 128MB. You should use the same conversion formula on the GB minimum here as in the computed maximum. Let's pick 0.1 for simplicity, since these are metric units.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -55,17 +63,16 @@ export class PreferenceDialogComponent extends React.Component {
...
+    }
+
+    @computed get showedPvPreviewCubeSizeLimit(): number {
+        const preference = PreferenceStore.Instance;
+        return this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? preference.pvPreviewCubeSizeLimit : preference.pvPreviewCubeSizeLimit / ConvertToGB.MB;
```
confluence: Again, use the same conversion formula here on the preference.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -99,9 +106,17 @@ export class PreferenceDialogComponent extends React.Component {
     }, 100);
 
     @action private handlePvPreviewCubeSizeUnitChange = _.throttle(unit => {
-        PreferenceStore.Instance.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT_UNIT, unit);
+        this.pvPreviewCubeSizeLimitUnit = unit;
+    }, 100);
+
+    @action private handlePvPreviewCubeSizeChange = _.throttle((size, unit) => {
+        const storedSize = unit === MemoryUnit.GB ? size : size * ConvertToGB.MB; // Convert to GB if necessary
```
confluence: Again, use the same formula here.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -99,9 +106,17 @@ export class PreferenceDialogComponent extends React.Component {
...
+        PreferenceStore.Instance.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT, storedSize);
     }, 100);
 
+    // variable for showing preview cube size unit in the dialog
+    @observable private pvPreviewCubeSizeLimitUnit = PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit;
```
confluence: This preference should not be used anywhere (except the preference upgrade function).

Unless you intend to have a preference for the unit to use in the actual preference dialog (which was not my impression from the discussion), just set this to GB by default.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -806,18 +835,18 @@ export class PreferenceDialogComponent extends React.Component {
...
-                            min={1e-12}
+                            min={this.pvPreviewCubeSizeMinValue}
                             max={this.pvPreviewCubeSizeMaxValue}
-                            value={preference.pvPreivewCubeSizeLimit}
-                            majorStepSize={1}
-                            stepSize={1}
-                            onValueChange={value => preference.setPreference(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT, value)}
+                            value={this.showedPvPreviewCubeSizeLimit}
+                            majorStepSize={this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 2 : 2048}
+                            stepSize={this.pvPreviewCubeSizeLimitUnit === MemoryUnit.GB ? 0.1 : 128}
```
confluence: 1. Use the same conversion formula for both step sizes
2. Why have a major step size of 2GB if 2GB is the current maximum?

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -806,18 +835,18 @@ export class PreferenceDialogComponent extends React.Component {
...
+                            onValueChange={value => this.handlePvPreviewCubeSizeChange(value, this.pvPreviewCubeSizeLimitUnit)}
                         />
-                        <HTMLSelect value={preference.pvPreivewCubeSizeLimitUnit} onChange={ev => this.handlePvPreviewCubeSizeUnitChange(ev.target.value)}>
-                            <option key={0} value={"MB"}>
+                        <HTMLSelect value={this.pvPreviewCubeSizeLimitUnit} onChange={ev => this.handlePvPreviewCubeSizeUnitChange(ev.target.value)}>
+                            <option key={0} value={MemoryUnit.MB}>
                                 MB
                             </option>
-                            <option key={1} value={"GB"}>
+                            <option key={1} value={MemoryUnit.GB}>
```
confluence: Eliminate the use of the old MemoryUnit enum here and just use the strings. If we were including all the units, we could iterate over the enum keys, but for only two options, hardcoding is fine.

src/components/PvGenerator/PvGeneratorComponent.tsx
```diff
@@ -51,20 +51,6 @@ export class PvGeneratorComponent extends React.Component<WidgetProps> {
         return {value, unit, bitValue: bitValue};
     };
 
-    public static getBitValueFromFormatted = (value: number, unit: string): number => {
```
confluence, line 54: You can also completely remove `formatBitValue`. It only exists for the now-obsolete comparison of values with different units.

src/components/PvGenerator/PvGeneratorComponent.tsx
```diff
@@ -157,7 +143,7 @@ export class PvGeneratorComponent extends React.Component<WidgetProps> {
     }
 
     @computed get isCubeSizeBelowLimit(): boolean {
-        return this.estimatedCubeSize?.bitValue <= PvGeneratorComponent.getBitValueFromFormatted(PreferenceStore.Instance.pvPreivewCubeSizeLimit, PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit);
+        return this.estimatedCubeSize?.bitValue <= PreferenceStore.Instance.pvPreviewCubeSizeLimit * 1e9; // Convert GB to bits
```
confluence: This is unnecessary. Instead of continuing to return an object from `estimatedCubeSize`, return only the size in GB (or undefined if there's no frame). Refactor the form in this component where it uses this size. Then compare directly to `pvPreviewCubeSizeLimit`.

src/components/PvGenerator/PvGeneratorComponent.tsx
```diff
@@ -302,7 +288,7 @@ export class PvGeneratorComponent extends React.Component<WidgetProps> {
                         <br />
                         1. Line region is selected.
                         <br />
-                        2. Preview cube size is less than the threshold ({`${PreferenceStore.Instance.pvPreivewCubeSizeLimit} ${PreferenceStore.Instance.pvPreivewCubeSizeLimitUnit}`}).
+                        2. Preview cube size is less than the threshold ({`${PreferenceStore.Instance.pvPreviewCubeSizeLimit} ${PreferenceStore.Instance.pvPreviewCubeSizeLimitUnit}`}).
```
confluence: You should not use the unit preference; this should be hardcoded to GB.

src/services/ApiService.ts
```diff
@@ -185,58 +185,69 @@ export class ApiService {
...
+        const key = PreferenceKeys.WCS_OVERLAY_WCS_TYPE;
+        if (/[A-Z]/.test(preferences[key])) {
+            preferences[key] = preferences[key].toLowerCase();
+            this.setPreference(key, preferences[key]);
+        }
+
+        // This is to ensure consistency in the unit used for the preview cube size limit
+        const cubeSizeUnitKey = PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT_UNIT;
+        const cubeSizeKey = PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT;
+        if (preferences[cubeSizeUnitKey] === "MB") {
```
confluence: * Other units were previously valid, so this should convert from *any* existing unit. Because this is done before validation, there should also be handling for possible invalid units or values.
* The unit preference should always be cleared from the database, and also deleted from the local preferences object.

Something like:

```typescript
const memoryUnits = ["TB", "GB", "MB", "kB", "B"];

if (memoryUnits.includes(preferences[cubeSizeUnitKey]) && typeof preferences[cubeSizeKey] === "number") {
  let gbSize = preferences[cubeSizeKey] * ConvertToGB[preferences[cubeSizeUnitKey]];

  if (gbSize != preferences[cubeSizeKey]) {
    preferences[cubeSizeKey] = gbSize;
    this.setPreference(cubeSizeKey, preferences[cubeSizeKey]);
  }

  delete preferences[cubeSizeUnitKey];
  this.clearPreferences([cubeSizeUnitKey]);
}
```

* Because numeric enums are compiled to bidirectional mappings, I created a local array of just the string keys here.
* If the size value is not a number, it will be reset in the validation step.
* After this point, the unit preference should not be used anywhere, and the size preference is guaranteed to be in GB.

src/services/ApiService.ts
```diff
@@ -185,58 +185,69 @@ export class ApiService {
...
+                red,
+                forest,
+                blue,
+                turquoise,
+                violet,
+                gold,
+                gray
+            }
+            const astColorKey = PreferenceKeys.WCS_OVERLAY_AST_COLOR;
+            const color = preferences[astColorKey] in ASTColors ? ASTColors[preferences[astColorKey]] : "blue";
```
confluence: I've realised that `in` is not sufficient here -- you also have to check that `preferences[astColorKey]` is a number. Otherwise a string value which is a valid enum key will result in the reverse lookup (and we'll end up with `auto-3`, etc.).

We should also not rely on implicit numbering in the enum, because the numeric values have meaning -- it would be safer to add them explicitly.

src/stores/PreferenceStore/PreferenceStore.ts
```diff
@@ -630,11 +629,11 @@ export class PreferenceStore {
         return this.preferences.get(PreferenceKeys.PERFORMANCE_STOP_ANIMATION_PLAYBACK_MINUTES) ?? DEFAULTS.PERFORMANCE.stopAnimationPlaybackMinutes;
     }
 
-    @computed get pvPreivewCubeSizeLimit(): number {
+    @computed get pvPreviewCubeSizeLimit(): number {
         return this.preferences.get(PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT) ?? DEFAULTS.PERFORMANCE.pvPreviewCubeSizeLimit;
     }
 
-    @computed get pvPreivewCubeSizeLimitUnit(): string {
+    @computed get pvPreviewCubeSizeLimitUnit(): string {
```
confluence: Since this is deprecated, this preference should never be accessed. I suggest removing this computed property entirely to ensure this. It should probably also be removed from the defaults.

If you hardcode the string name, you could actually also remove it from the preference keys, but perhaps it would be safer to keep it there for now and add a comment marking it as deprecated.

confluence (changes requested):

confluence:
@loveluthien I have changed the minimum in the schema, so this PR needs another submodule bump.

src/services/ApiService.ts
loveluthien: I didn't figure out why I can't use `ConvertToGB` which in PreferenceDialogComponent, so I created the same enum temporarily.

kswang1029 (approved):
looks good.

src/services/ApiService.ts
```diff
@@ -185,59 +186,82 @@ export class ApiService {
...
+        const cubeSizeUnitKey = PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT_UNIT;
+        const cubeSizeKey = PreferenceKeys.PERFORMANCE_PV_PREVIEW_CUBE_SIZE_LIMIT;
+
+        enum ConvertToGB {
+            TB = 1e3,
+            GB = 1,
+            MB = 1e-3,
+            kB = 1e-6,
+            B = 1e-9
+        }
```
confluence: What error were you getting when you tried to use the existing enum? Were you definitely importing it?

src/services/ApiService.ts
```diff
@@ -185,59 +186,82 @@ export class ApiService {
...
+        if (ConvertToGB[preferences[cubeSizeUnitKey]] && typeof preferences[cubeSizeKey] === "number") {
```
confluence: Here you need to check that `ConvertToGB[preferences[cubeSizeUnitKey]]` is a number, not that `preferences[cubeSizeKey]` is a number. The size will always be a number; the check is to limit recognised old units to strings (in case the user has something weird in their preferences file).

I suggest adding a temporary variable for this to avoid repeating the verbose string:
```
let conversionFactor = ConvertToGB[preferences[cubeSizeUnitKey]];
if (typeof conversionFactor === "number") {
...
```

src/services/ApiService.ts
```diff
@@ -185,59 +186,82 @@ export class ApiService {
...
+            let gbSize = preferences[cubeSizeKey] * ConvertToGB[preferences[cubeSizeUnitKey] as string];
```
confluence: Adding ` as string` here probably doesn't help: if it's not a string, the result of the lookup won't be valid, and you'll get some kind of error or undefined value.

(Checking directly if the conversion factor is a number is the correct thing to do -- it will also handle the user e.g. mistyping a unit or using a unit we don't recognise.)

src/services/ApiService.ts
```diff
@@ -185,59 +186,82 @@ export class ApiService {
...
+            if (gbSize !== preferences[cubeSizeKey]) {
+                preferences[cubeSizeKey] = gbSize;
+                this.setPreference(cubeSizeKey, preferences[cubeSizeKey]);
             }
-            preferences.version = 2;
-            this.setPreferences(preferences);
         }
```
confluence, line 259: We should also add an `else if (cubeSizeUnitKey in preferences)` here to check if the unit key is in the preferences but is *not* a recognised unit -- and delete the *size* preference with a message if that is the case. If the user put in a weird unit, then we don't know how to interpret the size, and we need to reset it.

We could also set it to an invalid value (like -1) here, and let the validation step take care of the deletion and message.

(The reason that we can't just leave it for the validation step as-is is that the size value in isolation may be valid -- we don't want to delete the invalid unit and then treat the size as valid GB when we have no idea what it's supposed to be.)

confluence (changes requested):

src/services/ApiService.ts
confluence: I don't know what is happening with GitHub's anchoring of these comments! I mean that the `if (typeof conversionFactor === "number")` should have an `else if (cubeSizeUnitKey in preferences)`.

src/services/ApiService.ts
loveluthien: `ReferenceError: Cannot access uninitialized variable.` in the console when starting CARTA. My guess is that `ConvertToGB` in the `PreferenceDialogComponent` is not yet called when processing ApiService.

confluence: Maybe move it into `models`? e.g. `models/DataSize/DataSize.ts`?

loveluthien: Good idea. That works.

src/services/ApiService.ts
```diff
@@ -184,59 +185,79 @@ export class ApiService {
...
+                        // Remove the invalid preference from the preferences object
+                        delete preferences[errorKey];
+                        this.clearPreferences([errorKey]);
+                        deletedKeys.push(errorKey);
                     }
                 }
+
+                if (deletedKeys.length > 0) {
+                    // Show an alert to the user about the deleted preferences
+                    AppStore.Instance.alertStore.showAlert(`Deleted invalid preferences: ${deletedKeys.join(", ")}`);
```
confluence: Minor point: instead of "deleted" here I would say: `Invalid preferences reset to defaults: ...`

src/services/ApiService.ts
```diff
@@ -184,59 +185,79 @@ export class ApiService {
...
+        const conversionFactor = ConvertToGB[preferences[cubeSizeUnitKey]];
+        if (typeof conversionFactor === "number") {
+            let gbSize = preferences[cubeSizeKey] * conversionFactor;
+            if (gbSize !== preferences[cubeSizeKey]) {
+                preferences[cubeSizeKey] = gbSize;
+                this.setPreference(cubeSizeKey, preferences[cubeSizeKey]);
+            }
+        } else if (cubeSizeUnitKey in preferences) {
+            delete preferences[cubeSizeUnitKey];
+            this.clearPreferences([cubeSizeUnitKey]);
```
confluence: The unit pref should *always* be deleted (even if we were able to do the conversion) -- these two lines should go back outside the conditional.

confluence (changes requested):

src/services/ApiService.ts
loveluthien: okay

confluence (approved):
I think this is all good now.

PR #2565, opened 2025-07-14
strict null check migration 2

loveluthien:
**Description**

This PR is a sequential work of https://github.com/CARTAvis/carta-frontend/issues/2343 following this https://github.com/CARTAvis/carta-frontend/pull/2344. This PR migrates all `Stores`.

This PR closes #2592.

**Checklist**

For linked issues (if there are):
- [ ] assignee and label added
- [ ] GitHub Project estimate added

For the pull request:
- [ ] reviewers and assignee added
- [ ] GitHub Project estimate added
- [ ] changelog updated / no changelog update needed
- [ ] unit test added (for functions with no dependenies)
- [ ] API documentation added (for public variables and methods in stores)

For dependencies:
- [x] e2e test passing / corresponding fix added / new e2e test created
- [ ] protobuf version bumped / no protobuf version bumped needed
- [ ] protobuf updated to the latest dev commit / no protobuf update needed
- [ ] corresponding ICD test fix added (`BackendService` changed) / no ICD test fix needed (`BackendService` unchanged)
- [ ] user manual prepared (for large new features)

PR #2577, opened 2025-08-13
strict null check migration 3

### LINKED ISSUE #2343: Enable strictNullChecks compiler option

loveluthien:
**Description**

This PR is the final part of https://github.com/CARTAvis/carta-frontend/issues/2343 and closes #2343, following #2344 and #2565. This PR migrates the rest of the files. 

**Checklist**

For the pull request:

For dependencies:


src/components/Dialogs/StokesDialog/StokesDialogComponent.tsx
```diff
@@ -121,19 +126,23 @@ export class StokesDialogComponent extends React.Component {
                 name={"Polarization"}
                 cellRenderer={rowIndex => {
                     const file = files[rowIndex];
+                    const stoke = this.stokes.get(file);
```
kswang1029: better rename as stokes (it is a name to memorize [George Gabriel Stokes](https://en.wikipedia.org/wiki/George_Gabriel_Stokes))

src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
```diff
@@ -553,14 +708,14 @@ export class FileBrowserDialogComponent extends React.Component {
...
+        return <InputGroup autoFocus={true} placeholder="Enter file name" value={fileBrowserStore.saveFilename || ""} onChange={this.handleSaveFileNameChanged} rightElement={fileTypeMenu} />;
     }
 
     private renderOpenFilenameInput(browserMode: BrowserMode) {
         const preferenceStore = PreferenceStore.Instance;
 
-        let filterName: string;
-        let filterDescription: string;
+        let filterName: string = "Unknown";
+        let filterDescription: string = "Filter by filename";
```
kswang1029, line 711: should this be a const?

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -402,34 +454,60 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
 
     private updateHeaderTableColumnSize = (index: number, size: number) => {
         const widgetsStore = this.widgetStore;
-        if (widgetsStore.headerTableColumnWidts) {
-            widgetsStore.headerTableColumnWidts[index] = size;
+        if (widgetsStore) {
+            // Ensure the array exists and has the correct length (5 columns)
+            if (!widgetsStore.headerTableColumnWidts) {
```
kswang1029: please fix typo in variable naming as well

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -402,34 +454,60 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
...
+                widgetsStore.headerTableColumnWidts = new Array(5).fill(undefined);
```
kswang1029: should 5 be replaced by expectedColumnCount to avoid using a magic number directly?

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -402,34 +454,60 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
...
+            } else if (widgetsStore.headerTableColumnWidts.length !== 5) {
```
kswang1029: should 5 be replaced by expectedColumnCount to avoid using a magic number directly?

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -402,34 +454,60 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
...
+                // Resize array to match expected column count
+                const newArray = new Array(5).fill(undefined);
```
kswang1029: should 5 be replaced by expectedColumnCount to avoid using a magic number directly?

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -402,34 +454,60 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
...
+                for (let i = 0; i < Math.min(widgetsStore.headerTableColumnWidts.length, 5); i++) {
```
kswang1029: should 5 be replaced by expectedColumnCount to avoid using a magic number directly?

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -656,12 +787,24 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
...
+        const expectedColumnCount = profileStore.displayedColumnHeaders.length;
+        let tableColumnWidths = profileStore.tableColumnWidths;
+        if (!tableColumnWidths || tableColumnWidths.length !== expectedColumnCount) {
+            tableColumnWidths = new Array(expectedColumnCount).fill(undefined);
+        }
+
+        // Filter out undefined values to match expected Array<number> type
+        const validColumnWidths = tableColumnWidths.filter((w): w is number => w != null || w !== undefined);
+
+        console.log("qq", profileStore.sortingInfo.columnName);
```
kswang1029: what does "qq" mean?

src/components/Dialogs/RegionDialog/RectangularRegionForm/RectangularRegionForm.tsx
```diff
@@ -101,23 +105,29 @@ export class RectangularRegionForm extends React.Component<{region: RegionStore;
 
     private handleCenterWCSXChange = (wcsString: string): boolean => {
         if (isWCSStringFormatValid(wcsString, AppStore.Instance.overlaySettings.numbers.formatTypeX)) {
-            const newPoint = getPixelValueFromWCS(this.props.wcsInfo, {x: wcsString, y: this.centerWCS.y});
-            const existingValue = this.props.region.center.x;
-            if (isFinite(newPoint?.x) && !closeTo(newPoint.x, existingValue, RectangularRegionForm.REGION_PIXEL_EPS)) {
-                this.props.region.setCenter(newPoint);
-                return true;
+            const centerWCS = this.centerWCS;
+            if (centerWCS) {
```
kswang1029: can centerWCS be 0?

src/components/ImageView/RegionView/InvariantShapes.tsx
```diff
@@ -130,7 +135,10 @@ export const Anchor = (props: AnchorProps) => {
     };
 
     const handleCircleDraw = (ctx, shape) => {
-        const inverseScale = 1 / shape.getStage().scaleX();
+        const stage = shape.getStage();
+        if (!stage) return;
```
kswang1029: code style inconsistency in Line:19 and Line:139. Have you run prettier?

src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
loveluthien: please skip all this kind of issue. using `let` for the unchanged value will be picked up in another PR.

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
loveluthien: done

src/components/Dialogs/RegionDialog/RectangularRegionForm/RectangularRegionForm.tsx
loveluthien: The type of `centerWCS` is not a number; it can not be 0, but I corrected this kind to condition in other places

src/components/ImageView/RegionView/InvariantShapes.tsx
loveluthien: `prettier` cannot pick up this

kswang1029 (approved):

kswang1029:
@veggiesaurus could you give this a quick review before merging? It involves quite some changes that e2e may not be able to catch if there is regression. We will need to rely on beta tests and "time" to ensure it is stable.

src/components/Animator/AnimatorComponent.tsx
```diff
@@ -224,15 +224,15 @@ export class AnimatorComponent extends React.Component<WidgetProps> {
                     {!hideSliders && (
                         <React.Fragment>
                             <Slider value={imageIndex} min={0} max={numImages - 1} showTrackFill={false} labelValues={imageTick} labelPrecision={0} onChange={this.onImageChanged} disabled={appStore.animatorStore.animationActive} />
-                            <div className="slider-info">{appStore.activeImage.store.filename}</div>
+                            <div className="slider-info">{appStore.activeImage?.store.filename ?? ""}</div>
```
veggiesaurus: makes more sense to move the check up along with `!hideSliders`

src/components/CatalogOverlay/CatalogOverlayPlotSettingsPanelComponent/CatalogOverlayPlotSettingsPanelComponent.tsx
```diff
@@ -105,44 +108,54 @@ export class CatalogOverlayPlotSettingsPanelComponent extends React.Component<Wi
...
             catalogFileItems.push(value);
         });
         this.catalogFileNames = CatalogStore.Instance.getCatalogFileNames(catalogFileIds);
-        const fileName = this.catalogFileNames.get(this.catalogFileId);
+        const catalogFileId = this.catalogFileId;
+        const fileName = catalogFileId !== undefined ? this.catalogFileNames.get(catalogFileId) : undefined;
         let activeFileName = "";
-        if (fileName !== undefined) {
-            activeFileName = `${this.catalogFileId}: ${fileName}`;
+        if (fileName !== undefined && catalogFileId !== undefined) {
```
veggiesaurus, line 157: we're a bit inconsistent with our use of `catalogFileId !== undefined` versus just `catalogFileId` (line 75)

src/components/CatalogOverlay/CatalogOverlayPlotSettingsPanelComponent/CatalogOverlayPlotSettingsPanelComponent.tsx
```diff
@@ -820,15 +836,18 @@ export class CatalogOverlayPlotSettingsPanelComponent extends React.Component<Wi
...
-        this.widgetStore.setSizeAxisTab(Number.parseInt(newTabId.toString()));
+        this.widgetStore?.setSizeAxisTab(Number.parseInt(newTabId.toString()));
     };
 
     private getCatalogShape = (shape: CatalogOverlayShape) => {
         const widgetStore = this.widgetStore;
+        if (!widgetStore) {
+            return <Icon icon="circle" color={Colors.TURQUOISE3} />;
+        }
         let color = widgetStore.catalogColor;
```
veggiesaurus: would make sense to just do `const color = widgetStore?.catalogColor ?? Colors.TURQUOISE3;`

src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
```diff
@@ -311,50 +354,61 @@ export class CatalogPlotComponent extends React.Component<WidgetProps> {
     @computed get numBinsX(): number {
         const widgetStore = this.widgetStore;
         const profileStore = this.profileStore;
+        if (!widgetStore || !profileStore || !widgetStore.xColumnName) {
+            return 10; // default fallback
```
veggiesaurus: store this as a constant somewhere. e.g. `DEFAULT_NUM_BINS`

src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
```diff
@@ -311,50 +354,61 @@ export class CatalogPlotComponent extends React.Component<WidgetProps> {
...
+        }
         const coords = profileStore.get1DPlotData(widgetStore.xColumnName);
-        const nBinx = Math.ceil(Math.sqrt(coords.wcsData?.length));
+        const nBinx = Math.ceil(Math.sqrt(coords.wcsData?.length ?? 100));
```
veggiesaurus: `coords.wcsData?.length ? Math.ceil(Math.sqrt(coords.wcsData.length)): DEFAULT_NUM_BINS` would be more understandable

src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
```diff
@@ -512,28 +596,33 @@ export class CatalogPlotComponent extends React.Component<WidgetProps> {
...
+        const widgetStore = this.widgetStore;
+        if (!widgetStore) {
+            return;
+        }
+
+        const inDragmode = selectionMode.includes(widgetStore.dragmode);
         const profileStore = this.profileStore;
         const catalogWidgetStore = this.catalogWidgetStore;
-        if (event?.points?.length > 0 && inDragmode) {
+        if (event?.points?.length > 0 && inDragmode && profileStore && catalogWidgetStore) {
```
veggiesaurus, line 619: makes more sense to move these checks up to line 610 as well

`if (!widgetStore || !profileStore || !catalogWidgetStore)`

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -666,12 +747,23 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
...
+
+        // Ensure columnWidths matches the number of displayed columns
+        const expectedColumnCount = profileStore.displayedColumnHeaders.length;
+        let tableColumnWidths = profileStore.tableColumnWidths;
+        if (!tableColumnWidths || tableColumnWidths.length !== expectedColumnCount) {
+            tableColumnWidths = new Array(expectedColumnCount).fill(undefined);
+        }
+
+        // Filter out undefined values to match expected Array<number> type
+        const validColumnWidths = tableColumnWidths.filter((w): w is number => w != null || w !== undefined);
```
veggiesaurus: can't we use `isFinite` here?

veggiesaurus (changes requested):
reviewed up to and including `ContourDialogComponent`, will do a next iteration soon

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
loveluthien: after testing, null is acceptable here, so we should just filter out undefined values.

src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
loveluthien: I found another way to deal with this checking.

src/stores/Frame/FrameStore.ts
```diff
@@ -683,15 +683,15 @@ export class FrameStore {
             }
             // convert frequency to velocity
             const velocityVal = this.astSpectralTransform(SpectralType.VRAD, SpectralUnit.KMS, this.spectralSystem, freqVal);
-            if (velocityVal !== undefined && isFinite(velocityVal)) {
+            if (velocityVal && isFinite(velocityVal)) {
```
izkgao: If `velocityVal` (also for `freqGHz` and `velocityKMS`) is `0`, it will not enter the `if` block.

src/stores/Frame/FrameStore.ts
```diff
@@ -1644,7 +1644,7 @@ export class FrameStore {
 
     private convertSpectral = (values: Array<number>): Array<number> => {
         const N = values?.length;
-        if (!N || !this.spectralFrame || !this.spectralType || !this.spectralUnit || !this.spectralSystem) {
+        if (!N || !this.spectralFrame) {
```
izkgao, line 1647: Just curious about the reason for not keeping the other three.

src/components/LayerList/LayerListComponent.tsx
```diff
@@ -358,23 +363,29 @@ export class LayerListComponent extends React.Component<WidgetProps> {
     private restFreqShortCutOnClick = (selectedFrameIndex: number) => {
         const widgetsStore = AppStore.Instance.widgetsStore;
         const layerListWidget = widgetsStore.layerListWidgets?.get(this.props.id);
-        widgetsStore.createFloatingSettingsWidget(LayerListComponent.WIDGET_CONFIG.title, this.props.id, LayerListComponent.WIDGET_CONFIG.type);
+        const title = LayerListComponent.WIDGET_CONFIG.title;
+        if (title) {
```
izkgao, line 367: Do we need to check `LayerListComponent.WIDGET_CONFIG.type` as well?

src/components/RenderConfig/RenderConfigComponent.tsx
```diff
@@ -58,35 +58,43 @@ export class RenderConfigComponent extends React.Component<WidgetProps> {
         return new RenderConfigWidgetStore();
     }
 
-    @computed get plotData(): {values: Array<Point2D>; xMin: number; xMax: number; yMin: number; yMax: number} {
+    @computed get plotData(): {values: Array<Point2D>; xMin: number; xMax: number; yMin: number; yMax: number} | null {
         const frame = AppStore.Instance.activeFrame;
-        if (frame && frame.renderConfig.histogram && frame.renderConfig.histogram.bins && frame.renderConfig.histogram.bins.length) {
+        if (frame) {
             const histogram = frame.renderConfig.histogram;
+            const bins = histogram.bins;
```
izkgao, line 68: Adding this before `const bin = historgram.bins;` can avoid `[mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[Autorun@82]' TypeError: Cannot read properties of undefined (reading 'bins')` when opening a new image.
```
if (!histogram) {
    return null;
}
```

src/components/SpectralLineQuery/SpectralLineQueryComponent.tsx
```diff
@@ -382,13 +387,16 @@ export class SpectralLineQueryComponent extends React.Component<WidgetProps> {
...
             updateSortRequest: this.updateSortRequest,
-            sortingInfo: widgetStore.sortingInfo,
+            sortingInfo: {
+                columnName: widgetStore.sortingInfo.columnName ?? "",
+                sortingType: widgetStore.sortingInfo.sortingType
+            },
             disableSort: false,
             updateColumnFilter: widgetStore.setColumnFilter,
-            columnWidths: widgetStore.resultTableColumnWidths,
+            columnWidths: widgetStore.resultTableColumnWidths?.filter((width): width is number => width != null || width !== undefined),
```
izkgao: `width !== undefined` should be removed (see https://github.com/CARTAvis/carta-frontend/pull/2590#discussion_r2625619259).

An element in `resultTableColumnWidths` can be (number | null | undefined). So we only need `width != null` here.

src/components/CatalogOverlay/CatalogPlotComponent/CatalogPlotComponent.tsx
```diff
@@ -402,51 +464,71 @@ export class CatalogPlotComponent extends React.Component<WidgetProps> {
...
+            const initBorder = this.initScatterBorder;
+            if (initBorder) {
+                widgetsStore.setScatterborder(initBorder);
+            }
         } else {
-            widgetsStore.setHistogramXBorder(this.initHistogramXBorder);
+            const initBorder = this.initHistogramXBorder;
+            if (initBorder) {
+                widgetsStore?.setHistogramXBorder(initBorder);
+            }
```
izkgao:
```
if (!widgetStore) {
    return;
}
```
We also need this here, or if `widgetsStore?.plotType` is falsy it jumps to the `else` block.

src/components/Shared/Tables/FilterableTableComponent.tsx
izkgao, line 191: I am not sure if this is a relevant issue. I loaded an image and a catalog, and this catalog contains `NaN`. After toggling "Display" of a column, console showed ```Warning: Received NaN for the `children` attribute. If this is expected, cast the value to a string.```. So I added this block before `let cell = cellContext;` and no warning happened.

```
if (typeof cellContext === "number" && !Number.isFinite(cellContext)) {
    cellContext = String(cellContext);
}
```

The reason that I use `String` instead of `.toString()` is to avoid errors in the cases that `cellContext` is `null` or `undefined`.

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -666,12 +747,23 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
         }
 
         const catalogTable = this.catalogDataInfo;
+
+        // Ensure columnWidths matches the number of displayed columns
+        const expectedColumnCount = profileStore.displayedColumnHeaders.length;
+        let tableColumnWidths = profileStore.tableColumnWidths;
+        if (!tableColumnWidths || tableColumnWidths.length !== expectedColumnCount) {
+            tableColumnWidths = new Array(expectedColumnCount).fill(undefined);
```
izkgao: Any `undefined` in `tableColumnWidths` will be filtered in L759. We can just create an empty array here.

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -666,12 +747,23 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
...
+        const validColumnWidths = tableColumnWidths.filter((w): w is number => w !== undefined);
+
         const dataTableProps: FilterableTableComponentProps = {
-            dataset: catalogTable.dataset,
+            dataset: catalogTable.dataset ?? new Map(),
             filter: profileStore.catalogControlHeader,
             columnHeaders: profileStore.displayedColumnHeaders,
             numVisibleRows: catalogTable.numVisibleRows,
-            columnWidths: profileStore.tableColumnWidths,
+            columnWidths: validColumnWidths.length === expectedColumnCount ? validColumnWidths : undefined,
```
izkgao, line 768: **L754-756:** Any `undefined` in `tableColumnWidths` will be filtered in L759. We can just create an empty array here.

**L759:** An element in `tableColumnWidths` can be `(number | null | undefined)`, and I also see `tableColumnWidths` being something like `[null, null, null]` when I printed it out in console. So, to match the expected `Array<number>` type, we should use `w != null` instead of `w !== undefined` here.

The current code is a bit complicated. I suggest that we combine these lines into
```
const tableColumnWidths = profileStore.tableColumnWidths;
const columnWidths =
    tableColumnWidths.length === expectedColumnCount &&
    tableColumnWidths.every((w): w is number => w != null)
        ? tableColumnWidths
        : undefined;
```

src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
```diff
@@ -103,7 +107,8 @@ export class FileBrowserDialogComponent extends React.Component {
             }
             appStore.setLoadingMultipleFiles(false);
         } else {
-            await this.loadFile({fileInfo: fileBrowserStore.selectedFile, hdu: fileBrowserStore.selectedHDU});
+            const selectedFile = fileBrowserStore.selectedFile;
+            await this.loadFile({fileInfo: selectedFile || undefined, hdu: fileBrowserStore.selectedHDU || undefined});
```
izkgao: It looks better if we use `fileBrowserStore.selectedFile` directly in L111 so that we do not need L110.

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -398,7 +426,7 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
                 columnWidths={this.widgetStore.headerTableColumnWidths}
```
izkgao: Should be `columnWidths={columnWidths}`.

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
```diff
@@ -407,20 +435,41 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
 
     private updateHeaderTableColumnSize = (index: number, size: number) => {
         const widgetsStore = this.widgetStore;
-        if (widgetsStore.headerTableColumnWidths) {
-            widgetsStore.headerTableColumnWidths[index] = size;
+        if (widgetsStore) {
```
izkgao: Make this an early escape so that we do not need indents below.
```
if (!widgetsStore) {
    return;
}

File: src/components/CatalogOverlay/CatalogOverlayComponent.tsx [hunk 29]
```diff
@@ -407,20 +435,41 @@ export class CatalogOverlayComponent extends React.Component<WidgetProps> {
...
+                // Resize array to match expected column count
+                const newArray = new Array(expectedColumnCount).fill(undefined);
+                for (let i = 0; i < Math.min(widgetsStore.headerTableColumnWidths.length, expectedColumnCount); i++) {
+                    newArray[i] = widgetsStore.headerTableColumnWidths[i];
+                }
+                widgetsStore.headerTableColumnWidths = newArray;
+            }
+
+            if (index >= 0 && index < widgetsStore.headerTableColumnWidths.length) {
+                widgetsStore.headerTableColumnWidths[index] = size;
```
izkgao: Can be simplified like
```
let widths = widgetsStore.headerTableColumnWidths;
if (!widths || widths.length !== expectedColumnCount) {
    widths = Array.from({length: expectedColumnCount}, (_, i) => widths?.[i]);
    widgetsStore.headerTableColumnWidths = widths;
}
if (index >= 0 && index < expectedColumnCount) {
    widths[index] = size;
}
```

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -64,23 +64,26 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
...
         return null;
     }
 
     @computed get exportHeaders(): string[] {
-        let headerString = [];
-        const regionProperties = this.widgetStore.effectiveFrame?.getRegionProperties(this.widgetStore.effectiveRegionId);
-        regionProperties?.forEach(regionProperty => headerString.push(regionProperty));
+        let headerString: string[] = [];
+        const regionId = this.widgetStore.effectiveRegionId;
+        if (regionId !== null) {
```
izkgao, line 83: `(regionId != null)` can additionally exclude `undefined`.

izkgao (changes requested):
I was wondering whether we should consider unifying the usage of `||` and `??`, and perhaps let `||` be used purely for its logical OR semantics rather than as a default-value operator.

For example, in SpatialProfilerComponent.tsx:
**L27:** `AppStore.Instance.getFrame(this.widgetStore.fileId) || undefined;`
**L506:** `this.profileStore.x ?? undefined`

Similar patterns also appear with `|| false`, `|| null`, and `|| undefined`. Since `||` treats all falsy values the same, this can unintentionally mask valid values (e.g., 0 or false). In contrast, `??` is usually a better fit when the intention is to handle only `null` or `undefined`.

If this refactor may take some time, we could address it in a separate follow-up PR, similar to what you are planning for changing variables declared with `let` but never reassigned to `const`.

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
loveluthien: let's leave this as a refactoring

src/components/CatalogOverlay/CatalogOverlayComponent.tsx
loveluthien: let this to be a refactoring issue

src/components/LayerList/LayerListComponent.tsx
loveluthien: It doesn't complain, so maybe we don't.

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
loveluthien: regionId is either `number` or `null`

src/stores/Frame/FrameStore.ts
loveluthien: some arguments in transformSpectralPointArray allow `null`, so we don't need to check them.

izkgao (approved):

wasm_src/ast_wrapper/post.ts
```diff
@@ -307,7 +306,7 @@ Module.transformSpectralPoint = function (spectralFrameFrom: number | null, spec
     return zOut[0];
 };
 
-Module.transformSpectralPointArray = function (spectralFrameFrom: number, specType: string, specUnit: string, specSys: string, zIn: Float64Array | Array<number>, forward: boolean = true) {
+Module.transformSpectralPointArray = function (spectralFrameFrom: number | null, specType: string | null, specUnit: string | null, specSys: string | null, zIn: Float64Array | Array<number>, forward: boolean = true) {
```
veggiesaurus: we're using `?` in the line below which implies to me that `zIn` can also be undefined.

src/App.tsx
```diff
@@ -79,7 +79,9 @@ export class App extends React.Component {
 
     private setAppContainerRef = (ref: HTMLDivElement | null) => {
         this.appContainerRef.current = ref;
-        AppStore.Instance.setAppContainer(ref);
+        if (ref) {
```
veggiesaurus: I think there might be a good reason why we want to assign a null ref. In fact, `appStore.appContainer` would be undefined until this line is called 🤔

src/stores/Frame/FrameStore.ts
```diff
@@ -2531,17 +2531,19 @@ export class FrameStore {
         }
 
         for (const contourSet of processedData.contourSets ?? []) {
-            if (contourSet.level) {
+            if (contourSet.level != null) {
```
veggiesaurus, line 2556: level could also be `undefined`

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -713,17 +726,21 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
             let qSmoothedDic = this.assembleLinePlotData(compositeProfile.qProfileSmoothed, compositeProfile.qSmoothedX, StokesCoordinate.LinearPolarizationQ);
             let uSmoothedDic = this.assembleLinePlotData(compositeProfile.uProfileSmoothed, compositeProfile.uSmoothedX, StokesCoordinate.LinearPolarizationU);
 
+            // Create fallback empty datasets for null results
+            const emptyDataset = {dataset: [] as Point2D[], border: {xMin: 0, xMax: 0, yMin: 0, yMax: 0}};
```
veggiesaurus: I think `new Array<Point2D>()` is more readable (same with the line below)

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -617,16 +630,16 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
         var hi = data.length - 1;
         while (hi - lo > 1) {
             mid = Math.floor((lo + hi) / 2);
-            if (data[mid].z < channel) {
+            if (data[mid].z !== undefined && data[mid].z! < channel) {
```
veggiesaurus: not a fan of `.z!`. Rather explicitly check that it exists. Otherwise what's the point in typing it with `z?: number`?

src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
```diff
@@ -781,16 +941,16 @@ export class FileBrowserDialogComponent extends React.Component {
                         <div className="file-info-pane">
                             <FileInfoComponent
                                 infoTypes={FileBrowserDialogComponent.GetFileInfoTypes(fileBrowserStore.browserMode)}
-                                HDUOptions={{HDUList: fileBrowserStore.HDUList, handleSelectedHDUChange: fileBrowserStore.selectHDU}}
-                                fileInfoExtended={fileBrowserStore.fileInfoExtended}
+                                HDUOptions={{HDUList: fileBrowserStore.HDUList || [], handleSelectedHDUChange: fileBrowserStore.selectHDU}}
+                                fileInfoExtended={fileBrowserStore.fileInfoExtended as any}
                                 regionFileInfo={fileBrowserStore.regionFileInfo ? fileBrowserStore.regionFileInfo.join("\n") : ""}
-                                catalogFileInfo={fileBrowserStore.catalogFileInfo}
+                                catalogFileInfo={fileBrowserStore.catalogFileInfo as any}
```
veggiesaurus: this is not a good idea. Why do we need all these `as any`? we give up all type safety that way

src/components/Dialogs/FileBrowser/FileBrowserDialogComponent.tsx
```diff
@@ -765,8 +925,8 @@ export class FileBrowserDialogComponent extends React.Component {
                                 fileProgress={fileProgress}
                                 fileList={fileBrowserStore.getfileListByMode}
                                 fileBrowserMode={fileBrowserStore.browserMode}
-                                selectedFile={fileBrowserStore.selectedFile}
-                                selectedHDU={fileBrowserStore.selectedHDU}
+                                selectedFile={fileBrowserStore.selectedFile as any}
```
veggiesaurus: why do we need these?

src/components/Dialogs/FileInfoDialog/FileInfoDialogComponent.tsx
```diff
@@ -44,9 +44,9 @@ export class FileInfoDialogComponent extends React.Component {
                 <div className={Classes.DIALOG_BODY}>
                     <FileInfoComponent
                         infoTypes={[FileInfoType.IMAGE_FILE, FileInfoType.IMAGE_HEADER]}
-                        fileInfoExtended={appStore.activeFrame ? appStore.activeFrame.frameInfo.fileInfoExtended : null}
+                        fileInfoExtended={appStore.activeFrame ? appStore.activeFrame.frameInfo.fileInfoExtended : (undefined as any)}
                         regionFileInfo={""}
-                        catalogFileInfo={null}
+                        catalogFileInfo={undefined as any}
```
veggiesaurus: this makes no sense

src/components/Dialogs/LayoutDialog/LayoutDialogComponent.tsx
```diff
@@ -229,7 +229,7 @@ export class LayoutDialogComponent extends React.Component {
                         <Tab
                             id={LayoutDialogMode.DynamicLayout}
                             title="Dynamic Layout"
-                            panel={<LayoutMappingComponent orderedLayoutNames={layoutStore.orderedLayoutNames} existLayoutMapping={preferenceStore.existLayoutMapping} activeFrame={appStore.activeFrame} />}
+                            panel={<LayoutMappingComponent orderedLayoutNames={layoutStore.orderedLayoutNames} existLayoutMapping={preferenceStore.existLayoutMapping} activeFrame={appStore.activeFrame as any} />}
```
veggiesaurus: also makes no sense.

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -564,8 +564,8 @@ export class PreferenceDialogComponent extends React.Component {
                             <Switch checked={preference.isHighDimPriority} onChange={() => preference.setPreference(PreferenceKeys.LAYOUT_IS_HIGH_DIM_PRIORITY, !preference.isHighDimPriority)} />
                         </Tooltip>
                     </FormGroup>
-                    <Collapse isOpen={appStore.dynamicLayoutStore.isMappingExisted || (appStore.activeFrame && appStore.activeFrame.dynamicLayout.ctype !== "")}>
-                        <LayoutMappingComponent orderedLayoutNames={layoutStore.orderedLayoutNames} existLayoutMapping={preference.existLayoutMapping} activeFrame={appStore.activeFrame} />
+                    <Collapse isOpen={appStore.dynamicLayoutStore.isMappingExisted || (appStore.activeFrame ? appStore.activeFrame.dynamicLayout.ctype !== "" : false)}>
+                        <LayoutMappingComponent orderedLayoutNames={layoutStore.orderedLayoutNames} existLayoutMapping={preference.existLayoutMapping} activeFrame={appStore.activeFrame as any} />
```
veggiesaurus: again, very bad idea

veggiesaurus (changes requested):
some critical changes needed, related to use of `any`.

src/stores/Frame/FrameStore.ts
loveluthien: do you mean we allow level to be undefined?

veggiesaurus: according to the type definition, `level` could be `null` or `undefined` so we should check for `undefined` as well, shouldn't we?

loveluthien: I had confirmed that `x != null` equals to `x !== null` + `x !== undefined`, so I use `!= null` for variables not `null` nor `undefined` everywhere.

veggiesaurus: ah, sorry I missed that it was just `!=` and not `!==` 🙈 another delightful JS quirk!

loveluthien:
@veggiesaurus, thanks a lot. I mended the `as any` problem and other type issues.

src/components/StokesAnalysis/StokesAnalysisComponent.tsx
```diff
@@ -727,7 +729,7 @@ export class StokesAnalysisComponent extends React.Component<WidgetProps> {
             let uSmoothedDic = this.assembleLinePlotData(compositeProfile.uProfileSmoothed, compositeProfile.uSmoothedX, StokesCoordinate.LinearPolarizationU);
 
             // Create fallback empty datasets for null results
-            const emptyDataset = {dataset: [] as Point2D[], border: {xMin: 0, xMax: 0, yMin: 0, yMax: 0}};
+            const emptyDataset = {dataset: [] as Array<Point2D>, border: {xMin: 0, xMax: 0, yMin: 0, yMax: 0}};
```
veggiesaurus: no, just `dataset: new Array<Point2D>()`

veggiesaurus (approved):
approved, one minor comment

kswang1029 (approved):

PR #2606, opened 2025-09-19
Reset window lock state when resuming a session

### LINKED ISSUE #2082: Reset window lock state when resuming a session
This is found by @pford 

What do you think about resuming a session without previous dialog progress?  Example: open image/images, do stuff, import region file, backend crashes.  When resume, Import Region dialog is still there (progress blocking everything else) waiting for a response which will never come.  Have to restart CARTA, reload image/images, and do stuff again to get back to where you were.

izkgao:
**Description**

Closes #2082.

After resuming a session, the window lock will be reset.

This works for importing regions, image fitting, PV generation, moment generation, and cube histogram (including that from contour generation) generation.

_For importing regions:_
The behavior of the file browser depends on whether **ImportRegionAck** has been received:
1. If **ImportRegionAck** has **_not_** been received: the file browser remains open.
2. If **ImportRegionAck** has been received: the file browser will be closed.

**How to test:**
1. Load an image
2. Importing regions 
2.1. Import regions using this file
[test_region_10000_point.crtf.zip](https://github.com/user-attachments/files/22419957/test_region_10000_point.crtf.zip)
2.2 Terminate the backend at one of the following stages:
    - After the progress bar is displayed, but before it starts to increase.
        - The window lock is cleared.
        - The file browser remains open.
    - After the progress bar has started to increase, but before all regions have finished importing.
        - The window lock is cleared.
        - The file browser is closed.
        - All regions are successfully imported.
3. Other tasks
3.1 Find a cube with enough channels to terminate the backend before it completes.
3.2 Terminate the backend during processing.
3.3 Restart the backend.
3.4 Ensure the screen is not frozen by the progress bar.

**Checklist**

For the pull request:

For dependencies:

### DIFF
--- a/src/stores/AppStore/AppStore.ts
+++ b/src/stores/AppStore/AppStore.ts
@@ -165,10 +165,6 @@ export class AppStore {
     // dynamic zIndex
     public zIndexManager = new FloatingObjzIndexManager();
 
-    // ImportRegion
-    // Track whether ImportRegionAck has been received to determine file browser behavior during session resume
-    private hasReceivedImportRegionAck = false;
-
     private appContainer: HTMLElement;
     private fileCounter = 0;
     private previousConnectionStatus: ConnectionStatus;
@@ -1258,7 +1254,7 @@ export class AppStore {
         try {
             const ack = yield this.backendService.importRegion(directory, file, type, frame.frameInfo.fileId);
             if (frame && ack.success && ack.regions) {
-                this.hasReceivedImportRegionAck = true;
+                this.fileBrowserStore.setHasReceivedImportRegionAck(true);
                 const regions = Object.entries(ack.regions);
                 const regionStyleMap = new Map<string, CARTA.IRegionStyle>(Object.entries(ack.regionStyles));
                 let startIndex = 0;
@@ -1270,13 +1266,11 @@ export class AppStore {
                 this.fileBrowserStore.setImportingRegions(false);
                 this.fileBrowserStore.resetLoadingStates();
                 this.fileBrowserStore.hideFileBrowser();
-                this.hasReceivedImportRegionAck = false;
             }
         } catch (err) {
             console.error(err);
             this.fileBrowserStore.setImportingRegions(false);
             this.fileBrowserStore.resetLoadingStates();
-            this.hasReceivedImportRegionAck = false;
             AppToaster.show(ErrorToast(err));
         }
     }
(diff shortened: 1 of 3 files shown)


kswang1029 (commented):
@izkgao the new behaviour for the region file browser looks good. 

Could you also investigate the following two cases as well?

1) moment map generator
When generating moment images, there is also a progress bar blocking the GUI. If when it is progressing the backend died and resumed, the frontend will be locked and it appears that clicking the cancel button does not help so the GUI is not in an ideal state for resume.

2) pv generator
Similarly, when a session is resumed, we still see the progress bar but click the cancel button works so the GUI is unlocked. However if we try to generate a PV image again, we will see the progress bar with "cancellation in progress" in the popup window.
Screenshot 2025-09-23 at 14 02 50

I am not sure if the frontend needs to send a cancel message again to the backend while a session is resuming. I guess the backend won't remember that there was an ongoing task to complete.

izkgao:
@kswang1029 Now `onSessionResumed` resets the states of importing region, image fitting, PV generation, moment generation,  and cube histogram (including that from contour) generation.

src/stores/AppStore/AppStore.ts
```diff
@@ -1254,6 +1258,7 @@ export class AppStore {
         try {
             const ack = yield this.backendService.importRegion(directory, file, type, frame.frameInfo.fileId);
             if (frame && ack.success && ack.regions) {
+                this.hasReceivedImportRegionAck = true;
```
loveluthien: Please check if we can use `this.fileBrowserStore.isImportingRegions` instead of `this.hasReceivedImportRegionAck`.

src/stores/AppStore/AppStore.ts
```diff
@@ -2540,6 +2547,53 @@ export class AppStore {
         this.initRequirements();
         this.resumingSession = false;
         this.backendService.connectionDropped = false;
+
+        // Reset file browser loading states
+        if (this.fileBrowserStore.isImportingRegions) {
```
loveluthien, line 2582: Do we have similar problem when loading large catalog?

izkgao: No, only tasks that use `TaskProgressDialogComponent` have this problem.

src/stores/AppStore/AppStore.ts
izkgao: `this.fileBrowserStore.isImportingRegions` can not be used to tell whether `ImportRegionAck` has been received within the importing region event so we need `this.hasReceivedImportRegionAck`.

src/stores/AppStore/AppStore.ts
```diff
@@ -1265,11 +1270,13 @@ export class AppStore {
                 this.fileBrowserStore.setImportingRegions(false);
                 this.fileBrowserStore.resetLoadingStates();
                 this.fileBrowserStore.hideFileBrowser();
+                this.hasReceivedImportRegionAck = false;
```
loveluthien: how about move `this.hasReceivedImportRegionAck = false;` to `resetLoadingStates`? If doing that we need to modify the order of `resetLoadingStates` in `onSessionResumed`.

loveluthien (approved):
😎

kswang1029 (approved):
looks good. 👍

PR #2608, opened 2025-10-07
carta v6 package upgrade

loveluthien:
**Description**

Closes #2605. Package upgrade.

BlueprintJS: 
- minor modifications for deprecated functions
- setup @blueprintjs/eslint-plugin
  -  use blueprintjs html-components, ex `<Pre>` and `<H1>` instead of `<pre>` and `<h1>`
  -  use blueprintjs `Classes` in the `classNames`
  - modify css for blueprintjs `<Pre>`
- e2e `Coordinate Grid Rendering` failed. It doesn't like a bug.
- dark theme color shifted. Modify `Colors.DARK_GRAY2` -> `Colors.DARK_GRAY1` in some places. Light theme has no change.
- doesn't support react 19 yet

Plotly.js:
- modify title input accordingly.

react-window:
- relace `FixedSizeList` with `List` and drop the `AutoSizer` since automatically sizing becomes build-in support in `List`

**Checklist**

Major upgrade: 
- blueprintjs 5 -> 6
- plotly.js  2 -> 3
- react-window  1 -> 2
- UUID 10 -> 13
- eslint 9
- typescript 5.3 -> 5.9

Other minor and patch upgrade

For linked issues (if there are):
- [x] assignee and label added
- [x] GitHub Project estimate added

For the pull request:
- [x] reviewers and assignee added
- [x] GitHub Project estimate added
- [x] ~changelog updated~ / no changelog update needed
- [x] ~unit test added (for functions with no dependenies)`
- [x] ~API documentation added (for public variables and methods in stores)~

For dependencies:
- [ ] e2e test passing / corresponding fix added / new e2e test created
- [x] ~protobuf version bumped~ / no protobuf version bumped needed
- [x] ~protobuf updated to the latest dev commit~ / no protobuf update needed
- [x] ~corresponding ICD test fix added (`BackendService` changed)~ / no ICD test fix needed (`BackendService` unchanged)
- [x] ~user manual prepared (for large new features)~

PR #2601, opened 2025-09-15
rsbuild migration

loveluthien:
**Description**

This PR closes #2479 and is to migrate the build tool from react-scripts to Rebuild. This migration is based on #2590. 

Migration points:
- create rsbuild related settings (done by Yu-Hsuan)
  - Added `NodePolyfill`, `glsl` plugins
  - Setup TypeScript decorators (legacy version)
  - Set TypeScript verbatimModuleSyntax to true and migrate the entire codebase
- Move `export enum` to the newly created `enums` folder (done by Yu-Hsuan)
- Give initial values to `mobx` observables
- Directly use `Jest` for unit tests
- Set up `eslint`
  - `prefer const`: force to use `const` if the variable was not reassigned
  - Use `typescript-eslint/consistent-type-imports` instead of turning on `verbatimModuleSyntax`. They might have some conflicts.
- Use `sass-embedded` instead of `sass` to support various platforms. Both packages are from the same team.
- frontend api documents
  - support the relocation of the export `enum`

**Checklist**

- check if CARTA can be built by running `npm install` and `npm run build`
- `npm run build` is successful on different platforms

For linked issues (if there are):
- [x] assignee and label added
- [x] GitHub Project estimate added

For the pull request:
- [x] reviewers and assignee added
- [x] GitHub Project estimate added
- [x] ~changelog updated~ / no changelog update needed
- [x] ~unit test added (for functions with no dependenies)~
- [x] ~API documentation added (for public variables and methods in stores)~

For dependencies:
- [x] e2e test passing / corresponding fix added / new e2e test created
- [x] ~protobuf version bumped~ / no protobuf version bumped needed
- [x] ~protobuf updated to the latest dev commit~ / no protobuf update needed
- [x] ~corresponding ICD test fix added (`BackendService` changed)~ / no ICD test fix needed (`BackendService` unchanged)
- [x] ~user manual prepared (for large new features)`

PR #2590, opened 2025-08-29
mobx react v9 and strict mode migration

loveluthien:
**Description**

This PR upgrades the mobx to v9 and adapts the React strict mode. Close #2432. The PR is blocked by https://github.com/CARTAvis/carta-frontend/pull/2565 and https://github.com/CARTAvis/carta-frontend/pull/2577.

Please do `npm install` before testing this PR.

**Checklist**
- no [mobx] errors in the console

For linked issues (if there are):
- [x] assignee and label added
- [x] GitHub Project estimate added

For the pull request:
- [x] reviewers and assignee added
- [x] GitHub Project estimate added
- [x] ~changelog updated~ / no changelog update needed
- [x] ~unit test added (for functions with no dependenies)~
- [x] ~API documentation added (for public variables and methods in stores)~

For dependencies:
- [x] e2e test passing / corresponding fix added / new e2e test created
- [x] ~protobuf version bumped~ / no protobuf version bumped needed
- [x] ~protobuf updated to the latest dev commit~ / no protobuf update needed
- [x] ~corresponding ICD test fix added (`BackendService` changed)~ / no ICD test fix needed (`BackendService` unchanged)
- [x] ~user manual prepared (for large new features)~

PR #2629, opened 2025-11-28
Correct colormap image and add colormap generation script

### LINKED ISSUE #2152: make a python script to generate the allmaps.png file
We would need a python script to generate the allmaps.png file in the codebase as a preparation of enhancing colormap support.

https://github.com/CARTAvis/carta-frontend/blob/dev/src/static/allmaps.png

### LINKED ISSUE #2628: Incorrect colormap–name pairing
**Describe the bug**

Several colormaps are paired with the wrong names.

CARTA:
Image

Matplotlib (expected):
Image
Reference: https://matplotlib.org/stable/users/explain/colors/colormaps.html#miscellaneous

The colormaps for gnuplot and gnuplot2 do not match Matplotlib's versions.
Additionally, PuBu and PuBuGn are also mismatched, though they do not appear in the currently selected list.

**To Reproduce**
1. Click Colormap in Render Configuration
2. Scroll down to gnuplot and gnuplot2
3. See the incorrect colormap–name pairing

**Expected behavior**
The gnuplot and gnuplot2 colormaps should match the versions provided by Matplotlib.

 - OS [e.g. macOS Monterey]: macOS Sequoia
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: v5.0.3
 - Frontend branch [e.g. dev, v3b2 release]: v5.0.3

izkgao:
**Description**

Closes #2152 and #2628.

This PR fixes mismatched colormap–name pairs by generating a corrected colormap image, and introduces a new Python script for generating the colormap image (`src/static/allmaps.png`).

**Checklist**

For the pull request:

For dependencies:


confluence:
@izkgao I commented on the schema issues in the schemas PR.

1. Changing the case of the colormap keys (and changing `Greys` to `Grays`) would require 1) adding code to the frontend to handle updating old preferences, 2) updating the Python wrapper to change the constants, and 3) possibly adding code to the Python wrapper to distinguish between frontend versions (although, looking at the wrapper code, I see some problems with the current implementation, which I'll summarise in an issue there).
   
   Instead I would suggest lowercasing all the keys used in the frontend, and accepting values in any case (normalising them to lowercase before looking them up in the maps).
   
   Matplotlib added aliases to normalize the grey / gray spelling [here](https://github.com/matplotlib/matplotlib/pull/26003), so either will work for fetching the colormap in the script. If we want to normalize the spelling in the frontend, I suggest using `Grays` to fetch / generate the texture (to ensure the correct ordering) and perhaps adding the same aliases to the frontend.

2. A potential issue with fixing the mappings: existing user preferences may now be referring to the wrong colormaps. But you can't just correct them unconditionally, because you'll have no way of knowing whether the preference was saved using the correct mapping or the incorrect mapping.
   
   I think that the correct way to handle this (which would really simplify future backwards compatibility issues) is to add a minor version to the current prefs schema (optional; defaulting to 0) and bump it whenever a change like this is made (which doesn't break the schema, but breaks old preferences). Then you could explicitly check preferences for the minor version on load, make the required fixes, and save the updated minor version.
   
   This would be much less fragile than the current piecemeal solutions, which rely on the ability to detect obsolete preference values from the values themselves (which is not always guaranteed, like in this case).

3. The Python script could read the full list of colormaps from matplotlib directly instead of hardcoding it here, and also output an updated list of keys (lowercased, if we want to normalize case) that could be pasted directly into `COLOR_MAPS_ALL` in the store code (replacing the matplotlib keys only, not the special monochrome keys). That would allow us to update our colormaps regularly to match matplotlib (without affecting the curated list that is selectable in the GUI). But that does raise some questions about what we do and don't want to include. This lists every single matplotlib colormap in alphabetical order: `import matplotlib.pyplot as plt; sorted(plt.colormaps(), key=lambda k: k.lower())`. To exclude all the reversed / inverted colormaps, we can do this: `sorted((k for k in plt.colormaps() if not k.endswith("_r")), key=lambda k: k.lower())`. To remove the "grey" aliases and keep only "gray", we can do this: `sorted({k for k in plt.colormaps() if not k.endswith("_r")} - {'grey', 'Greys', 'gist_grey', 'gist_yerg'}, key=lambda k: k.lower())`
   
   1. That leaves some colormaps that were more recently added to matplotlib: [`cividis`](https://github.com/matplotlib/matplotlib/pull/9871), [`twilight` and `twilight-shifted`](https://github.com/matplotlib/matplotlib/pull/6254), [`turbo`](https://github.com/matplotlib/matplotlib/pull/15275), and [`berlin`, `managua`, and `vanimo`](https://github.com/matplotlib/matplotlib/pull/28587). Should we add these new colormaps?
   
   2. Per [previous versions of the docs](https://matplotlib.org/3.4.3/api/pyplot_summary.html) (I went down a [rabbithole](https://github.com/matplotlib/matplotlib/issues/30796) trying to find out where this information went), `gist_gray`, `gist_yarg` and `binary` are identical to `gray`, `gray_r` and `gray_r`, respectively. So it's unnecessary for us to include these. I'm not sure if we care about existing preferences using these as defaults, since none of these are GUI-selectable: we could fix `gist_gray` easily, but we don't have preferences for inverting the default colormaps (should we? It's sensible for us not to duplicate textures when we can invert dynamically, but at the moment it's impossible to set a default colormap that is inverted). If this isn't important, we can reset these values to the default (it should be obvious to the user that this has happened because the default colormap isn't grayscale).

If we wanted to include the new colormaps and exclude the duplicate colormaps, the final list in the script could look like this:
```
COLOR_MAPS_ALL = sorted({k for k in plt.colormaps() if not k.endswith("_r")} - {'grey', 'Greys', 'gist_grey', 'gist_yerg', `gist_gray`, `gist_yarg`, `binary`}, key=lambda k: k.lower())
```

We could use that to print a new definition for COLOR_MAPS_ALL in the ts source, or even automate searching and replacing the old definition.

@kswang1029 what do you think about adding the newest matplotlib colormaps?

(Adding a note here that in several places we hardcode the number of non-monochrome colormaps to 79, and we really should replace that with a constant that's calculated from the map sizes!)

kswang1029:
@confluence regarding

> @kswang1029 what do you think about adding the newest matplotlib colormaps?

The ultimate goal is to 1) update new matplotlib colormaps and 2) let user choose which colormaps to be displayed in the colormap dropdown menu (in render config settings), 3) support custom colormaps.

As for v6-beta, if possible, can we fix bugs only (mismatched names and colormaps) and leave the rest enhancement+refactoring to beta2? Or, we do refactoring+bug fix in beta1 and to the "ultimate goal" in beta2? @loveluthien any suggestions?

loveluthien:
@kswang1029 
For beta1, keep the plan focused on bug fixes, not refactoring, and make only the minimal necessary changes related to the Python interface. Any refactoring should wait till merging #2577, #2590, #2601, and #2608. 

As @confluence pointed out, we need to handle the preferences that include the "wrong" colormap name when we correct the colormap name.

izkgao:
@confluence Thank you for the detailed feedback. It will be very helpful for the follow-up PRs.

Since this PR is intended as a quick fix for v6-beta1 (without additional enhancements) and updating the colormap image is sufficient to resolve the issue, I have removed the changes to `RenderConfigStore.ts` and closed the related schema PR.

kswang1029 (approved):

scripts/make_colormaps.py
```diff
@@ -0,0 +1,155 @@
+# /// script
+# requires-python = ">=3.13"
```
loveluthien: is the requirement of python3.13 too strict?

scripts/make_colormaps.py
```diff
@@ -0,0 +1,155 @@
...
+"""Generate a stacked colormap PNG from CARTA's colormap definitions.
+
+This script reads COLOR_MAPS_ALL and COLOR_MAPS_MONO from color.ts,
+filters out monochrome colormaps, and generates a single PNG image with each
+colormap rendered as a horizontal stripe. The output is used by the CARTA
+frontend for both colormap previews (in the dropdown selector) and as a lookup
+texture for WebGL rendering of image tiles, contours, and overlays.
+
+Usage:
+    uv run scripts/make_colormaps.py              # Output to src/static/allmaps.png
```
loveluthien: I think `uv` is not the standard way to use python yet. Please add regular execution using `python`.

scripts/make_colormaps.py
izkgao: It does not matter if a user uses uv. `uv run` will handle the requirements automatically according to this header, which only works for uv. On the other hands, lower versions of python may limit the available versions of matplotlib and thus have no access to some new APIs. Typically the second latest stable version is recommended.

scripts/make_colormaps.py
izkgao: It might be more convenient to use uv to run this script, since it requires only a single command:
`uv run scripts/make_colormaps.py`.

If users prefer to run the script with plain Python, one possible approach would be to first create and activate a virtual environment, then install the required dependencies from the provided requirements.txt, and finally run:
`python scripts/make_colormaps.py`. This ensures that the system environment remains clean and avoids introducing additional packages or potential version conflicts.

Would you like me to add the plain-Python usage instructions to the script as well?

loveluthien: Yes, please add a plain-Python usage instruction.

loveluthien (approved):
😎

PR #2660, opened 2026-02-02
overlay vector pixel averaging width step correction

### LINKED ISSUE #2612: Incorrect handling of default vector overlay pixel averaging width preference
**Describe the bug**

* The default vector overlay pixel averaging width can be set to zero in the preferences, which is interpreted as averaging being disabled by default.
* The vector overlay dialog normalizes averaging widths to positive (nonzero) even numbers, but it only applies this logic if the value is changed in the dialog.
* When averaging is disabled by default, the greyed-out width value displayed in the dialog is 2 (probably because that's the minimum set in the input element), but when averaging is then re-enabled, it changes to zero (and is not corrected).
* Clicking "apply" while the width is zero causes no vector field to be displayed (but the dialog enters into the state in which a field is being displayed).
* If the dialog is closed while the width is zero and averaging is enabled, when it is reopened, the value *appears* to be corrected to 2, but it's actually still zero (as you can see if you click "apply"). Only changing the value in the dialog causes it to "stick" in that image (opening a new image will cause the defaults to be applied again).

In the same way, it's possible to bypass the normalization and set the width to an odd number by manually entering the number as the default preference. This appears to display valid vector fields consistent with the appearance of widths above and below (and with a width of 1 being identical to "no smoothing"). I can't see any reason in the backend why odd numbers should not be used (although better smoothing algorithms can potentially be used for multiples of 4 and multiples of 8). If we want to avoid arbitrary values that can't take advantage of these algorithms, then perhaps we should restrict these inputs to a selection of multiples of 4 and 8.

**To Reproduce**

1. Set the averaging value to zero in the preferences
2. Open a new image
3. Open the vector overlay dialog
4. Enable pixel averaging
5. Click "apply"

or

5. Close the dialog
6. Open the dialog
7. Click "apply"

or

1. Set the averaging value to 3 in the preferences; etc.
  
**Expected behavior**

I would expect the value to be corrected when the averaging is enabled and when the dialog is opened.

 - OS [e.g. macOS Monterey]: Ubuntu Jammy
 - Browser [e.g. chrome, safari, electron app]: Firefox
 - Browser version [e.g. 22]: 143.0.4
 - Backend branch [e.g. dev, v3b2 release]: v5 release
 - Frontend branch [e.g. dev, v3b2 release]: v5 release

loveluthien:
**Description**

Close #2612. 

- Make the input behavior of vector overlay pixel-averaging width consistent between the vector and the preference dialogs. 
- Update the description in the [protobuf](https://github.com/CARTAvis/carta-protobuf/pull/107). Notice that no protobuf message is modified.
- Remove `pixelAveragingEnabled` variable. Notice that `pixelAveragingEnabled` in workspace_schema_1.json should be cleaned at some point.

**Checklist**

For the pull request:

For dependencies:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 * Set the default aspect ratio to unity when images lacked valid WCS information ([#2604](https://github.com/CARTAvis/carta-frontend/issues/2604)).
 * Changed to retrieve all columns instead of default columns from vizier ([#1958](https://github.com/CARTAvis/carta-frontend/issues/1958)).
-* Vector overlay averaging pixel input UI ([#2612](https://github.com/CARTAvis/carta-frontend/issues/2612))..
 
 ## [5.0.3]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Removed the incorrect WCS information when in image coordinate system ([#2614](https://github.com/CARTAvis/carta-frontend/issues/2614)).
 * Fixed incorrect colormap–name pairing ([#2628](https://github.com/CARTAvis/carta-frontend/issues/2628)).
 * Fixed vector overlay not updating after session resume ([#2217](https://github.com/CARTAvis/carta-frontend/issues/2217)).
+* Made vector overlay pixel averaging consistent between the Vector Overlay dialog and Preferences dialog ([#2612](https://github.com/CARTAvis/carta-frontend/issues/2612)).
 ### Changed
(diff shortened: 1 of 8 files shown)


kswang1029 (changes requested):
Please keep using the original spinbox as the input for the pixel averaging factor. Usually the pixel averaging factor is small but the slider approach makes it harder to have a finer control because small values are squeezed to the left end. In addition, mostly in CARTA the slider action is bound to backend computations such as the sliders in the animator to change displayed image, the slider in the channel map view control widget to change the starting channel, or the slider in the histogram settings dialog where the bin number can be changed and result is updated accordingly. Here the pixel averaging factor is not bound to any other backend computation. If the backend can support caching the Stokes Q and U images and cancellation (@confluence ?), then binding the computation should be okay as there will be no repeated data loading process. If caching is not supported, then having the slider binding to backend computation would result in excessive image load, computation, and websocket backpressure at the backend.

The protobuf description needs an update to have a consistent behaviour. https://carta-protobuf.readthedocs.io/en/latest/messages.html#setvectoroverlayparameters (not limited to an even number anymore).

izkgao (approved):

kswang1029 (approved):

PR #2706, opened 2026-03-02
Address issue with upgradePreferences

### LINKED ISSUE #2705: upgradePreferences attempts to make changes even if none are required
**Describe the bug**
An attempt is made to delete `cubeSizeUnitKey` even if no such key exists was the originally noticed symptom.  A further evaluation suggests that additional unnecessary modifications may be getting attemped

**To Reproduce**
1. Launch carta with browser inspector open to network tab and notice that an attempt is made to delete `cubeSizeUnitKey`
2. Launch carta again with browser inspector open to network tab and notice that another attempt is made to delete `cubeSizeUnitKey`

**Expected behavior**
Upgrade preferences doesn't attempt to make any changes if the preferences version is already 2

daikema:
**Description**

This attempts to resolve #2705 by moving all the remaining code from the `upgradePreferences` function into the `if (preferences["version"] === 1) {` block as it seems likely that this was what was intended.  (Otherwise I'd expect to see upgrade code for versions other than 1).

### DIFF
--- a/src/services/ApiService.ts
+++ b/src/services/ApiService.ts
@@ -240,34 +240,34 @@ export class ApiService {
             preferences[astColorKey] = `auto-${color}`;
             this.setPreference(astColorKey, preferences[astColorKey]);
 
+            // Normalize case of wcsType value, which may have been saved incorrectly in existing preferences
+            const key = PreferenceKeys.WCS_OVERLAY_WCS_TYPE;
+            if (/[A-Z]/.test(preferences[key])) {
+                preferences[key] = preferences[key].toLowerCase();
+                this.setPreference(key, preferences[key]);
+            }
+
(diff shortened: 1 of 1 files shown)


veggiesaurus (approved):

loveluthien:
@daikema, thanks for fixing the bug. Is it an argent PR? If not, we could merge it after beta1 release.

daikema:
@loveluthien I don't think this is particularly urgent.   It just trims out a bit of code that was being run at startup each time rather than only when a change is made.  Will let you decide when it's best to merge.

loveluthien:
ok, let's merge this after beta1.

loveluthien (approved):
😎

kswang1029 (approved):

PR #2808, opened 2026-05-20
Improve GSL download retry and mirror fallback

### LINKED ISSUE #2807: [CI] Limit wget retry attempts and timeout when downloading GSL
Recently, `http://ftpmirror.gnu.org/` has been unavailable at least twice in the past few weeks, causing CI jobs to be blocked for a long time while downloading GSL.

The delay happens because the default `wget` settings are relatively large:

```bash
tries = 20
timeout = 60 seconds
```

As a result, when a selected GNU mirror is unavailable, `wget` may spend a long time retrying the same mirror before switching to another one. In our CI, this has caused the GSL download step to take more than one hour.

To reduce the impact of unavailable mirrors, we can explicitly limit `wget` to a single attempt per mirror, for example:

```bash
wget --tries=1
```

Then, we can rely on the existing `max_retries=2` logic in `build_gsl.sh` to handle retries at the script level. This should allow CI to fail over to another mirror more quickly and avoid long stalls during the GSL download step.

izkgao:
**Description**

Closes #2807.

This PR improves the GSL download step in `wasm_libs/build_gsl.sh` by making retry and mirror fallback behavior more robust.

**What is implemented:**
- Replace nested mirror fallback logic with a maintainable URL list loop.
- Use `wget` built-in retry support with `--tries` and `--waitretry`.
- Remove partial or invalid downloads before trying the next mirror.
- Verify the downloaded GSL tarball with `md5sum` before accepting it.

**How to test:**
Need to activate emcc first.
```bash
npm install
npm run build-libs
npm run build
```

**Checklist**

For the pull request:
- [x] ~~unit test added (for functions with no dependenies)~~
- [x] ~~API documentation added (for public variables and methods in stores)~~

For dependencies:
- [x] ~~user manual prepared (for large new features)~~

### DIFF
--- a/wasm_libs/build_gsl.sh
+++ b/wasm_libs/build_gsl.sh
@@ -16,20 +16,12 @@ if ! [[ $(find gsl-2.6.tar.gz -type f 2>/dev/null && md5sum -c gsl.md5 &>/dev/nu
 
     rm -f "${gsl_tar}"
     for gsl_url in "${gsl_urls[@]}"; do
-        retry_count=0
-        while (( retry_count < max_retries )); do
-            if wget --tries=1 -O "${gsl_tar}" "${gsl_url}" && md5sum -c gsl.md5 &>/dev/null; then
-                downloaded=true
-                break 2
-            fi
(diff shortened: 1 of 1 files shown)


loveluthien (approved):
😎

kswang1029 (approved):

PR #2840, opened 2026-06-18
preferences schema refactoring

### LINKED ISSUE #2568: Outstanding issues related to the preferences schema
These are issues that I found in the preferences schema while working on the [range and enum constraint fix](https://github.com/CARTAvis/schemas/pull/10):

### Definition of constraints and defaults

Some constraint definitions are replicated in the schema and in the frontend. This can lead to (and has previously caused) duplication at best and inconsistencies at worst.

* Ideally, the schemas should be treated as the canonical source of these constraints, and where the frontend currently hardcodes duplicate values it should instead read them from the schema. I suggest adding functionality to the preferences store which reads the actual schema file and can query it for these constraints.

* In some cases, hardcoded constraints and defaults are also duplicated in the frontend between the preferences code and other parts of the code where preference values are applied -- these duplicated values should be read from a single source.

* There are some cases where additional constraints are imposed by the GUI interface for entering preferences, but aren't actual limitations on what values are valid. For example, the preferences dialog only allows the default raster rendering percentile to be set to a preset choice from a dropdown, but any percentile value is legal and will work. Also, various data size limits and defaults have changed over time and are likely to change again. In cases like this it's fine for the frontend to maintain a separate set of constraints (as long as it's a strict subset of what the schema allows).
 
* In the v1 schema, a lot of the properties had defaults defined, but these were removed from the v2 schema. My suggestion would be to remove all remaining defaults from the schema and continue to manage these in the frontend only (since they are choices subject to change, just like the GUI input selections). These should be hardcoded in one location which is used both by the preferences store and by other stores that set the properties.

### Specific preferences

* Currently `pvPreviewCubeSizeLimit` has a corresponding `pvPreviewCubeSizeLimitUnit` preference, and both are needed to determine the value. This makes it difficult to define limits for this value. The unit preference should be eliminated, one unit should be picked for the size preference, and unit conversion should be performed in the frontend *only* when the user enters a value or a value is displayed to the user.
* The NaN colour is controlled by two properties: `nanColorHex`, which despite its name can be any valid HTML colour string, and `nanAlpha`. We convert `nanColorHex` to an RGB hex, and save the alpha value separately, and reconstruct the preference in reverse when we read it. This is unnecessary: we can eliminate `nanAlpha` entirely and include the alpha value directly in `nanColorHex`  by saving it in the [four-parameter RGB notation](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb) (e.g. `rgb(31 120 50 / 0.5)`). (There are many other places where we disable alpha values in colour pickers, but colours with alpha values work fine if they're set directly in the file -- could we use this notation consistently and allow alpha values everywhere that they can be used?)
* Unreachable setting: `textAnnotationLineWidth`. Added [here](https://github.com/CARTAvis/carta-frontend/commit/44be88adb666ba98257ce1006e883eeca9912101) together with other annotation preferences, but omitted from the preferences dialog. Was this omission an error?
* Unreachable setting: `colormapHexStart`. Added [here](https://github.com/CARTAvis/carta-frontend/commit/e3b41f411c2db00189cca8345719d2375b68c64c) with the custom colourmap preferences. This seems to be a limitation of the `ColormapComponent`, which only allows one colour (the end colour) to be selected (a custom start can be passed to the component but cannot be set). Logically, this component should allow the user to set both endpoints of the custom colourmap (various UX options possible).

loveluthien:
**Description**

Close #2568. Refactor the preferences schema, remove default values of schema v1 residual, and drop the `pvPreviewCubeSizeLimitUnit` and `nanAlpha`.

Companion [schema PR](https://github.com/CARTAvis/schemas/pull/15). 

**Checklist**

For the pull request:

For dependencies:


src/services/ApiService.ts
```diff
@@ -261,6 +262,17 @@ export class ApiService {
...
+
+        // Migrate nanAlpha to nanColorHex if present
+        if ("nanAlpha" in preferences) {
+            const nanColorHex = preferences["nanColorHex"] || "#137CBD";
+            const nanAlpha = preferences["nanAlpha"];
+            const combinedColor = tinycolor(nanColorHex).setAlpha(nanAlpha).toRgbString();
+            preferences["nanColorHex"] = combinedColor;
+            this.setPreference("nanColorHex", combinedColor);
+            delete preferences["nanAlpha"];
+            this.clearPreferences(["nanAlpha"]);
```
izkgao:
```typescript
if (await this.setPreference("nanColorHex", combinedColor)) {
    await this.clearPreferences(["nanAlpha"]);
}

delete preferences["nanAlpha"];
```

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -608,11 +609,13 @@ export class PreferenceDialogComponent extends React.Component {
...
                         placeholder="Region size"
-                        min={10}
-                        max={100}
+                        min={preference.getMinConstraint(PreferenceKeys.REGION_SIZE) ?? 10}
+                        max={preference.getMaxConstraint(PreferenceKeys.REGION_SIZE) ?? 100}
                         value={preference.regionSize}
                         stepSize={1}
-                        onValueChange={(value: number) => preference.setPreference(PreferenceKeys.REGION_SIZE, Math.max(10, Math.min(100, value)))}
+                        onValueChange={(value: number) =>
+                            preference.setPreference(PreferenceKeys.REGION_SIZE, Math.max(preference.getMinConstraint(PreferenceKeys.REGION_SIZE) ?? 10, Math.min(preference.getMaxConstraint(PreferenceKeys.REGION_SIZE) ?? 100, value)))
```
izkgao: Would `clamp` help improve readability here?

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -677,11 +680,31 @@ export class PreferenceDialogComponent extends React.Component {
...
-                        max={100}
+                        min={preference.getMinConstraint(PreferenceKeys.POINT_ANNOTATION_WIDTH) ?? 1}
+                        max={preference.getMaxConstraint(PreferenceKeys.POINT_ANNOTATION_WIDTH) ?? 100}
                         value={preference.pointAnnotationWidth}
                         stepSize={1}
-                        onValueChange={(value: number) => preference.setPreference(PreferenceKeys.POINT_ANNOTATION_WIDTH, Math.max(1, Math.min(100, value)))}
+                        onValueChange={(value: number) =>
+                            preference.setPreference(
+                                PreferenceKeys.POINT_ANNOTATION_WIDTH,
+                                Math.max(preference.getMinConstraint(PreferenceKeys.POINT_ANNOTATION_WIDTH) ?? 1, Math.min(preference.getMaxConstraint(PreferenceKeys.POINT_ANNOTATION_WIDTH) ?? 100, value))
```
izkgao: also here

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
```diff
@@ -677,11 +680,31 @@ export class PreferenceDialogComponent extends React.Component {
...
+                    <SafeNumericInput
+                        placeholder="Text line width"
+                        min={preference.getMinConstraint(PreferenceKeys.TEXT_ANNOTATION_LINE_WIDTH) ?? 0.5}
+                        max={preference.getMaxConstraint(PreferenceKeys.TEXT_ANNOTATION_LINE_WIDTH) ?? 10}
+                        value={preference.textAnnotationLineWidth}
+                        stepSize={0.5}
+                        onValueChange={(value: number) =>
+                            preference.setPreference(
+                                PreferenceKeys.TEXT_ANNOTATION_LINE_WIDTH,
+                                Math.max(preference.getMinConstraint(PreferenceKeys.TEXT_ANNOTATION_LINE_WIDTH) ?? 0.5, Math.min(preference.getMaxConstraint(PreferenceKeys.TEXT_ANNOTATION_LINE_WIDTH) ?? 10, value))
```
izkgao: and here

src/components/Dialogs/PreferenceDialog/PreferenceDialogComponent.tsx
loveluthien: 👍

izkgao (approved):

kswang1029 (approved):

PR #2946, opened 2026-09-04
List files before folders in the file browser

markccchiang:
**Description**

This feature was suggested by @pshnghng0318. The file browser (and the catalog, region and image-save browsers, which share the same table component) lists the sub-folders of a directory first, followed by the files. In image directories with many CASA `.image` or `.miriad` folders, the files a user is usually looking for are pushed below the fold and need scrolling to reach.

This PR reverses the default grouping: files are listed first, then folders. Sorting within each group is unchanged, so the Filename, Type, Size, and Date column headers behave exactly as before; only the group order is different.

**What is implemented:**

- `FileListTableComponent.tableEntries` used to push directory rows and then file rows into a single array, which hard-coded the folders-first order. Directory rows are now collected in a separate array and appended after the file rows.

- Unit tests updated for the new order, plus a new test asserting that files precede folders for every sort column in both directions.

**Checklist**

For the pull request:
- [x] changelog updated ~/ no changelog update needed~
- [x] unit test added ~(for functions with no dependenies)~

For dependencies:
- [x] ~protobuf version bumped /~ no protobuf version bumped needed
- [x] ~protobuf updated to the latest dev commit /~ no protobuf update needed
- [x] ~corresponding ICD test fix added (`BackendService` changed) /~ no ICD test fix needed (`BackendService` unchanged)


loveluthien:
@markccchiang, please create a corresponding issue.

kswang1029:
The counter use case is who mostly work with CASA/MIRIAD images. This change will make them facing the same issue. Having folders listed first is also consistent what common OS adopts. 


I don't think we can do much about this. Better just encourage users to use filter feature, or sort by image types.

markccchiang:
How about adding an option in Preferences that allows users to choose whether to list files or folders first, like in the following screenshot?

截圖 2026-09-04 晚上10 26 07

loveluthien:
In the PR, we don't discuss which features to include. Please create an issue, and we can discuss it there.

markccchiang:
I don’t have a strong opinion about this feature. If it’s not an urgent issue, I would prefer not to implement it for now and simply close this PR.

PR #2955, opened 2026-09-08
test(tiling): stabilize performance benchmarks

loveluthien:
- Replace Jest retries for the tiling and tile-coordinate performance tests with a fastest-of-five benchmark pattern in each test file.
- Warm up each benchmark once to account for JIT compilation.
- Run five measured samples and use the fastest result to reduce parallel-worker timing noise.
- Keep the existing correctness checks and performance thresholds

### DIFF
--- a/src/models/Tile/TileCoordinate.test.ts
+++ b/src/models/Tile/TileCoordinate.test.ts
@@ -48,7 +48,7 @@ function fastestRun(run: () => void, samples = 5): number {
     return best;
 }
 
-test("encodes 10000 coordinates in less than 5 ms", () => {
+test("encodes 1M coordinates in less than 20 ms", () => {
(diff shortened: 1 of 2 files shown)


kswang1029 (commented):
Might be worth by checking system overall cpu usage percentage and set a threshold (eg < 50%) before running the test as tiles are generated in parallel and tests are run in parallel(?).

loveluthien:
> Might be worth by checking system overall cpu usage percentage and set a threshold (eg < 50%) before running the test as tiles are generated in parallel and tests are run in parallel(?).

Right. It seems retry doesn't work.

izkgao (approved):

kswang1029 (approved):

PR #2956, opened 2026-09-08
feat: support string-type coordinate columns with sexagesimal and degree formats

### LINKED ISSUE #2954: Coordinate columns in hmsdms format do not show in dropdown menu
**Describe the bug**
Coordinate columns with hmsdms format do not show in dropdown menu.

**To Reproduce**
1. Open an image and a catalog with RA and Dec in hmsdms format
2. See no those columns in the RA, Dec dropdown menu of catalog widget

**Expected behavior**
Coordinate columns in hmsdms format should also show in the dropdown menu.

**Screenshots or videos**
If applicable, add screenshots/videos to help explain your problem.

 - OS [e.g. macOS Monterey]: macOS Tahoe
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: dev
 - Frontend branch [e.g. dev, v3b2 release]: dev

**Additional context**

izkgao:
**Description**

Closes #2954.

Catalog overlays can now use string-valued coordinate columns. A column's format is taken from its declared units, or sniffed from its values when it declares none.

**Formats accepted**

Sexagesimal separated by colons, whitespace or `h`/`m`/`s` and `d`/`m`/`s` letters; decimal degrees; decimal hours; radians; the separator-free `HHMMSS.s` / `±DDMMSS.s` form; CASA's dot-separated declinations (`-021.57.15.4625`); and Unicode `°′″` marks including U+2212 minus. Partial values (`15h17m`, `20:54`) are read; AST's omitted-field spellings (`:45:33`, `6::2.5`) are rejected rather than given a second meaning.

**What decides what**

Whether a column *can* be read as a number is decided from its type, units and values. Which column the user probably *meant* is decided from its name. The two are kept apart: the parser is never shown a column name, and the ranking is never shown a value. That is what lets a column called `dec_deg` be put on the RA axis by hand, and what stops a column called `RA` being read as hours when its values are plainly degrees.

The one deliberate exception is hours-versus-degrees, which a bare `12:30:00` cannot settle on its own. Detection leaves it ambiguous and the axis resolves it, after the column has been assigned somewhere and never before.

**Reading a column that is still arriving**

A large catalog arrives a page at a time, and a column that declares no units cannot be read until some of its values have turned up. What that means in practice:

- **A first page that is blank, or full of `N/A`, does not condemn the column.** The answer is deferred and a later page can settle it. Once it is settled, a stray unreadable row among hundreds of good ones does not remove the column from the axis menu either; that row is dropped on its own.
- **Rows that arrived before the format was known still get plotted.** They are re-read once it settles, rather than left as gaps in the overlay.
- **A column that is nothing but placeholders stops being chased.** Auto-selection gives up after a thousand rows rather than pulling down the whole file looking for one good value.

**Behaviour changes worth a reviewer's attention**

- **A column name no longer disqualifies a column.** A column called `dec_deg` can be put on the RA axis by hand; names decide menu order and what auto-select picks, nothing more.
- **Axis menus list string columns too, ranked rather than filtered**, so a mislabelled catalog stays usable.
- **Sources with a latitude beyond a pole are dropped.**
- **Numeric columns are scaled by their declared units.** A `Double` column in `h` or `rad` was plotted as though its values were degrees.
- **A catalog's declared equinox and epoch are now used.**
- **`COOSYS/@system` is matched against the VOTable enumeration rather than by substring.** `"ECL_FK5"` contains `"FK5"`, so every standards-conforming ecliptic catalog was read as equatorial, which also scaled its longitude by fifteen.
- **Leaving the image overlay clears a string coordinate axis.** Scatter plots and histograms take numeric columns only, so the plot button would otherwise stay enabled on an empty plot.

**How to test**

The test data is the `catalog_variants_2954` set, attached below. Its `README.md` says what each file is for and carries a step-by-step UI procedure of about ten minutes, one image throughout. Every file that is expected to work holds the same 20 sources, so the check is: load one, load another, and no marker should move.

[2954_files_v2.zip](https://github.com/user-attachments/files/32232217/2954_files_v2.zip)

**Checklist**

For the pull request:

For dependencies:


kswang1029 (commented):
CASA adopted the following formatting style for dec, `17:56:21.76513, -021.57.15.4625`. Would this implementation be able to parse it correctly?

loveluthien (approved):
😎

kswang1029 (approved):

PR #2965, opened 2026-09-10
Fix catalog overlay color mapping after replotting with a filter

### LINKED ISSUE #2849: incorrect catalog color mapping after data filtering
**Describe the bug**
When there is already a catalog image overlay rendered with color mapping, if we set a new data filter and replot, we see colors are incorrectly set to sources. In the example, distance is set for color mapping so we expect to see colors in rings if rendered correctly.

Not sure if size and orientation have the same issue.

Initial (correct):

Image

after setting a new filter (incorrect)

Image

**To Reproduce**
1. load an image with a catalog
2. tigger image overlay with color mapping (easier to see the issue if distance column is used)
3. now apply a column filter in the table (easier to see the issue if distance column is used)
4. replot image overlay and see issue. 

**Expected behavior**
color should be assigned correctly. 

 - OS [e.g. macOS Monterey]: macOS Tahoe
 - Browser version [e.g. 22]: 
 - Backend branch [e.g. dev, v3b2 release]: dev, v6.0
 - Frontend branch [e.g. dev, v3b2 release]: dev, v6.0

markccchiang:
**Description**

Fixes #2849. With a catalog image overlay colored by a column (e.g., `ANG_DIST` with `tab10`), applying a column filter and clicking Plot again leaves the sources with the wrong colors. The same applies to size and orientation mapping.

Root cause: the map textures uploaded to WebGL (`colorArray`, `sizeArray`, `sizeMinorArray`, `orientationArray` in `CatalogDisplayStore`) are derived from `@computed` getters that read the column values from the profile store's catalog data. That data is a plain, non-observable `Map` which the filter response mutates in place, so nothing invalidates the computed values: the color texture keeps the unfiltered values while the overlay positions are rebuilt for the filtered sources. The shader looks both textures up by source index, so filtered source *i* is drawn with the color of unfiltered row *i*.

**How to fix**

- `CatalogDisplayStore` reads the number of plotted sources (`CatalogStore.catalogCounts`) as an observable dependency of the four map-data computations, through a shared `getMapColumnData` helper. That count is reset and refilled whenever the overlay positions are rebuilt (Plot, a filter, each streamed chunk in view-update mode), which is exactly when the map textures have to follow the current data. The same trick is already used in `CatalogPlotComponent` (`numVisibleRows` as a dummy dependency) for the same non-observable data.

- The reactions that recompute the default min/max of each map now only reset the range when the data range actually changed. Without this, every replot of unchanged data would have discarded a user's clipped min/max, which was not the case before the fix.

- Unit tests for the color and size textures being recomputed from the current data on replot, and for a clipped color range surviving a replot of the same sources.

**Checklist**

For the pull request:
- [x] changelog updated ~/ no changelog update needed~
- [x] unit test added ~(for functions with no dependenies)~

For dependencies:
- [x] ~protobuf version bumped /~ no protobuf version bumped needed
- [x] ~protobuf updated to the latest dev commit /~ no protobuf update needed
- [x] ~corresponding ICD test fix added (`BackendService` changed) /~ no ICD test fix needed (`BackendService` unchanged)

### DIFF
--- a/src/stores/Catalog/CatalogDisplayStore.ts
+++ b/src/stores/Catalog/CatalogDisplayStore.ts
@@ -139,16 +139,23 @@ export class CatalogDisplayStore {
         this.catalogFileId = catalogFileId;
         makeObservable(this);
 
+        // In world (angular size) mode the size and orientation columns are used as they are: the output ranges are kept equal to the
+        // data range, as when the column is selected, so that the mapping stays an identity when the data changes (e.g. a filter)
         this.disposers.push(
             reaction(
                 () => this.sizeMapData,
                 column => {
                     const {minVal, maxVal} = getDefaultRange(column);
+                    const isRangeChanged = minVal !== this.sizeColumnMin.default || maxVal !== this.sizeColumnMax.default;
                     if (minVal !== this.sizeColumnMin.default) {
                         this.setSizeColumnMin(minVal, "default");
                     }
                     if (maxVal !== this.sizeColumnMax.default) {
                         this.setSizeColumnMax(maxVal, "default");
+                    }
+                    if (isRangeChanged && column.length && this.catalogDisplayMode === CatalogDisplayMode.WORLD) {
+                        this.setSizeMax(maxVal);
+                        this.setSizeMin(minVal);
                     }
                 }
             )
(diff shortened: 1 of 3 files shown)


izkgao (approved):

loveluthien (approved):
😎

kswang1029 (changes requested):
It seems rendering with angular size mapping is not correct if there are filters applied.
before filtering
Image

after filtering
Image

The PA mapping is affected.

markccchiang:
@kswang1029  Thanks, good catch! In angular size mode, the size and P.A. columns are used as they are, which the store achieved by copying the data range into the output range once at column selection. The new range reaction moved the data range after a filter but not the output range, so the mapping stopped being an identity and the angles were rescaled. The reactions now keep both in step in world mode, with tests for size and P.A.

src/stores/Catalog/CatalogDisplayStore.ts
```diff
@@ -134,13 +139,24 @@ export class CatalogDisplayStore {
         this.catalogFileId = catalogFileId;
         makeObservable(this);
 
+        // In world (angular size) mode the size and orientation columns are used as they are: the output ranges are kept equal to the
```
loveluthien: code is clear, so no need this comment

kswang1029 (approved):

